PR #1, opened 2021-08-31
Added documention on design patterns and routing

nishraptor:
Included Documentation on Basic Design Patterns on the AIE and flows and routing.

The design patterns includes information on:
Single Buffered Communication
Double Buffered Communication
Controlling From the ARM processor
Static and Dynamic DDR Configuration (this might have to be changed)

The flows section describes the routing, flows, and shimmux


docs/AIE Routing.md.md
```diff
@@ -0,0 +1,76 @@
+﻿# AIE Flows and Routing
```
stephenneuendorffer, line 1: The name of this file is weird. (.md.md?)
In general, I'd avoid spaces in file names, if possible.

docs/AIE Routing.md.md
```diff
@@ -0,0 +1,76 @@
...
+
+Introduction on how connect tiles in the AIE physical dialect
```
stephenneuendorffer, line 3: how to connect

stephenneuendorffer (approved):
This is a great start, Nishant!  In general, can you add pointers to the full examples, so that people can see it in context?  Then I suggest we commit this and we can continue to refine from here.  One thing I suspect will help is some nice figures to graphically indicate what the code is describing.

docs/AIE Routing.md.md
nishraptor: I updated these docs with the changes you requested - the only thing I did not add was the dynamic ddr example since we don't have an example of that.

PR #16, opened 2021-09-20
chess_compiler_tests in unti_tests

arkhodamoradi:
Some naming corrections plus readme.


test/unit_tests/chess_compiler_tests/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## 1) Precompiled kernel (tests 02 and 04)
+
+In this method, the *kernel.cc* manages the locks and is compiled to an elf file, which can be used to program the AIE.
+
+You can modify and recompile the *kernel.cc* without recompiling the host.
+
+## 2) Precompiled core function (tests 01, 03, 05, and 07)
+
+The *kernel.cc* does not manage the locks in this method and is compiled to an objective file.
```
stephenneuendorffer, line 11: What *does* manage the locks?

test/unit_tests/chess_compiler_tests/README.md
```diff
@@ -0,0 +1,13 @@
+There are two ways to use precompiled functions in your code.
```
stephenneuendorffer, line 1: What are the pros and cons of each approach?  I think generally we should aggressively steer people to #2, unless there's no other way to accomplish what they are trying to do.

test/unit_tests/chess_compiler_tests/README.md
arkhodamoradi: Locks are managed in aie.mlir. ReadMe is adjusted to clarify this.

test/unit_tests/chess_compiler_tests/README.md
arkhodamoradi: The second line of each point has a very brief explanation for each method.
Calling a function vs. loading an elf file for a core. Recompiling also differs in each case.

I see why method two is cleaner and may be preferred when changing code. But I don't know why we should aggressively steer away from method one. Does it harm user experience or future plans/changes? We can remove it to just have method two.

stephenneuendorffer: The main reason to avoid precompiled kernels is that information is replicated in multiple places.  For instance, the code for both processors that access a particular buffer need to be kept synchronized.  We use this mechanism internally, but only as the very last stage of compilation.

stephenneuendorffer (approved):

test/unit_tests/chess_compiler_tests/README.md
arkhodamoradi: It makes sense. I tried to capture your comments in the updated readme. I think it is more clear now.

PR #27, opened 2021-09-30
Improved error message when routing to or from a shim tile with no noc

### LINKED ISSUE #26: Pathfinder: routing to or from a shim tile without a NOC connection leads to assertion
Assertion leads to core dump instead of providing hints for debugging. Should change assertion to error message for clarity.

erwei-xilinx:
Previously this used assertion which would lead to core dump with little information for debugging. Now it uses MLIR's diagnostic (emitError) for displaying error message. Closes #26

### DIFF
--- a/lib/AIECreatePathfindFlows.cpp
+++ b/lib/AIECreatePathfindFlows.cpp
@@ -207,7 +207,6 @@ public:
     if(coordToShimMux.count(std::make_pair(col, row))) {
       return coordToShimMux[std::make_pair(col, row)];
     } else {
-      assert(getTile(builder, col, row).isShimNOCTile());
       ShimMuxOp switchboxOp =
         builder.create<ShimMuxOp>(builder.getUnknownLoc(),
                                     getTile(builder, col, row));
@@ -217,6 +216,9 @@ public:
       coordToShimMux[std::make_pair(col, row)] = switchboxOp;
       maxcol = std::max(maxcol, col);
       maxrow = std::max(maxrow, row);
+      // check if the shim tile has noc connection
+      if(!getTile(builder, col, row).isShimNOCTile())
+        switchboxOp.emitError("Attempting to route to or from ShimTile (") << col << ", " << row << "), which has no NOC connection";
       return switchboxOp;
     }
   }


lib/AIECreatePathfindFlows.cpp
```diff
@@ -217,6 +216,10 @@ class DynamicTileAnalysis {
       coordToShimMux[std::make_pair(col, row)] = switchboxOp;
       maxcol = std::max(maxcol, col);
       maxrow = std::max(maxrow, row);
+      // check if the shim tile has noc connection
+      if (!getTile(builder, col, row).isShimNOCTile())
+        switchboxOp.emitError("Attempting to route to or from ShimTile (")
+            << col << ", " << row << "), which has no NOC connection";
```
stephenneuendorffer, line 222: If it's a precondition check, then why move it after the associated code?

stephenneuendorffer (changes requested):
Please add a unit test.

lib/AIECreatePathfindFlows.cpp
erwei-xilinx: Agreed. This check has now been moved to immediately after when switchboxOp (a ShimMuxOp) has just been created. This error message is linked to that op in MLIR diagnostic, so as to take advantage of the "note" which points to this shim mux.

lib/AIECreatePathfindFlows.cpp
```diff
@@ -207,10 +207,13 @@ class DynamicTileAnalysis {
...
       return coordToShimMux[std::make_pair(col, row)];
     } else {
-      assert(getTile(builder, col, row).isShimNOCTile());
       ShimMuxOp switchboxOp =
         builder.create<ShimMuxOp>(builder.getUnknownLoc(),
                                     getTile(builder, col, row));
+      // check if the shim tile has noc connection
+      if (!getTile(builder, col, row).isShimNOCTile())
+        switchboxOp.emitError("Attempting to route to or from ShimTile (")
+            << col << ", " << row << "), which has no NOC connection";
```
stephenneuendorffer, line 216: So this is still somewhat problematic, because the ShimMuxOp is not really visible to the user.  Maybe it should emit the error on the FlowOp being routed and then have  a note that points to the shimmux or tile op where the error was encountered?

SamuelBayliss: I agree with this - is there a way to put the error on the flow-op that failed.

erwei-xilinx: Agreed. This test has now been moved to where the src and dst TileOps of every FlowOp are first created. Also linked this error to the FlowOp instead of ShimMuxOp.

PR #30, opened 2021-09-30
Edits to correctly implement flows to/from ShimDMAs

jgmelber:
Changed pathfinder flow writing from North to South direction and updated the Mux/Demux logic to check for the DMA WireBundle.


stephenneuendorffer (commented):
Do you have associated unit tests?

PR #34, opened 2021-10-05
Add DialectFoldInterface to AIE dialect

fifield:
This is needed to contain the materialization of constants in
mlir regions that are not isolated from above. It is used here
for the AIE.core op. Without it the folder will move constants
from AIE.core regions into the enclosing module.


stephenneuendorffer (approved):
Please add a unit test.

PR #44, opened 2021-11-01
aiecc multi-threaded flow

swang203:
1. apply a thread-pool for multi-threaded aie-core compilation
2. dispatch the arm code-gen steps to be run parallel with process_core()
3. create a new #16 prime_sieve unit test for multi-thread benchmark and profile.
4. add an option for multi-threaded compilation, the default is "false" before the onnx-mlir issues are resolved

### DIFF
--- a/test/unit_tests/16_prime_sieve_large/code_gen.py
+++ b/test/unit_tests/16_prime_sieve_large/code_gen.py
@@ -1,3 +1,10 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2021 Xilinx Inc.
+
 import sys
 import re
 import math
@@ -45,6 +52,11 @@ def main():
 // (c) Copyright 2021 Xilinx Inc.
 //
 //===----------------------------------------------------------------------===//
+//
+// Note:
+// this large prime-sieve pattern is generated from code_gen.py,
+// it contains 360 cores in this pattern, user could change the core numbers
+// by specifying different rows and cols value in code_gen.py
 
 // RUN: aiecc.py --sysroot=%VITIS_SYSROOT% %s -I%aie_runtime_lib% %aie_runtime_lib%/test_library.cpp %S/test.cpp -o test.elf
 // RUN: %run_on_board ./test.elf \n\n\n""")
--- a/test/unit_tests/16_prime_sieve_large/code_gen.py
+++ b/test/unit_tests/16_prime_sieve_large/code_gen.py
@@ -1,0 +1,231 @@
+import sys
+import re
+import math
+import random
+
+def prime_gen():
+    lower = 2
+    upper = 10000
+
+    prime_cnt = 0
+    results = [1]
+    for num in range(lower, upper + 1):
+        # all prime numbers are greater than 1
+        if num > 1:
+            for i in range(2, num):
+                if (num % i) == 0:
+                    break
+            else:
+                results.append(num)
+                prime_cnt = prime_cnt + 1
+        if (prime_cnt == 500):
+            break
+
+    print("prime_cnt = %d" %(prime_cnt))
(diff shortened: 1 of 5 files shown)


test/unit_tests/16_prime_sieve_large/code_gen.py
```diff
@@ -0,0 +1,231 @@
+import sys
```
stephenneuendorffer, line 1: Needs copyright header

test/unit_tests/16_prime_sieve_large/code_gen.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+    f.write("""//===- aie.mlir ------------------------------------------------*- MLIR -*-===//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2021 Xilinx Inc.
+//
+//===----------------------------------------------------------------------===//
```
stephenneuendorffer, line 47: It's generally good practice to have a comment that automatically generated files are automatically generated.

tools/aiecc/aiecc/main.py
```diff
@@ -120,31 +120,42 @@ def process_core(core):
...
+      cmd += ['-fuse-ld=lld','-rdynamic','-lxaiengine','-lmetal','-lopen_amp','-ldl']
+
+      if(len(opts.arm_args) > 0):
+        do_call(cmd + opts.arm_args)
+
+
+    if (opts.nthreads == True):
+      with ThreadPool() as thdpool:
+          thdpool.map(process_core, (cores))
+          thdpool.apply_async(process_arm_cgen)
```
stephenneuendorffer, line 150: How does this code prioritize the arm_cgen task, which we expect to be long?

tools/aiecc/aiecc/cl_arguments.py
```diff
@@ -65,6 +65,11 @@ def parse_args():
             action='store',
             help='arguments for ARM compiler',
             nargs=argparse.REMAINDER)
+    parser.add_argument('--nthreads',
+            dest="nthreads",
+            default=False,
+            action='store_false',
+            help='Compile with multi-thread pool')
```
stephenneuendorffer, line 72: "-j" would be the conventional option here...

tools/aiecc/aiecc/main.py
swang203: I cannot test now, but if I remember correctly, putting the apply_async() before the map() has almost the same performance. But you are right, spawning the arm_code_gen() in advance should be theoretically faster. Will change that.

test/unit_tests/16_prime_sieve_large/code_gen.py
swang203: ok will do it!

test/unit_tests/16_prime_sieve_large/code_gen.py
swang203: ok will do it!

tools/aiecc/aiecc/main.py
```diff
@@ -15,7 +15,7 @@
 import sys
 import time
 from subprocess import PIPE, run, call
-# from joblib import Parallel, delayed
```
stephenneuendorffer, line 18: Remove comment?

tools/aiecc/aiecc/main.py
stephenneuendorffer: It could be worth a comment like "prefer to run process_arm_cgen first, since it typically takes longer than process_core".

stephenneuendorffer (approved):

stephenneuendorffer:
Looks like you need to run git clang-format..

stephenneuendorffer:
Can you resolve the conflicts so we can commit this?

swang203:
> Can you resolve the conflicts so we can commit this?

Hello,
I don't notice the upstream also change the main.py recently. I fix it, can you help to commit it? Thanks.

PR #45, opened 2021-11-02
Fixed prepend path bug in lit.cfg.py and added libxaiev2 support for aiecc.py

jackl-xilinx:


tools/aiecc/aiecc/cl_arguments.py
```diff
@@ -61,6 +61,16 @@ def parse_args():
...
+    parser.add_argument('--aie-generate-xaie',
+            dest="xaie",
+            default=1,
+            action='store_const', const=1,
+            help='Generate libxaie v1 drivers (default is v1)')
+    parser.add_argument('--aie-generate-xaiev2',
+            dest="xaie",
+            default=1,
+            action='store_const', const=2,
+            help='Generate libxaie v2 drivers (default is v1)')
```
stephenneuendorffer, line 73: I think we will probably have other targets, too...  can I suggest "--target=libxaiev1", "--target=libxaiev2" for these options?

jackl-xilinx: Good idea. I'll make that change on the next commit.

PR #49, opened 2021-11-07
Start of .gitignore

hanchenye:
Added the .gitignore file to the root. Not sure how to deal with the platforms directory though. Does it make sense to factor out a new repository to maintain the platforms?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,13 @@
+/build*
+.vscode
+__pycache__
+
+/platforms/vck190_bare/petalinux/build
+/platforms/vck190_bare/petalinux/components
+/platforms/vck190_bare/petalinux/images
+/platforms/vck190_bare/petalinux/project-spec/configs/*.old
+/platforms/vck190_bare/petalinux/sysroot
+
+/platforms/vck190_bare/vivado/*.xpe
+/platforms/vck190_bare/vivado/*.xsa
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -0,0 +1,5 @@
+/build*
+.vscode
+__pycache__
+
+/platforms
```
stephenneuendorffer, line 5: I think platforms are a submodule, so shouldn't be a problem?

stephenneuendorffer (approved):

.gitignore
hanchenye: The vck190_bare platform is not a submodule, which generates build files of petalinux and vivado after making.

stephenneuendorffer: It generates them in the platforms directory? That seems.. wrong...

hanchenye: Right. Based on the [document](https://xilinx.github.io/mlir-aie/Platform.html), we do `make all` under `platforms/vck190_bare` to built the petalinux and vivado project. The generated files are all under `platforms/vck190_bare`.

stephenneuendorffer: Can you make the ignore pattern more selective then?  Ultimately, this seems like something that should be fixed, though. so we're not generating things in the source tree.

stephenneuendorffer (approved):

PR #53, opened 2021-11-09
Added the ability to extend pathfinder flows to/from PLIO in shim tiles.

jgmelber:
Added flows to/from all shim tiles for PLIO connections. Most PLIOs just configure the stream switch to direct streams to/from PLIO, shim noc tiles require shimmux configuration for some PLIOs as well.

### DIFF
--- a/test/create-flows/more_flows_shim.mlir
+++ b/test/create-flows/more_flows_shim.mlir
@@ -1,0 +1,93 @@
+//===- more_flows_shim.mlir ------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2021 Xilinx Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt --split-input-file --aie-create-pathfinder-flows %s | FileCheck %s
+// CHECK: module
+// CHECK: %[[T70:.*]] = AIE.tile(7, 0)
+// CHECK: %[[T71:.*]] = AIE.tile(7, 1)
+// CHECK:  %{{.*}} = AIE.switchbox(%[[T70]])  {
+// CHECK:    AIE.connect<North : 0, South : 2>
+// CHECK:  }
+// CHECK:  %{{.*}} = AIE.switchbox(%[[T71]])  {
+// CHECK:    AIE.connect<North : 0, South : 0>
+// CHECK:  }
+// CHECK:  %{{.*}} = AIE.shimmux(%[[T70]])  {
+// CHECK:    AIE.connect<North : 2, PLIO : 2>
+// CHECK:  }
+
+module {
+  %t70 = AIE.tile(7, 0)
+  %t71 = AIE.tile(7, 1)
+  AIE.flow(%t71, North : 0, %t70, PLIO : 2)
+}
+
+// -----
+
+// CHECK: module
+// CHECK: %[[T60:.*]] = AIE.tile(6, 0)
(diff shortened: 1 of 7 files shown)


test/create-flows/more_flows_shim.mlir
```diff
@@ -0,0 +1,93 @@
+//===- more_flows_shim.mlir ------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2021 Xilinx Inc.
+//
+//===----------------------------------------------------------------------===//
+
```
stephenneuendorffer, line 10: Because of the differences depending on the shim tiles, can you comment these test cases more?

test/create-flows/more_flows_shim.mlir
```diff
@@ -0,0 +1,93 @@
...
+// CHECK: %[[T41:.*]] = AIE.tile(4, 1)
+// CHECK:  %{{.*}} = AIE.switchbox(%[[T40]])  {
+// CHECK:    AIE.connect<North : 0, South : 3>
+// CHECK:    AIE.connect<South : 4, North : 0>
+// CHECK:  }
+// CHECK:  %{{.*}} = AIE.switchbox(%[[T41]])  {
+// CHECK:    AIE.connect<North : 0, South : 0>
+// CHECK:    AIE.connect<South : 0, North : 0>
+// CHECK:  }
+
```
stephenneuendorffer, line 65: and here "Tile 4,0 has no shimmux."  Actually this case is somewhat inconsistent since there is no 'connect' to a PLIO bundle...

stephenneuendorffer (approved):

test/create-flows/more_flows_shim.mlir
jgmelber: Thanks for the comments, I will add more comments to the tests. What do you mean by inconsistent here? 

The concept is that every PLIO in a shim tile (both NoC and PL) can be routed to/from with consistent syntax. The PLIOs are numbered sequentially and hide the details of whether that particular PLIO passes through a shimmux or not. Even within a NoC tile, with the shimmux, not every PLIO passes through the shimmux.

stephenneuendorffer: Under normal circumstances if you have:
`flow<%t1, DMA:0, %t2, CORE:0>`
Then there will eventually result in:
```
connect<DMA:0, ....>
connect<...., CORE:0>
```
This is also true for tiles with a ShimMux
```
flow<%shimTileWithMux, PLIO:0, %otherShimTileWithMux, PLIO:1>
AIE.shimmux(%shimTileWithMux)  {
connect<PLIO:0, NORTH:something>
}
AIE.shimmux(%otherShimTileWithMux)  {
connect<NORTH:something, PLIO:1>
}

```
But it's not true for tiles without a shimMux,  since there is no shimMux and the south port of the stream switch 'just connects' to the PL.  Is this confusing/misleading?  probably... I'm not sure that there is a good fix though...  This is something that an AIE.wire could help with, since it would make explicit the non-configurable connection.  But it's not clear that wires actually add any information... they are just 'documentation' at some level.

jgmelber: Ahh, I understand now. It is somewhat confusing that the result of some flows to/from PLIO just connect South from/to the PL. Would AIE.plio ops help? It seems like they are also just a form of documentation and don't have any information about which PLIO in the column.

stephenneuendorffer: Maybe...  We should probably look more closely at some of the Vail architecture plans to come up with a scheme that works there too..

jgmelber:
Comments added to tests and updated the documentation AIERouting.md

PR #54, opened 2021-11-10
Added initial set of benchmarks for AIE

nishraptor:
The runtime lib is updated with the performance counter class that allows the use of performance counters. The benchmarks and calibration tests have been added to the test/ folder, along with a README that discusses the individual benchmarks and explains the measurement tools.


stephenneuendorffer:
Is this ready to move towards being committed?  There are some conflicts that need to be resolved and looks like 'git clang-format' needs to be run.

PR #55, opened 2021-11-15
Tile DMA 2D X and Y address generation 

jgmelber:
Introduced the dmaBd2D op supporting X and Y addressing mode in the tile DMA. This allows for many different addressing schemes to reorganize data at the 32-bit stream beat granularity. For example, transposed matrix addressing.

### DIFF
--- a/include/aie/AIE.td
+++ b/include/aie/AIE.td
@@ -739,6 +739,53 @@ def DMAChan: I32EnumAttr<"DMAChan", "DMA Channel number",
   let cppNamespace = "xilinx::AIE";
 }
 
+def AIE_DMABD2DOp: AIE_Op<"dmaBd2D", []> {
+  let summary = "Enable 2D (X and Y) addressing scheme for a dma block descriptor";
+  let description = [{
+    This operation enables 2 dimensional address generation for a block descriptor for DMA operations. 
+    In particular, it specifies the offset, increment and wrap in the X and Y dimension.
+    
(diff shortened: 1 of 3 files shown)


jgmelber:
This operation enables 2 dimensional address generation for a block descriptor for DMA operations. In particular, it specifies the offset, increment and wrap in the X and Y dimension.

    AIE.dmaBd2d(X_offset, X_increment, X_wrap, Y_offset, Y_increment, Y_wrap)

Offset = value to add to the address for each increment
Wrap = how many increments before wrapping
Increment = how many stream beats before increment

The address is generated according to the following formulas: 
    Address = Base Address + Xn + Yn
    Xn = Xoffset * (( n / Xincrement) mod Xwrap)
    Yn = Xoffset * (( n / Yincrement) mod Ywrap)

This operation must be used in an MLIR block that lives inside a MemOp's region, and before AIE.dmaBd.
The block descriptor specifies what lock to use and the buffer configuration.

Example:
      ^bd5:
        AIE.useLock(%lck, "Acquire", 0, 0)
        AIE.dmaBd2D(4, 1, 8, 1, 8, 4)
        AIE.dmaBd(<$buf0 : memref<4x8xi32>, 0, 32>, 1)
        AIE.useLock(%lck, "Release", 1, 0)
        br ^bd6 // point to the next Block

stephenneuendorffer (commented):
I see that my comments were still pending: sorry about that.

PR #59, opened 2021-11-24
Teach aiecc.py to use elf_file attribute

fifield:
The libxae backend does the right thing for the 'elf_file' attribute but it's ignored by aiecc.py. I'm not super thrilled about using an eval of a string to pass information but that's how the existing code worked.

### DIFF
--- a/lib/Targets/AIETargets.cpp
+++ b/lib/Targets/AIETargets.cpp
@@ -314,7 +314,11 @@ SECTIONS
           int col = tileOp.colIndex();
           int row = tileOp.rowIndex();
           if (auto coreOp = tileOp.getCoreOp()) {
-            output << '(' << std::to_string(col) << ',' << std::to_string(row) << "),";
+            std::string elf_file = "None";
+            if (auto fileAttr = coreOp->getAttrOfType<StringAttr>("elf_file"))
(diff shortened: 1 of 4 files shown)


stephenneuendorffer:
I've been trying to push to remove this and focus on 'link_with' instead.  Is this something that's really useful?

FWIW, I'm not super happy about evaluating strings either. Once the python extension parts in MLIR are stable, I'd like to dump this and use a python extension in aiecc.py to access the IR directly.

fifield:
I needed a mechanism to control the naming and/or location of the elf files at compile time (where they're written to) and at runtime (where they're loaded from). This attribute already did half of what I needed so I just added the other half.

lib/Targets/AIETargets.cpp
```diff
@@ -314,7 +314,11 @@ SECTIONS
           int col = tileOp.colIndex();
           int row = tileOp.rowIndex();
           if (auto coreOp = tileOp.getCoreOp()) {
-            output << '(' << std::to_string(col) << ',' << std::to_string(row) << "),";
+            std::string elf_file = "None";
```
stephenneuendorffer, line 317: I wonder if aiecc.py is going to rely on this mechanism to compute the names of the elf files, then maybe it should always use it?  basically this would push the 'corefile' function from aiecc.py into C++.

PR #65, opened 2021-11-30
Roll back shimmux fix - change was intended. Fixed chess compiler shimmux tests.

jackl-xilinx:


stephenneuendorffer:
If you're reverting a previous change, please point to the change that you're reverting.

PR #66, opened 2021-12-04
Update target-triple unit test using new APIs

hanchenye:
Accommodate the change in #57


hanchenye:
This has been solved in #68

PR #70, opened 2021-12-16
Bump LLVM to 4b553297ef3ee4dc2119d5429adf3072e90fac38

hanchenye:
This patch aligns the LLVM version with CIRCT to make use of the `handshake` dialect in CIRCT.

Several main changes:
1. Arithmetic dialect:
    - Rename `constant` to `arith.constant`, `addi` to `arith.addi`, etc. in all test cases
    - Register Arithmetic dialect in `aie-opt` and `aie-translate`
2. OpConversionPattern class:
    - `matchAndRewrite` and `rewrite` methods now take `OpAdaptor` as input instead of operands list `ArrayRef<Value>`
    - Overridable `match` method now takes the concrete operation as input instead of `Operation *`

All tests passed on my machine, hope this can pass the CI 🙏As most of changes come from clang-format, I'll file another formatting PR to make this patch cleaner.

### DIFF
--- a/include/aie/AIENormalizeAddressSpaces.td
+++ b/include/aie/AIENormalizeAddressSpaces.td
@@ -19,8 +19,8 @@ def toDefaultAddressSpace : NativeCodeCall<"TypeAttr::get(memRefToDefaultAddress
 def hasNonDefaultAddressSpace : Constraint<
     CPred<"$0.getValue().cast<MemRefType>().getMemorySpace() != 0">,
     "has non-default address space">;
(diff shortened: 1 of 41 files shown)


hanchenye:
This patch has also been rebased and ready for review!

stephenneuendorffer:
I've bumped LLVM to be what we use internally... Can you check and see that this works for you?  I've cleaned things up for our internal repo, so it should be easier to stay close to HEAD.

hanchenye:
@stephenneuendorffer The new LLVM version works, thanks! I'll close this PR.

Issue #75, opened 2022-01-12
[RoutePacketFlow] Issue when merging packet rules


hanchenye:
Hi, I'm trying to route a GEMM design with 40 AIEs using the `-create-packet-flows` pass, but encountered some issue there.

I dig a little and found it may be caused by the packet rules merging. See the test case below, both packet0 and packet1 from West0 should be connected to Core0 and they are merged into the same packet rule. Based on my understanding, the correct packet rule should be `AIE.rule(30, 0, %2)` rather than `AIE.rule(30, 1, %2)` in order to match both packet0 and packet1.

From:
https://github.com/Xilinx/mlir-aie/blob/9fe602308bc889a15684515480eefe5d639c7e85/test/create-packet-flows/test_create_packet_flows5.mlir#L31-L44
To:
https://github.com/Xilinx/mlir-aie/blob/9fe602308bc889a15684515480eefe5d639c7e85/test/create-packet-flows/test_create_packet_flows5.mlir#L15-L24

To verify this observation, I disabled the packet rules merging in RoutePacketFlow, which means the test case above is transformed to:
```mlir
AIE.packetrules(West : 0) { 
  AIE.rule(31, 0, %2)
  AIE.rule(31, 1, %2) 
} 
```

And it works on some small test cases (for example, a scale-down GEMM design with 8 AIEs)!

Then, I tried the modified pass on my original GEMM design, unfortunately, the design still doesn't work - it stalls somewhere due to incorrect streaming. Note that the same GEMM design routed with circuit-mode works correctly on board.

I attached two versions of designs for your reference, the first is connected through FlowOp - it works after `-create-pathfinder-flows` pass, and the second one is connected through PacketFlowOp - it doesn't work even with the modified `-create-packet-flows` pass.

Hanchen

[tmp-circuit.mlir.txt](https://github.com/Xilinx/mlir-aie/files/7858225/tmp-circuit.mlir.txt)
[tmp-packet.mlir.txt](https://github.com/Xilinx/mlir-aie/files/7858226/tmp-packet.mlir.txt)

stephenneuendorffer:
Fix pending

PR #76, opened 2022-01-16
[RoutePacketFlows] Fix a bug, NFC.

hanchenye:
This solves another issue found in RoutePacketFlows other than #75.


stephenneuendorffer (approved):
ouch... surprised that didn't trigger a warning?

PR #78, opened 2022-01-20
[TokenAnalysis][CreateLocks] Support ShimDMAOp

hanchenye:


lib/AIECreateLocks.cpp
```diff
@@ -51,9 +55,12 @@ struct Token2LockLowering : public OpConversionPattern<UseTokenOp> {
...
       srcRow = mem.rowIndex();
       IsParentMemOp = true;
+    } else if (auto shimDma = dyn_cast<ShimDMAOp>(parentOp)) {
+      srcCol = shimDma.colIndex();
+      srcRow = shimDma.rowIndex();
     } else {
-      llvm_unreachable(
-          "A parent operation of UseTokenOp must be either CoreOp or MemOp");
+      llvm_unreachable("A parent operation of UseTokenOp must be either CoreOp "
+                       "or MemOp or ShimDMAOp");
```
stephenneuendorffer, line 63: probably this should be moved to a verifier?  Actually I was thinking we should extract all of the 'token' stuff into a separate dialect entirely.

stephenneuendorffer (changes requested):
Add a test please?

lib/AIECreateLocks.cpp
hanchenye: Added a verifier for UseTokenOp.

test/create-locks/test_lock_shimdma.mlir
```diff
@@ -0,0 +1,98 @@
...
+module @test_lock_shimdma {
+  AIE.token(0) {sym_name = "token0"}
+  %buf_ext = AIE.external_buffer 0 : memref<256xi32>
+
+  %t60 = AIE.tile(6, 0)
+  %c60 = AIE.core(%t60) {
+    AIE.useToken @token0(Acquire, 2)
+    AIE.useToken @token0(Release, 3)
+    AIE.end
+  }
```
hanchenye, line 67: Shim tile should not have a CoreOp in theory. The token uses here should be located in a runtime dialect operation. I'm experimenting on a runtime dialect locally and will file a PR when it is more mature.

hanchenye: Based on my experiments, a CoreOp associated with a shim tile will not break the aiecc.

stephenneuendorffer: I don't understand the intention here... can you elaborate?

hanchenye: The token uses here should happen on the host CPU sides such that we can avoid data pollution & enable double buffering between DDR and host CPU.

For example in this case, only if the host CPU has acquired token0 with value 2, it should begin to use the data on the corresponding external buffer. Similarly, if value 3 is not released by the host CPU, the shim tile should not launch the DMA. Does this make sense?

stephenneuendorffer: Got it.. one thing I've pondered is having a host CPU exist in the MLIR, which might be a better place for this.. In the meantime, this is probably good enough..  Maybe you can capture this as a TODO in the file.

hanchenye: Added a TODO note here!

hanchenye:
Hi @stephenneuendorffer, could you help to merge this PR?

hanchenye:
Thanks!

PR #81, opened 2022-01-25
Add aievec dialect, associated optimizations, and aievec-to-cpp transl…

pssrawat:
…ation. This allows generation of autovectorized AIE kernels from scalar kernels written in affine dialect


stephenneuendorffer (approved):
It doesn't seem like you have any true unit-tests... At best you're testing combinations of affine-super-vectorize and --aie-vectorize.  Can you add some tests that are pure unit tests?

Please try to address as much of the feedback as you have time to.

PR #107, opened 2022-06-02
findflows antenna detection initial commit

iavssw:
This is the initial implementation of the antenna detection solution in the findflows pass.

### DIFF
--- a/lib/AIEFindFlows.cpp
+++ b/lib/AIEFindFlows.cpp
@@ -25,6 +25,100 @@ typedef std::pair<int, int> MaskValue;
 typedef std::pair<Operation *, Port> PortConnection;
 typedef std::pair<Port, MaskValue> PortMaskValue;
 typedef std::pair<PortConnection, MaskValue> PacketConnection;
+typedef std::pair<PortConnection, PacketConnection> Source_PacketConnection;
+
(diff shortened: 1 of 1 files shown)


lib/AIEFindFlows.cpp
```diff
@@ -133,20 +228,115 @@ class ConnectivityAnalysis {
       // If there is no wire to follow then bail out.
       if (!nextConnection.hasValue())
         continue;
-
+      PortConnection sour = std::make_pair(switchOp, nextPort);
+      PacketConnection dest = std::make_pair(nextConnection.getValue(), newMaskValue);
```
stephenneuendorffer, line 232: generally prefer full words "source" and "destination"

iavssw:
Thanks for the feedback. I’ll try to improve the code based on your feedback.
Also, I’ve tracked down the reason for my email not working.  It was due to the mail being forwarded to the amd account. I’ve now logged in using amd credentials and will be able to respond to you via that email.

Greg
________________________________

iavssw:
Version 2 of the antenna detection feature update. Removed bugs. Simplified the code to minimize redundant operations. Now using the MLIR Diagnostic Infrastructure to output messages.

Issue #101, opened 2022-05-10
Replace !aievec.acc<> with vector<> 


stephenneuendorffer:
At the level of MLIR, modeling the accumulator separately doesn't add much and complicates the code significantly, since it has to deal with these types independently.

PR #109, opened 2022-06-02
Use vector instead of AccType for float Mul + FMA

brandon-t-nguyen:
The AIEVec Mul and FMA ops use the AIEVec AccType to model the
accumulators of the integer path. These accumulators are only present
in the integer path. In order to move data between vector registers
and accumulators AIEVec UPS and SRS ops are inserted.

Prior to this commit both integer and floating point AIEVec Mul and FMAs
use the AIEVec AccType. However, as a result of the floating point
AIEVec Mul and FMA ops outputting AIEVec AccTypes, AIEVec UPS
and SRS ops are unnecessarily inserted (and later ignored).

This commit has floating point AIEVec Muls and FMAs output normal
vectors instead of AIEVec AccTypes, and in doing so clean up the
floating point path so that the unnecessary AIEVec UPS and SRS ops
are not inserted.

### DIFF
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -117,7 +117,7 @@ def AIEVec_FMAOp :
   AIEVec_Op<"mac", [
     NoSideEffect
   ]>,
-  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, AIEVec_Acc:$acc,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, AnyTypeOf<[AIEVec_Acc, VectorOf<[F32]>]>:$acc,
                DefaultValuedStrAttr<StrAttr, "">:$xstart,
                DefaultValuedStrAttr<StrAttr, "">:$xoffsets,
                DefaultValuedStrAttr<StrAttr, "">:$xoffsets_hi,
@@ -129,7 +129,7 @@ def AIEVec_FMAOp :
                DefaultValuedStrAttr<StrAttr, "">:$zstep,
                DefaultValuedStrAttr<StrAttr, "">:$zsquare,
                DefaultValuedAttr<BoolAttr, "false">:$fmsub)>,
-  Results<(outs AIEVec_Acc:$result)> {
+  Results<(outs AnyTypeOf<[AIEVec_Acc, VectorOf<[F32]>]>:$result)> {
   let summary = "AIE vector fused multiply-add";
   let description = [{
     Xilinx-specific multiply-add operation. It multiplies two 1-D vectors,
@@ -199,7 +199,7 @@ def AIEVec_MulOp:
                DefaultValuedStrAttr<StrAttr, "">:$zoffsets_hi,
                DefaultValuedStrAttr<StrAttr, "">:$zstep,
                DefaultValuedStrAttr<StrAttr, "">:$zsquare)>,
-  Results<(outs AIEVec_Acc:$result)> {
+  Results<(outs AnyTypeOf<[AIEVec_Acc, VectorOf<[F32]>]>:$result)> {
   let summary = "AIE vector multiply";
   let description = [{
     Xilinx-specific multiply operation that multiplies two 1-D vectors.
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -117,7 +117,8 @@ def AIEVec_FMAOp :
   AIEVec_Op<"mac", [
     NoSideEffect
   ]>,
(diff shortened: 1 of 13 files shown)


include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -117,7 +117,7 @@ def AIEVec_FMAOp :
   AIEVec_Op<"mac", [
     NoSideEffect
   ]>,
-  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, AIEVec_Acc:$acc,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, AnyTypeOf<[AIEVec_Acc, VectorOf<[F32]>]>:$acc,
```
stephenneuendorffer, line 120: I think we can just remove AIEVec_Acc

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -389,30 +394,40 @@ void aievec::FMAOp::print(OpAsmPrinter &p) {
...
+  bool isInt = lhsType.getElementType().isa<IntegerType>();
+
+  VectorType resultVecType =
+      op.result().getType().template dyn_cast<VectorType>();
+  aievec::AccType resultAccType =
+      op.result().getType().template dyn_cast<aievec::AccType>();
+  if (isInt && !resultAccType)
     return op.emitError("requires accumulator type");
+  else if (!isInt && !resultVecType)
+    return op.emitError("requires vector type");
```
stephenneuendorffer, line 413: It seems like these checks should just be type constraints on the op?

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -389,30 +394,40 @@ void aievec::FMAOp::print(OpAsmPrinter &p) {
...
-      op.result().getType().template dyn_cast<aievec::AccType>();
   VectorType lhsType = op.lhs().getType().template dyn_cast<VectorType>();
   VectorType rhsType = op.rhs().getType().template dyn_cast<VectorType>();
 
   if (!lhsType || !rhsType)
     return op.emitError("requires vector type");
-  if (!resultType)
+
+  // Determine if it is an integer or float operation
+  bool isInt = lhsType.getElementType().isa<IntegerType>();
```
stephenneuendorffer, line 404: Perhaps it would make sense to also separate the integer and floating point cases into separate ops?

stephenneuendorffer:
Looks like a good start!  Does this generate the same C code as before?

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
brandon-t-nguyen: For the sake of keeping the scope of this PR to removing the AIEVec_Acc (and thus the SRS and UPS) from just the float path, I opted to keep the AIEVec_Acc to reflect the current state of how the rest of the integer operations still use it to model the accumulator. AIEVec_Acc is still needed here due to how UPS is specified to have an AIEVec_Acc result: removing the AIEVec_Acc here would require additional changes to UPS that are more inline with the PR for #101 that I'm going to make that removes AIEVec_Acc entirely.

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
brandon-t-nguyen: Huh, yeah it would seem that these sorts of checks should already be handled by the tablegen constraints: I'm seeing them all over the file. I'll take a look to see if these sorts of checks actually catch anything the tablegen constraints don't. Would clearing out these dead type checks constitute another issue/PR?

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
brandon-t-nguyen: This is something that I was contemplating as well. However, with the upcoming PR for #101, I end up not actually needing to differentiate between integer and float all that much since I mostly did it for the sake of dealing with AccType, the exception being to determine whether a UPS is needed when the AccType isn't around to serve as the trigger. This does seem like something to investigate as part of figuring out how abstract we want the dialect to be.

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
stephenneuendorffer: Yes, handling it separately is fine.  Having logically atomic and separately verifiable changes is good practice.

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
stephenneuendorffer: Good point, forgot there was a separate issue for it.

stephenneuendorffer (approved):

brandon-t-nguyen:
> Looks like a good start! Does this generate the same C code as before?

`diff` shows no differences between `main` and this branch as far as the generated code (`gen.cc`) for the `unit_tests/aievec_test`s go; would this be comprehensive enough?

PR #110, opened 2022-06-15
antenna detection v2 intial commit

iavssw:
Antenna Detection feature update. Generates warnings using the MLIR Diagnostic Infrastructure when antennas are in the design. 
(reupload due to branch error).


lib/AIEFindFlows.cpp
```diff
@@ -100,15 +181,20 @@ class ConnectivityAnalysis {
...
 
   std::vector<PacketConnection>
-  maskSwitchboxConnections(Operation *switchOp,
+  maskSwitchboxConnections(Graph &g, std::vector<int> &antennaIndex,
+                           Operation *switchOp,
                            std::vector<PortMaskValue> nextPortMaskValues,
                            MaskValue maskValue) const {
     std::vector<PacketConnection> worklist;
-    bool matched = false;
+    // bool matched = false;
```
stephenneuendorffer, line 197: remove?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
         // Incoming packets cannot match this rule. Skip it.
         continue;
       }
-      matched = true;
+      // matched = true;
```
stephenneuendorffer, line 213: remove?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
+      antenna_nonvalid_path.clear();
+      for (auto &node_idx : path_buffer) {
+        if (std::count(valid_path.begin(), valid_path.end(), node_idx)) {
+          antenna_valid_path.push_back(node_idx);
+        } else {
+          antenna_nonvalid_path.push_back(node_idx);
+        }
+      }
+      // emit warning message for antennas
+      for (auto &anvp : antenna_nonvalid_path) {
```
stephenneuendorffer, line 268: anvp -> switch?  vertex?  e? v?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
+      }
+    }
+
+    // output antenna path
+    std::vector<int> antenna_valid_path;
+    std::vector<int> antenna_nonvalid_path;
+    for (auto &antenna_idx : antennaIndexList) {
+      path_buffer = g.get_path_to_root(antenna_idx);
+      antenna_valid_path.clear();
+      antenna_nonvalid_path.clear();
```
stephenneuendorffer, line 259: bring the declarations inside the loop?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
 
+      g.add_child(switchOp, nextPort, nextConnection.getValue().first,
+                  nextConnection.getValue().second, false);
       worklist.push_back(
           std::make_pair(nextConnection.getValue(), newMaskValue));
     }
     return worklist;
   }
 
+  void detect_antenna(Graph &g, std::vector<int> connectedTilesIndexList,
```
stephenneuendorffer, line 234: document?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
+                      std::vector<int> antennaIndexList) const {
+
+    std::vector<int> path_buffer;
+    // create valid path from connected tiles
+    std::vector<int> valid_path;
+    for (auto &Tile_idx : connectedTilesIndexList) {
```
stephenneuendorffer, line 240: tile_idx?

lib/AIEFindFlows.cpp
```diff
@@ -26,6 +26,75 @@ typedef std::pair<Operation *, Port> PortConnection;
...
+        else if (!checkBundleChannel)
+          idx = i;
+      }
+    }
+    return idx;
+  }
+
+  std::vector<int> get_path_to_root(int start_node) {
+    std::vector<int> valid_path;
+    std::vector<int> worklist;
```
stephenneuendorffer, line 82: Seems like worklist can only have at most one element?  Why have a list?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
     }
     return worklist;
   }
 
+  void detect_antenna(Graph &g, std::vector<int> connectedTilesIndexList,
+                      std::vector<int> antennaIndexList) const {
+
+    std::vector<int> path_buffer;
+    // create valid path from connected tiles
+    std::vector<int> valid_path;
```
stephenneuendorffer, line 239: Isn't this just the set of connected tiles?  Would a set-like container be more appropriate?  See: https://llvm.org/docs/ProgrammersManual.html#picking-the-right-data-structure-for-a-task

lib/AIEFindFlows.cpp
```diff
@@ -176,26 +349,20 @@ class ConnectivityAnalysis {
...
-        // append to the worklist
-        worklist.insert(worklist.end(), newWorkList.begin(), newWorkList.end());
-        if (nextPortMaskValues.size() > 0 && newWorkList.size() == 0) {
-          // No rule matched some incoming packet.  This is likely a
-          // configuration error.
-          LLVM_DEBUG(llvm::dbgs() << "No rule matched incoming packet here: ");
-          LLVM_DEBUG(other->dump());
-        }
-      } else if (auto switchOp = dyn_cast_or_null<ShimMuxOp>(other)) {
+        connectedTilesIndex.push_back(
```
stephenneuendorffer, line 352: So, here the code is creating a new graph data structure with the goal of giving good error messages.  If the data structure is only used to create error messages, maybe we should optimize for the 'there is no error case' and traverse the connections again if we find any antennas?  Only then would we create the graph?
We might still need some bookkeeping in order to detect other cases (like "dangling island" paths that don't start at a FlowEndPoint), I guess?

lib/AIEFindFlows.cpp
```diff
@@ -124,29 +210,108 @@ class ConnectivityAnalysis {
...
+    std::vector<int> valid_path;
+    for (auto &Tile_idx : connectedTilesIndexList) {
+      if (valid_path.empty()) {
+        valid_path = g.get_path_to_root(Tile_idx);
+      } else {
+        path_buffer = g.get_path_to_root(Tile_idx);
+        for (auto &node_idx : path_buffer) {
+          if (!std::count(valid_path.begin(), valid_path.end(), node_idx)) {
+            valid_path.push_back(node_idx);
+          }
```
stephenneuendorffer, line 248: Else break?  (since other elements in the path would have been accounted for with some other path, so we don't need to keep calling std::count, which is expensive for a vector i.e. O(n)).

stephenneuendorffer (changes requested):

stephenneuendorffer:
This is a good start!

iavssw:
Thanks for the feedback!

lib/AIEFindFlows.cpp
```diff
@@ -92,26 +153,40 @@ class ConnectivityAnalysis {
...
-                    std::make_pair(masterSetOp.destPort(), maskValue));
+                PortConnection portconnection = std::make_pair(
+                    masterSetOp.getOperation(), masterSetOp.destPort());
+                opportSet.push_back(std::make_pair(portconnection, maskValue));
               }
             }
       }
     }
-    return portSet;
+    return opportSet;
```
stephenneuendorffer, line 177: CamelCase?

lib/AIEFindFlows.cpp
```diff
@@ -63,15 +104,35 @@ class ConnectivityAnalysis {
...
+
+    // can use begin()/front() to access the Region/Block due to
+    // swichbox/shimmux being defined to only have one Region/Block
+    Block &b = op->getRegions().begin()->front();
     for (auto connectOp : b.getOps<ConnectOp>()) {
       if (connectOp.sourcePort() == sourcePort) {
+        // remove accessed connectOp from dangling island set
+        connections_set.erase(connectOp);
+        // add to graph if in detection mode and op is connectOp
+        if (g.isNotEmpty())
```
stephenneuendorffer, line 128: using an empty graph to track this state seems fragile?

test/find-flows/antenna_detection.mlir
```diff
@@ -0,0 +1,86 @@
...
+// CHECK: %[[T23:.*]] = AIE.tile(2, 3)
+// CHECK: %[[T22:.*]] = AIE.tile(2, 2)
+// CHECK: AIE.flow(%[[T23]], Core : 0, %[[T22]], Core : 1)
+// CHECK: AIE.flow(%[[T22]], Core : 0, %[[T22]], Core : 0)
+// CHECK: AIE.flow(%[[T22]], Core : 1, %[[T23]], Core : 1)
+// CHECK: AIE.packet_flow(0) {
+// CHECK:   AIE.packet_source<%[[T22]], DMA : 0>
+// CHECK:   AIE.packet_dest<%[[T23]], DMA : 1>
+// CHECK: }
+// CHECK: AIE.flow(%2, Core : 0, %9, DMA : 0)
```
stephenneuendorffer, line 21: Check for generated error messages?

test/find-flows/antenna_detection.mlir
```diff
@@ -4,13 +4,48 @@
...
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2021 Xilinx Inc.
+// (c) Copyright 2022 Xilinx Inc.
 //
 //===----------------------------------------------------------------------===//
 
-// RUN: aie-opt -aie-find-flows %s | FileCheck %s
-// CHECK: antenna_detection.mlir:54:5: warning: Dangling Island Antenna
+// RUN: aie-opt -aie-find-flows %s 2>&1 | FileCheck %s
```
iavssw, line 11: not sure if this is the best way to pipe error messages

iavssw, line 11: not sure if this is the best way to pipe error messages?

keryell:
What is the status of this PR?

stephenneuendorffer:
@andrej is this obsolete with your recent work?

PR #113, opened 2022-06-27
Remove aievec.acc and replace it with vector

brandon-t-nguyen:
The aievec.acc type does not offer much benefit in modeling the
integer accumulator while at the same time complicating the code as
ops need to handle conversions between aievec.acc and vector as well
as needing special cases to deal with aievec.acc and vector separately.

Issue #101

### DIFF
--- a/include/aie/Dialect/AIEVec/AIEVecUtils.h
+++ b/include/aie/Dialect/AIEVec/AIEVecUtils.h
@@ -61,40 +61,29 @@ inline int32_t getVectorSizeInBits(VectorType type) {
   return veclen;
 }
 
-// Return true if the incoming type is accumulator type
-inline bool isAccType(Type type) { return type.isa<AccType>(); }
-
 // Return true if this is an operation defined in AIE dialect
 inline bool isAIEOp(Operation *op) {
   return llvm::isa<AIEVecDialect>(op->getDialect());
 }
 
-// Create an AIE accumulator type. If the incoming type is vector, then the
-// #lanes are copied from it. Otherwise, a scalar accumulator type is created.
-inline AccType getAccType(Type type) {
-  Type stype = type;
-  unsigned lanes = 1;
-  // If the type was vector type, get the number of lanes and the underflying
-  // element type
-  if (VectorType vtype = type.dyn_cast<VectorType>()) {
-    stype = vtype.getElementType();
-    lanes = getVectorLaneSize(vtype);
-  }
+// Determine the output type for a vector operation based on whether
+// it operates on integer or floating point data.
+inline VectorType getVectorOpDestType(VectorType type) {
+    Type stype = type.getElementType();
 
-  // Now create the accumulator type. We currently only support i48 and i80
-  // type accumulator
-  AccType acc;
-  if (IntegerType itype = stype.dyn_cast<IntegerType>()) {
-    assert(itype.getWidth() <= 64);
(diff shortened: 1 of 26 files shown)


include/aie/Dialect/AIEVec/AIEVecUtils.h
```diff
@@ -61,40 +61,29 @@ inline int32_t getVectorSizeInBits(VectorType type) {
...
+      unsigned width = itype.getWidth() <= 16 ? 48 : 80;
+
+      Type ctype = mlir::IntegerType::get(itype.getContext(), width);
+      return VectorType::get(type.getShape(), ctype);
+    } else if (stype.isa<FloatType>())
+      // Floating point vector types are returned as is since the floating point
+      // operations write back to registers and not accumulators
+      return type;
+    else
+      llvm_unreachable("Unsupported destination type");
```
brandon-t-nguyen: In the AccType-less world I opted to restrict this counterpart to the `getAccType()` helper to `VectorType`s: in all the situations `getAccType()` was called, the callers are already dealing with vectors since the AIEVec ops, which have operand/result types constrained to vectors, are inserted for MLIR Vector dialect operations, which also have operand/result types constrained to vectors. I'm a bit unsure of why the original `getAccType()` function supported scalar types, so I dropped it in favor of simplifying replacement function.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1960,9 +1981,8 @@ static void insertSRSOp(Operation *Op, VectState *state) {
...
     return;
 
-  // The operation must be in AIE dialect, and its result should be accumulator
-  // type.
-  assert(isAIEOp(Op) && llvm::any_of(Op->getResultTypes(), isAccType));
+  // The operation must write to an accumulator
+  assert(writesToAccumulator(Op));
 
   // Check if any user of this operation is a non-AIE op. If any user of this
   // operation is non-AIE op, then we need to generate SRS op to move value
```
brandon-t-nguyen, line 1984: This is where it's determined that SRS is inserted only when we go from accumulator to non-AIEVec op

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -327,6 +327,28 @@ static bool isWellFormedVectorOp(Operation *Op) {
...
+// based on operation type and operand types
+static bool writesToAccumulator(Operation *op) {
+  // Integer muls and FMAs write to accumulator
+  if (!isAIEOp(op)) {
+    return false;
+  } else if (isa<aievec::MulOp>(op)) {
+    auto mulOp = dyn_cast<aievec::MulOp>(op);
+    auto lhsType = mulOp.lhs().getType().cast<VectorType>().getElementType();
+    auto rhsType = mulOp.rhs().getType().cast<VectorType>().getElementType();
+    return !lhsType.isa<FloatType>() && !rhsType.isa<FloatType>();
```
stephenneuendorffer, line 340: One think that might simplify the logic is to have a different op for floating point operands and for integer operands.  Is there a constraint that the left and right operands are both float or both int?

stephenneuendorffer (approved):

dotkrnl:
@brandon-t-nguyen 
`aievec.acc` is not removed from several Targets tests yet, which leads to `check-aie` errors.
https://github.com/Xilinx/mlir-aie/search?q=aievec.acc

PR #114, opened 2022-06-27
Emit error if a DMA block uses multiple locks/states

dotkrnl:
This PR performs a sanity check and emits an error if the user uses multiple locks in one single (Shim)DMA block, which is not supported in the XAIE library.  It also emits an error if the user attempts to acquire/release the lock from/to multiple states in a single block.

### DIFF
--- a/lib/AIEDialect.cpp
+++ b/lib/AIEDialect.cpp
@@ -503,10 +503,42 @@ template <typename... ParentOpTypes> struct HasSomeParent {
         return success();
       operation = operation->getParentOp();
     }
-    return op->emitOpError()
-           << "expects some parent op "
-           << (sizeof...(ParentOpTypes) != 1 ? "to be one of '" : "'")
-           << llvm::makeArrayRef({ParentOpTypes::getOperationName()...}) << "'";
+    return failure();
+  }
+};
+
+struct UsesOneLockInDMABlock {
+  static LogicalResult verifyTrait(Operation *op) {
+    auto block = op->getBlock();
+    int lockID = -1;
+    for (auto op : block->getOps<xilinx::AIE::UseLockOp>()) {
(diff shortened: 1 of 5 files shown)


jgmelber:
Do you have tests for the checks for multiple states in a single block?

dotkrnl:
@jgmelber Done.

jgmelber (approved):
Looks good.

lib/Targets/AIETargetXAIEV1.cpp
```diff
@@ -279,14 +279,26 @@ mlir::LogicalResult AIETranslateToXAIEV1(ModuleOp module, raw_ostream &output) {
...
       StringRef acqEnable = disable;
       StringRef relEnable = disable;
-      int lockID;
+      int lockID = -1;
       for (auto op : block.getOps<UseLockOp>()) {
         LockOp lock = dyn_cast<LockOp>(op.lock().getDefiningOp());
+        if (lockID != -1 && lockID != lock.getLockID()) {
+          op->emitError("Each DmaBd can use up to one lock.");
+          return failure();
+        }
```
stephenneuendorffer, line 288: Seems like these should be verifer conditions?

stephenneuendorffer (changes requested):
My vision for this was originally that blocks would not represent hard boundaries on BDs.  This would require something like an 'instruction selection' stage that determined legal BD boundaries. Given the current state, it's probably better to check this and come back and make it more featured later.  Thanks for this!

lib/Targets/AIETargetXAIEV1.cpp
dotkrnl: Done.

dotkrnl:
Formatting issue addressed.  Ready for review.

PR #115, opened 2022-06-27
Support initialization of locks in MLIR

dotkrnl:
This PR supports the initialization of locks when `AIE.useLock` is written in the top-level module.  This feature was previously partially implemented, but unable to run end-to-end, and has some minor issues.

An example usage could be seen at: https://github.com/Xilinx/mlir-aie/blob/7fe1988fa99d93a5ce337f1deed06865d93a2a5f/test/xaie_target/test_lock_init.mlir

### DIFF
--- a/lib/AIEDialect.cpp
+++ b/lib/AIEDialect.cpp
@@ -497,9 +497,9 @@ int xilinx::AIE::SwitchboxOp::rowIndex() { return getTileOp().rowIndex(); }
 
 template <typename... ParentOpTypes> struct HasSomeParent {
   static LogicalResult verifyTrait(Operation *op) {
-    Operation *operation = op;
+    Operation *operation = op->getParentOp();
     while (operation) {
-      if (llvm::isa<ParentOpTypes...>(operation->getParentOp()))
+      if (llvm::isa<ParentOpTypes...>(operation))
         return success();
       operation = operation->getParentOp();
     }
@@ -511,6 +511,10 @@ template <typename... ParentOpTypes> struct HasSomeParent {
 };
 
 LogicalResult xilinx::AIE::UseLockOp::verify() {
+  // AIE.useLock may be used in a module to set the lock's default value
+  if (llvm::isa<mlir::ModuleOp>((*this)->getParentOp()))
+    return success();
+
   return HasSomeParent<xilinx::AIE::CoreOp, xilinx::AIE::MemOp,
                        xilinx::AIE::ShimDMAOp>::verifyTrait(*this);
 
--- a/test/xaie_target/test_lock_init.mlir
+++ b/test/xaie_target/test_lock_init.mlir
@@ -1,0 +1,22 @@
+//===- test_lock_init.mlir -----------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2022 Xilinx Inc.
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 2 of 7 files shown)


jgmelber:
Can you add some comments to your tests about what they are doing?

dotkrnl:
@jgmelber Done.

lib/AIEDialect.cpp
```diff
@@ -511,6 +511,10 @@ template <typename... ParentOpTypes> struct HasSomeParent {
 };
 
 LogicalResult xilinx::AIE::UseLockOp::verify() {
+  // AIE.useLock may be used in a module to set the lock's default value
+  if (llvm::isa<mlir::ModuleOp>((*this)->getParentOp()))
+    return success();
+
   return HasSomeParent<xilinx::AIE::CoreOp, xilinx::AIE::MemOp,
                        xilinx::AIE::ShimDMAOp>::verifyTrait(*this);
```
stephenneuendorffer, line 519: I think all of should be implemented by tablegen using ParentOneOf:  see https://mlir.llvm.org/docs/Traits/#hasparent

test/xaie_target/test_lock_init.mlir
```diff
@@ -0,0 +1,22 @@
...
+// CHECK: XAieTile_LockAcquire(&(ctx->TileInst[3][3]), 0, 1, 1);
+
+module @test_lock_init {
+  %t33 = AIE.tile(3, 3)
+  %l33_0 = AIE.lock(%t33, 0)
+  // When written in the top-level module, the useLock is treated as
+  // an initialization to the lock.  For example, in the operation below,
+  // the lock %l33_0 is to be initialized as Acquired 1 when the host function
+  // mlir_aie_initialize_locks(ctx) is invoked.
+  AIE.useLock(%l33_0, Acquire, 1)
```
stephenneuendorffer, line 21: Is there a unit_test that tests the generated functionality?

lib/AIEDialect.cpp
dotkrnl: This cannot be done in tablegen as the verification here requires two different kinds of parent verification.  One is the custom `HasSomeParent` which verifies parents recursively, and the other is the `isa` test for checking the *immediate* parent.

test/xaie_target/test_lock_init.mlir
dotkrnl: This is done.

lib/AIEDialect.cpp
stephenneuendorffer: The immediate parent check could be done in tablegen, leaving the verify method to just do things that no good tablegen support exists for.   (And HasSomeParent might be a good thing to refactor and upstream tablegen support for anyway...)

stephenneuendorffer:
> Code format of `lib/AIECoreToStandard.cpp` was not changed and only one line of code is added, which adheres to the format requirements. Please override the `Build and Test / Check code format (pull_request)` test.

The big problem is your test file, which needs to be formatted.  If you think it's necessary, in lib/AIECoreToStandard.cpp you can add code comments like:
```
// clang-format off
...
// clang-format on
```
to disable the formatting.

See https://github.com/Xilinx/mlir-aie/runs/7097267788?check_suite_focus=true
under "clang format and tidy patches display" for details.

lib/AIEDialect.cpp
dotkrnl: Sorry, I’m quite new to mlir.  Please bear with me if this question sounds naive.  Does writing a trait in tablegen express an “or” semantics to the verify function?  It sounds “and” to me.  If not, how do I write it to implement: immediate parent “or” some parent.

stephenneuendorffer: Ah, I see your concern now :)  yeah, maybe easier to just leave it as is...

stephenneuendorffer: The closest you could get would be Or<HasSomeParent<...>, ParentOneOf<....>>

PR #116, opened 2022-06-27
Use system version of opt and llvm-link

dotkrnl:
This PR prefers the system (or, `PATH`) version of `opt` and `llvm-link`, when installed or built, to the Vitis version.  This PR also fixes issues raised by Jinming and Andra.  When the Vitis environment is set up, the Vitis version of `llvm-link` will be used, which leads to errors.


stephenneuendorffer (approved):
Would the more correct thing be to avoid modifying the PATH entirely?

dotkrnl:
> Would the more correct thing be to avoid modifying the PATH entirely?

Definitely.  I am not very familiar with the `PATH` requirements of `aietools` yet so this might be a quick fix for now.  Maybe an issue can be created for removing `PATH` changes.

PR #120, opened 2022-06-28
Replace tests aievec.acc with vector

dotkrnl:
#113 breaks `check-aie` since `aievec.acc` is still used in tests.  This PR replaces the usages with `vector`.


brandon-t-nguyen:
Oh cool, you updated them! The fixed tests just unveiled a bug from #113, working on a fix for them now.

PR #123, opened 2022-06-29
Bump llvm to bc07634

fifield:


dotkrnl:
This version bump broke xchesscc tests:  `@__chess_separator_dummy = external dso_local global ptr, align 4` where `ptr` is not a type.  Originally, `i8*` was generated in place of `ptr`.  I am taking a look into the reason.

PR #85, opened 2022-02-02
[AIETarget] Support packet stream for ShimDMAs

hanchenye:
This patch sets the packet bit when DMABDPacketOp is found in the ShimDMAOp region.

This patch also adds a 32x32x32 GEMM unit test. Unfortunately, currently this test case doesn't work on board -- the AIE stalls waiting for the shim DMA to send out the packet.

The layout of the design:
image

PR #125, opened 2022-07-06
Add packet headers to shim DMA BDs

jgmelber:
This extends the packet header insertion to Shim DMA BDs previously implemented for tile DMA BDs


jgmelber:
@hanchenye You are correct. I will take a look at #85, there is also some work to be done for packet_flows to/from ShimDMAs

PR #126, opened 2022-07-06
Reimplement AIE.token logical locks

dotkrnl:
This PR introduces a major reimplementation of the `AIE.token`.  As described in `AIE.td`:  `AIE.token` creates a logical lock with a symbol so that it can be referenced globally.  Unlike physical locks, logical locks are unlimited, and we can specify any integer value associated with a lock. The logical lock is used to manually specify the dependence of tasks, or core executions.

The following design principles are used in this re-implementation:

* For `useToken` in a DMA block, the acquired token must be mapped to `0`/`1`, and the released token must be its opposite within the same physical lock.  In the original implementation, this is not adhered to.
*  For other tokens, the mapped value can be any `0`/`1` in any physical lock – instead of in the original implementation, only `1` is used for implementing a physical chain.
*  Token release is implemented as releasing the corresponding physical lock to the values mapped state.
*  Token acquired is implemented without distinguishing between tokens.  When all locks are acquired, the token is acquired.

The major changes from the original implementation follow:

1. One token value is now implemented as a state in one physical lock, instead of the chaining approach in the original design.  (Example: `test/create-locks/test_lock0.mlir`, `token0` value `0` is mapped to lock `%1` state `0`).
2. All states in MLIR are implemented, compared to the previous method that requires modeling of the host code.  (Example: `test/create-locks/test_lock_shimdma.mlir`, a dummy core is removed).
3. DMA block's one physical lock rule is observed and enforced (Example: `test/create-locks/test_lock_shimdma.mlir`, only `%2` is used in `shimDMA`, even when the token is released back to value `0` at line 69).
4. For the same behavior, uses fewer physical locks than the previous approach.  (Example: `test/create-locks/test_lock0_reuse.mlir`).
5. Avoid using existing user-defined locks.   (Example: `test/create-locks/test_lock0_avoid_existing_locks.mlir`)
6. Avoid generating false physical release-acquire chain.   (Example: `test/create-locks/test_locks_5.mlir`, previously: https://github.com/Xilinx/mlir-aie/blob/main/test/create-locks/test_lock5.mlir#L30 and https://github.com/Xilinx/mlir-aie/blob/main/test/create-locks/test_lock5.mlir#L50=)

Behavior changes:

*  A token value cannot be acquired by multiple locations.  Previously, this was "implemented" as ignoring lock chains to SS2M DMAs.  However, sematically, SS2M does not depend on the token at all.  This might affect @hanchenye's current project.


dotkrnl:
Requesting a review from @stephenneuendorffer, the author of the original token.   And requesting a review from @hanchenye as the main current user of `AIE.token`.

dotkrnl:
Token initialization added.

dotkrnl:
Requested by @stephenneuendorffer, here is a summary of what this model redefining is.

#### What it intends to do

`AIE.token` intends to allow users to create logical locks that:

1. does not need a specification of its physical location and physical lock ID.
2. allows multiple states instead of 0/1 in the physical locks.
3. has the potential of implementing multiple states of one value into different physical locations.

#### How it works today

It uses release/acquire pair chaining to implement physical locks.

It does not work properly.  It attempts to allow acquiring of a token value in multiple places and hopes that those could be triggered when the token is released to a specific value.  To implement that, the token usage chain is analyzed and for each release/acquire pair, a release of a physical lock and an acquisition of that physical lock is generated.  However, in the examples, it attempts to use that for non-neighbor tiles and workaround by ignoring lock chains of DMA-cores.

The previous model makes it difficult to implement and prone to errors, as I discussed in the PR body.

#### How it works in this PR

It mapped each logical token value to a 0/1 state in a physical lock.  Each token value can be acquired only once -- acquiring the corresponding physical state.  Releasing a token value is to release the corresponding physical lock state.  The physical locks are placed according to their users.  This model is simpler than previously.

PR #127, opened 2022-07-06
Workaround xchesscc by disabling opaque pointers

dotkrnl:
#123 has several `xchesscc` tests failing with the error message:

```
acdc_project/core_7_3.chesslinked.ll:11:54: error: expected type
@__chess_separator_dummy = external dso_local global ptr, align 4
                                                     ^
```

This is because `xchesscc` does not support [Opaque Pointers](https://llvm.org/docs/OpaquePointers.html) introduced in LLVM 15.  A workaround is proposed in this PR to disable opaque-pointers in `mlir-translate` temporarily.


stephenneuendorffer (approved):
Thanks Jason!

PR #134, opened 2022-07-13
Disable xchesscc and host compilation in CI and enable unit tests

dotkrnl:
#121 disables all unit tests in GitHub CI due to the usage of `xchesscc`.  However, the compiling and linking with `xchesscc` can be disabled individually.  In this PR, the `xchesscc` usage is disabled while the unit tests are enabled.


.github/workflows/buildAndTest.yml
```diff
@@ -65,6 +65,8 @@ jobs:
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=Debug \
+            -DAIE_COMPILER=NONE \
+            -DAIE_LINKER=NONE \
```
stephenneuendorffer, line 69: So, if we don't compile and link with xchesscc, then we need a different compiler.  What are you expecting to happen here?

test/unit_tests/00_itsalive/aie.mlir
```diff
@@ -8,7 +8,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-// REQUIRES: valid_xchess_license
```
stephenneuendorffer, line 11: The requirement here depends on the default compiler set for aiecc.py.   I don't think we can/should just remove this.

stephenneuendorffer (changes requested):
Can you clarify the problem you're trying to solve here?

test/unit_tests/00_itsalive/aie.mlir
dotkrnl: @jackl-xilinx requested this change.  I will leave to him to elaborate.

.github/workflows/buildAndTest.yml
dotkrnl: Still working in progress.  I forgot to convert it to a draft.  I was expecting it to generate llvm code and stop there.  @jackl-xilinx is this your expectation as well?

stephenneuendorffer: hmm.. good idea, but that's not what aiecc.py does today, I think...  Want to look into aiecc.py to add that?

dotkrnl: Will do this tomorrow.

test/unit_tests/00_itsalive/aie.mlir
stephenneuendorffer: Thinking about this further, maybe it is reasonable, since I've been arguing that we do this check at configuration time, anyway.

jackl-xilinx: @stephenneuendorffer This check doesn't work when we run unit tests on the board with a cross compiled build since it will have peano but not xchesscc. The main issue to resolve is what the publicly built llvm (not peano) does in conjunction with aiecc.py since aiecc.py seems to require opt/llc/clang with some arch=aie options which I had previously assumed the non-peano llvm provides. Regardless, we need a different way to "disable" these tests for CI if we can't run it properly without both peano and xchesscc.

.github/workflows/buildAndTest.yml
jackl-xilinx: This was added to prevent aiecc.py from trying to run with the default linker (--xbridge) which calls xchess_wrapper which then needs xchesscc which is not present for CI. I think the check for xchesscc should affect the xbridge option as well and should guide whether --xchesscc and --xbridge is on or not.

PR #139, opened 2022-07-25
Testupdate (rebase)

stephenneuendorffer:


stephenneuendorffer:
@jackl-xilinx  I'm curious what you think about this, The goal is to check for the license file, but not spend an expensive execution to test-run xchesscc...

PR #141, opened 2022-07-27
Add broadcast_packet support

JinmingZhuang:
Including:
1.  A new data movement abstraction that combines broadcast together with packet switch.
2. A pass that lowers broad_packet operation to packet_flow.
3. An IR example and its corresponding output.


jgmelber:
Can you add some documentation and examples in the [docs](https://github.com/Xilinx/mlir-aie/blob/main/docs/AIEDesignPatterns.md)?

JinmingZhuang:
I add a unit test, which successfully passes the on-broad test, as well as descriptions and an example in the docs.

PR #144, opened 2022-08-01
Multicast

JinmingZhuang:
1) Add "multicast" operation which represents connections from one source to multiple destinations. 
2) Add the "create-multicast" pass that lowers "multicast" operation to "flow" operation.


AndraBisca:
It seems that this PR contains all of the changes from PR #141 as well. It might be good to separate them.

PR #145, opened 2022-08-01
Add multicast operation

JinmingZhuang:
1) Add "multicast" operation which represents connections from one source to multiple destinations.
2) Add the "lower-multicast" pass that lowers "multicast" operation to "flow" operation.

### DIFF
--- a/include/aie/AIE.td
+++ b/include/aie/AIE.td
@@ -660,6 +660,63 @@ def AIE_PacketRuleOp: AIE_Op<"rule", [HasParent<"PacketRulesOp">]> {
   // ];
 }
 
+def AIE_MulticastOp: AIE_Op<"multicast", [SingleBlockImplicitTerminator<"EndOp">]> {
+  let arguments = (
+    ins Index:$tile,
+        WireBundle:$bundle,
+        I32Attr:$channel
+  );
+  let regions = (region AnyRegion:$ports);
+  let summary = "An abstraction of multicast";
+  let description = [{
+    An abstraction of broadcast. During place and
+    route, it will be replaced by multiple flows.
+
+    Example:
+    ```
+      %70 = AIE.tile(7, 0)
+      %73 = AIE.tile(7, 3)
+      %74 = AIE.tile(7, 4)
+      %63 = AIE.tile(6, 3)
+      %64 = AIE.tile(6, 4)
+      AIE.multicast(%70, "DMA" : 0){
+        AIE.multi_dest<%73, "DMA" : 0>
+        AIE.multi_dest<%74, "DMA" : 0>
+        AIE.multi_dest<%63, "DMA" : 0>
+        AIE.multi_dest<%64, "DMA" : 0>
+      }
+    ```
+  }];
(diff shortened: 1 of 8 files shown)


include/aie/AIE.td
```diff
@@ -660,6 +660,63 @@ def AIE_PacketRuleOp: AIE_Op<"rule", [HasParent<"PacketRulesOp">]> {
...
 }
 
+def AIE_MulticastOp: AIE_Op<"multicast", [SingleBlockImplicitTerminator<"EndOp">]> {
+  let arguments = (
+    ins Index:$tile,
+        WireBundle:$bundle,
+        I32Attr:$channel
+  );
+  let regions = (region AnyRegion:$ports);
+  let summary = "An abstraction of multicast";
```
AndraBisca, line 670: I think this should be a bit more descriptive, something along the lines of "An operation used to abstract multiple AIE.flow operations that are created during broadcast."

include/aie/AIE.td
```diff
@@ -660,6 +660,63 @@ def AIE_PacketRuleOp: AIE_Op<"rule", [HasParent<"PacketRulesOp">]> {
...
+def AIE_MultiDestOp: AIE_Op<"multi_dest", [HasParent<"MulticastOp">]> {
+  let arguments = (
+    ins Index:$tile,
+        WireBundle:$bundle,
+        I32Attr:$channel
+  );
+  let summary = "A destination port of multicast flow";
+  let description = [{
+    An object representing the destination of a multicast flow. This must exist
+    within an [AIE.multicast] operation. There can multiple destinations with in the Op.
```
AndraBisca, line 707: Quick typo correction: "There can be multiple destinations within an [AIE.multicast] operation."

include/aie/AIEPasses.td
```diff
@@ -175,6 +175,19 @@ def AIERoutePathfinderFlows : Pass<"aie-create-pathfinder-flows", "ModuleOp"> {
   ];
 }
 
+def AIEMulticast : Pass<"aie-create-multicast", "ModuleOp"> {
+  let summary = "Replace multicast flow by flow";
+  let description = [{
+    An abstraction of multicast flow
```
AndraBisca, line 181: This could be a bit more descriptive, maybe something like "This pass replaces AIE.multicast with the equivalent number of AIE.flow operations."

lib/AIEMulticast.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "aie/AIENetlistAnalysis.h"
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/Location.h"
+#include "mlir/IR/PatternMatch.h"
+#include "mlir/Pass/Pass.h"
+#include "mlir/Tools/mlir-translate/MlirTranslateMain.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "llvm/ADT/Twine.h"
+
+#define DEBUG_TYPE "aie-create-multicast"
```
AndraBisca, line 11: I think "aie-lower-multicast" makes a bit more sense as the pass lowers aie.multicast into aie.flow operations.

PR #158, opened 2022-08-29
[VITIS-5359] Add legality checks for unaligned loads in AIEVectorize

linay-xsj:
Check all the transfer_read memrefs:
1) If the lowest dim has iv, check whether its corresponding loop step is divisible by the vector lanes.
2) For the higher dimension, check whether the lower dimensions' shape sizes is divisible by the vector lanes.
3) Skip checking the higher dimensions with dynamic size.
4) Avoid generating the code with wrong results due to unaligned upper bound's affine_map offset and vector lanes.


test/aievec/conv2d_f32_unaligned_loads.mlir
```diff
@@ -0,0 +1,100 @@
...
+                     %c11 = arith.addf %ci, %p11 : f32
+                     //Store accumulated sum
+                     affine.store %c11, %C[%arg3, %arg4] : memref<2046x2046xf32>
+                }
+            }
+        }
+    }
+    return
+}
+//CHECK: %0 = vector.transfer_read %arg2[%arg3, %arg4], %cst {in_bounds = [true]} : memref<2046x2046xf32>, vector<8xf32>'s dim 1's shape size cannot be divisible by the vector lanes.
```
stephenneuendorffer, line 27: usually we put the check lines first, consistent with llvm/utils/update_mlir_test_checks.py

test/aievec/conv2d_f32_unaligned_loads.mlir
```diff
@@ -0,0 +1,100 @@
+// This test case will directly return the result generated from -affine-super-vectorize
+// because in transfer_read %arg2[%arg3, %arg4], dim 1's memref shape size(2046) cannot be
+// divisible by the vector lanes(8).
+
+// RUN: aie-opt %s --affine-loop-unroll="unroll-full unroll-full-threshold=3" --canonicalize -affine-super-vectorize="virtual-vector-size=8" --aie-vectorize -split-input-file -unaligned-loads-check=true --debug-only=aie-vect 2>&1 | FileCheck %s
```
stephenneuendorffer, line 5: This is a pretty complex run line... It might be better to have a focused unit test that just looks at one aspect... i.e. start with a minimal loop nest and a single load/store and show that it's not handled?

stephenneuendorffer: Also, if this is a negative test, it should probably check for the message as output?

linay-xsj: > This is a pretty complex run line... It might be better to have a focused unit test that just looks at one aspect... i.e. start with a minimal loop nest and a single load/store and show that it's not handled?

I reuse the test conv2d_f32.mlir to show this is an unaligned lit test when we turn -unaligned-loads-check to true. I can write a small test case by your request.

linay-xsj: > Also, if this is a negative test, it should probably check for the message as output?

The debug message check is at line 27.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -35,6 +35,10 @@ using namespace xilinx;
 using namespace xilinx::aievec;
 
 #define DEBUG_TYPE "aie-vect"
+static llvm::cl::opt<bool>
+    unalignedLoadsCheck("unaligned-loads-check",
+                        llvm::cl::desc("Enable the unaligned loads check"),
+                        llvm::cl::init(false));
```
david-vc, line 41: I recommend to flip the default value so that the check will be on by default.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2282,6 +2372,34 @@ static void computeReuseInFunc(func::FuncOp func, VectState *state) {
...
+// transfer_read operations capture the vector load. Since AIE only allows for
+// aligned vector loads, we need to compose multiple transfer reads together to
+// form intervals of certain width (128, 256, 512, or 1024), and create an AIE
+// vector from each interval.
+static void computeReuseInFunc(func::FuncOp func, VectState *state) {
+  if (!unalignedLoadsCheck) {
+    for (AffineForOp forOp : func.getOps<AffineForOp>()) {
+      SmallVector<Operation *, 8> enclosingLoops;
+      enclosingLoops.push_back(forOp);
+      computeEnclosingLoopsPerBlock(forOp, state, enclosingLoops);
```
david-vc, line 2399: Function computeEnclosingLoopsPerBlock() modifies "state".
The original code has only 1 calls site of this function.
The new code has 2 call sites. In case that unalignment-check is enabled computeEnclosingLoopsPerBlock is called "early" during isUnalignedLoad(). Otherwise it is still called in computeReuseInFunc.
This is unfortunate, because it makes the new code harder to follow, and moreover the isUnalignedLoad() function now has the side effect of changing "state".

test/aievec/conv2d_f32_unaligned_loads.mlir
david-vc: Test conv2d_f32_unaligned_load duplicates much of conv2d_f32. The conv2d_f32_unaligned_load should not check the MLIR that is produced but instead just check that the error message was emitted.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,12 +2272,98 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
...
+      AffineForOp affineForOp = cast<AffineForOp>(loop);
+      auto iv = affineForOp.getInductionVar();
+      auto invariants = getInvariantAccesses(iv, indices);
+
+      if (!invariants.count(index)) {
+        step = affineForOp.getStep();
+        if (step % lanes) {
+          LLVM_DEBUG(llvm::dbgs() << "\n\n"
+                                  << *readOp
+                                  << "'s lowest dim's loop step cannot be "
```
david-vc, line 2315: "cannot be divisible" => "is not divisible". There are several more places where this is needed.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
linay-xsj: After computeEnclosingLoopsPerBlock(), the only change is state->blockToEnclosingLoops, which was an empty map before computeEnclosingLoopsPerBlock(). I don't think this is a negative side effect. The key thing is we can have an early exit when it hits an unalignment load and we don't need to do any more analysis and transformation. If you think the code between line 2396-2400 is hard to follow, I can just use it once and put it before line 2465.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
linay-xsj: Will revise these errors.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,12 +2272,98 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
...
+    // index.
+    for (auto loop : enclosingLoops) {
+      AffineForOp affineForOp = cast<AffineForOp>(loop);
+      auto iv = affineForOp.getInductionVar();
+      auto invariants = getInvariantAccesses(iv, indices);
+
+      if (!invariants.count(index)) {
+        step = affineForOp.getStep();
+        if (step % lanes) {
+          LLVM_DEBUG(llvm::dbgs() << "\n\n"
```
david-vc, line 2313: Is this message printed only if the code was compiled for debug? If so, then the user will have no way to figure out what went wrong in case the the alignment check first.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
david-vc: Yes. Moving 2396-2399 to before line 2465 would address my concern.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,20 +2272,117 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
   }
 }
 
+static LogicalResult isUnalignedLoad(TransferReadOp readOp, VectState *state) {
```
david-vc, line 2275: Rename function to 'checkLoadAlignment'.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,20 +2272,117 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
...
+      auto invariants = getInvariantAccesses(iv, indices);
+
+      if (!invariants.count(index)) {
+        step = affineForOp.getStep();
+        if (step % lanes) {
+          LLVM_DEBUG(llvm::dbgs() << "\n\n"
+                                  << *readOp
+                                  << "'s lowest dim's loop step is not "
+                                     "divisible by the vector lanes.\n\n");
+          return failure();
```
david-vc, line 2317: Change lines 2313-2317 to: 
```suggestion
return readOp->emitError() 
             << "Loop step of inner index of " 
            << readOp->getName() 
            << " is not divisible by number of vector lanes.";
```

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,20 +2272,117 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
...
+        // the vector lanes.
+        AffineBound ub = affineForOp.getUpperBound();
+        AffineMap origUbMap = ub.getMap();
+        if (!origUbMap.isEmpty() && !origUbMap.isConstant()) {
+          AffineExpr origUbMapResult = origUbMap.getResult(0);
+          AffineExpr base;
+          int32_t offset;
+          std::tie(base, offset) = getBaseAndOffset(origUbMapResult);
+          if (offset % lanes) {
+            LLVM_DEBUG(llvm::dbgs()
```
david-vc, line 2332: See my suggestion for 2313-2317. Do same here for 2332-2338.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2268,20 +2272,117 @@ static void computeReuse(TransferReadOp readOp, VectState *state) {
...
+  // For the higher dimension, check whether the lower dimensions' shape sizes
+  // is divisible by the vector lanes.
+  for (int i = 1; i < numDims; ++i) {
+    // Skip checking the higher dimensions with dynamic size.
+    if (sizes[i] == -1) {
+      continue;
+    }
+
+    if (sizes[i] % lanes) {
+      LLVM_DEBUG(llvm::dbgs()
```
david-vc, line 2354: See my suggestion for 2313-2317. Do same here for 2354-2358.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2344,6 +2445,20 @@ void AIEVectorize::runOnOperation() {
...
+    // necessary to identify loads/stores that are nested together.
+    for (AffineForOp forOp : func.getOps<AffineForOp>()) {
+      SmallVector<Operation *, 8> enclosingLoops;
+      enclosingLoops.push_back(forOp);
+      computeEnclosingLoopsPerBlock(forOp, state, enclosingLoops);
+    }
+
+    // Check whether there is any unalignment loads.
+    if (unalignedLoadsCheck && failed(hasUnalignedLoads(func, state))) {
+      func.emitError("Unaligned loads\n");
```
david-vc, line 2458: See my suggestion for 2313-2317. 
Perhaps change 2458 to func.emitError() << "Cannot apply aie-vectorize to " 
<< func->getName() << " because alignment check has failed.\n";

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
linay-xsj: I don't think this is a good function name, because it will return a logical result to show whether the readOp is an unaligned load. Furthermore, 'checkLoadAlignment' is too similar as the commandline option llvm::cl::opt<bool> unalignedLoadsCheck at line 38

test/aievec/conv2d_uij_f32.mlir
```diff
@@ -106,3 +106,12 @@ func.func @conv2d (%A: memref<2048x2048xf32>, %B: memref<9xf32>, %C: memref<2046
...
 //CHECK-NEXT: %20 = aievec.mac %18, %1, %19 {xoffsets = "0x76543210", xstart = "2", zoffsets = "0x00000000", zstart = "0"} : vector<16xf32>, vector<8xf32>, vector<8xf32>
 //CHECK-NEXT: vector.transfer_write %20, %arg2[%arg3, %arg4] {in_bounds = [true]} : vector<8xf32>, memref<2046x2046xf32>
+
+// This test case will directly return the result generated from -affine-super-vectorize when
+// -unaligned-loads-check=true. The reason is that in transfer_read %arg2[%arg3, %arg4],
+// dim 1's memref shape size(2046) is not divisible by the vector lanes(8).
+
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=8" --aie-vectorize="shift=0" -split-input-file 2>&1 | FileCheck %s -check-prefix=ALIGNMENT
+
+// ALIGNMENT: vector.transfer_read's shape size of index 1 is not divisible by number of vector lanes.
```
david-vc, line 116: Should this be CHECK-LABEL instead of ALIGNMENT?

linay-xsj: > Should this be CHECK-LABEL instead of ALIGNMENT?

No, ALIGNMENT is for the second RUN command only. CHECK-LABEL is for the first RUN command.

PR #160, opened 2022-08-30
Allow useLock operations in functions

stephenneuendorffer:
With the --localize-locks pass converting locks into integer indexes,
we no longer need to be restricted to useLock operations existing
primarily in CoreOps.  Instead, we can simply pass around the correct
integer indexes.


lau-amd (approved):
This is great.  Thank you, Steve!  I wonder if you want to patch the `clang-format` diff before merging so I will leave it unmerged for now, but the changes looks great.

PR #168, opened 2022-09-22
[AIE][AIEVec] Structure aie-vectorize as a pipeline

jsetoain:
Right now, the AIEVectorize pass is monolithic and makes poor use of MLIR infrastructure. This is the first of a series of patches aimed at modernizing the pass and making it easier to tweak and extend.

I've changed the name from aie-vectorize to aie-affine-vectorize in anticipation of the next patch, which will embed the affine supervectorizer into the pipeline.

In the future, AIEVectorize should also be renamed to something like LowerVectorToAIEVector, which matches better the semantics of the pass.

Also in the future, the pass should be split into a series of conversion and analysis passes that compose better, so we can build more vectorization options reusing the existing code.

### DIFF
--- a/test/aievec/gemm64_tile448_unroll8_unaligned_loads.mlir
+++ b/test/aievec/gemm64_tile448_unroll8_unaligned_loads.mlir
@@ -3,7 +3,7 @@
 // (affine.for %arg7 = #map0(%arg4) to #map1(%arg4))'s upper bound’s affine_map(<(d0) -> (d0 + 4)>)
 // result's offset(4) is not divisible by the vector lane size(8).
 
-// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=8" --aie-vectorize 2>&1 | FileCheck %s
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=8" --aie-affine-vectorize 2>&1 | FileCheck %s
 
 // CHECK-LABEL: Loop upper bound's affine map offset of inner index of vector.transfer_read is not divisible by number of vector lanes.
 // CHECK-LABEL: Cannot apply aie-vectorize to func.func because alignment check has failed.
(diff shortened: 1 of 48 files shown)


test/aievec/gemm64_tile448_unroll8_unaligned_loads.mlir
```diff
@@ -75,77 +75,83 @@ module {
...
-//CHECK-LABEL:#map0 = affine_map<(d0) -> (d0)>
-//CHECK-NEXT: #map1 = affine_map<(d0) -> (d0 + 4)>
-//CHECK-NEXT: #map2 = affine_map<(d0, d1) -> (0)>
-//CHECK-NEXT: #map3 = affine_map<(d0) -> (d0 + 1)>
-//CHECK-NEXT: #map4 = affine_map<(d0) -> (d0 + 2)>
-//CHECK-NEXT: #map5 = affine_map<(d0) -> (d0 + 3)>
-//CHECK-NEXT: #map6 = affine_map<(d0) -> (d0 + 5)>
-//CHECK-NEXT: #map7 = affine_map<(d0) -> (d0 + 6)>
-//CHECK-NEXT: #map8 = affine_map<(d0) -> (d0 + 7)>
+//CHECK-LABEL:#map = affine_map<(d0, d1) -> (0)>
```
linay-xsj, line 78: I did not see any functional change in this change set, but why are gemm64_tile448_unroll8_unaligned_loads.mlir and gemm64_unroll4_unaligned_loads.mlir changed so much after --aie-affine-vectorize? They should return the results after -affine-super-vectorize. Why do they have scf.for loops instead of affine.for loops?

jsetoain: I believe it's because the AIEVectorize pass is not "failing properly", and so it doesn't interrupt the pipeline. One of the next stages lowers affine, hence the conversion. When I'm done with the refurbish, we will have proper failure modes. That will also mean that this test won't output any code, just the error and its location.

david-vc (approved):
Nice work.

test/aievec/gemm64_tile448_unroll8_unaligned_loads.mlir
linay-xsj: > I believe it's because the AIEVectorize pass is not "failing properly", and so it doesn't interrupt the pipeline. One of the next stages lowers affine, hence the conversion. When I'm done with the refurbish, we will have proper failure modes. That will also mean that this test won't output any code, just the error and its location.
I found the reason now. After alignment check, original code will directly return the module, but the new code, you added postCanonicalizeIR step to it at line AIEVectorize.cpp:2497-2500. Do you think this change is necessary?

jsetoain: The change is necessary if we want to organise the pipeline using MLIR infrastructure for pipelines, as opposed to having a pass that has embedded passes, which is what we have now.

Also, the current behaviour is less than ideal. We are giving an error and producing partially lowered code. Given the current state of the pass, I understand that that's better than nothing, but the appropriate thing to do is interrupting the compilation process when there's a failure of this nature.

When the refurbishment is over, your alignment check will become a legality check, and a misaligned vector load with stop the compilation and simply generate an error, a failure to legalise vector.transfer_read/write. That is, in my opinion, the proper way to handle this. I've changed the file just to make sure the test passes for now, but in the near future all the lines I've changed will be unnecessary, so I thought it was an acceptable compromise 🙂

linay-xsj (approved):

jsetoain:
I've updated this patch to align with [PR #169](https://github.com/Xilinx/mlir-aie/pull/169)

PR #169, opened 2022-09-29
[AIE] Align source code with llvm bebc9695

jsetoain:
This patch aligns the mlir-aie code with the llvm version currently used by torch-mlir (bebc96956b76bdbc36f1d82a788c810e5b12e2c5).

The main issue in the update is a change in the name of the autogenerated (tablegen) "getter" methods for dialect ops. Further cleaning up of naming might be desirable.

### DIFF
--- a/include/aie/AIE.td
+++ b/include/aie/AIE.td
@@ -114,22 +114,22 @@ def AIE_TileOp: AIE_Op<"tile", [FlowEndPoint]>, Results<(outs Index:$result)> {
   let extraClassDeclaration = [{
     int getNumSourceConnections(WireBundle bundle);
     int getNumDestConnections(WireBundle bundle);
-    int colIndex() { return col(); }
-    int rowIndex() { return row(); }
-    bool isShimTile() { return row() == 0; }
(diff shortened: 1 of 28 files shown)


jsetoain:
This is a work in progress. All chess-based tests are failing for me, but they've been failing all day, so I suspect it's an issue with my environment and not necessarily an issue with the patch. If somebody with a working environment could test it, that should be enough. Otherwise, while I solve that, this is a good reference for what's coming. There was an "easier" alternative to this change: adding the old methods as extra class declaration. I decided against that and in favour of repaying the technical debt now, instead of just kicking the can.

fifield:
You probably need to update the commit hash in utils/clone-llvm.sh for the build to succeed.

jsetoain:
> You probably need to update the commit hash in utils/clone-llvm.sh for the build to succeed.

Yes! I realised that right after I PR'd :-)

fifield (approved):
Looks good to me. I get the same 3 test failures I always see.

jsetoain:
Rebase to main

PR #172, opened 2022-10-07
Merge supervectorizer into vectorization pass

jsetoain:
AIEVectorize assumes it's taking the output of the supervectorizer, might as well make it part of the pipeline


jsetoain:
This PR and the ones hanging from it (#173 & #174) depend on [D134632](https://reviews.llvm.org/D134632), which has landed in main but it's not in our version of llvm yet. I leave these PR's here waiting for the next llvm bump.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2410,8 +2397,9 @@ static void reassociateOpsInFunc(func::FuncOp func, VectState *state) {
   reassociateAddOpInFunc(func, state);
 }
 
-struct AIEVectorize : public AIEVectorizeBase<AIEVectorize> {
-  AIEVectorize() = default;
+struct AIEVectorize : public aievec::impl::AIEVectorizeBase<AIEVectorize> {
+  using Base::Base;
```
linay-xsj, line 2401: Where do we use this?

jsetoain: Line 2493:
`
pm.addPass(createAIEVectorize(options.getAIEVectorizeOptions()));
`
createAIEVectorize (autogenerated) eventually calls the inherited constructor in AIEVectorize.

PR #173, opened 2022-10-07
[AIEVec] Split load/store optimization as a pass

jsetoain:
This stage might not always be needed, and we might want to use it in different vectorization strategies. It works better on its own.


jsetoain:
This PR, #172 and #174 depend on [D134632](https://reviews.llvm.org/D134632), which has landed in main but it's not in our version of llvm yet. I leave these PR's here waiting for the next llvm bump.

PR #174, opened 2022-10-07
[AIEVec] Make AIEVectorize a function pass

jsetoain:
There is no reason for this pass to be a module pass, and for bigger code bases a function pass can run in parallel, significantly speeding-up the compilation process.


jsetoain:
This PR, #173 and #174 depend on [D134632](https://reviews.llvm.org/D134632), which has landed in main but it's not in our version of llvm yet. I leave these PR's here waiting for the next llvm bump.

PR #177, opened 2022-10-07
[VITIS-5359]Extend the vectorization capability to support for AIE-ML

linay-xsj:
The current PR version can support int32 and int16 types for AIE-ML.


lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -697,28 +728,42 @@ static Operation *generateFMAOp(vector::FMAOp fmaOp, AIEOpAttributes &opAttr,
...
   if (!isSimpleVectIntrinsic(fmaOp, state)) {
     AIEVecAttributes lstat = getOperandVecStats(fmaOp, state, 0);
     assert(lstat.vecSizeInBits % 256 == 0);
-    if (lstat.vecSizeInBits == 256) {
-      VectorType concatType =
-          createVectorType(512 / lstat.elementSizeInBits, lstat.elementType);
-      SmallVector<Value> sources = {lhs, lhs};
-      lhs = generateConcatOp(sources, state, fmaOp->getLoc(), concatType);
+    if (AIEML) {
+      rhs = generateBroadcastOp(rhs, stoi(opAttr.start[1]), state,
```
david-vc, line 739: Do we need additional checking here to determine that "broadcast" is the right semantics?
The condition !isSimpleVecIntrinsic does not imply that it must be a broadcast.
(see definition of isSimpleVecIntrinsic())

david-vc (commented):
Hi Lina: 
This looks reasonable. Please also ask for review from Javier and Brandon. Thank you,
David

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -40,6 +40,9 @@ static llvm::cl::opt<bool>
                         llvm::cl::desc("Enable the unaligned loads check"),
                         llvm::cl::init(true));
 
+static llvm::cl::opt<bool> AIEML("aieml", llvm::cl::desc("AI Engine-ML"),
```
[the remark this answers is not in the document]
linay-xsj: > In the future we may need to consider more variants. I suggest an enum-style command line option "aiearch" for with (currently) 2 legal values "aie" and "aie-ml". This will make it more consistent with other tool chains. The default value would be "aie"

Agree.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
linay-xsj: Yes, this is on my cleanup list. I will add canGenerateBroadcast() before generateBroadcastOp(). We need to check whether the zbuff is a splat and zoffset is 0x00000000, and xbuffer's accIncr is 1. This will be done after I finish i8 type test, because we might need to add more checks.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -357,17 +357,9 @@ static bool writesToAccumulator(Operation *op) {
...
+    return true;
   } else if (auto mul_elemOp = dyn_cast<aievec::MulElemOp>(op)) {
-    return mul_elemOp.getResult()
-        .getType()
-        .cast<VectorType>()
-        .getElementType()
-        .isa<IntegerType>();
+    return true;
   } else if (isa<aievec::UPSOp>(op))
     return true;
```
jsetoain, line 364: You can merge all of this in a single "if" statement with a multiple "isa" check. Something like "isa<FMAElemOp, MulElemOp, UPSOp>(op)".

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -575,46 +576,46 @@ static LogicalResult printOperation(CppEmitter &emitter, aievec::UPSOp upsOp) {
   raw_indented_ostream &os = emitter.ostream();
 
   // Generate the initialization for the accumulator
-  if (failed(emitter.emitAssignPrefix(*upsOp)))
+  if (failed(emitter.emitAssignPrefix(*upsOp, true)))
```
jsetoain, line 579: Nit: consider writing `/*isAcc =*/true` for clarity.

Maybe even change "isAcc" for "isAccumulator" or something else a bit more descriptive, but that's just a suggestion, we're not being very consistent in this regard.

jsetoain (changes requested):

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -690,35 +713,53 @@ static Operation *generateFMAOp(vector::FMAOp fmaOp, AIEOpAttributes &opAttr,
...
     LLVM_DEBUG(llvm::dbgs()
                << "\n\nCreated UPS op " << acc << " to move the output of "
                << fmaOp << " into accumulator");
   }
   // If the lhs operand vector is not >= twice the rhs operand vector, then use
-  // concat operator.
+  // concat operator. If targeting for AIE-ML intrinsics, use broadcast operator
+  // for rhs.
   Value lhs = fmaOp.getLhs();
+  Value rhs = fmaOp.getRhs();
```
david-vc, line 726: Hi @linay-xsj : 
I see little commonality between AIEML and !AIEML from 727 till the end of the function.
I suggest that 727-764 be recoded as:
`Operation *xfmaOp = nullptr;
 if (AIEML) {
   .. 
  } else { 
   .. 
  }
`
This will make it clear which cases are currently supported for AIE and which cases are currently supported for AIEML.

david-vc (changes requested):
Hi @linay-xsj : this looks good, but I think that my comment on AIEVectorize.cpp:726 needs to be addressed before I can approve. Thank you. David

jsetoain (changes requested):

david-vc (approved):

PR #187, opened 2022-10-26
Update llvm to f8b84268

fifield:


fifield:
I don't understand the reason for the build error:
https://github.com/Xilinx/mlir-aie/actions/runs/3331939238/jobs/5514627847#step:12:129
I can't reproduce it.

fifield:
Replaced by https://github.com/Xilinx/mlir-aie/pull/192

PR #191, opened 2022-10-28
Set pathfinder router as default router.

AndraBisca:


AndraBisca:
I've modified aiecc's main.py such that the default router is the pathfinder. However, I've not deleted the source files of the old router. Should I @jgmelber ?

tools/aiecc/aiecc/main.py
```diff
@@ -173,10 +173,7 @@ async def process_arm_cgen(self):
 
       # Generate the included host interface
       file_physical = os.path.join(self.tmpdirname, 'input_physical.mlir')
-      if(self.opts.pathfinder):
-        await self.do_call(task, ['aie-opt', '--aie-create-pathfinder-flows', '--aie-lower-broadcast-packet', '--aie-create-packet-flows', '--aie-lower-multicast', self.file_with_addresses, '-o', file_physical]);
-      else:
-        await self.do_call(task, ['aie-opt', '--aie-create-flows', '--aie-lower-broadcast-packet', '--aie-create-packet-flows', '--aie-lower-multicast', self.file_with_addresses, '-o', file_physical]);
+      await self.do_call(task, ['aie-opt', '--aie-create-pathfinder-flows', '--aie-lower-broadcast-packet', '--aie-create-packet-flows', '--aie-lower-multicast', self.file_with_addresses, '-o', file_physical]);
```
jgmelber, line 176: If you remove the option to use the basic router, the --pathfinder command line option should be removed from cl_arguments.py, and the files implementing the basic router. Alternatively, you could revert this and change the default for the --pathfinder cl_argument to be True and add a --no-pathfinder option. My vote is for the first option.

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -121,7 +121,7 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
   for (auto tileOp : module.getOps<TileOp>()) {
     int col = tileOp.colIndex();
     int row = tileOp.rowIndex();
-    if (tileOp.isShimTile()) {
+    if (tileOp.isShimNOCorPLTile()) {
```
jgmelber, line 124: Thanks!

jgmelber (changes requested):
See my comment about the --pathfinder command line option and possibly removing the basic router entirely. Removing the option might impact dependent repos that are in the air...

tools/aiecc/aiecc/main.py
AndraBisca: Set pathfinder as default router, removed --pathfinder option and removed source files for old router.

AndraBisca:
@jgmelber I have removed the command line --pathfinder option and also removed the files implementing the old default router. 
Ready for review.

stephenneuendorffer (approved):
I'm so glad we've gotten to this point :)

jgmelber (approved):

PR #193, opened 2022-11-01
Make --host_target="" the default for aiecc

fifield:
The user should not need to specify a target if they are not trying to cross-compile. This is the default behavior of most compiler drivers.


jgmelber (approved):
This makes sense. It's worth noting the only platform provided for MLIR-AIE has an ARM processor for its host.

stephenneuendorffer:
I've been wanting to do this for a while... I wonder if we should go the next step and have the tests not force an arm host either?  Or, maybe a better question is: what scenarios do we want to test?  It would be nice to test:
1) X86 build, X86 host W/VCK5000
2) X86 build, Arm host W/VCK190 (e.g. aiecc is a cross-compiler, only scenario where --host-target is needed)
3) Arm build, Arm host W/VCK190 (e.g. using a cross-compiled toolchain)

fifield:
> I've been wanting to do this for a while... I wonder if we should go the next step and have the tests not force an arm host either? Or, maybe a better question is: what scenarios do we want to test? It would be nice to test:
> 
> 1. X86 build, X86 host W/VCK5000
> 2. X86 build, Arm host W/VCK190 (e.g. aiecc is a cross-compiler, only scenario where --host-target is needed)
> 3. Arm build, Arm host W/VCK190 (e.g. using a cross-compiled toolchain)

Downstream I have eliminated the arm target flags from tests to better support (1) and (3) and that is the motivation for this PR. We still have (2) downstream but I've already tried to remove it once and I question its utility. It seems to be a convenient way to build the lit tests to be copied over to a vck190. I'd rather have a good way to build and run the lit tests on the vck190 board using and installation package of (3).

stephenneuendorffer:
> > I've been wanting to do this for a while... I wonder if we should go the next step and have the tests not force an arm host either? Or, maybe a better question is: what scenarios do we want to test? It would be nice to test:
> > 
> > 1. X86 build, X86 host W/VCK5000
> > 2. X86 build, Arm host W/VCK190 (e.g. aiecc is a cross-compiler, only scenario where --host-target is needed)
> > 3. Arm build, Arm host W/VCK190 (e.g. using a cross-compiled toolchain)
> 
> Downstream I have eliminated the arm target flags from tests to better support (1) and (3) and that is the motivation for this PR. We still have (2) downstream but I've already tried to remove it once and I question its utility. It seems to be a convenient way to build the lit tests to be copied over to a vck190. I'd rather have a good way to build and run the lit tests on the vck190 board using an installation package of (3).

At least with the aie tests, if you have cross-compiled tools installed on the VCK190, you can run cmake in the test directory and they should discover the installed tools and use them locally (e.g., the test directory can be a toplevel cmake project)  There are 2 challenges with this:

1) the install is still quite big, even as a release build with libLLVM.so, so shlepping the whole thing onto a laptop and then onto the SD card takes a bit.  It might be easier if my VCK190 could connect to the regular network.  
2) compiling any non-trivial design will take a while, since the parallelism is limited.

PR #194, opened 2022-11-07
[VITIS-5359]Add new aievec dialect operations (shift, mul_conv, mac_cov) to extend the vectorization capability for AIE-ML

linay-xsj:


include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -494,4 +494,69 @@ def AIEVec_UnpackOp:
...
+    Pure
+  ]>,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs,
+             DefaultValuedAttr<SI32Attr, "0">:$M,
+             DefaultValuedAttr<SI32Attr, "0">:$N)>,
+  Results<(outs AnyVector:$result)> {
+  let summary = "AIE2 multiply convolution";
+  let description = [{
+    Xilinx-specific multiply convolution intrinsic. Multiply convolution operation
+    of (M x N)matrix with (N x P)kernel. Currently we only support P=1.
```
david-vc, line 525: (N x P) kernel => (N x 1) kernel.
Remove "Currently .. P=1.'

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -1284,5 +1284,272 @@ ParseResult UnpackOp::parse(OpAsmParser &parser, OperationState &result) {
...
+  unsigned ltypeWidth = ltype.getIntOrFloatBitWidth();
+  unsigned rtypeWidth = rtype.getIntOrFloatBitWidth();
+  unsigned atypeWidth = atype.getIntOrFloatBitWidth();
+
+  // Checks on the number of lanes
+  unsigned accLanes = getVectorLaneSize(resultType);
+  unsigned rhsLanes = getVectorLaneSize(rhsType);
+  unsigned lhsLanes = getVectorLaneSize(lhsType);
+
+  // lane size must match
```
david-vc, line 1465: There are some basic requirements on M,N:
M > 0, 
N > 0
2 * M >= M + N - 1 (because we will be shifting lhs over N-1 positions

Furthermore: 
accLanes == M
lhsLanes == M * 2

david-vc (changes requested):
Hi @linay-xsj 
This is nice work. I understand that you've attempted to achieve vectorization for the conv2d i16 case with minimal code changes. I do have some small suggestions in comments that need to be addressed.
I do have a larger concern about the overall approach, but I'm not going to hold back your code changes because of it.

linay-xsj:
> Hi @linay-xsj This is nice work. I understand that you've attempted to achieve vectorization for the conv2d i16 case with minimal code changes. I do have some small suggestions in comments that need to be addressed. I do have a larger concern about the overall approach, but I'm not going to hold back your code changes because of it.

Thanks, David. I will make a revision according to your suggestions.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -222,7 +222,8 @@ static inline std::pair<int32_t, int32_t> getNumRowsAndCols(Operation *op) {
                   : (lsize == 16 && rsize == 8) ? 64
                                                 : 32;
 
-  if (AIEML) {
+  int32_t vectorSize = lanes * rsize;
```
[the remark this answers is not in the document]
linay-xsj: Line 221 - line 223 is the original setting which I want to maintain it to the original settings.  For AIEML, we need to double the size of width(for example, extend 256b to 512b), but for conv2d case, super vector size is 16 for int16 type, so we don't need to double the size of width.

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
linay-xsj: This is a good point!

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1824,6 +1828,289 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+  Value rhs = nullptr;
+  Value acc = nullptr;
+  bool isMulOp = false;
+
+  // If the acc operand is a mul op, we will try to generate mul_conv operation
+  // If the acc operand is a fma op, we will try to generate fma_conv operation
+  if (auto mulOp = dyn_cast<aievec::MulOp>(mulOrFMAOp)) {
+    isMulOp = true;
+    // Check 7. mulOp must have two operands and one result
+    assert(mulOp->getNumOperands() == 2 && mulOp->getNumResults() == 1);
```
jsetoain, line 1910: Is it even possible to hit this assert? Wouldn't the MulOp parser fail to parse the Op if it doesn't have two operands and one result?

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1824,6 +1828,289 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+
+    // Determine the lhs and rhs values for the mul_conv
+    lhs = mulOp->getOperand(0);
+    rhs = mulOp->getOperand(1);
+
+    assert(lhs && rhs && "Failed to find the two operands of the mul_conv op");
```
jsetoain, line 1916: Same with this assert, this should be guaranteed by the parser.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1824,6 +1828,289 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+  } else {
+    auto fmaOp = cast<aievec::FMAOp>(mulOrFMAOp);
+    // Check 7. fmaOp must have three operands and one result
+    assert(fmaOp->getNumOperands() == 3 && fmaOp->getNumResults() == 1);
```
jsetoain, line 1920: Same

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1824,6 +1828,289 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+
+    // Determine the lhs, rhs and acc values for the fma_conv
+    lhs = fmaOp->getOperand(0);
+    rhs = fmaOp->getOperand(1);
+    acc = fmaOp->getOperand(2);
+
+    assert(lhs && rhs && acc &&
```
jsetoain, line 1927: Same

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/conv2d_uij_i16_noinit.mlir
```diff
@@ -0,0 +1,70 @@
+// REQUIRES: valid_xchess_license
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" -aieml=true --aie-vectorize="shift=10 zero-offset=4" | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
```
jsetoain, line 2: As a note: at some point we need to change this "floating option" (aieml) and make it a pass option for "aie-vectorize". Same in aie-translate, it should be an option for "aievec-to-cpp". In fact, we might want two translate passes, one for AIEv1 and one for AIEv2.

test/aievec/gemm64_i32_unroll16.mlir
```diff
@@ -125,82 +125,82 @@ module {
 //CHECK-NEXT:        scf.for %arg5 = %c0_2 to %c64_3 step %c16_4 {
 //CHECK-NEXT:          %2 = aievec.upd %arg0[%arg3, %arg5] {index = 0 : i8, offset = 0 : si32} : memref<64x64xi32>, vector<16xi32>
 //CHECK-NEXT:          %3 = aievec.upd %arg1[%arg5, %arg4] {index = 0 : i8, offset = 0 : si32} : memref<64x64xi32>, vector<16xi32>
-//CHECK-NEXT:          %4 = aievec.broadcast %2 {idx = 0 : i8} : vector<16xi32>, vector<16xi32>
+//CHECK-NEXT:          %4 = aievec.broadcast %2 {idx = 0 : si32} : vector<16xi32>, vector<16xi32>
```
jsetoain, line 128: Same with all these integers with signedness, it's better to keep them signless. Also, consider to maybe use `index` for these attributes.

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/kernel.cc
```diff
@@ -0,0 +1,7 @@
+#include "testbench.h"
+
+#if __AIEARCH__ == 10
+#include "gen_aie.cc"
+#else
+#include "gen_aie-ml.cc"
+#endif
```
jsetoain, line 7: The way to enable and disable tests in using lit.

jsetoain (changes requested):

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/conv2d_uij_i16_noinit.mlir
linay-xsj: Yes, we will do this after we finish the ops for aie2.

PR #199, opened 2022-11-15
Add named locks and external buffers.

stephenneuendorffer:
Note that the generic format of the 'mlir_aie_set_addr' call has been removed.


AndraBisca (approved):
Great changes! They will really make writing tests much easier. Especially with the external_buffers now decoupled from the buffer descriptors of the shim DMA.

denolf (approved):

jgmelber (approved):

PR #201, opened 2022-11-28
Improve aiecc build speed.

stephenneuendorffer:
It turns out that lots of compile jobs isn't always the right thing.  
For the prime sieve example (336 cores), a unified build runs in about 3 seconds, compared to 9 seconds for a parallel build. (on xsjsda131, with a release build of acdc)  This is essentially how long it takes to compile the ARM code, so making it run any faster becomes pointless.

Using xchesscc + xbridge, a unified build is about 2 minutes, while a parallel build is over 3 minutes. (on xsjrdevl167, 56 cores, debug build of acdc)

These are relatively beefy machines.  Running with less cores only makes a unified build an even better idea.


fifield (approved):
I tried out `aiecc.py -j 0` and it worked great.

PR #203, opened 2022-11-30
[VITIS-5359]Add new aievec dialect operations (shift, shuffle, mul_conv, mac_cov) to extend the vectorization capability for AIE-ML

linay-xsj:
The new operations are demonstrated to work by conv2d kernel test cases with i8 type in this change set.


test/aievec/conv2d_uij_i8_noinit_aie-ml.mlir
```diff
@@ -0,0 +1,94 @@
...
+//CHECK-NEXT:      %c32 = arith.constant 32 : index
+//CHECK-NEXT:      scf.for %arg4 = %c0_2 to %c256 step %c32 {
+//CHECK-NEXT:        %6 = aievec.upd %arg0[%arg3, %arg4] {index = 0 : i8, offset = 0 : si32} : memref<18x288xi8>, vector<64xi8>
+//CHECK-NEXT:        %7 = aievec.mul_conv %6, %1 {M = 32 : si32, N = 8 : si32} : vector<64xi8>, vector<64xi8>, vector<32xi32>
+//CHECK-NEXT:        %8 = aievec.upd %arg0[%4, %arg4] {index = 0 : i8, offset = 0 : si32} : memref<18x288xi8>, vector<64xi8>
+//CHECK-NEXT:        %9 = aievec.fma_conv %8, %2, %7 {M = 32 : si32, N = 8 : si32} : vector<64xi8>, vector<64xi8>, vector<32xi32>
+//CHECK-NEXT:        %10 = aievec.upd %arg0[%5, %arg4] {index = 0 : i8, offset = 0 : si32} : memref<18x288xi8>, vector<64xi8>
+//CHECK-NEXT:        %11 = aievec.fma_conv %10, %3, %9 {M = 32 : si32, N = 8 : si32} : vector<64xi8>, vector<64xi8>, vector<32xi32>
+//CHECK-NEXT:        %12 = aievec.srs %11 {shift = 0 : i8} : vector<32xi32>, vector<32xi8>
+//CHECK-NEXT:        vector.transfer_write %12, %arg2[%arg3, %arg4] {in_bounds = [true]} : vector<32xi8>, memref<16x256xi8>
```
jsetoain, line 94: To future-proof the test, use FileCheck regex matching/substitution blocks as opposed to a code dump. This test fixes the output, rather than ensuring semantic correctness.

jsetoain (changes requested):
I can't add a comment to the file, but please notice that you pushed a vim swapfile in `unit_tests/aievec_tests/i8xi8_static_sized_memref_aie-ml`

david-vc (changes requested):
Hi @linay-xsj : 
This is nice work. I have commented on lack of explanation of certain areas. I'd appreciate if you could address this with comments in the code.
I also would like the input data ([kernelAndInImage_256x16_k3_gaussblur.txt) changed to have more lines with valid data. Currently all values after line 70 of the file are '2'.

test/aievec/conv2d_uij_i8_noinit_aie-ml.mlir
jsetoain: You can use regex matching and substitutions everywhere, and still check the details of each operation. You just avoid relying on naming specifics, and ideally the order of operations, focusing on dependencies and semantics only. Check, for instance, this MLIR test: [vector-contract-matvec-transforms.mlir](https://github.com/llvm/llvm-project/blob/main/mlir/test/Dialect/Vector/vector-contract-matvec-transforms.mlir).

We should do this in all our tests, we haven't. At some point we need to bring that up to standard but, for now, I think a good step in the right direction is making all new tests to align with MLIR standards.

david-vc (approved):
Nice work Lina.

test/aievec/conv2d_uij_i8_noinit_aie-ml.mlir
```diff
@@ -0,0 +1,94 @@
...
+            %c33 = arith.addi %c32, %p33 : i8
+
+            //Store accumulated sum
+            affine.store %c33, %C[%arg3, %arg4] : memref<16x256xi8>
+        }
+    }
+    return
+}
+
+//CHECK-NEXT:    %[[C00]] = arith.constant 0 : index
```
jsetoain, line 70: If you run a canonicalization pass at the end, you won't end up with constant index = 0 defined everywhere, which will simplify the test.

test/aievec/conv2d_uij_i8_noinit_aie-ml.mlir
```diff
@@ -0,0 +1,94 @@
...
+//CHECK-NEXT:    %[[T0]] = aievec.upd %[[A1]][%[[C00]]] {index = 0 : i8, offset = 0 : si32} : memref<48xi8>, vector<64xi8>
+//CHECK-NEXT:    %[[T1]] = aievec.shuffle %[[T0]] {mode = 0 : i32} : vector<64xi8>, vector<64xi8>
```
jsetoain, line 72: There is no need to use `CHECK-NEXT`, a simple `CHECK` should do and if, for whatever reason, at some point something decides to add casts or non-relevant ops in between, the test won't break.

jsetoain (approved):

PR #211, opened 2022-12-06
Minor fix for xchesscc compilation

jgmelber:


AndraBisca:
Tried it out and it indeed fixes the compilation issues I was having.

tools/aiecc/aiecc/main.py
```diff
@@ -94,7 +94,7 @@ async def chesshack(self, task, llvmir):
       # Note that chess-clang comes from a time before opaque pointers
       #await self.do_call(task, ['clang', "-Xclang -no-opaque-pointers", llvmir_chesshack, self.chess_intrinsic_wrapper, '-S', '-emit-llvm', '-o', llvmir_chesslinked])
       global llvmlink
-      await self.do_call(task, [llvmlink, '--opaque-pointers=0', llvmir_chesshack, self.chess_intrinsic_wrapper, '-S', '-o', llvmir_chesslinked])
+      await self.do_call(task, ['llvm-link', '--opaque-pointers=0', llvmir_chesshack, self.chess_intrinsic_wrapper, '-S', '-o', llvmir_chesslinked])
```
fifield, line 97: shouldn't a fix go here instead?
https://github.com/Xilinx/mlir-aie/blob/c438678442c52677f80fc8df0d638eb8ad9ca67e/tools/aiecc/aiecc/main.py#L333

stephenneuendorffer: It's not clear to me what this is solving?

utils/single_unit_test_compile.sh
```diff
@@ -29,7 +29,8 @@ export RUNTIME_LIB_DIR=$2
 #aiecc.py --sysroot=$SYSROOT_DIR aie.mlir -I${RUNTIME_LIB_DIR} \
 # ${RUNTIME_LIB_DIR}/test_library.cpp ./test.cpp -o test.elf
 
-aiecc.py --sysroot=$SYSROOT_DIR --aie-generate-xaiev2 ./aie.mlir \
+aiecc.py --sysroot=$SYSROOT_DIR --aie-generate-xaiev2 --no-unified \
+		--host-target=aarch64-linux-gnu ./aie.mlir \
     -DLIBXAIENGINEV2 -I${RUNTIME_LIB_DIR} ${RUNTIME_LIB_DIR}/test_library.cpp \
```
stephenneuendorffer, line 34: Does this mean that --unified is not working?

tools/aiecc/aiecc/main.py
jgmelber: Using chess, we should be using llvm-link from the llvm build, not llvm-aie

stephenneuendorffer: Actually, this is a specific work around for a patch that is upstreamed but not yet merged into our local llvm build.  llvm-link didn't properly respect the --opaque-pointer option.  So, I expect that this will only work with peano/bin/llvm-link for the moment.

utils/single_unit_test_compile.sh
jgmelber: --unified still errors out for xchesscc

stephenneuendorffer: > --unified still errors out for xchesscc

What's the error?

stephenneuendorffer: OK, do you have peano and it can't find it, or are you trying to make it work without peano?

jgmelber: Work without peano, ensure support for users that only have access to xchesscc

stephenneuendorffer: got it. I'll take a look at cleaning up the unified flow

jgmelber:
@stephenneuendorffer would using `-DCLANG_ENABLE_OPAQUE_POINTERS=OFF` in the LLVM build be useful/sufficient? 

https://github.com/llvm/llvm-project/issues/54970

stephenneuendorffer:
> @stephenneuendorffer would using `-DCLANG_ENABLE_OPAQUE_POINTERS=OFF` in the LLVM build be useful/sufficient?
> 
> [llvm/llvm-project#54970](https://github.com/llvm/llvm-project/issues/54970)

I think that only affects the LLVM code emitted by the clang frontend.

utils/single_unit_test_compile.sh
stephenneuendorffer: See https://github.com/Xilinx/mlir-aie/pull/213

stephenneuendorffer:
For llvm-link, I'd rather jump bump llvm to this and remove the workaround, which should remove the last dependence on peano.

commit a2dc9f367c6be90589626a575f61ebea5409eee0 (origin/main, origin/HEAD, llvm-link)
Author: Stephen Neuendorffer <stephen.neuendorffer@xilinx.com>
Date:   Mon Nov 21 18:03:30 2022 -0800

    [llvm-link] Fix options of llvm-link

PR #213, opened 2022-12-07
Don't call opt when compiling with xchesscc

stephenneuendorffer:
The opt call adds an unnecessary dependence on peano, and should be redundant with the chess optimizations anyway.


AndraBisca (approved):

AndraBisca (changes requested):
I tried it out on my fork and it works, but only in combination with "--host-target=aarch64-linux-gnu" added to the single_unit_test_compile.sh, as in #211

jgmelber:
@AndraBisca @stephenneuendorffer Fixed the unit test compile script in #215

jgmelber (approved):

PR #217, opened 2022-12-08
Update llvm to 798fa4b4

fifield:
No changes needed to build.

This update is still missing fixes to chess-clang which are required due to upstream changes to the .ll format


stephenneuendorffer (commented):
Chess works with my commit.

PR #220, opened 2022-12-08
Add aie dialect python bindings

fifield:
`cmake -DAIE_ENABLE_BINDINGS_PYTHON=ON` to get python bindings for AIE dialect. MLIR must be compiled with `MLIR_ENABLE_BINDINGS_PYTHON=ON`


abisca:
It would probably be good to add the MLIR_ENABLE_BINDINGS_PYTHON=ON flag to the build-mlir-aie.sh script in /utils as well.

fifield:
> It would probably be good to add the MLIR_ENABLE_BINDINGS_PYTHON=ON flag to the build-mlir-aie.sh script in /utils as well.

done.

PR #221, opened 2022-12-08
Autocorrelation

stephenneuendorffer:
Hopefully this would get simpler with the shimDMA objectFifo.


denolf (approved):

AndraBisca (approved):
With object FIFOs you wouldn't need to write all of the dataflow movement explicitly. However, you would need to change the code inside the cores to access buffer elements through the respective FIFOs. I can make an equivalent of this!

PR #223, opened 2022-12-12
[VITIS-5359] Enable mul_conv and mac_conv to int16 type in AIE-ML

linay-xsj:


lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1954,6 +1954,248 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
     op->erase();
 }
 
+// We go through each fma operation and try to find the patten like this-
```
david-vc, line 1957: patten => pattern

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1954,6 +1954,248 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+// Transform -
+// %5 = aievec.mul %4, %0 {xoffsets = "0x03020100", xoffsets_hi = "0x07060504",
+// xsquare = "0x2110", xstart = "0", zoffsets = "0", zoffsets_hi = "0", zstart =
+// "0", zstep = "1"}
+//
+// %6 = aievec.mac %4, %0, %5 {xoffsets = "0x03020100",
+// xoffsets_hi = "0x07060504", xsquare = "0x2110", xstart = "2", zoffsets = "0",
+// zoffsets_hi = "0", zstart = "2", zstep = "1"}
+//
+// to-
```
david-vc, line 1969: Replace the values of xoffsets, xoffsets_hi, xsquare, with symbolic constants. This would make it easier to spot which attribute among %5 and %6 that need to have the same value, and which attributes need to differ.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1954,6 +1954,248 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+    lhs = fmaOp->getOperand(0);
+    rhs = fmaOp->getOperand(1);
+    acc = fmaOp->getOperand(2);
+  }
+
+  // Check 6. The def of two operands are upd operations
+  auto lUpdOp = dyn_cast<aievec::UPDOp>(lhs.getDefiningOp());
+  auto rUpdOp = dyn_cast<aievec::UPDOp>(rhs.getDefiningOp());
+
+  if (!lUpdOp || !rUpdOp) {
```
david-vc, line 2048: Why do these need to be UPDs?

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1954,6 +1954,248 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+
+  if (acc && rhs.getParentBlock() != acc.getParentBlock())
+    return false;
+
+  // Check 8. xstart and zstart distance between two operations should be
+  // 2. offsets, offsets_hi, square and step of two operations should be same.
+  if (isMulOp) {
+    aievec::MulOp defOp = cast<aievec::MulOp>(mulOrFMAOp);
+
+    if (curOp.getOffset(0) != defOp.getOffset(0) ||
```
david-vc, line 2065: What about the **zstep** attribute?

david-vc (changes requested):
Hi @linay-xsj :
As we discussed earlier, I do not think that the overall approach of handling int16 for AIE-ML as a post-processing step is appropriate. However for sake of moving forward, I'm willing to approve the change after you address my comments. Thank you,
David

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -1954,6 +1954,254 @@ static void fuseFMAOpsForColumnTopology(func::FuncOp func, VectState *state) {
...
+                stoi((std::string)defOp.getStart(0)) !=
+            2 ||
+        stoi((std::string)curOp.getStart(1)) -
+                stoi((std::string)defOp.getStart(1)) !=
+            2)
+      return false;
+  }
+
+  return true;
+}
```
jsetoain, line 2102: A cleaner way to do this, write a new "satisfiesProperty" (templated) method, and end the method with:

```cpp
return isMulOp && satisfiesProperty(defOp, cast<aievec::MulOp>(mulOrFMAOp)) ||
                  satisfiesProperty(defOp, cast<aievec::FMAOp>(mulOrFMAOp));
```

Not sure how I'd call that property, though. `matchOffsetAndXDistanceForFusion`, `mulOpAttributesAllowFusion`?

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/conv2d_uij_i16_noinit.mlir
```diff
@@ -0,0 +1,70 @@
+// REQUIRES: valid_xchess_license
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" -aieml=true --aie-vectorize="shift=10 zero-offset=4" | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
+// RUN: xchesscc -f -g +s -p me -P ${XILINX_VITIS}/aietools/data/aie_ml/lib/ +w work +o work -I%S -I. %S/testbench.cc %S/kernel.cc
```
jsetoain, line 3: Notice that `XILINX_VITIS` is not defined in our "lit" config files. If you're running this test on its own on your computer, it will pick it up from your environment, but `lit` will wipe out before running. I'm already tripping over another test that does this, if I run `ninja check-aie` it warns me that `XILINX_VITIS` is being unset and proceeds to fail.

There is a substitution that replaces "aietools" (defined as `-DVITIS_AIETOOLS_DIR` on cmake configuration) with what I'd expect "${XILINX_VITIS}/aietools" to point at, so you should probably replace what you have there with:

```bash
xchesscc -f -g +s -p me -P %aietools/data/aie_ml/lib/ +w work +o work -I%S -I. %S/testbench.cc %S/kernel.cc
```

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/testbench.cc
```diff
@@ -0,0 +1,135 @@
+#include "testbench.h"
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#define MAX_IMAGE_SIZE                                                         \
+  32768 // Test image are small (enough for 128x128 or 256x16)
```
jsetoain, line 9: Extra nl

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/conv2d_uij_i16_noinit.mlir
```diff
@@ -0,0 +1,70 @@
...
+// RUN: cp -r %S/data . && xme_ca_udm_dbg -qf -T -P ${XILINX_VITIS}/aietools/data/aie_ml/lib/ -t "%S/../profiling.tcl ./work/a.out"
```
jsetoain, line 4: Same here.

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/kernel.cc
```diff
@@ -0,0 +1,7 @@
+#include "testbench.h"
+
+#if __AIEARCH__ == 10
+#include "gen_aie.cc"
+#else
+#include "gen_aie-ml.cc"
+#endif
```
jsetoain, line 7: Still don't like this, this should be handled by lit.

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/gen_aie-ml.cc
```diff
@@ -0,0 +1,38 @@
+void conv2d(int16_t * restrict v1, int16_t * restrict v2, int16_t * restrict v3) {
```
jsetoain, line 1: I may be misunderstanding this, but should this file be generated by the test itself? Where it says: `aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc`

jsetoain (changes requested):

test/unit_tests/aievec_tests/i16xi16_static_sized_memref_aie-ml/gen_aie-ml.cc
linay-xsj: Yes, it is generated by aie-translate. I put it here just for the reference.

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
linay-xsj: Thanks! This is a good point!

david-vc (approved):

PR #234, opened 2022-12-15
[AIE][aievec] Remove the use of environment variable in test

jsetoain:


stephenneuendorffer:
Thanks!  Hadn't gotten to fixing that :)

PR #237, opened 2022-12-15
Disable next bd in libXAIEV2 driver calls when bd branches to end

AndraBisca:


stephenneuendorffer:
Please add a test.

AndraBisca:
Done. Please tell me if you'd rather the test file be placed anywhere else.

test/unit_tests/24_dma_endBlock/test.cpp
```diff
@@ -0,0 +1,107 @@
...
+    mlir_aie_start_cores(_xaie);
+
+    mlir_aie_acquire_lock_a14_6(_xaie, 0, 0);
+    mlir_aie_acquire_lock_a34_7(_xaie, 1, 0);
+
+    mlir_aie_release_lock_a14_6(_xaie, 1, 0);
+    mlir_aie_release_lock_a34_7(_xaie, 0, 0);
+
+    // Wait time for cores to run. Number used here is much larger than needed.
+    usleep(100);
```
stephenneuendorffer: This should have a lock, rather than rely on sleep()

test/unit_tests/24_dma_endBlock/test.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+    // Check buffer at index 3 again for expected value of 14 for tile(1,4)    
+    mlir_aie_check("After start cores:", mlir_aie_read_buffer_a14(_xaie, 3), 14,
+                   errors);
+    // Check buffer at index 5 again for expected value of 114 for tile(3,4)    
+    mlir_aie_check("After start cores:", mlir_aie_read_buffer_a34(_xaie, 5), 114,
+                   errors);
```
stephenneuendorffer: It's not clear to me what this is testing, exactly.  How does this fail without the patch?

test/unit_tests/24_dma_endBlock/test.cpp
jackl-xilinx: That's true. That could be changed to check for lock status instead of using a sleep.

test/unit_tests/24_dma_endBlock/test.cpp
jackl-xilinx: Without the patch, the cf.br ^end becomes and cf.br ^bd0. This means that the test.cpp calls to the lock functions will cause the DMAs to run again thereby overwriting the output with a 0. The test will fail if that's the case.

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -330,12 +330,15 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
...
+          if (!nextBlock->getOps<EndOp>().empty())
+            enableNextBd = 0;
+
           int nextBdNum = blockMap[nextBlock];
           output << "XAie_DmaSetNextBd(" << tileDMAInstRefStr(col, row, bdNum)
                  << ", "
                  << " /* nextbd */ " << nextBdNum << ", "
-                 << " /* enableNextBd */ 1);\n"; // TODO Check if br ^end: to
-                                                 // disable this?
+                 << " /* enableNextBd */ " << enableNextBd << ");\n";
```
jgmelber, line 341: It would be good to include a simple test to check for 0 and 1 for just the mem blocks with cr.br ^end and cf.br ^bd0 respectively

jgmelber (changes requested):

jackl-xilinx:
This change has been recreated in https://github.com/Xilinx/mlir-aie/pull/446/

PR #243, opened 2022-12-20
Squashed changes for AIRBIN target

jgmelber:
Creates a new target to lower mlir-aie code to a fat binary format, this is supported in mlir-air by runtime functions, test and demonstration cases. The target is derived from `AIRTargetLibXAIEv1.cpp`. The pass is enabled in `aiecc.py` with the flag `--aie-generate-airbin`

### DIFF
--- a/lib/Targets/AIETargetAirbin.cpp
+++ b/lib/Targets/AIETargetAirbin.cpp
@@ -232,6 +232,9 @@ static constexpr auto SHIM_BD_END = 0x1D15Cu;
 // SHIM resets are handled by the runtime.
 static void generateShimConfig(TileOp &tileOp) {
 
+  assert(tileOp.isShimTile() &&
+         "The tile must be a Shim to generate Shim Config");
+
   TileAddress tileAddress{tileOp};
 
   if (tileOp.isShimNOCTile()) {
--- a/lib/Targets/AIETargetAirbin.cpp
+++ b/lib/Targets/AIETargetAirbin.cpp
@@ -26,6 +26,7 @@
 
 #include "aie/AIENetlistAnalysis.h"
 #include "aie/Dialect/AIE/IR/AIEDialect.h"
+#include "aie/Dialect/AIEX/IR/AIEXDialect.h"
 
 #include "AIETargets.h"
 
@@ -628,7 +629,7 @@ static void configure_dmas(mlir::ModuleOp module, NetlistAnalysis &NL) {
       int acqValue = 0, relValue = 0;
       auto acqEnable = disable;
       auto relEnable = disable;
-      Optional<int> lockID = llvm::None;
+      Optional<int> lockID = std::nullopt;
 
       for (auto op : block.getOps<UseLockOp>()) {
         LockOp lock = dyn_cast<LockOp>(op.getLock().getDefiningOp());
@@ -678,7 +679,7 @@ void XAieDma_TileBdSetLock(XAieDma_Tile *DmaInstPtr, u8 BdNum, u8 AbType, u8 Loc
         LockRelEn = relEnable
         LockRelVal = relValue
    clang-format on */
-          bdData.addr_a = bdAddressLockID(lockID.value()) |
+          bdData.addr_a = bdAddressLockID(*lockID) |
                           bdAddressReleaseEnable(relEnable) |
                           bdAddressAcquireEnable(acqEnable);
 
@@ -1155,7 +1156,7 @@ static void configure_switchboxes(mlir::ModuleOp &module) {
         if (!isEmpty) {
           result.emplace(switchboxOp);
         }
-      } else if (AIE::SelectOp sel = dyn_cast<AIE::SelectOp>(
+      } else if (AIEX::SelectOp sel = dyn_cast<AIEX::SelectOp>(
                      switchboxOp.getTile().getDefiningOp())) {
         // TODO: Use XAIEV1 target and translate into write32s
         TODO;
@@ -1419,7 +1420,7 @@ void XAieTile_StrmConfigSlv(XAieGbl_Tile *TileInstPtr, u8 Slave, u8 Enable,
     }
   }
 
-  Optional<TileAddress> currentTile = llvm::None;
+  Optional<TileAddress> currentTile = std::nullopt;
   for (auto op : module.getOps<ShimMuxOp>()) {
     Region &r = op.getConnections();
     Block &b = r.front();
@@ -1469,7 +1470,7 @@ void XAieTile_StrmConfigSlv(XAieGbl_Tile *TileInstPtr, u8 Slave, u8 Enable,
         }();
 
         // We need to add to the possibly preexisting mask.
-        Address addr{currentTile.value(), 0x1F004u};
+        Address addr{*currentTile, 0x1F004u};
         auto currentMask = read32(addr);
 
         write32(addr, currentMask |
@@ -1497,7 +1498,7 @@ void XAieTile_StrmConfigSlv(XAieGbl_Tile *TileInstPtr, u8 Slave, u8 Enable,
           }
         }();
 
-        Address addr{currentTile.value(), 0x1F000u};
+        Address addr{*currentTile, 0x1F000u};
         auto currentMask = read32(addr);
 
         write32(addr, currentMask |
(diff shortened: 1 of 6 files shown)


lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+  bool operator<(const Write &rhs) const {
+    return static_cast<uint64_t>(address) < static_cast<uint64_t>(rhs.address);
+  }
+
+private:
+  Address address;
+  uint32_t data;
+};
+
+static std::vector<Write> writes;
```
stephenneuendorffer, line 154: Seems like all this static stuff should be a class, hiding the implementation of 'writes'?

stephenneuendorffer: At the very least, it would seem that sorting this by write address would be a good thing?

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+  }
+}
+
+static constexpr auto BD_BASE = 0x1D000u;
+static constexpr auto BD_END = 0x1D1F8u;
+static constexpr auto SHIM_BD_END = 0x1D15Cu;
+
+// The SHIM row is always 0.
+// SHIM resets are handled by the runtime.
+static void generateShimConfig(TileOp &tileOp) {
```
stephenneuendorffer, line 233: assert that it is a shim tile?

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+  static constexpr auto shifted_mask = unshifted_mask << low_bit;
+
+  [[nodiscard]] constexpr uint32_t operator()(uint32_t value) const {
+    return (value << low_bit) & shifted_mask;
+  }
+};
+
+// LLVM may have a way to read ELF files,
+// but for our purposes, manually parsing is fine for now.
+static bool loadElf(TileAddress tile, const std::string &filename) {
```
stephenneuendorffer, line 272: Um... yeah... There's enough magic stuff here to use the LLVM way... you want:

  Expected<std::unique_ptr<ObjectFile>> ELFObjOrErr =
      object::ObjectFile::createELFObjectFile(
          MemoryBufferRef(toStringRef(D.Data), "dummyELF"));

See llvm/object/ObjectFile.h and llvm/object/ELF.h

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+      generateNonShimConfig(tileOp);
+    }
+  }
+}
+
+// Start execution of all the cores.
+// This is by the runtime. We do not need to record it.
+/*
+static void start_cores(mlir::ModuleOp module) {
+
```
stephenneuendorffer, line 436: Remove?

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+static constexpr auto TILE_ADDR_ROW_SHIFT = TILE_ADDR_OFF_WITDH;
+static constexpr auto TILE_ADDR_ROW_WIDTH = 5u;
+
+static constexpr auto TILE_ADDR_COL_SHIFT =
+    TILE_ADDR_ROW_SHIFT + TILE_ADDR_ROW_WIDTH;
+static constexpr auto TILE_ADDR_COL_WIDTH = 7u;
+
+static constexpr auto TILE_ADDR_ARR_SHIFT =
+    TILE_ADDR_COL_SHIFT + TILE_ADDR_COL_WIDTH;
+
```
stephenneuendorffer, line 66: Perhaps this magic stuff should come out of libXAIE?

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1746 @@
...
+  };
+
+  Field<1> dmaChannelReset;
+  Field<0> dmaChannelEnable;
+
+  for (auto memOp : module.getOps<MemOp>()) {
+    /* clang-format off
+    output << "XAieDma_TileInitialize(" << tileInstStr(col, row) << ", " << tileDMAInstStr(col, row) << ");\n";
+    ----
+    DmaInstPtr->BaseAddress = TileInstPtr->TileAddr;
```
stephenneuendorffer, line 523: Remove this commented out code?

tools/aiecc/aiecc/main.py
```diff
@@ -320,10 +324,12 @@ async def run_flow(self):
         progress.update(progress.task,advance=0,visible=False)
         progress.task_completed = progress.add_task("[green] AIE Compilation:", total=len(cores)+1, command="%d Workers" % nworkers)
 
-        processes = [self.process_arm_cgen()]
+        self.progress_bar = progress
+        processes = []
         for core in cores:
           processes.append(self.process_core(core))
         await asyncio.gather(*processes)
+        await self.process_arm_cgen()
```
stephenneuendorffer, line 333: Why this change?  The intention here is to start running ARM code generation ASAP, since it is slower than everything else.  If we can't do that here because the generated ELF files need to be parsed, then we should probably generate a header and separately compile and link the host code in 2 parts.

tools/aiecc/aiecc/main.py
```diff
@@ -221,11 +221,15 @@ async def process_arm_cgen(self):
...
-      file_inc_cpp = os.path.join(self.tmpdirname, 'aie_inc.cpp')
-      if(opts.xaie == 1):
-          await self.do_call(task, ['aie-translate', '--aie-generate-xaie', '--xaie-target=v1', file_physical, '-o', file_inc_cpp])
+      if(opts.airbin):
+        file_airbin = os.path.join(self.tmpdirname, 'air.bin')
+        await self.do_call(task, ['aie-translate', '--aie-generate-airbin', file_physical, '-o', file_airbin])
       else:
-          await self.do_call(task, ['aie-translate', '--aie-generate-xaie', '--xaie-target=v2', file_physical, '-o', file_inc_cpp])
+        file_inc_cpp = os.path.join(self.tmpdirname, 'aie_inc.cpp')
+        if(opts.xaie == 1):
```
stephenneuendorffer, line 229: As I read this, you're no longer generating aie_inc.cpp... But there's a bunch of host APIs in there we probably can't just drop.

stephenneuendorffer (changes requested):
Overall, I'm a little surprised by the result here..  I was expecting something that still depended on libXAIE... instead it seems that the modified libXAIE backend was used to reverse engineer what it was doing, and the logic has been replicated here.... This seems like a significant piece of code that will need to be maintained indefinitely and also will change for AIEv2, etc.

tools/aiecc/aiecc/main.py
jgmelber: The change here is that the airbin generation occurs in `self.process_arm_cgen()` and it needs to have the AIE elfs already in order to generate the airbin

tools/aiecc/aiecc/main.py
jgmelber: For AIR we don't need those APIs when we are using airbin, air_inc.cpp is only used for configuration via libXAIE

stephenneuendorffer: So all the tests run with airbin enabled?  There are more host APIs generated than just configuration.

lib/Targets/AIETargetAirbin.cpp
jgmelber: Agreed, it would be best if the architecture specific values were factored out of all of the AIETargets

lib/Targets/AIETargetAirbin.cpp
jgmelber: The writes are sorted at the end before generating the binary

tools/aiecc/aiecc/main.py
jgmelber: Which tests? This works for the mlir-air tests using airbin, it has not been tested on every air test/example yet. It is not intended as a back end for mlir-aie tests without the air runtime

tools/aiecc/aiecc/main.py
stephenneuendorffer: I'd prefer to see this separated out.

tools/aiecc/aiecc/main.py
stephenneuendorffer: But seems like it very easily could be...

keryell:
Status of this PR?

PR #247, opened 2022-12-22
Nextbd

stephenneuendorffer:
Jeff pointed out that cf.br caused problems with canonicalization... this should fix it.


AndraBisca:
I think having an AIE specific branch-to-next-bd operation inside AIE.mem and AIE.shimDMA is a good change anyway! I will also apply this to #237.

AndraBisca (approved):

fifield (approved):

PR #248, opened 2022-12-23
RFC: AIEX Dialect

stephenneuendorffer:
This PR is a rather significant refactoring of the AIE Dialect, with the goal of moving more experimental concepts into a separate dialect.  I've taken an initial stab at what operations should move with the goal of making sure there weren't extra dependencies: I was glad to see that for the most part, there weren't.  I'd like to get wider input from the rest of you on what you think should be considered a 'core' concept.

Primarily there are three things that I propose to move into the AIEX Dialect:
1) the 'Token' concept as an abstraction of locks
2) the 'multicast' and 'broadcast packet' concepts
3) the 'herd' routing and placement concept

One thing I hope that that the tutorial will cover all of the 'core' concepts and mostly avoid the 'experimental' ones until they are more mature.


AndraBisca:
I think this is a good separation since there's quite a lot of experimental operations in AIE. This would also allow to better identify which areas need to be tested further for the general use-case.

As for deciding which are 'core' concepts, a rule of thumb I personally use is: operations which are present after all the lowerings have executed are physical-level core operations. Everything else is an optimisation or abstraction on top of that, so I would call it a logical-level core concept. And for these core concepts I think they are either widely used, ex. AIE.FlowOp, or allow to combine the physical-level core concepts in a particular way, ex. AIE.objectFifo ops.

AndraBisca:
I would also recommend as experimental the AIE.objectFifo.registerProcess operation and associated lowering pass, AIEObjectFifoRegisterProcess.cpp. While the main idea of what they're supposed to do is implemented, there's still room for more work before I think they're ready for general use.

AndraBisca (approved):

fifield (approved):
It make sense to me to remove experimental or abandoned code from the core dialect and passes.

erwei-xilinx (approved):
Make sense to me. The core dialect and passes are now easier to maintain.

test/create-multicast/test_multicast.mlir
```diff
@@ -18,10 +18,10 @@ module @test_multicast {
...
-    AIE.multi_dest<%74, "DMA" : 0>
-    AIE.multi_dest<%63, "DMA" : 0>
-    AIE.multi_dest<%64, "DMA" : 0>
+  AIEX.multicast(%70, "DMA" : 0){
+    AIEX.multi_dest<%73, "DMA" : 0>
+    AIEX.multi_dest<%74, "DMA" : 0>
+    AIEX.multi_dest<%63, "DMA" : 0>
+    AIEX.multi_dest<%64, "DMA" : 0>
   }
 }
```
jsetoain, line 27: Missing EOL

tools/aie-opt/CMakeLists.txt
```diff
@@ -10,13 +10,13 @@ llvm_map_components_to_libnames(llvm_libs support core irreader)
...
-  MLIRAIEPassIncGen
-  MLIRAIEVecPassIncGen
-)
+# add_dependencies(aie-opt 
+#   MLIRADFIncGen
+#   MLIRAIEIncGen
+#   MLIRAIEEnumsIncGen
+#   MLIRAIEPassIncGen
+#   MLIRAIEVecPassIncGen
+# )
```
jsetoain, line 19: Is there a reason to leave them as comments? Otherwise, I think it's preferable to just delete those lines.

stephenneuendorffer:
> I would also recommend as experimental the AIE.objectFifo.registerProcess operation and associated lowering pass, AIEObjectFifoRegisterProcess.cpp. While the main idea of what they're supposed to do is implemented, there's still room for more work before I think they're ready for general use.

I think we'll take this up separately because it's tricky to have a DialectType (e.g. ObjectFifoType) that is referenced from another dialect.

denolf (approved):

PR #253, opened 2023-01-10
Modified mlir-aie doc for clearer build instructions.

AndraBisca:


denolf (commented):
Add an extra note to make clear steps 2-4 have the be run from within the mlir-aie folder. Thanks. Kristof

PR #254, opened 2023-01-10
Added exit checks to all bash build commands

jackl-xilinx:


fifield:
A different way to do this is:
```
set -o pipefail
...
myCommand |& tee myCommand.log
ec=$?
if [ $ec -ne 0 ]; then
    echo "myCommand error"
    exit $ec
fi
...
```

jackl-xilinx:
Thanks @fifield . I considered that option as it lets you print a message but I was trying to keep it more concise. BTW, do you know why the Build and Test check is failing for this?

PR #267, opened 2023-01-17
Change cf.br to AIE.nextBd

jackl-xilinx:


searsm8:
cf.br also needs to be changed in at least one README (tutorial 4). I'll include this in my review of the tutorials.

PR #269, opened 2023-01-18
Updated build prerequisites and added script to install python packages.

AndraBisca:


denolf (commented):
I never needed joblib, so it can be removed from the requirements.txt

denolf (approved):

PR #270, opened 2023-01-18
[AIE][aievec] Add conversion pass from Vector to AIEVec

jsetoain:
Very rough patch with the initial ideas of the new approach.


test/Conversion/VectorToAIEVec/test-upd.mlir
```diff
@@ -0,0 +1,49 @@
...
+    // CHECK: vector.transfer_write %[[LD]], {{.*}}
+    vector.transfer_write %0, %arg1[%arg2] : vector<16xi32>, memref<256xi32>
+  }
+  return
+}
+
+func.func @veccopy_i64(%arg0: memref<256xi64>, %arg1: memref<256xi64>) {
+  %c0_i64 = arith.constant 0 : i64
+  affine.for %arg2 = 0 to 256 step 16 {
+    // CHECK: %[[LD0:.*]] = aievec.upd {{.*}} {index = 0 : i8, offset = 0 : si32} : memref<256xi64>, vector<16xi64>
```
david-vc, line 40: AIE1 does not have a v16int64 data type.
AIEML does not have any int64 vector data types.

test/Conversion/VectorToAIEVec/test-upd.mlir
```diff
@@ -0,0 +1,49 @@
...
+  affine.for %arg2 = 0 to 256 step 16 {
+    // CHECK: %[[LD:.*]] = aievec.upd {{.*}} {index = 0 : i8, offset = 0 : si32} : memref<256xi32>, vector<16xi32>
+    %0 = vector.transfer_read %arg0[%arg2], %c0_i32 : memref<256xi32>, vector<16xi32>
+    // CHECK: vector.transfer_write %[[LD]], {{.*}}
+    vector.transfer_write %0, %arg1[%arg2] : vector<16xi32>, memref<256xi32>
+  }
+  return
+}
+
+func.func @veccopy_i64(%arg0: memref<256xi64>, %arg1: memref<256xi64>) {
```
david-vc, line 37: My suggestion is to remove this since AIE1/AIE-ML don't really support vector operations on int64 vector types.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+    : public OpConversionPattern<vector::TransferReadOp> {
+  using OpConversionPattern<vector::TransferReadOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(vector::TransferReadOp readOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // TODO: Linearlize accesses when needed
+    SmallVector<Value, 4> indices(adaptor.getIndices().begin(),
+                                  adaptor.getIndices().end());
+    rewriter.replaceOpWithNewOp<xilinx::aievec::UPDOp>(
```
david-vc, line 79: IMO, this pattern as is should fire only if the transfer read is reading a slice that is continguous and vector-aligned.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/// Convert incoming vector operations into their corresponding AIE vector
+/// intrinsics. The pass assumes the IR is in a state
+void ConvertVectorToAIEVec::runOnOperation() {
+  auto func = getOperation();
+  MLIRContext *context = &getContext();
+  RewritePatternSet patterns(context);
+  ConversionTarget target(*context);
+  int8_t aieVersion = 1;
```
david-vc, line 217: It's probably worthwhile defining an enum type for specifying the AIE architecture. e.g.,
enum class AieArch {
 aie,
 aie_ml
};

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+  if (!aieTarget.empty()) {
+    std::string target = aieTarget;
+    if (target == "1" || target == "v1" || target == "V1") {
```
david-vc, line 220: IMO, ambiguity is to be avoided. I would prefer to accept only 2 values "aie" and "aie_ml". If none is specified, it should default to "aie".

david-vc (changes requested):
Hi Javier,
This is a good start. I think that the patterns should more more restrictive on the operand types and sizes to ensure that we do not create aievec operations that cannot be turned into valid intrinsics downstream.
I expect that the lowering of transfer reads will need to become a lot more sophisticated to handle the convolution test cases, but that is OK. 
Thanks.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+using namespace xilinx;
+using namespace xilinx::aievec;
+
+#define DEBUG_TYPE "aievec-lowering"
+
+//===----------------------------------------------------------------------===//
+// Analyses
+//===----------------------------------------------------------------------===//
+
+struct UPDOpEffectiveAccessSizeAnalysis {
```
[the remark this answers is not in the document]
jsetoain: Yes, it's a terrible name. I'll try to find something easier to read, but I'm using this just as an example of an analysis, as a reference. We don't really need this, and once we find a more sensible use case, we can probably get rid of this one entirely.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+    rewriter.replaceOpWithNewOp<xilinx::aievec::UPDOp>(
+        readOp, readOp.getVector().getType(), adaptor.getSource(), indices, 0,
+        0, TypedValue<VectorType>(nullptr));
+    return success();
+  }
+
+};
+
+template <typename SrcOpTy, typename DstOpTy>
+struct OneToOneVectorOpToAIEVecOpPattern : public OpConversionPattern<SrcOpTy> {
```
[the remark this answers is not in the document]
jsetoain: It's a very simple pattern, I'm sure I'm missing many cases. That said, if this should not work for 64-bit addition, the current version is not doing anything about it. The test source MLIR generates the same code both with the current infra and the new one. Or am I misunderstanding something?

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                  adaptor.getIndices().end());
+    rewriter.replaceOpWithNewOp<xilinx::aievec::UPDOp>(
+        readOp, readOp.getVector().getType(), adaptor.getSource(), indices, 0,
+        0, TypedValue<VectorType>(nullptr));
+    return success();
+  }
+
+};
+
+template <typename SrcOpTy, typename DstOpTy>
```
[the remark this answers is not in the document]
jsetoain: Agree. Also, a more careful coding. I'll take care of this.

include/aie/Dialect/AIEVec/Pipelines/Passes.h
```diff
@@ -0,0 +1,70 @@
+//===- Passes.h - AIE Vector pipeline entry points --------------*- C++ -*-===//
```
linay-xsj, line 1: Why do you comment line 17-68?

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,244 @@
...
+struct UPDOpEffectiveAccessSizeAnalysis {
+  UPDOpEffectiveAccessSizeAnalysis(aievec::UPDOp updOp) {
+    auto vecType = cast<VectorType>(updOp.getResult().getType());
+    unsigned sizeInBits =
+        cast<ShapedType>(vecType).getSizeInBits() - updOp.getOffset();
+    for (Operation *user : updOp->getUsers()) {
+      auto userUpdOp = dyn_cast<xilinx::aievec::UPDOp>(user);
+      if (userUpdOp)
+        sizeInBits -= userUpdOp.getOffset();
+    }
```
linay-xsj, line 59: Not quite understand line 55-59 for the upd which is the user of a upd , could you please add a simple case in the comment?

test/Conversion/VectorToAIEVec/test-upd.mlir
jsetoain: I'll take care of this. The right thing to do is probably to split v16int64 ops down to two v8int64, for instance, and failing to legalize int64 for AIE1. Beware, though, current infra doesn't complain about this. These tests also pass if you use `-aie-vectorize` instead of `-convert-vector-to-aievec`.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
jsetoain: Yes, I was thinking the same.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
jsetoain: Agree.

include/aie/Dialect/AIEVec/Pipelines/Passes.h
jsetoain: We don't have a pipeline yet, this was here as a reference for me. I meant to remove all of these, this one slipped through. Good catch, thanks!

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
jsetoain: This code is a bit hackish, and it's supposed to go eventually. It's just a little trick to evaluate whether two or more UPD ops are linked, but it's just easier to generate them when lowering transfer_read. I will add an explanation, but I don't think it's worth thinking too hard about this.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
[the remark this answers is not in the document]
jsetoain: And that's fine. If the conversion is not one to one, then we add a pattern for that operation. We templatize if there's common code, but there's nothing preventing us from adding an op-specific pattern. If the same op can be lowered in two different ways depending on the specifics of the IR, you check those specifics at the start of the pattern and bail if it's the wrong case. As long as there's no overlap (i.e.: only one of the patterns can apply for any given op), you can have as many of them as you need, and they can be as generic or as specific as the use case calls for.

linay-xsj:
> I'm working on all of this, but I wanted to give you a heads up of what I'm thinking.

Thank for get this started, Javier.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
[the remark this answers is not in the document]
david-vc: The old implementation should not have allowed this either.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
jsetoain: I've added a comment for the time being. Fixing this properly will require a bit more thought (I've already started and have some code downstream).

test/Conversion/VectorToAIEVec/test-upd.mlir
jsetoain: I've added an alternative test that checks the "needs to split" case.

david-vc (approved):
Thanks!

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -0,0 +1,291 @@
...
+//===----------------------------------------------------------------------===//
+
+// Calculates the effective size of the load operation (in bits).
+// If a long UPD is followed by another one with an offset, we count
+// its effective size as the number of bits loaded up to that offset.
+// E.g.:
+//  As is, the effective size of:
+//     %0 = aievec.upd %m[%i] {index = 0 : i8, offset = 0 : si32}
+//                            : memref<256xi32>, vector<32xi32>
+//  would be `8 * sizeof(i32) * 32` (i.e: 1024 bits).
```
linay-xsj, line 64: I don't think we need 8 * here and at line 70.

PR #271, opened 2023-01-18
remove requirement absolute paths

denolf:
- avoid requiring absolute paths
- mlir-aie build script takes in build llvm folder in case other name than build was used while building llvm


stephenneuendorffer (approved):

denolf:
done.

PR #273, opened 2023-01-19
lsp-server builds

SamuelBayliss:
Initial support for building lsp-server


keryell:
Interesting.
I tried a few months ago the generic MLIR `lsp-server` in my Emacs environment but it was not even launched, while it works like a charm for me with `clangd` for C++.
I need to debug deeper my configuration when you make some progress on this PR.

keryell:
I completely forgot about this when I started https://github.com/Xilinx/mlir-aie/pull/1112 :-(

fifield:
added in https://github.com/Xilinx/mlir-aie/pull/1112

PR #274, opened 2023-01-19
Resolve build warnings.

AndraBisca:


AndraBisca:
Warnings from AIEDialect.cpp and AIETargets.cpp are already resolved here #268, @stephenneuendorffer

denolf (approved):

PR #280, opened 2023-01-20
Update requirements.txt: added cmake

denolf:
added cmake to make sure we have recent version


jgmelber (approved):
Independently tested. ✅

denolf:
merged #280

PR #283, opened 2023-01-20
Add configuration detail to the tutorial README

jgmelber:
Add a bit of detail about what mlir-aie generates for configuring the device, and how that is used in the context of the host code.


AndraBisca (approved):
A very useful addition, thank you!

PR #288, opened 2023-01-24
Allow passing locks to C code.

stephenneuendorffer:


test/unit_tests/chess_compiler_tests/01_precompiled_core_function/kernel.cc
```diff
@@ -10,13 +10,15 @@
 
 #include "kernel.h"
 
-void func(int32_t *a, int32_t *b)
+void func(int32_t *a, int32_t *b, int64_t lock)
```
keryell, line 13: Bad for the carbon footprint to have 64 bits here?

test/unit_tests/chess_compiler_tests/01_precompiled_core_function/aie.mlir
```diff
@@ -22,12 +22,12 @@ module @test_chesss_01_precompiled_core_function {
   %lock13_3 = AIE.lock(%tile13, 3) { sym_name = "input_lock" }
   %lock13_5 = AIE.lock(%tile13, 5) { sym_name = "output_lock" }
 
-  func.func private @func(%A: memref<256xi32>, %B: memref<256xi32>) -> ()
+  func.func private @func(%A: memref<256xi32>, %B: memref<256xi32>, %lock: index) -> ()
```
keryell, line 25: Perhaps `index` could be a more precise type depending on the architecture?
Or at least lowered to a more precise concrete type before calling the real AIE kernel?

test/unit_tests/chess_compiler_tests/01_precompiled_core_function/kernel.cc
stephenneuendorffer: Yes.. this is the main reason why this hasn't been checked in.  Today, it's relatively easy to lower from locks to a constant 'index' type in MLIR, but the size of the index type is something 'implementation defined' (by default 64 bits).   This really should be cleaner in the lowering to result in something more convenient.

keryell:
To be merged?

PR #299, opened 2023-01-26
[CI] Enable -Werr

stephenneuendorffer:


AndraBisca:
It might be a good idea to add this to the build scripts as well for consistency.

stephenneuendorffer:
> It might be a good idea to add this to the build scripts as well for consistency.

My understanding is that the build scripts are mostly there for people to use in their personal builds.  Are we using them in pre-commit CIs?  In this case, I'm primarily concerned with gate-keeping what is committed to the repo.

PR #311, opened 2023-01-30
ObjectFifo Host-side API (Locks, Shim)

AndraBisca:
This PR introduces symbolic names to objectFifos which are used to name the elements generated by the lowering (AIE buffers and locks). These names are further used in the case of locks to generate accessor functions that the host processor can use to acquire/release shim DMA locks generated by objectFifos without looking at the lowering.

Future improvements to this objectFifo host-side API will include generating accessor functions for the buffers generated by objectFifos, as well as for locks (other than those for the shim DMAs).


lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -844,6 +847,79 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
...
+      return;
+
+    if (row != 0)
+      return;
+
+    // naming scheme is: <objFifo_name>_<'prod' or 'cons'>_<lock_number>
+    std::string lockName(lock.name().getValue());
+    std::string prodLock = "_prod_lock_";
+    std::string consLock = "_cons_lock_";
+
```
stephenneuendorffer, line 868: can we avoid abbreviation here?

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -844,6 +847,79 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
...
+    int releaseLockValue = -1;
+
+    auto pos = lockName.find(prodLock);
+    if (pos != std::string::npos) {
+      // Producer
+      objFifoName = lockName.substr(0, pos);
+      objIndex = lockName.substr(pos + prodLock.length());
+      accessor_type = "producer";
+      acquireLockValue = 0;
+      releaseLockValue = 1;
```
stephenneuendorffer, line 882: I'm not a fan of parsing the name here.   Why isn't the full lock name put the sym_name of the lock, so that the regular handling of locks can happen at this level?

reference_designs/MM_2x2/objectFifo_circuit_switched_version/test.cpp
```diff
@@ -97,19 +97,19 @@ int main(int argc, char *argv[]) {
...
 
   mlir_aie_start_cores(_xaie);
 
   usleep(sleep_u);
 
-  mlir_aie_acquire_of_17_lock_0(_xaie, 1, 0);
-  mlir_aie_acquire_of_15_lock_0(_xaie, 1, 0);
+  mlir_aie_acquire_consumer_objFifo_out0_0(_xaie, 10000);
+  mlir_aie_acquire_consumer_objFifo_out1_0(_xaie, 10000);
   mlir_aie_sync_mem_cpu(_xaie, 6); // only used in libaiev2
```
stephenneuendorffer, line 113: I suspect what we want at this level is code a combined operation where the host pushes something into the objectFifo, rather than individual locks?

lib/Targets/AIETargetXAIEV2.cpp
AndraBisca: Yes I can change that!

lib/Targets/AIETargetXAIEV2.cpp
AndraBisca: I tried to do that initially but I hit a roadblock when passing the aie context from the function definition into the lock accessor function call.

reference_designs/MM_2x2/objectFifo_circuit_switched_version/test.cpp
AndraBisca: Yes, that is correct. Due to high demand for an API for the locks already I made this PR so that at least it won't be necessary to check the lowering every time for the shim DMA locks.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -226,9 +226,8 @@ struct AIEObjectFifoStatefulTransformPass
         BufferOp buff = builder.create<BufferOp>(builder.getUnknownLoc(),
                                                  elemType, creation_tile);
         buff.getOperation()->setAttr(
-            "sym_name",
-            builder.getStringAttr("of_" + std::to_string(of_index) + "_buff_" +
-                                  std::to_string(of_elem_index)));
+            "sym_name", builder.getStringAttr(op.name().getValue() + "_buff_" +
```
fifield, line 229: We should be using `SymbolTable::getSymbolAttrName()` instead of `"sym_name"` here and several other places.

include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -1205,6 +1205,22 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectFifo.createObjectFifo", []> {
...
+    }
+    StringAttr name() {
+      if(auto attr = getOperation()->getAttrOfType<StringAttr>(SymbolTable::getSymbolAttrName())) {
+        return attr;
+      } else {
+        emitOpError("does not have '") << SymbolTable::getSymbolAttrName() <<
+          "' attribute specified";
+      }
+      llvm_unreachable("unreachable");
+    }
```
fifield, line 1223: Can we have one function returning `Optional<StringAttr>`?  Instead of aborting the program when there's no name can we emit an error message where the error happens?  e.g. "error: could not perform X because Y is un-named"

lib/Targets/AIETargetXAIEV2.cpp
fifield: I'm also -1 on parsing the symbol name. Consider the case where a non-objfifo user has your string in their symbol name. Can you use some metadata (attribute) other than the symbol name?

AndraBisca:
Thank you for all the valuable feedback! I'll mark the PR as a draft for now until I've made a better version of it.

PR #312, opened 2023-01-30
Remove V2 column reset

jgmelber:
Removes the explicit column resets by register writes in the libXAIEV2 target. The writes are blocked by the libXAIEV2 backend for vck190, and no longer required for vck5000 after the device initialization packet implementation in MLIR-AIR.


AndraBisca (approved):
Thank you for the change! This will also resolve the errors some people have noticed when running the tutorial designs.

fifield:
Thanks for this, I kept reverting the original change to avoid the vck190 error messages :)

PR #313, opened 2023-01-31
Update tutorial Makefiles

jgmelber:
Use aie-opt installation directory to find runtime_lib installation rather than relative paths for test_library.h/cpp in tutorial Makefiles.

### DIFF
--- a/tutorials/tutorial-1/Makefile
+++ b/tutorials/tutorial-1/Makefile
@@ -3,6 +3,7 @@
 
 # TODO - Change this to env variable?
 SYSROOT = /group/xrlabs/platforms/vck190-pynq-v2.7/sysroot
+AIE_INSTALL = $(dir $(shell which aie-opt))/..
 
 all: core_1_4.elf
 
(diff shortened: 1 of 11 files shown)


AndraBisca (approved):
Definitely a welcome improvement. This might be a change we'd like to add to all of the mlir source files, not just the tutorial ones (to avoid inconsistency). That would be the files in the reference_designs/ and test/unit_tests.

jackl-xilinx:
I'm planning on using runtime lib to store some statis aiesim/Work related files. Do those get copied into install/runtime_lib as well? Or maybe I should just put those files elsewhere.

denolf (approved):

jgmelber:
You need to update the [CMakeLists.txt](https://github.com/Xilinx/mlir-aie/blob/bb3b4e8b28a8a24ea00c2a5d6cb8d621d5ce69f0/runtime_lib/CMakeLists.txt#L8-L11) file to include whatever files from runtime_lib in the installation directory.

PR #326, opened 2023-02-03
[aie][aievec] Add lowering for splat transfer read

jsetoain:
Takes a `vector.transfer_read` using a splat permutation map and replaces it with a regular `vector.transfer_read` followed by a `aievec.broadcast`. The index for the broadcast is extracted from the offset of the inner-most index of the `vector.transfer_read`.


david-vc (approved):
Hi @jsetoain : 
I understand that this is work in progress and that we'll add tests to exercise the patterns.
Thanks,
David

PR #330, opened 2023-02-07
[aievec] Convert splat reads into extract + broadcast ops

jsetoain:
In preparation to lower to aievec::mac with broadcast or aievec::broadcast (depending on the AIE version), we need to separate vector::transfer_read with a splat permutation map into a contiguous load followed by a vector::extract + vector::broadcast.

In subsequent patches, we will either fold these two into an aievec::mac or replace them with an aievec::broadcast.


david-vc (approved):
Thumb up :)

PR #332, opened 2023-02-07
[aievec][NFC] Make canonicalization pass public

jsetoain:
This patch also adds a test for the current canonicalization for AIEVec


test/dialect/AIEVec/transfer-read-splat.mlir
```diff
@@ -0,0 +1,52 @@
+// RUN: aie-opt %s -canonicalize-for-aievec -split-input-file | FileCheck %s
```
linay-xsj, line 1: Javier, could you please also add a test case for folding mul and add, and a test case for broadcastOp?

linay-xsj: Sorry, I missed one word "separately."

linay-xsj (approved):

PR #335, opened 2023-02-08
Minor fixes to tutorials and build instructions

andrej:
Just hoping to contribute something useful while I'm getting started with MLIR-AIE. I've spotted a couple typos in the first couple tutorials and clarified some wordings that weren't super clear to me. I also ran into some issues during the build process that were resolved by adding some dependencies -- I figured it might be useful for other people to have these dependencies listed right with the build instructions.

Happy to incorporate any feedback.


stephenneuendorffer (approved):

stephenneuendorffer:
Thanks Andre!

PR #338, opened 2023-02-08
[AIE][AIEVec]Add lowering for FMA_Elem

linay-xsj:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -89,6 +113,115 @@ struct UPDOpEffectiveAccessSizeAnalysis {
 //===----------------------------------------------------------------------===//
 // Lowering patterns
 //===----------------------------------------------------------------------===//
+// This pattern replaces `vector.fma` with `aievec.mac_elem`.
```
david-vc, line 116: 1. This description is insufficient. 
2. The requirement that one of the operands of the vector.fma is a vector.broadcast is unnecessary.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -89,6 +113,115 @@ struct UPDOpEffectiveAccessSizeAnalysis {
...
+    auto pos = extOp.getPosition();
+
+    Value rhsVal = rewriter.create<aievec::BroadcastOp>(
+        fmaOp->getLoc(), resultType, newRhs,
+        cast<IntegerAttr>(pos[0]).getInt());
+
+    auto fmaElemOp = rewriter.create<aievec::FMAElemOp>(
+        fmaOp->getLoc(), lhs, rhsVal, acc, /*fmsub=*/false);
+
+    for (auto user : fmaOp->getUsers()) {
```
david-vc, line 175: IMO, this pattern should replace the vector.fma with a elem.mac_elem if the elem type of the mac_elem and the fma are the same.
Otherwise, if the elem type of the mac_elem is wider than that of the vector.fma, it should replace the vector.fma with a mac_elem followed by an SRS.
The pattern should not check the users of the vector.fma; this is not relevant.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -369,6 +502,8 @@ static void populateAIEVecV2ConversionPatterns(RewritePatternSet &patterns,
                                                AnalysisManager &am) {
   patterns.add<LowerVectorTransferReadToAIEUPD, SplitUPDOpOnAccPattern>(
       patterns.getContext(), am, 512);
+
+  patterns.add<LowerVectorFMAToAIEVecFMAElem>(patterns.getContext());
```
david-vc, line 506: It seems to me that this pattern should be enabled only for AIE-ML.

david-vc (changes requested):

test/Conversion/VectorToAIEVec/test_mac_elem.mlir
```diff
@@ -0,0 +1,217 @@
+// RUN: aie-opt %s --convert-vector-to-aievec="aie-target=aieml" | FileCheck %s
+
+module {
```
david-vc, line 3: IMO this example should be cut down. For example you could change the code so that only 2 products are summed in the inner loop. The transform should apply equally.

`
module {
  func.func @matmul(%arg0: memref<64x64xi32>, %arg1: memref<64x64xi32>, %arg2: memref<64x64xi32>) {
    %c0_i32 = arith.constant 0 : i32
    affine.for %arg3 = 0 to 64 {
      affine.for %arg4 = 0 to 64 step 16 {
        affine.for %arg5 = 0 to 64 step 16 {
          %0 = vector.transfer_read %arg0[%arg3, %arg5], %c0_i32 {in_bounds = [true], permutation_map = affine_map<(d0, d1) -> (0)>} : memref<64x64xi32>, vector<16xi32>
          %1 = vector.transfer_read %arg1[%arg5, %arg4], %c0_i32 {in_bounds = [true]} : memref<64x64xi32>, vector<16xi32>
          %2 = arith.muli %0, %1 : vector<16xi32>
          %3 = vector.transfer_read %arg2[%arg3, %arg4], %c0_i32 {in_bounds = [true]} : memref<64x64xi32>, vector<16xi32>
          %4 = arith.addi %3, %2 : vector<16xi32>
          %5 = affine.apply affine_map<(d0) -> (d0 + 1)>(%arg5)
          %6 = vector.transfer_read %arg0[%arg3, %5], %c0_i32 {in_bounds = [true], permutation_map = affine_map<(d0, d1) -> (0)>} : memref<64x64xi32>, vector<16xi32>
          %7 = affine.apply affine_map<(d0) -> (d0 + 1)>(%arg5)
          %8 = vector.transfer_read %arg1[%7, %arg4], %c0_i32 {in_bounds = [true]} : memref<64x64xi32>, vector<16xi32>
          %9 = arith.muli %6, %8 : vector<16xi32>
          %10 = arith.addi %4, %9 : vector<16xi32>
          vector.transfer_write %94, %arg2[%arg3, %arg4] {in_bounds = [true]} : vector<16xi32>, memref<64x64xi32>
        }
      }
    }
    return
  }
}
`

david-vc (changes requested):
@linay-xsj : 
IMO, it would be better if LowerVectorFMAToAIEVecFMAElem would **not** require that one of the operands is a broadcast.
Also, please try to reduce the size of the test case. instead of summing 16 products in the inner loop, you can cut down the example to sum just 2 products.

david-vc (approved):
Thanks for addressing my earlier concerns.

PR #344, opened 2023-02-09
[tutorial] typo fix in `tutorial-10`

cabreraam:
Just reading through one of the tutorials, and I think the `.mlir` extension might have been dropped on this step.


jackl-xilinx:
Thanks for the correction!

PR #349, opened 2023-02-10
horizontal_diff

arkhodamoradi:


denolf (changes requested):
Thanks Ali. Can you review all commented out code so that comments are really used only as comments.

PR #361, opened 2023-02-14
Fix platform and tutorial Makefiles

andrej:
Executive summary: These changes should allow an outside user that strictly follows the "Getting Started" instructions in the documentation to eventually build and simulate the tutorials in the `tutorials` folder. For this, some hard-coded paths, typos, and missing search paths had to be fixed in several Makefiles. I have tested and suceeded at building all tutorials except 2c (because I don't have the physical hardware available) on a fairly bare-bones Ubuntu 20.04.

Note: This changes the _MLIR_AIE_SYSROOT_ in the tutorial Makefiles. Previously, this path was set to `/group/xrlabs/platforms/vck190-pynq-v2.7/sysroot` -- I assume this is a Xilinx-internal path, that most outside users that are getting started with the tutorials would not have. If you would like to build the tutorials using this previous value for the sysroot path, it is still possible, but you will now have to call make as `MLIR_AIE_SYSROOT=/group/xrlabs/platforms/vck190-pynq-v2.7/sysroot make`, not just `make`.

I think this change is justified, since it will make it easier for most users who follow the instructions in the documentation, by using the sysroot installed by following the instructions in `Platform.md`. (That sysroot will be installed in a subdirectory of `/platforms/vck190_bare_prod` when calling `make` from there.)

------

More details (to elaborate a little in the changes of the commits):

Building the sysroot in `platforms/vck190_bare_prod` and any of the tutorials in `tutorials` did not succeed for me on a fresh Ubuntu 20.04 install. 

The following were the issues stopped me from building the platform, specifically `make aie_platform_build` in `platforms/vck190_bare_prod` was broken:

1. A typo (missing letter A) for the sysroot path
2. A guard for an undefined variable that is not actually used anywhere (PLATFORM_REPO_PATHS, seems to be a remnant of previous versions of this file)
3. A hard-coded path to a user employee (`/group/xrlabs2/jackl ...`) that will not be present on users systems
4. At least on my system, clang++ was not able to find the necessary C++ includes and libraries without some help, so I added some `-I` and `-L` flags; these should be unproblematic on a system where the files are located elsewhere, but can be found by clang, since those flags just append to the search path.

With these issues fixed, I was able to move on to compiling the tutorials and simulating them on my machine. Again, for cross-compilation, my version of clang++ did not find the includes and libraries, so I added some flags for the appropriate search paths. These search paths are set up such that a PetaLinux sysroot version 2021.1, which includes libstdc++ 10.2.0, should work.

I also took the liberty of reorganizing the makefiles so they would be a little easier to read by moving redundant/repeated definitions into a separate new file, `tutorials/makefile-common`.

Finally, I also took the liberty of a miniscule change in `utils/env_setup.sh`. Since this script will mostly be `source`d, calling `exit` in there will quit an user's terminal if they accidentally run the script without arguments, before they can read the error message. I think a `return` is more appropriate -- it quits the script, shows the user the error message, but allows the user to re-try in the same terminal.

----

I'm very happy about any feedback. Please feel free to disregard this if I clearly misunderstood something. This is simply the only way I could get things to run, from an outsider perspective.

### DIFF
--- a/platforms/vck190_bare_prod/aie_platform/Makefile
+++ b/platforms/vck190_bare_prod/aie_platform/Makefile
@@ -11,18 +11,9 @@
 #limitations under the License.
 #
 
-#SYSROOT = /proj/xbuilds/2021.1_daily_latest/internal_platforms/sw/versal/xilinx-versal-common-v2021.1/sysroots/aarch64-xilinx-linux/
-#IMAGE = /proj/xbuilds/2021.1_daily_latest/internal_platforms/sw/versal/xilinx-versal-common-v2021.1/Image
-#ROOTFS = /proj/xbuilds/2021.1_daily_latest/internal_platforms/sw/versal/xilinx-versal-common-v2021.1/rootfs.ext4
-#CXX = aarch64-linux-gnu-g++
-
-#SYSROOT = /group/xrlabs2/jackl/proj/acdc/vitis_2021.2_bram_platform_prod/sysroots/cortexa72-cortexa53-xilinx-linux
-#IMAGE = /group/xrlabs2/jackl/proj/acdc/vitis_2021.2_bram_platform_prod/build/petalinux/images/linux/Image
-#ROOTFS = /group/xrlabs2/jackl/proj/acdc/vitis_2021.2_bram_platform_prod/build/petalinux/images/linux/rootfs.ext4
-
-SYSROOT = ../petalinux/build/petalinux/images/linux/sdk/sysroots/cortexa72-cortex53-xilinx-linux
-IMAGE   = ../petalinux/build/petalinux/images/linux/Image
-ROOTFS  = ../petalinux/build/petalinux/images/linux/rootfs.ext4
+SYSROOT ?= ../petalinux/build/petalinux/images/linux/sdk/sysroots/cortexa72-cortexa53-xilinx-linux
+IMAGE    = ../petalinux/build/petalinux/images/linux/Image
+ROOTFS   = ../petalinux/build/petalinux/images/linux/rootfs.ext4
 
 
 #------------------------------------------------------------------------------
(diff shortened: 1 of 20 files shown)


jackl-xilinx:
@andrej Thanks for making these much need consistency changes. Can you also apply the Makefile updates to the subfolders in the different tutorials (e.g. tutorial-3/objectFifo, tutorial-4/flow, tutorial-4/switchbox, etc)? Thanks!

andrej:
@jackl-xilinx Thanks for the quick response! I have applied the changes to the sub-folders as well, my apologies for that oversight. I've rebased on current main so merging should be easy. Please let me know if you would like me to squash the commits.

All the sub-folders compile as well, except for `tutorial-7/switchbox`, which gives me below attached error. I do not think this error is related to my changes, as it is compiled with the same flags:

```
aie.mlir:45:5: error: custom op 'AIE.broadcast_packet' is unknown
    AIE.broadcast_packet(%tile14, DMA: 0) {
    ^
Error encountered while running: aie-opt --lower-affine --aie-assign-lock-ids 
--aie-register-objectFifos --aie-objectFifo-stateful-transform 
--aie-lower-broadcast-packet --aie-create-packet-flows --aie-lower-multicast 
--aie-assign-buffer-addresses -convert-scf-to-cf aie.mlir -o 
acdc_project/input_with_addresses.mlir
```

This is with MLIR-AIE built early yesterday, so if this is a change that has occured/been fixed since, I didn't see it.

jgmelber:
The broadcast_packet OP is now in the AIEX dialect as of #248

jackl-xilinx (approved):
These changes look good. Thanks for organizing and consolidating the env variables.

jackl-xilinx:
I'm going to leave tutorial-7/switchbox failing. It does need to be moved to AIEX but the entire design actually needs to be finished as it was copied from tutorial-7/flow and had not been completed. I'll fix that along with making it complete on the next commit.

andrej:
Sounds good. Thank you! Have a good evening.

PR #364, opened 2023-02-14
Fix aiecc to not hard code project directory

fifield:
The hard coded directory name breaks mlir-air.


fifield:
There's a similar issue with the `sim` directory.

stephenneuendorffer (approved):

PR #365, opened 2023-02-14
Add redundant read and store optimization pass

linay-xsj:


david-vc (commented):
Looks good to me. Please also seek feedback from Javier.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -635,8 +635,29 @@ populateAIEv1CanonicalizeConversionPatterns(RewritePatternSet &patterns) {}
...
+
+  funcOp.walk([&](Operation *Op) {
+    if (auto readOp = dyn_cast<TransferReadOp>(Op)) {
+      if (!readOp.getInBounds())
+        setInBounds<TransferReadOp>(readOp);
+    } else if (auto writeOp = dyn_cast<TransferWriteOp>(Op)) {
+      if (!writeOp.getInBounds())
+        setInBounds<TransferWriteOp>(writeOp);
+    }
+  });
```
jsetoain, line 657: Navigating the code ourselves and making in-place modifications leads to bad failure modes and potentially incorrect transformations. If we need to do this, it should be done in its own "TransferReadOp"-bound rewrite/conversion. For something of this nature (marking transfer_reads as in-bound), a proper pre-analysis is probably the right approach.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -635,8 +635,29 @@ populateAIEv1CanonicalizeConversionPatterns(RewritePatternSet &patterns) {}
...
+
+  transferOpflowOpt(funcOp);
```
jsetoain, line 659: This would fit better within its own pass, rather than an annex to this one.

test/Conversion/VectorToAIEVec/test_redundant_store.mlir
```diff
@@ -0,0 +1,59 @@
...
+          %13 = affine.load %arg2[%arg3, %arg4] : memref<64x64xi32>
+          %14 = arith.addi %13, %12 : i32
+          affine.store %14, %arg2[%arg3, %arg4] : memref<64x64xi32>
+        }
+      }
+    }
+    return
+  }
+}
+
```
jsetoain, line 31: As a test for "redundant store elimination", this test is too complex and probably in the wrong place. This should be testable on its own, and since it's not really testing "vector to aievec", probably living in the Dialect test suite. Of course, we should also test this end-to-end, but those should live in the Integration test suite, and this test (gemm) is probably already there.

jsetoain (changes requested):
There's still a bit of the "old approach" in this code. Other than that, I see we're really close to matching the aie-vectorize output for gemm, nice job! 😀

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -684,6 +706,8 @@ void xilinx::aievec::buildConvertVectorToAIEVec(
   // Add post-lowering canonicalization passes
   pm.addPass(createCanonicalizerPass());
   pm.addPass(createCSEPass());
+  pm.addPass(createLoopInvariantCodeMotionPass());
```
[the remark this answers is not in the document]
linay-xsj: We still need this pass to hoist invariant out of the loop, for example %3 in test/unit_tests/aievec_tests/i32xi32_gemm/gemm64_int_unroll16.mlir will be hoisted out before the innermost loop.

test/Conversion/VectorToAIEVec/test_mac_elem.mlir
```diff
@@ -29,19 +29,23 @@ module {
...
+//      CHECK:          %[[T3:.*]] = aievec.broadcast %[[T2:.*]] {idx = 0 : i8} : vector<16xi32>, vector<16xi32>
+//      CHECK:          %[[T4:.*]] = aievec.upd %[[A1]][%[[A5:.*]], %[[A4:.*]]] {index = 0 : i8, offset = 0 : si32} : memref<64x64xi32>, vector<16xi32>
+//      CHECK:          %[[T5:.*]] = aievec.mac_elem %[[T3:.*]], %[[T4:.*]], %[[T1:.*]] : vector<16xi32>, vector<16xi32>, vector<16xi64>
+//      CHECK:          %[[T6:.*]] = aievec.broadcast %[[T2:.*]] {idx = 1 : i8} : vector<16xi32>, vector<16xi32>
+//      CHECK:          %[[T7:.*]] = arith.addi %[[A5:.*]], %[[C1:.*]] : index
+//      CHECK:          %[[T8:.*]] = aievec.upd %[[A1]][%[[T7:.*]], %[[A4:.*]]] {index = 0 : i8, offset = 0 : si32} : memref<64x64xi32>, vector<16xi32>
+//      CHECK:          %[[T9:.*]] = aievec.mac_elem %[[T6:.*]], %[[T8:.*]], %[[T5:.*]] : vector<16xi32>, vector<16xi32>, vector<16xi64>
+//      CHECK:          %[[T10:.*]] = aievec.srs %[[T9:.*]] {shift = 0 : i8} : vector<16xi64>, vector<16xi32>
+//      CHECK:          vector.transfer_write %[[T10:.*]], %[[A2]][%[[A3:.*]], %[[A4:.*]]] {in_bounds = [true]} : vector<16xi32>, memref<64x64xi32>
```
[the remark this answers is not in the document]
linay-xsj: We also want to test whether broadcast index is correct here, so I use two sets of mul and add to test the correctness.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -668,11 +692,54 @@ void CanonicalizeForAIEVecPass::runOnOperation() {
...
+  void runOnOperation() override;
+};
+
+static void
+configureRedundantLoadStoreOptimizationLegalizations(ConversionTarget &target) {
+  target.addLegalDialect<vector::VectorDialect>();
+  target.addDynamicallyLegalOp<TransferReadOp>(
+      [](vector::TransferReadOp op) { return !(!op.getInBounds()); });
+  target.addDynamicallyLegalOp<TransferWriteOp>(
+      [](vector::TransferWriteOp op) { return !(!op.getInBounds()); });
```
jsetoain, line 709: As a rewrite pass, we wouldn't need legality checks.

jsetoain (commented):
It's looking much better. Thank you, Lina 🙂

test/Conversion/VectorToAIEVec/test_redundant_store.mlir
```diff
@@ -0,0 +1,48 @@
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" --convert-vector-to-aievec="aie-target=aieml" -lower-affine -canonicalize | FileCheck %s
```
jsetoain, line 1: We don't need to test the optflow optimization, it's already being tested in LLVM (mlir/test/Dialect/Vector/vector-transferop-opt.mlir), together with more complex cases.

test/unit_tests/aievec_tests/i16xi16_gemm/gemm64_int16_unroll32.mlir
```diff
@@ -1,5 +1,6 @@
 // REQUIRES: valid_xchess_license
 // RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=32" -aieml=true --aie-vectorize | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=32" -redundant-load-store-optimization --convert-vector-to-aievec="aie-target=aieml" -lower-affine | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
```
jsetoain, line 3: Notice that "-redundant-load-store-optimization" doesn't exist any more, since it's a private pass. Trying to invoke the pass from the command line should fail. For some reason, though, if you run "check-aie" it doesn't, which makes me think this test is not being run, for some reason. Could you look into it?

test/unit_tests/aievec_tests/i32xi32_gemm/gemm64_int_unroll16.mlir
```diff
@@ -1,5 +1,6 @@
 // REQUIRES: valid_xchess_license
 // RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" -aieml=true --aie-vectorize | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" -redundant-load-store-optimization --convert-vector-to-aievec="aie-target=aieml" -lower-affine | aie-translate -aieml=true --aievec-to-cpp -o gen_aie-ml.cc
```
jsetoain, line 3: Same here, there's something odd going on with this test.

jsetoain (changes requested):

test/Conversion/VectorToAIEVec/test_redundant_store.mlir
linay-xsj: Not sure what you mean here. How do you want me to test this test case.

test/unit_tests/aievec_tests/i16xi16_gemm/gemm64_int16_unroll32.mlir
linay-xsj: Sorry, I forgot to add the changes for these two unit test cases.

test/Conversion/VectorToAIEVec/test_broadcast.mlir
```diff
@@ -0,0 +1,10 @@
+// RUN: aie-opt %s --convert-vector-to-aievec="aie-target=aieml" | FileCheck %s
+
+// CHECK-LABEL: func @vector_extract_broadcast_to_aievec(
+// CHECK-SAME: %[[A:[A-Za-z0-9]+]]: vector<16xi32>
+func.func @vector_extract_broadcast_to_aievec(%a : vector<16xi32>) -> vector<16xi32> {
+  // CHECK: aievec.broadcast %[[A]] {idx = 0 : i8} : vector<16xi32>, vector<16xi32>
+  %0 = vector.extract %a[0] : vector<16xi32>
+  %1 = vector.broadcast %0 : i32 to vector<16xi32>
+  return %1 : vector<16xi32>
+}
```
jsetoain, line 10: Maybe add a second test for non-zero index, like so:
```suggestion
func.func @vector_extract_broadcast_to_aievec(%a : vector<16xi32>) -> (vector<16xi32>, vector<16xi32>) {
  // CHECK: aievec.broadcast %[[A]] {idx = 0 : i8} : vector<16xi32>, vector<16xi32>
  %0 = vector.extract %a[0] : vector<16xi32>
  %1 = vector.broadcast %0 : i32 to vector<16xi32>
  // CHECK: aievec.broadcast %[[A]] {idx = 2 : i8} : vector<16xi32>, vector<16xi32>
  %2 = vector.extract %a[2] : vector<16xi32>
  %3 = vector.broadcast %2 : i32 to vector<16xi32>
  return %1, %3 : vector<16xi32>, vector<16xi32>
}
```

jsetoain (approved):

jsetoain (approved):

PR #367, opened 2023-02-17
Updates for AIE2

stephenneuendorffer:
Most of this is just removing AIE1 assumptions at this point.


jackl-xilinx:
Seems reasonable. Are you adding an option to xchesscc_wrapper to choose between aie1 and aie2? That's a local change I had to make so it target aie2.

stephenneuendorffer:
> Seems reasonable. Are you adding an option to xchesscc_wrapper to choose between aie1 and aie2? That's a local change I had to make so it target aie2

My plan was something like that.  Most of this is just refactoring to isolate such changes.

PR #373, opened 2023-02-28
Aiesim install path fix

denolf:


denolf:
More changes included then intended, please ignore

PR #375, opened 2023-02-28
[aievec] Remove muli+addi -> vector.fma pattern

jsetoain:
https://reviews.llvm.org/D141711 upstream disallows `vector.fma` for integral types. In order to work around this, we need to merge the identification of `arith.muli` + `arith.addi` into the patterns that take `vector.fma` and lower it to the equivalent operation in `aievec`.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
+++ b/lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
@@ -55,30 +55,34 @@ using namespace xilinx::aievec;
 
 #define DEBUG_TYPE "aievec-lowering"
 
-// Given an AIEOp, determines if an operation writes to an accumulator
-// based on operation type and operand types
-static bool writesToAccumulator(Operation *op) {
-  // Integer muls and FMAs write to accumulator
-  if (!op || !isAIEOp(op))
-    return false;
-  if (auto mulOp = dyn_cast<aievec::MulOp>(op))
-    return mulOp.getResult()
-        .getType()
-        .cast<VectorType>()
-        .getElementType()
-        .isa<IntegerType>();
-  if (auto fmaOp = dyn_cast<aievec::FMAOp>(op))
-    return fmaOp.getResult()
-        .getType()
-        .cast<VectorType>()
-        .getElementType()
-        .isa<IntegerType>();
+//===----------------------------------------------------------------------===//
+// Utility functions
+//===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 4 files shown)


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -55,30 +55,34 @@ using namespace xilinx::aievec;
...
+// defined by an `arith::MulIOp` and, therefore, the `lhs`, `rhs`, and `acc`
+// operands of the MAC operation that can replace them.
+static bool extractMACOperandsFromAddOperands(Value addLhs, Value addRhs,
+                                              Value &macLhs, Value &macRhs,
+                                              Value &macAcc) {
+  auto lhsDefOp = addLhs.getDefiningOp();
+  auto rhsDefOp = addRhs.getDefiningOp();
+  arith::MulIOp mulOp = nullptr;
+  if (lhsDefOp) {
+    mulOp = dyn_cast<arith::MulIOp>(lhsDefOp);
```
linay-xsj, line 72: I found a bug here when I tested conv2d case. We need to test mul at rhs first. like -
if (rhsDefOp) {
    mulOp = dyn_cast<arith::MulIOp>(rhsDefOp);
    macAcc = addLhs;
  }
  if (!mulOp && lhsDefOp) {
    mulOp = dyn_cast<arith::MulIOp>(lhsDefOp);
    macAcc = addRhs;
  }
In the test case 
c = a[0] * b[0];
d = a[1] * b[1];
e = c + d;
we need to transform it to 
c = a[0] * b[0]; 
e = c + a[1] * b[1]
if we test whether lhs is a mul first, the result will be
d = a[1] * b[1];
e = a[0] * b[0] + d;
we want to keep the array with less index at the front.

jsetoain: Something must be off in my environment, I didn't see any tests failing. Can you point me to the test file, please? If it needs to happen in a specific order, we have to do it properly, We shouldn't rely on the operations being in a specific IR order.

linay-xsj: Yes, we need to reassoicate add and mul. You can try aie-opt conv2d_uij_i8_noinit.mlir -affine-super-vectorize="virtual-vector-size=32" --convert-vector-to-aievec -lower-affine under test/unit_tests/aievec_tests/i8xi8_static_sized_memref_aie-ml

PR #386, opened 2023-03-14
[aievec] Add lhs concat when lowering to mac

jsetoain:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -301,12 +301,12 @@ struct FoldBroadcastToFMAOp : public OpConversionPattern<aievec::FMAOp> {
...
 
-    auto newLhs = extOp.getVector();
+    auto lhs = extOp.getVector();
     // XXX: We assume a 1D vector
     auto pos = extOp.getPosition();
     int64_t zstart = cast<IntegerAttr>(pos[0]).getInt();
     rewriter.replaceOpWithNewOp<aievec::FMAOp>(
-        fmaOp, fmaOp.getResult().getType(), newLhs, rhs, adaptor.getAcc(),
+        fmaOp, fmaOp.getResult().getType(), lhs, rhs, adaptor.getAcc(),
         /*xstart =*/"0", /*xoffsets =*/"0x76543210", adaptor.getXoffsetsHi(),
```
linay-xsj, line 310: We would better to generate offsets and compute start by a function due to different data type, because i8 and i16 types would have different xoffsets from i32 type. It seems we need to get xstart, xoffset, xoffset_hi, xstep, xsquare, zstart, zoffset, zoffset_hi, zstep, zsquare when generate FMAOp. The order here looks not correct to me.

linay-xsj:
Hi, Javier, the change set looks good. Could you please also add a command line in a test case in /test/unit_tests/aievec_tests to verify the correctness?

jsetoain:
> Hi, Javier, the change set looks good. Could you please also add a command line in a test case in /test/unit_tests/aievec_tests to verify the correctness?

Hi Lina, I'm working on that and a couple of extra tests as well. I still haven't figured out how to push changes to my branch downstream without updating the PR 😅

linay-xsj:
Hi, Javier, 
     The attributes pattern of mac for splat are correct, because it concats a zero vector for lhs which I did not noticed. Please add a unit test case for i32 type and you can use the testbench under mlir-aie/test/unit_tests/aievec_tests/i32xi32_mul_elem for your i16 and i32 unit tests. Thanks!

PR #387, opened 2023-03-14
Move aiecc.py implementation to library

fifield:
This allows access to aiecc from other python. It also removes the `AIE_ENABLE_BINDINGS_PYTHON` flag (it's always on).

for example,
```
>>> import aie.compiler.aiecc.main
>>> aie.compiler.aiecc.cl_arguments.parse_args(['-h'])
usage: aiecc [-h] [--sysroot sysroot] [--tmpdir tmpdir] [-v] [--vectorize] [--xbridge] [--no-xbridge] [--aiesim] [--no-aiesim] [--xchesscc] [--no-xchesscc]
             [--peano PEANO_INSTALL_DIR] [--compile] [--no-compile] [--host-target HOST_TARGET] [--aie-target AIE_TARGET] [--compile-host] [--no-compile-host] [--link]
             [--no-link] [--aie-generate-xaiev2] [--aie-generate-xaie] [-j NTHREADS] [--profile] [--unified] [--no-unified] [-n] [--progress]
             file ...

positional arguments:
  file                  MLIR file to compile
  arm_args              arguments for ARM compiler

optional arguments:
  -h, --help            show this help message and exit
  --sysroot sysroot     sysroot for cross-compilation
  --tmpdir tmpdir       directory used for temporary file storage
  -v                    Trace commands as they are executed
  --vectorize           Enable MLIR vectorization
  --xbridge             Link using xbridge
  --no-xbridge          Link using peano
  --aiesim              Generate aiesim Work folder
  --no-aiesim           Do no generate aiesim Work folder
  --xchesscc            Compile using xchesscc
  --no-xchesscc         Compile using peano
  --peano PEANO_INSTALL_DIR
                        Root directory where peano compiler is installed
  --compile             Enable compiling of AIE code
  --no-compile          Disable compiling of AIE code
  --host-target HOST_TARGET
                        Target architecture of the host program
  --aie-target AIE_TARGET
                        Target architecture of the AIE program
  --compile-host        Enable compiling of the host program
  --no-compile-host     Disable compiling of the host program
  --link                Enable linking of AIE code
  --no-link             Disable linking of AIE code
  --aie-generate-xaiev2
                        Generate libxaie v2 drivers (default is v2)
  --aie-generate-xaie   Generate libxaie v1 drivers (default is v2)
  -j NTHREADS           Compile with max n-threads in the machine (default is 1). An argument of zero corresponds to the maximum number of threads on the machine.
  --profile             Profile commands to find the most expensive executions.
  --unified             Compile all cores together in a single process
  --no-unified          Compile cores independently in separate processes
  -n                    Disable actually executing any commands.
  --progress            Show progress visualization
```


fifield:
This also adds a `aie.compiler.aiecc.configure.install_path()` function to replace the ad-hoc use `os.path.realpath(__file__))` in aiecc to find the build or install location.  `os.path.realpath(__file__))` stopped working when `aiecc/main.py` was moved to the python library because the python files are soft links in the build area which means that during testing `os.path.realpath(__file__))` gives a location in the source tree instead of in the build tree which results in failing tests. The `install_path` function lives in configure.py which is a generated file living in the build tree so it doesn't suffer the same problem.

CMakeLists.txt
```diff
@@ -102,37 +102,30 @@ add_subdirectory(docs)
 add_dependencies(docs mlir-doc)
 
 # python install directory
-if (AIE_ENABLE_BINDINGS_PYTHON)
```
stephenneuendorffer, line 105: I'm a little nervous about python being a hard dependency here.  MLIR practice has been for that to be optional, if only because getting it to build has lots of pitfalls.

python/aie/compiler/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
```
stephenneuendorffer, line 2: Everything else in aie is apache licensed and the recommendation we've had is to not change licenses.  I think not mixing is also a good idea, unless there is a particular reason.

CMakeLists.txt
stephenneuendorffer: I guess one advantage of requiring python bindings is that aiecc.py could use them to walk the IR, rather than relying on aie-translate in a few places to extract information.

stephenneuendorffer (approved):

keryell:
Status of this PR?

PR #392, opened 2023-03-17
hierarchical toolchain and other fixes

denolf:
* added submodule for https://github.com/Xilinx/cmakeModules and inherit its generic cross compile toolchain files to temporary work around pynq or Vitis specific sysroot structures
* Relying on VitisSysroot variable to restrict defaulting to Vitis Sysroot (see updated comment in top-level CMakeList.txt file). If we want to maximize out of the box lit compile test, we currently need a sysroot since many lit tests hard code aarch64 as target. The defaults in the makefiles for the tutorial also do this. So this PR is an intermediate step.
* Fix to respect -DLibXAIE_DIR


utils/build-mlir-aie.sh
```diff
@@ -40,11 +42,12 @@ cmake -GNinja \
     -DCMAKE_CXX_COMPILER=clang++ \
     -DLLVM_DIR=${LLVM_BUILD_DIR}/lib/cmake/llvm \
     -DMLIR_DIR=${LLVM_BUILD_DIR}/lib/cmake/mlir \
-    -DCMAKE_MODULE_PATH=${CMAKEMODULES_DIR}/ \
+    -DCMAKE_MODULE_PATH=${CMAKEMODULES_DIR}/modulesXilinx \
     -DCMAKE_INSTALL_PREFIX="../${INSTALL_DIR}" \
     -DCMAKE_BUILD_TYPE=Release \
     -DLLVM_ENABLE_ASSERTIONS=ON \
     -DAIE_ENABLE_BINDINGS_PYTHON=ON \
+    -DVitisSysroot="" \
```
AndraBisca, line 50: We should change the Building on X86 instructions section, maybe add a section at the beginning, to explain the possible build scenarios based on combinations of using(or not) a sysroot and Vitis. Depending on what exactly you want to build mlir-aie for you don't require Vitis or you don't require a sysroot but there are combinations for which the lit tests will fail.

That should also make it easier for people starting with mlir-aie to know what to expect and which prerequisites they really need.

AndraBisca (approved):
I think these changes add more flexibility to the current build path, i.e., Vitis isn't a hard requirement anymore and we can now point to a specific sysroot . Apart from my comments, I think this looks good!

utils/build-mlir-aie.sh
denolf: yes indeed, I was waiting for some feedback on this PR before touching the readme

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "platforms/boards"]
 	path = platforms/boards
 	url = https://github.com/Xilinx/XilinxBoardStore.git
+[submodule "cmake/modulesXilinx"]
```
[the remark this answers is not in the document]
denolf: the cmake/modules folder should contain all necessary .cmake files specific for the mlir-aie project
the cmake/modulesXilinx is the subfolder where we clone the more generic .cmake files

utils/clone-llvm.sh
```diff
@@ -14,9 +14,6 @@
 
 export commithash=5e111eb275eee3bec1123b4b85606328017e5ee5
 
-git clone --depth 1 https://github.com/Xilinx/cmakeModules cmakeModules/cmakeModulesXilinx
```
keryell, line 17: The root cause is that this should have used https://cmake.org/cmake/help/latest/module/FetchContent.html instead of this ad-hoc solution.

denolf:
merged #392

utils/build-mlir-aie.sh
fifield: the comments in the util scripts could also use an update after this pr

Issue #396, opened 2023-03-24
Some AIE Dialect verifiers call "assert" to check correctness... should return failure() to enable graceful shutdown. 


SamuelBayliss:
Code like AIEDialect.cpp 

xilinx::AIE::CoreOp::verify()

AndraBisca:
Fix proposed in PR #397.

There are other Transforms in mlir-aie which also rely on the usage of assert but the change to a more graceful failure is less obvious so I will leave that for future PRs/issues. I am at least currently working on refactoring the ObjectFifo related passes to correct this. Other passes are:
- AIECreatePacketFlows.cpp
- AIECreatePathfindFlows.cpp
- AIEPathfinder.cpp

PR #397, opened 2023-03-25
Small fixes in AIEDialect.cpp

AndraBisca:
In response to issue #396 this PR replaces asserts in AIEDialect.cpp with emitOpError. This is also applied to one of the Transforms, namely AIECoreToStandard.cpp, where the change is trivial.

Additionally, some of the verify() methods in AIEDialect.cpp didn't correctly check if the region of an operation was empty, as these operations had implicit terminator ops added to them in AIE.td. This PR addresses that as well.


fifield (changes requested):
Why do we disallow empty swithboxes?  They are produced by at least on mlir-aie pass. After building this PR I cannot run aiecc.py:
```
$ aiecc.py -v --sysroot ../../../../DockerArm/sysroot --host-target x86_64-amd-linux-gnu --tmpdir air_project/segment_0 --aie-generate-xaiev2 --no-aiesim --no-xbridge --no-xchesscc air_project/aiecc.segment_0.mlir
Found Vitis at /opt/Xilinx/Vitis/2021.2

compiling air_project/aiecc.segment_0.mlir
created temporary directory air_project/segment_0
aie-opt --lower-affine --aie-canonicalize-device --aie-assign-lock-ids --aie-register-objectFifos --aie-objectFifo-stateful-transform 
--aie-lower-broadcast-packet --aie-create-packet-flows --aie-lower-multicast --aie-assign-buffer-addresses -convert-scf-to-cf 
air_project/aiecc.segment_0.mlir -o air_project/segment_0/input_with_addresses.mlir
 MLIR compilation: ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:-- 0:00:00 0/1 1 Worker<unknown>:0: error: 'AIE.switchbox' op cannot have an empty body
<unknown>:0: note: see current operation: 
%1 = "AIE.switchbox"(%0) ({
  "AIE.end"() : () -> ()
}) : (index) -> index
Done in 0.050 sec: aie-opt --lower-affine --aie-canonicalize-device --aie-assign-lock-ids --aie-register-objectFifos 
--aie-objectFifo-stateful-transform --aie-lower-broadcast-packet --aie-create-packet-flows --aie-lower-multicast --aie-assign-buffer-addresses 
-convert-scf-to-cf air_project/aiecc.segment_0.mlir -o air_project/segment_0/input_with_addresses.mlir
Error encountered while running: aie-opt --lower-affine --aie-canonicalize-device --aie-assign-lock-ids --aie-register-objectFifos 
--aie-objectFifo-stateful-transform --aie-lower-broadcast-packet --aie-create-packet-flows --aie-lower-multicast --aie-assign-buffer-addresses 
-convert-scf-to-cf air_project/aiecc.segment_0.mlir -o air_project/segment_0/input_with_addresses.mlir
 Error ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% -:--:-- 0:00:00 0/1 1 Worker

**Comment by AndraBisca** [comment]
I removed the changes to the Switchbox ops verifiers. The initial purpose of the PR was to reduce uses of assert() so I want to keep it that. 

Thank you @fifield for your help! Could you please review the PR again?

**Review by fifield — APPROVED** [review]

File: lib/Dialect/AIE/IR/AIEDialect.cpp [hunk 1]
```diff
@@ -808,10 +815,10 @@ LogicalResult xilinx::AIE::BufferOp::verify() {
 
 // MemOp
 LogicalResult xilinx::AIE::MemOp::verify() {
+  Region &body = getBody();
   DenseSet<xilinx::AIE::DMAChannel> used_channels;
-
-  assert(getOperation()->getNumRegions() == 1 && "MemOp has zero region!");
-  assert(!getBody().empty() && "MemOp should have non-empty body");
+  if (body.empty())
+    return emitOpError("cannot have an empty body");
```
stephenneuendorffer, line 821: Seems strange that this message differs from CoreOp... why not use the same language?

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -65,7 +65,8 @@ struct AIEDebugOpToStdLowering : public OpConversionPattern<DebugOp> {
 
     std::string funcName = "debug_i32";
     auto func = module.lookupSymbol<func::FuncOp>(funcName);
-    assert(func && "Could not find the intrinsic function!");
+    if (!func)
+      return module.emitOpError("Could not find the intrinsic function!");
```
stephenneuendorffer, line 69: It's not clear to me that this is a user-visible OpError?  I assume your code is creating the debug_i32 intrinsic somewhere else?  In that case, an assert is actually more appropriate, since there's nothing about the user's input that they could change to fix it.

lib/Dialect/AIE/IR/AIEDialect.cpp
AndraBisca: The emitOpError() will actually start out by printing the Op at which the error occurred, in this case the CoreOp. 
Typically the syntax is:
CoreOp core;
core.emitOpError(...);
But inside the verify() we don't need to refer to the CoreOp.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -735,8 +733,9 @@ int xilinx::AIE::ShimMuxOp::rowIndex() { return getTileOp().rowIndex(); }
 
 // ShimDMAOp
 LogicalResult xilinx::AIE::ShimDMAOp::verify() {
-  assert(getOperation()->getNumRegions() == 1 && "ShimDMAOp has zero region!");
-  assert(!getBody().empty() && "ShimDMAOp should have non-empty body");
+  if (getBody().empty())
+    return emitOpError("cannot have an empty body");
```
stephenneuendorffer, line 737: Sorry, this is the oddball one... All the rest say "should have non-empty body"

lib/Dialect/AIE/IR/AIEDialect.cpp
stephenneuendorffer: Most of the rest say "should have non-empty body".  Seems like consistency would be good here?

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
AndraBisca: I am not actually very familiar with this transform so I don't know where the intrinsic function is created. My reasoning for removing the assert() within the pass was to use MLIR's conversion pattern failure system for more graceful failing and also ensure that it would fail even without asserts() being enabled.

lib/Dialect/AIE/IR/AIEDialect.cpp
AndraBisca: Ah indeed! I will correct that. Thank you Steve!

lib/Dialect/AIE/IR/AIEDialect.cpp
AndraBisca: Same as earlier, I will correct the message.

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
stephenneuendorffer: Got it... Basically, if this is an internal consistency check, then assert() is appropriate.  If its an assumption about valid inputs, then opError is appropriate.  Unfortunately, between passes, this can become tricky, since pass A may generate something that pass B needs:  So is that an 'internal consistency check', because we always run pass A before pass B, or is an assumption?  Both types of messaging can be confusing, but I'd probably err on the side of emitOpError.

AndraBisca: I see! I will check whether there is another pass that always runs before this one, in which case I will switch back to using assert(). But if the pass can be used as a standalone then I will leave emitOpError. Thank you for the insights!

PR #399, opened 2023-03-28
Support folding mul add chain to conv ops for i8 and i16 types in AIEVec

linay-xsj:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -562,6 +568,586 @@ struct SetInboundsToReadStoreOpPattern : public RewritePattern {
...
+  return defLhs ? defLhs : defRhs;
+}
+
+// The mul add chain pattern is valid if the defining ops of a mul op consist of
+// a broadcast op and an upd op.
+static bool checkChainPattern(arith::MulIOp mulOp, MulDefMapTy &macChainMap,
+                              SmallVectorImpl<Value> &bcastOpSourceVec) {
+  // Get the mul op's lhs and rhs defining ops. We keep splat op at
+  // rhs.
+  if (isa<aievec::BroadcastOp>(mulOp.getOperand(0).getDefiningOp())) {
```
david-vc, line 704: checkChainPattern must not modify the IR for cases where the pattern does not apply!

david-vc (changes requested):
Hi Lina:
I'm going to need more time to review this.
However, this change adds 500 lines to ConvertVectorToAIEVec.cpp.
It adds lots of static functions that only are relevant for the new pattern and 
these functions have names that give no indication that this is for the new pattern.
Therefore, I recommend putting the new pattern code in a separate cpp file.

david-vc:
Hi Lina:
I also see lots of asserts in code called at the time that we are still checking whether the new pattern applies.
It seems to me that at the time we are checking whether the pattern applies, we should only assert conditions that
can fail only if the IR is malformed. Conditions that indicate that a situations that the code is currently not handling
should be handled gracefully by just not applying the pattern rather than asserting.

include/aie/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.h
```diff
@@ -0,0 +1,511 @@
...
+        refreshFusedGroups(defTuple, nextMulOp, fusedOps, groupFusedOps, curIdx,
+                           curUpdOp, curMulOp);
+        continue;
+      }
+
+      xDist = dists[0];
+      zDist = dists[1];
+      dupFactor = dists[0];
+
+      fusedOps.push_back(nextMulOp);
```
david-vc, line 257: Can we follow the same pattern as you have earlier, e,g.
`if (fusedOps.size() == groupSize) {
  refreshFusedGroups(..):
  continue;
}`

include/aie/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.h
```diff
@@ -0,0 +1,511 @@
...
+          return false;
+        }
+        refreshFusedGroups(defTuple, nextMulOp, fusedOps, groupFusedOps, curIdx,
+                           curUpdOp, curMulOp);
+        continue;
+      }
+
+      xDist = dists[0];
+      zDist = dists[1];
+      dupFactor = dists[0];
```
david-vc, line 255: For the conv2d case, it could be that for the first line we determing dupFactor=1, but the second line could set dupFactor=2. My point is that this code is overwriting the dupfactor computed in previous iterations. Is this leading to problems in the caller? Seems to me that we need to check that the dupfactor computed in sucessive iterations is the same.

include/aie/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.h
```diff
@@ -0,0 +1,511 @@
...
+  groupFusedOps.push_back(fusedOps);
+  fusedOps.clear();
+  fusedOps.push_back(nextMulOp);
+  std::tie(curIdx, curUpdOp, curMulOp) = defTuple;
+}
+
+// Check whether mul add chain is valid for the transformation and classify the
+// fused ops into different groups with valid constant memref distances.
+static bool
+collectFusedOps(unsigned groupSize, unsigned &dupFactor,
```
david-vc, line 139: "groupSize" => "maxGroupSize"

include/aie/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.h
```diff
@@ -0,0 +1,511 @@
...
+
+      auto notMulOrAddOp = [&](Operation *op) {
+        return !isa<arith::MulIOp, arith::AddIOp>(op);
+      };
+
+      Operation *convOp = nullptr;
+      arith::AddIOp lastAddOp =
+          cast<arith::AddIOp>(*(fusedOps.back()->getUsers().begin()));
+
+      if (llvm::any_of(lastAddOp->getUsers(), notMulOrAddOp)) {
```
david-vc, line 484: I would expected that we care only whether the last add has a user that is not an add.
Why do you instead check whether the last add has a user that is not a add and not a mul?

david-vc (commented):
Hi Lina: 
This looks good. I found a number of small issues and questions that should be straightforward to address or answer. I will approve after you do this. Thanks.

david-vc (approved):
Great job Lina!

PR #400, opened 2023-03-29
Fix memory model for aie2

stephenneuendorffer:


lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -211,8 +211,8 @@ bool AIE2TargetModel::isMemSouth(int srcCol, int srcRow, int dstCol,
 
 bool AIE2TargetModel::isLegalMemAffinity(int coreCol, int coreRow, int memCol,
                                          int memRow) const {
-  bool IsMemWest = isInternal(coreCol, coreRow, memCol, memRow);
-  bool IsMemEast = isEast(coreCol, coreRow, memCol, memRow);
+  bool IsMemWest = isWest(coreCol, coreRow, memCol, memRow);
```
keryell, line 214: While you are here, I have the feeling all these should be `constexpr` if we want to reuse these definitition for metaprogramming for example, in another context?

stephenneuendorffer: Hrm.. doesn't that require c++23?  We only have c++17 (aligned with LLVM)

keryell: They look like static member functions, so that should work, But anyway, there is no use case yet, so, wait and see...

PR #401, opened 2023-03-31
Update linker script generation with stack size control

### LINKED ISSUE #209: Need MLIR representation for where the stack goes
Currently AssignMemBuffers reserves some space for the stack of each core and the backend targets will eventually generate a linker script that puts the stack in the corresponding place, however this is somewhat wasteful since it means we lose some explicit control over how the stack is allocated.  In addition, we also lack any validation that the stack size is correct WRT the requirements of the code, which should be based on some static analysis. 

So, probably the right thing is to have an AIE.buffer-like object that represents the stack, which would be generated explicitly in AssignMemBuffers, then serialized in the linker scripts appropriately.

stephenneuendorffer:
CoreOp now supports a 'stackSize' attribute, enabling the control of the stack allocation.

Along the way, fix a bug where memory areas were not properly being reserved (either with gnu linker scripts or bcf files).  This allowed static allocations to potentially overlap with allocated communication buffers.  Note that in order to keep static allocations independent on different cores, we allow them only to be mapped into the local memory of a core.

Fixes #209

### DIFF
--- a/lib/Targets/AIETargets.cpp
+++ b/lib/Targets/AIETargets.cpp
@@ -77,6 +77,9 @@ void writeBCFMap(raw_ostream &output, BufferOp buf, int offset,
          << "0x" << llvm::utohexstr(offset + bufferBaseAddr) << " "
          << "0x" << llvm::utohexstr(numBytes) << '\n';
   output << "_extern " << bufName << "\n";
+  output << "_reserved DMb "
+         << "0x" << llvm::utohexstr(offset + bufferBaseAddr) << " "
+         << "0x" << llvm::utohexstr(numBytes) << '\n';
 }
 // Output the memorymap in gnu linker format for the given buffer operations,
 // with the given offset. The offset is different depending on where the buffers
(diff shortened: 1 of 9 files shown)


lib/Targets/AIETargets.cpp
```diff
@@ -389,11 +416,15 @@ SECTIONS
...
 
+            int stacksize = 0;
+            if (auto core = tile.getCoreOp())
+              stacksize = core.getStackSize();
             output << "_stack    DM_stack 0x"
                    << llvm::utohexstr(
                           target_model.getMemInternalBaseAddress(srcCoord))
-                   << "  0x400 //stack for core\n"; // TODO Needs to match
-                                                    // stack size!!!
+                   << "  0x" << llvm::utohexstr(stacksize)
```
keryell, line 425: Did you find a way to put the stack at a symbolic location to maximize the stack + heap usage? It looks like Bridge linker is very limited and does not have the concept of "." for example.

PR #406, opened 2023-04-04
Make pathfinder pass DeviceOp aware

AndraBisca:
Designs compiled with the current version of mlir-aie main branch fail to execute both in simulation and on hardware (tested on vck190). This is due to missing connection and wire operations in the input_physical.mlir files. This PR solves that issue by making the pathfinder and find-flow passes DeviceOp aware such that the missing operations are generated.


denolf (approved):

denolf:
merged #406

PR #409, opened 2023-04-05
Bugfixes for AIE2 target

fifield:
* For now assume row 1 is memtile in `AIE2TargetModel::getMemSouth`
* Fix logic in bcf gen

### DIFF
--- a/lib/Dialect/AIE/IR/AIETargetModel.cpp
+++ b/lib/Dialect/AIE/IR/AIETargetModel.cpp
@@ -159,7 +159,7 @@ Optional<TileID> AIE2TargetModel::getMemNorth(TileID src) const {
 Optional<TileID> AIE2TargetModel::getMemSouth(TileID src) const {
   Optional<TileID> ret = std::make_pair(src.first, src.second - 1);
   // The first row doesn't have a tile memory south
-  if (!isValidTile(*ret) || ret->second == 0)
+  if (!isValidTile(*ret) || ret->second == 0 || ret->second == 1)
     ret.reset();
   return ret;
 }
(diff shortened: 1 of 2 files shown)


lib/Targets/AIETargets.cpp
```diff
@@ -398,17 +398,13 @@ SECTIONS
               }
             };
 
-            if (auto tile = target_model.getMemSouth(srcCoord))
-              doBuffer(tile, target_model.getMemSouthBaseAddress(),
+            doBuffer(target_model.getMemSouth(srcCoord), target_model.getMemSouthBaseAddress(),
```
stephenneuendorffer, line 401: Hmm... probably the ld script generation should have similar structure?

lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -159,7 +159,7 @@ Optional<TileID> AIE2TargetModel::getMemNorth(TileID src) const {
 Optional<TileID> AIE2TargetModel::getMemSouth(TileID src) const {
   Optional<TileID> ret = std::make_pair(src.first, src.second - 1);
   // The first row doesn't have a tile memory south
-  if (!isValidTile(*ret) || ret->second == 0)
+  if (!isValidTile(*ret) || ret->second == 0 || ret->second == 1)
```
stephenneuendorffer, line 162: Sure.. I'll work on modeling the memtiles more explicitly.

lib/Targets/AIETargets.cpp
fifield: Yeah looks like it. I'll add the change to this PR

stephenneuendorffer (approved):

PR #411, opened 2023-04-06
Support folding unsorted mul add chain to conv ops and enable conv2d test case with initialization

linay-xsj:


david-vc (approved):
Good work, but too bad that this pattern has gotten very big and likely hard to maintain.

PR #416, opened 2023-04-07
Enable converting arith.mul to aievec.mul_elem operation for AIE-ML i16 and i32 types

linay-xsj:


test/unit_tests/aievec_tests/i32xi32_mul_elem/i32xi32_mul_elem.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: aie-opt %s -affine-super-vectorize="virtual-vector-size=16" --convert-vector-to-aievec="aie-target=aieml" -lower-affine | aie-translate -aieml=true --aievec-to-cpp -o convert_aie-ml.cc
+// RUN: xchesscc -f -g +s -p me -P %aietools/data/aie_ml/lib/ +w work +o work -I%S -I. %S/testbench.cc %S/kernel.cc
+// RUN: xme_ca_udm_dbg -qf -T -P %aietools/data/aie_ml/lib/ -t "%S/../profiling.tcl ./work/a.out"
```
david-vc, line 3: Can you please verify that if the test case would compute an incorrect result, the test will fail?
My reading of the code is that even if the test computes incorrect values, the test will still pass.
Please fix.

david-vc (changes requested):
Could you please make sure that the unit_tests fail if the test case would compute incorrect results?

test/unit_tests/aievec_tests/i32xi32_mul_elem/i32xi32_mul_elem.mlir
linay-xsj: testbench.cc line51-55  shows if there is any error, the test will fail. Not sure why you think the test will pass if the test computes incorrect values.

david-vc (approved):

PR #417, opened 2023-04-07
restructure xaiengine install rules to match sysroot structure

denolf:

PR #419, opened 2023-04-10
[xaiengine] Consistently structure installed directory

stephenneuendorffer:
The xaiengine install area has separate lib and include directories. However, it's unclear how this should work in the presence of multiple target architectures.


stephenneuendorffer:
Many users still need to get updated.

stephenneuendorffer:
> I prefer to compile it myself then give the architecture specific `LibXAIE_ROOT` in my toolchain file

Then we can only have one target at a time...

stephenneuendorffer:
See #417

PR #422, opened 2023-04-10
refactor runtime_lib compilation

stephenneuendorffer:
This patch enables compiling the host runtime library for multiple host architectures (e.g., x86_64, aarch64).  This is done by specifying AIE_RUNTIME_TARGETS as a list of targets and ${target}_TOOLCHAIN_FILE for each target.

Along the way, we separate out the aie_runtime_lib directory to contain libraries for different aie target architectures.  The intention here is to be able to build a toolchain which can be used to target different host/device combinations at runtime, particularly, ARM + VCK190 (e.g. aarch64 host and AIE1 target) and X86 + VCK5000 (e.g. x86_64 host and AIE2 target).


runtime_lib/CMakeLists.txt
```diff
@@ -13,22 +13,71 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
...
 project("aie runtime libs")
 
-# include(ExternalProject)
+include(ExternalProject)
 
-# set(INSTALLS
-#     test_library.cpp
-#     test_library.h)
+# test_library
+# FIXME: don't ship as source... compile them instead.
```
stephenneuendorffer, line 19: @denolf The FIXME here was somewhat important... I was trying to limit this to having one external project for all of the runtime_lib directory.  Do you have a better idea for how to accomplish that without having the external project in the toplevel CMakeLists.txt?

denolf (approved):

runtime_lib/CMakeLists.txt
```diff
@@ -5,81 +5,86 @@
...
 
-add_subdirectory(AIE)
-add_subdirectory(AIE2)
+  if (NOT EXISTS ${${target}_TOOLCHAIN_FILE})
+    message(FATAL_ERROR "Toolchain file ${${target}_TOOLCHAIN_FILE} not found! Cannot build target ${target}.")
+  endif()
+  message("Building AIE runtime for ${target} with TOOLCHAIN=${${target}_TOOLCHAIN_FILE}")
+  
+  # xaiengine
+  ExternalProject_Add(xaiengine_${target}
```
stephenneuendorffer, line 42: @denolf Would really like to avoid having 2 ExternalProjects here.

test/CMakeLists.txt
```diff
@@ -26,6 +26,9 @@ endif()
 
 project(aie-test LANGUAGES CXX C)
 
+# default to host system architecture for runtime target
+set(AIE_RUNTIME_TEST_TARGET ${CMAKE_HOST_SYSTEM_PROCESSOR} CACHE STRING "Runtime architecture to test with.")
```
stephenneuendorffer, line 30: This doesn't work if CMAKE_HOST_SYSTEM_PROCESSOR isn't in AIE_RUNTIME_TARGETS

test/CMakeLists.txt
```diff
@@ -120,14 +114,13 @@ set(TEST_DEPENDS
...
 add_lit_testsuite(check-aie "Running the aie regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS ${TEST_DEPENDS}
-  ARGS "-sv --timeout 300 --time-tests"
+  ARGS "-a --timeout 300 --time-tests"
   )
 set_target_properties(check-aie PROPERTIES FOLDER "Tests")
 
-add_lit_testsuites(AIE ${CMAKE_CURRENT_BINARY_DIR} DEPENDS ${TEST_DEPENDS} ARGS "-sv --timeout 300 --time-tests")
+add_lit_testsuites(AIE ${CMAKE_CURRENT_BINARY_DIR} DEPENDS ${TEST_DEPENDS} ARGS "-a --timeout 300 --time-tests")
```
stephenneuendorffer, line 126: Was this intentional? It causes quite a bit of noise.

denolf: @stephenneuendorffer  Yes it was and remains helpful when redirecting the output of the lit tests to a .log file. #448 moves it back to -sv to reduce the noice.

runtime_lib/CMakeLists.txt
denolf: @stephenneuendorffer since these are two different targets that need to get cross compiled, they need their own external project. As soon as 2023.1 is released, I plan to no longer get the code from the Vitis environment but from https://github.com/Xilinx/aie-rt/ and that will also motivates having a separate external project for xaiengine.

stephenneuendorffer: We don't really need an external project per sub-directory: it's quite heavyweight.  I think in this case, we're not using cmake stuff from the Vitis environment (we are providing all the cmake stuff).  Will that change with aie-rt?

test/CMakeLists.txt
denolf: @stephenneuendorffer see  #455 for a fix

PR #427, opened 2023-04-20
[aiecc] Generate unique project directories by default.

stephenneuendorffer:
If running multiple designs in the same directory in parallel, then they can collide when writing to our temporary project directory. This patch generates a unique project directory by default based on the mlir file name.  (Note that this will still fail when compiling the same file name with different options)
In addition, move sim directory generation into the project directory.


fifield (approved):
This is good cleanup. We might still have a problem with parallel builds and intermediate outputs from chess getting generated in the current working directory and colliding. They should get generated into the project folder.

stephenneuendorffer:
> This is good cleanup. We might still have a problem with parallel builds and intermediate outputs from chess getting generated in the current working directory and colliding. They should get generated into the project folder.

good point

stephenneuendorffer:
@fifield Perhaps more fundamentally, when we generate separate .elf files for each core, the names are all the same, so running different designs in the same directory is going to collide anyway

PR #436, opened 2023-04-25
Enable converting arith.mul to aievec.mul_elem operation for AIE-ML bfloat16 input type with bfloat16 output

linay-xsj:


david-vc (approved):
This pattern will need to be augmented to handle a float32 result with bfloat16 operands.

linay-xsj:
> This pattern will need to be augmented to handle a float32 result with bfloat16 operands.

Yes, it will be in the next changeset.

PR #439, opened 2023-04-26
Cleanup warnings

AndraBisca:
PR to clean up the build warnings.

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIECreatePacketFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePacketFlows.cpp
@@ -447,14 +447,14 @@ struct AIERoutePacketFlowsPass
     LLVM_DEBUG(llvm::dbgs() << "CHECK mastersets\n");
     for (auto map : mastersets) {
       Operation *tileOp = map.first.first;
-      WireBundle bundle = map.first.second.first;
-      int channel = map.first.second.second;
+      [[maybe_unused]] WireBundle bundle = map.first.second.first;
+      [[maybe_unused]] int channel = map.first.second.second;
       assert(tileOp);
-      TileOp tile = dyn_cast<TileOp>(tileOp);
+      [[maybe_unused]] TileOp tile = dyn_cast<TileOp>(tileOp);
       LLVM_DEBUG(llvm::dbgs()
                  << "master " << tile << " " << stringifyWireBundle(bundle)
                  << " : " << channel << '\n');
-      for (auto value : map.second)
+      for ([[maybe_unused]] auto value : map.second)
         LLVM_DEBUG(llvm::dbgs() << "amsel: " << value << '\n');
     }
 
@@ -533,9 +533,9 @@ struct AIERoutePacketFlowsPass
     LLVM_DEBUG(llvm::dbgs() << "CHECK Slave Masks\n");
     for (auto map : slaveMasks) {
       auto port = map.first.first;
-      TileOp tile = dyn_cast<TileOp>(port.first);
-      WireBundle bundle = port.second.first;
-      int channel = port.second.second;
+      [[maybe_unused]] TileOp tile = dyn_cast<TileOp>(port.first);
+      [[maybe_unused]] WireBundle bundle = port.second.first;
+      [[maybe_unused]] int channel = port.second.second;
       int ID = map.first.second;
       int mask = map.second;
 
@@ -629,7 +629,7 @@ struct AIERoutePacketFlowsPass
         int ID = group.front().second & mask;
 
         // Verify that we actually map all the ID's correctly.
-        for (auto slave : group) {
+        for ([[maybe_unused]] auto slave : group) {
           assert((slave.second & mask) == ID);
         }
         Value amsel = amselOps[slaveAMSels[group.front()]];
--- a/lib/Dialect/AIE/Transforms/AIECreatePathfindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathfindFlows.cpp
@@ -301,11 +301,11 @@ struct ConvertFlowsToInterconnect : public OpConversionPattern<AIE::FlowOp> {
     TileOp srcTile = cast<TileOp>(flowOp.getSource().getDefiningOp());
     TileOp dstTile = cast<TileOp>(flowOp.getDest().getDefiningOp());
     TileID srcCoords = std::make_pair(srcTile.colIndex(), srcTile.rowIndex());
-    TileID dstCoords = std::make_pair(dstTile.colIndex(), dstTile.rowIndex());
+    [[maybe_unused]] TileID dstCoords = std::make_pair(dstTile.colIndex(), dstTile.rowIndex());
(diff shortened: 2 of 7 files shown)


AndraBisca:
There are still a few warnings left. Could you please take a quick look @stephenneuendorffer?

lib/Dialect/AIEVec/Transforms/IntervalReuse.cpp
```diff
@@ -33,8 +33,8 @@ bool IntervalReuse::sameEnclosingLoops(
   // reference.
   Operation *ref = extentMap.begin()->first;
 
-  Block *opBlock = op->getBlock();
-  Block *refBlock = ref->getBlock();
+  [[maybe_unused]] Block *opBlock = op->getBlock();
+  [[maybe_unused]] Block *refBlock = ref->getBlock();
```
jsetoain, line 37: The solution to this is not to mark them as `maybe unused` but to actually use them in the return statement.

lib/Dialect/AIEX/Transforms/AIECreateCores.cpp
```diff
@@ -173,7 +173,7 @@ struct AIECreateCoresPass : public AIECreateCoresBase<AIECreateCoresPass> {
           // the function argument's type is a scalar, we promote it to a
           // one-element memref, and do a load to the buffer at index 0
           for (auto pair : coreBufTypes) {
-            MemRefType t = pair.first;
+            [[maybe_unused]] MemRefType t = pair.first;
```
jsetoain, line 176: This should be inside the if statement. Also, maybe consider using `pair.first.getShape()` in the asserts instead of capturing `pair.first` in `t`.

lib/Dialect/AIEX/Transforms/AIECreateLocks.cpp
```diff
@@ -160,8 +160,8 @@ struct AIECreateLocksPass : public AIECreateLocksBase<AIECreateLocksPass> {
       Operation *acqUser = TA.getTokenUserOp(acquire);
       bool IsRelUserCore = isa<CoreOp>(relUser);
       bool IsAcqUserCore = isa<CoreOp>(acqUser);
-      std::pair<int, int> relUserCoord = TA.getCoord(relUser);
-      std::pair<int, int> acqUserCoord = TA.getCoord(acqUser);
+      [[maybe_unused]] std::pair<int, int> relUserCoord = TA.getCoord(relUser);
+      [[maybe_unused]] std::pair<int, int> acqUserCoord = TA.getCoord(acqUser);
```
jsetoain, line 164: Have you tried moving these within the LLVM_DEBUG block?

lib/Dialect/AIEX/Transforms/AIECreateLocks.cpp
```diff
@@ -160,8 +160,8 @@ struct AIECreateLocksPass : public AIECreateLocksBase<AIECreateLocksPass> {
...
 
       Operation *tileOp = TA.getShareableTileOp(relUser, acqUser);
```
jsetoain, line 166: Nit: This would look better closer to first use, after the LLVM_DEBUG.

jsetoain (changes requested):
Thank you, Andra!

lib/Dialect/AIEVec/Transforms/IntervalReuse.cpp
AndraBisca: Yes, that's true. The reason I chose to do it like this was to avoid changing the code of the original designer too much. I'll make the change you suggested, I also prefer it.

lib/Dialect/AIEX/Transforms/AIECreateCores.cpp
AndraBisca: Same as above.

lib/Dialect/AIEX/Transforms/AIECreateLocks.cpp
AndraBisca: I have not, I will give it a try!

keryell (commented):
Ouch. The original code has a lot of copy-paste, which make think that there is some factorization possible.
Provide a macro `assert_all_range(range, lambda_to_be_verified_on_each_element)` while waiting for C++20 to have this as a function using `std::source_location` for the debug? Or just keep the code as is?
Otherwise there is also `std::ignore` I gave a presentation on at the AMD C++ user-group https://gitenterprise.xilinx.com/rkeryell/xilinx_beamer/tree/master/2022/2022-08-09-C%2B%2B-user-group-C%2B%2B-ignorance

keryell:
Rethinking to it with a more C-like solution.
Instead of adding a lot of `[[maybe_unused]]` you might put the code necessary only for `assert()` between
```c++
#ifndef NDEBUG
// Some code only required by the assert()
#endif
```

lib/Dialect/AIE/Transforms/AIECreatePacketFlows.cpp
```diff
@@ -629,7 +629,7 @@ struct AIERoutePacketFlowsPass
         int ID = group.front().second & mask;
 
         // Verify that we actually map all the ID's correctly.
-        for (auto slave : group) {
+        for ([[maybe_unused]] auto slave : group) {
           assert((slave.second & mask) == ID);
         }
```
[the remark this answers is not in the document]
AndraBisca: Yes that definitely looks better, thank you!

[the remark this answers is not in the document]
keryell: Well, better for some definition of "better". It is not great either. At the end it is longer than the original code, but at least it removes the dead code... Perhaps the solution with `#ifndef NDEBUG` is more straightforward?

jsetoain (approved):

lib/Dialect/AIE/Transforms/AIECreatePathfindFlows.cpp
```diff
@@ -32,11 +32,13 @@ static llvm::cl::opt<bool>
                llvm::cl::init(false));
 
 #define BOOST_NO_EXCEPTIONS
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Winvalid-noreturn"
 #include <boost/throw_exception.hpp>
 void boost::throw_exception(std::exception const &e) {
   // boost expects this exception to be defined.
-  assert(false);
```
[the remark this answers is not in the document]
AndraBisca: Yes, I put it right back! Thank you for noticing as well!

jsetoain (approved):

jsetoain (commented):
lgtm

lib/Targets/ADFGenerateCppGraph.cpp
```diff
@@ -92,8 +93,8 @@ struct GraphWriter {
       return "stream";
     else if (auto windowType = type.dyn_cast<ParameterType>())
       return "parameter";
-    else
-      assert(false);
+    assert(false);
```
keryell, line 96: Just curious on these trivial use cases.
Is it not a godd use case for `std::terminate()`? What else if the code goes on here?

PR #440, opened 2023-04-26
Enable converting arith.mul to aievec.mul_elem operation for AIE-ML bfloat16 input type with float output.

linay-xsj:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -174,16 +174,21 @@ static bool canFoldAIEShiftAndBroadcast(aievec::BroadcastOp op,
 }
 
 static aievec::MulElemOp
-createMulElemOpForI8(ConversionPatternRewriter &rewriter, Value lval,
-                     Value rval, VectorType srcType, unsigned bitWidth,
-                     Location loc) {
+createMulElemOpForI8OrBf16(ConversionPatternRewriter &rewriter, Value lval,
```
david-vc, line 177: I realize that Javier insisted on a specific name for this function, but it seems to me that down the road this function will handle more cases. Trying to encode the exact uses cases in the function name will become impractical. It seems to me the naming this function createMulElemAieML would then be more appropriate. I also would like to see a comment right before L176 that lists which cases this function tries to handle. Right now, I need to de-cypher the function body.

david-vc (approved):
Can you please rename createMulElemFor... as requested in my comment? Thanks.

PR #442, opened 2023-04-26
Fix tests after https://github.com/Xilinx/mlir-aie/pull/422

fifield:
Fix tests failing on internal runner.


denolf (approved):
Thanks for making these changes to all unit tests.

denolf:
Merged #422 in main.

PR #449, opened 2023-04-27
Better fix for AIE2 lock count

stephenneuendorffer:


lib/Targets/AIETargetXAIEV1.cpp
```diff
@@ -153,6 +152,7 @@ mlir::LogicalResult AIETranslateToXAIEV1(ModuleOp module, raw_ostream &output) {
       output << clear_range(col, row, 0x3F200, 0x3F3AC);
 
       // Release locks
+      int numLocks = target_model.getNumLocks(col, row);
       output << "for (int l=0; l<" << numLocks << "; l++)\n"
```
keryell, line 156: While you are here, apply the coding style to the generated code
```suggestion
      int numLocks = target_model.getNumLocks(col, row);
      output << "for (int l = 0; l < " << numLocks << "; ++l)\n"
```

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -139,6 +138,7 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
       output << "XAie_CoreDisable(" << deviceInstRef << ", "
              << tileLocStr(col, row) << ");\n";
       // Release locks
+      int numLocks = target_model.getNumLocks(col, row);
       output << "for (int l=0; l<" << numLocks << "; l++)\n"
```
keryell, line 142:
```suggestion
      int numLocks = target_model.getNumLocks(col, row);
      output << "for (int l = 0; l < " << numLocks << "; ++l)\n"
```

lib/Targets/AIETargetXAIEV1.cpp
stephenneuendorffer: strange that clang-format didn't catch it.

keryell: Because it is hidden in a string. :-) Perhaps you need a kind ChatGPT-powered clang-format... ;-)

stephenneuendorffer: Ah, right.

fifield (approved):
Makes sense to me since number of locks can vary by tile.

PR #450, opened 2023-04-28
Remove calls to MLIR passes in assert()

AndraBisca:


jsetoain (approved):

lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -2623,7 +2623,8 @@ static void redundantLoadStoreOptimization(ModuleOp module) {
 static void preCanonicalizeIR(ModuleOp module) {
   PassManager pm(module.getContext());
   pm.addPass(createCanonicalizerPass());
-  assert(!failed(pm.run(module)));
+  auto f = failed(pm.run(module));
+  assert(!f);
```
keryell, line 2627: Avoid double negation (is this sentence a triple negation? ;-) ).
I think using the member function is safer than using ADL to avoid name resolution conflicts in general https://en.cppreference.com/w/cpp/language/adl
```c++
bool success = pm.run(module).succeeded();
std::ignore = success;
assert(success);
```
or
```c++
[[maybe_unused]] bool success = pm.run(module).succeeded();
assert(success);
```
PS: leave more time to the reviewers before merging.

AndraBisca: @keryell Thank you for the review. This pass is actually being fully rewritten and will be deleted (@jsetoain) so these corrections will go along with it. But I agree that I should've done a better job to avoid creating new unused variable warnings, I will correct that.

AndraBisca: See: https://github.com/Xilinx/mlir-aie/pull/451

PR #452, opened 2023-04-28
Sim restructure

denolf:
Added 2 x86_64 specific static test_lib targets: test_lib_sim_aie and test_lib_sim_aie2 to support aie and aie2 simulation on x86_64. This should allow us to (finally) remove test_library.cpp from the install.
They are both installed in runtime_lib/x86_64/test_lib/lib for now. Maybe aie_runtime_lib/{AIE,AI2}/aiesim is a better location since they are simulation specific?

### DIFF
--- a/runtime_lib/test_lib/CMakeLists.txt
+++ b/runtime_lib/test_lib/CMakeLists.txt
@@ -31,9 +31,6 @@ install(TARGETS test_lib
     PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/runtime_lib/${AIE_RUNTIME_TARGET}/test_lib/include
 )
 
-install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/test_library.cpp ${CMAKE_CURRENT_SOURCE_DIR}/test_library.h DESTINATION ${CMAKE_INSTALL_PREFIX}/runtime_lib/${AIE_RUNTIME_TARGET}/test_lib)
-
-
 # create extra libs supporting simulation
 # AIE1 test_lib for aiesimulator
 add_library(test_lib_sim_aie STATIC test_library.cpp)
--- a/runtime_lib/test_lib/CMakeLists.txt
+++ b/runtime_lib/test_lib/CMakeLists.txt
@@ -36,7 +36,6 @@ install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/test_library.cpp ${CMAKE_CURRENT_SOURC
 
 # create extra libs supporting simulation if target is x86_64
 if(${AIE_RUNTIME_TARGET} STREQUAL "x86_64")
-message(STATUS "NOLF NOLF: sim static libs")
 
 # AIE1 test_lib for aiesimulator
 add_library(test_lib_sim_aie STATIC test_library.cpp)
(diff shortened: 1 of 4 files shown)


jackl-xilinx:
Looks good. I tested it in simulation and on the board for tutorial-2b.

jackl-xilinx (approved):

runtime_lib/xaiengine/CMakeLists.txt
```diff
@@ -8,10 +8,11 @@ cmake_minimum_required(VERSION 3.20.1)
...
 
-set(xaienginePath ${VITIS_ROOT}/data/embeddedsw/XilinxProcessorIPLib/drivers/aienginev2_v3_0/src) 
-#set(xaienginePath ${VITIS_AIETOOLS_DIR}/include/drivers/aiengine)
-file(GLOB libheaders ${xaienginePath}/*.h)
-file(GLOB libheadersSub ${xaienginePath}/*/*.h)
+set(xaienginePath ${VITIS_AIETOOLS_DIR}/include/drivers/aiengine)
+#tmp path to header files since we use an older (3_0) version of xaiengine. need to fix when we upgrade to 2023.1
+set(xaienginePathTmp ${VITIS_ROOT}/data/embeddedsw/XilinxProcessorIPLib/drivers/aienginev2_v3_0/src) 
+file(GLOB libheaders ${xaienginePathTmp}/*.h ${xaienginePath}/xioutils.h)
+file(GLOB libheadersSub ${xaienginePathTmp}/*/*.h)
```
stephenneuendorffer, line 16: I don't understand this change.  Seems like it does nothing?  or it's just picking up xioutils.h?

runtime_lib/test_lib/CMakeLists.txt
```diff
@@ -31,4 +31,32 @@ install(TARGETS test_lib
     PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_PREFIX}/runtime_lib/${AIE_RUNTIME_TARGET}/test_lib/include
 )
 
-install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/test_library.cpp ${CMAKE_CURRENT_SOURCE_DIR}/test_library.h DESTINATION ${CMAKE_INSTALL_PREFIX}/runtime_lib/${AIE_RUNTIME_TARGET}/test_lib)
\ No newline at end of file
+install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/test_library.cpp ${CMAKE_CURRENT_SOURCE_DIR}/test_library.h DESTINATION ${CMAKE_INSTALL_PREFIX}/runtime_lib/${AIE_RUNTIME_TARGET}/test_lib)
+
```
stephenneuendorffer, line 35: Hrm.. does seem odd that runtime_lib/test_lib is compiling test_library.cpp for each host_target, but the file lives and is used in aie_runtime_lib

stephenneuendorffer (approved):

runtime_lib/xaiengine/CMakeLists.txt
denolf: @stephenneuendorffer yes, we need to pick up  xioutils.h. This is a temp work around since we need to pick up older header files corresponding to the Vitis version that was used to generate the images for our boards (xaienginev2 v3_0). This will be cleaned up after we upgrade fully to 2023.1 once it is released.

PR #456, opened 2023-05-02
Make AIE::AIETargetModel::getTargetModel work on AIE.device ops

fifield:
Check if the op itself implements `AIETarget` before trying `getParentOfType<AIETarget>`


stephenneuendorffer (approved):
Out of curiosity, where did you need this?

fifield:
> Out of curiosity, where did you need this?

I was updating some mlir-air code to use `AIE.device` and didn't technically _need_ it because in that case I already knew I had a device op and could have used `op.getTargetModel()`. But it was unexpected and took some time to debug when it didn't work on the device op.

PR #459, opened 2023-05-03
Enable converting arith.add to aievec.add_elem and converting arith.sub to aievec.sub_elem for AIE-ML with different input and output types

linay-xsj:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -862,6 +863,156 @@ using LowerVectorSubIOpToAIEVecSubOp =
...
+
+template <typename SrcOpTy, typename DstOpTy>
+struct LowerVectorAddOrSubOpToAIEVecAddElemOrSubElemOp
+    : public OpConversionPattern<SrcOpTy> {
+  using OpConversionPattern<SrcOpTy>::OpConversionPattern;
+  using OpAdaptor = typename SrcOpTy::Adaptor;
+
+  LowerVectorAddOrSubOpToAIEVecAddElemOrSubElemOp(MLIRContext *context,
+                                                  unsigned shiftParam = 0)
+      : OpConversionPattern<SrcOpTy>(context), shiftParam(shiftParam) {}
```
david-vc, line 888: What are the use cases where shiftParam is not 0?
Are there any?
I would much prefer if this pattern would always use UPS with shift=0.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -862,6 +863,156 @@ using LowerVectorSubIOpToAIEVecSubOp =
...
+static void genAddElemAieML(ConversionPatternRewriter &rewriter, Value lval,
+                            Value rval, VectorType srcType, SrcOpTy srcOp) {
+  auto lCastOp = rewriter.create<aievec::CastOp>(srcOp.getLoc(), srcType, lval,
+                                                 /*isResAcc*/ true);
+  auto rCastOp = rewriter.create<aievec::CastOp>(srcOp.getLoc(), srcType, rval,
+                                                 /*isResAcc*/ true);
+  auto elemOp = rewriter.create<AIEv2ElemOp>(
+      srcOp.getLoc(), lCastOp->getResult(0).getType(), lCastOp->getResult(0),
+      rCastOp->getResult(0));
+  rewriter.replaceOpWithNewOp<aievec::CastOp>(
```
david-vc, line 876: Changing the return type to 'LogicalResult' and returning success() allows you to shorten the code in the caller, 
i.e.
`if (blah) {
  genAddElemAieML(..);
  return success();
}`
->
`if (blah)
  return genAddElemAieML(..);
`

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -862,6 +863,156 @@ using LowerVectorSubIOpToAIEVecSubOp =
...
+
+      if (laneSize == 32 && resultElWidth == 32) {
+        if (!lhsDefOp || !rhsDefOp) {
+          genAddElemAieML<SrcOpTy, DstOpTy>(rewriter, lhs, rhs, resultType,
+                                            srcOp);
+          return success();
+        }
+        auto lhsExt = dyn_cast<arith::ExtSIOp>(lhsDefOp);
+        auto rhsExt = dyn_cast<arith::ExtSIOp>(rhsDefOp);
+        if (!lhsExt || !rhsExt) {
```
david-vc, line 932: Since the 'then' branch always returns, there is an opportunity to reduce the nesting level of the 'else' part by 
`if (blah) {.. ; return succss();} else { ELSE_PART }`
->
`if (blah) {.. ; return succss();} 
  ELSE_PART`
(irgnore messed up formatting)

david-vc (approved):
Looks good; my only concern is about the use of 'shiftParam' in the pattern.
I would much prefer is the pattern would not need 'shiftParam'.

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -1050,6 +1047,102 @@ ParseResult SubOp::parse(OpAsmParser &parser, OperationState &result) {
...
+LogicalResult aievec::AddElemOp::verify() {
+  return verifyAddElemSubElemOp<aievec::AddElemOp>(*this);
+}
+
+LogicalResult aievec::SubElemOp::verify() {
+  return verifyAddElemSubElemOp<aievec::SubElemOp>(*this);
+}
+
+// Parse AddElem and SubElem op.
+ParseResult parseAddElemSubElemOp(OpAsmParser &parser, OperationState &result) {
```
jsetoain, line 1101: Is there a reason to be using a custom parser? Can't we use an auto-generated one?

jsetoain (changes requested):
As a general comment, this patch is unnecessarily large. A better split would be to send a patch adding the instructions and testing them, maybe even one patch per instruction; then another patch with the conversions and tests for those conversions, again, maybe even one patch per op conversion; and finally a patch with integration tests, or one patch per integration test set. Ideally, one patch does one self-contained thing: fix one issue, add one feature, etc.

The bigger the patch, the easier it is to miss things, and the harder it is to track changes in the logs. Patches this big risk overlooking significant things.

I recommend a careful read of LLVM's Developer Policy: https://www.llvm.org/docs/DeveloperPolicy.html#commit-messages

PR #461, opened 2023-05-03
reduce default link flags in aiecc

denolf:
-ldl was causing issues with air runtime linking and seems not needed 
-rdynamic seems not needed as well

check-aie and check-tutorials pass

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -123,7 +123,7 @@ set(TEST_DEPENDS
 add_lit_testsuite(check-aie "Running the aie regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS ${TEST_DEPENDS}
-  ARGS "-j8 -sv --timeout 300 --time-tests"
+  ARGS "-sv --timeout 300 --time-tests"
   )
 set_target_properties(check-aie PROPERTIES FOLDER "Tests")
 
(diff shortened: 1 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -123,7 +123,7 @@ set(TEST_DEPENDS
 add_lit_testsuite(check-aie "Running the aie regression tests"
   ${CMAKE_CURRENT_BINARY_DIR}
   DEPENDS ${TEST_DEPENDS}
-  ARGS "-sv --timeout 300 --time-tests"
+  ARGS "-j8 -sv --timeout 300 --time-tests"
```
stephenneuendorffer, line 126: Why -j8?  lit's generally pretty good about parallelizing things by default.

denolf: @stephenneuendorffer because the default is to try to have as many workers and that is pretty greedy and can kill your development server.

trying to pass in LIT_ARGS="-j8" prepending the ninja check-aie with this does not work for how our tests are set up

stephenneuendorffer: This isn't really the right place to have that per-user config.

denolf: OK, I will remove from this PR, we can tackle this in another PR

fifield (approved):
I agree we should solve the user setting `-j` another way

test/CMakeLists.txt
stephenneuendorffer: lit uses multiprocessing.Pool in python, which queries sched_getaffinity to get the number of available cores, which can apparently be configured on a per-process basis by this: https://linux.die.net/man/7/cpuset

It's unclear how convenient that is, however, if there are many processes.   Ideally one could do something equivalent to 'nice', or ninja check_aie -j8 would use this mechanism to limit the number of cores for all child processes or something.

denolf:
merged #461

PR #468, opened 2023-05-04
single lock test

denolf:
formerly know as multi-rate, this is a focused test when only a single lock is used to synchronize host and AIE. Added AIE2 variant as well.

### DIFF
--- a/test/unit_tests/27_single_L1_single_lock/Makefile
+++ b/test/unit_tests/27_single_L1_single_lock/Makefile
@@ -1,0 +1,19 @@
+include ../../../tutorials/makefile-common
+
+.PHONY: all clean
+
+all: SingleLock.exe SingleLock_FIX.exe
+
+SingleLock.exe: test.cpp aie.mlir 
+	aiecc.py -v -j4 $(AIECC_FLAGS) $(word 2,$^) $(AIECC_HOST_FLAGS) ./$< -o $@
+
+SingleLock_FIX.exe: test.cpp aieWithWorkaround.mlir 
+	aiecc.py -v -j4 $(AIECC_FLAGS) $(word 2,$^) $(AIECC_HOST_FLAGS) ./$< -o $@
+
+SingleLock_aie2.exe: test.cpp aie2.mlir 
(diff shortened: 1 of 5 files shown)


stephenneuendorffer:
@AndraBisca @denolf This looks like a bug in LibXAIE.

LockAcquire eventually calls MaskPoll, because acquiring and releasing locks are actually implement using memory read operations.  The processor just polls until the memory location returns a 1.   MaskPoll looks like this:

```
	while (Count > 0U) {
		XAie_LinuxIO_Read32(IOInst, RegOff, &RegVal);
		if((RegVal & Mask) == Value) {
			Ret = XAIE_OK;
			break;
		}
		usleep(MinTimeOutUs);
		Count--;
	}

	/* Check for the break from timed-out loop */
	XAie_LinuxIO_Read32(IOInst, RegOff, &RegVal);
	if((Ret == XAIE_ERR) && ((RegVal & Mask) ==
			 Value)) {
		Ret = XAIE_OK;
	}

```
So basically rather than storing a variable to determine that the loop needs to exit, they just perform the (expensive) read again.  This works fine if reads are idempotent, but lock acquire/release are not.

denolf:
@stephenneuendorffer thanks for looking into this. I am not sure I follow your remark. It seems no interrupt driven mechanism is implemented in libXAIE. As such, the expensive polling for the lock though `XAie_LinuxIO_Read32` needs to happen until the timeout value is reached. 
Or are you trying to say that the compiler might have optimized something away here?

stephenneuendorffer:
The way that the lock is grabbed is solely through read operations.  You read the register, and if a bit is set, then you've acquired/released the lock.  The lock is released successfully in the loop, at which point it is possible that the AIE core completes its work and releases the lock.  The host then executes the final Read32 operation, releasing the lock a second time.

stephenneuendorffer:
I patched this locally and got it to work.  Unfortunately, because of linux driver version issues, it seems challenging to pick up the latest version of aie-rt and use it with the current board image.

denolf:
Got it now, thanks.
Upgrading to 2023.1 for both our board images and the Vitis tools (including aie-rt) should solve those conflicts.

AndraBisca (approved):
Looks good! Just one question: do we still want to keep the work around version as well if it will now work without it?

denolf:
merged #468

PR #472, opened 2023-05-05
[aievec] Split unaligned vector.transfer_read ops

jsetoain:
Add a pattern that splits unaligned transfer reads into two aligned transfer reads, two `vector.extract_strided_slice`, and two `vector.insert_strided_slice` ops.


david-vc (approved):
NIce work.

PR #473, opened 2023-05-05
Remove valid_xchess_license requirement from tests not using xchess

fifield:
Also fixup some error code handling logic.


fifield:
@abisca When I run the object fifo tests I see test 24 does not work more than 1 iteration of the while loop and 25 also fails.

AndraBisca:
Thank you for pointing that out, I will take a look!

AndraBisca (approved):

PR #474, opened 2023-05-08
Enable aievec.max and aievec::min for AIE-ML with different input and output types

linay-xsj:
Enable converting arith.minsi and arith.minf to aievec.min and converting arith.maxsi and arith.maxf to aievec.max for AIE-ML with different input and output types


david-vc (approved):

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2520,7 +2590,8 @@ LogicalResult CppEmitter::emitOperation(Operation &op, bool trailingSemicolon) {
                 aievec::SubElemOp, aievec::UPDOp, aievec::UPSOp,
                 aievec::FMAElemOp, aievec::MulElemOp, aievec::BroadcastOp,
                 aievec::BroadcastScalarOp, aievec::MulConvOp, aievec::FMAConvOp,
-                aievec::ShiftOp, aievec::ShuffleOp, aievec::CastOp>(
+                aievec::ShiftOp, aievec::ShuffleOp, aievec::CastOp,
+                aievec::MinOp, aievec::MaxOp>(
```
jsetoain, line 2594: Why are we using custom printers/parsers/verifiers for these ops? Table gen should be able to generate them for us.

jsetoain (changes requested):

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
jsetoain: > We need to translate ops into different intrinsics according to data types. If you want to use table gen only, you can change them in a batch.

That won't prevent us from using auto-generated parsers, printers, and verifiers. It would simplify our code significantly.

PR #478, opened 2023-05-09
[aievec] Auto-generate parser for MinOp

jsetoain:
Use tablegen to auto-generate parser, printer, and verifier for `aievec.min`.

### DIFF
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -632,7 +632,8 @@ def AIEVec_FMAConvOp:
 
 def AIEVec_MinOp:
   AIEVec_Op<"min", [
-    Pure
+    Pure,
+    AllTypesMatch<["lhs", "rhs", "result"]>
   ]>,
   Arguments<(ins AnyVector:$lhs, AnyVector:$rhs)>,
   Results<(outs AnyVector:$result)> {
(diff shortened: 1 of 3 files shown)


lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -1926,10 +1922,6 @@ template <typename T> LogicalResult verifyMinMaxOp(T op) {
   return success();
 }
 
-LogicalResult aievec::MinOp::verify() {
-  return verifyMinMaxOp<aievec::MinOp>(*this);
```
jsetoain, line 1930: Same for `verifyMinMaxOp`

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -1972,10 +1964,6 @@ ParseResult parseMinMaxOp(OpAsmParser &parser, OperationState &result) {
   return parser.addTypeToList(resultType, result.types);
 }
 
-ParseResult MinOp::parse(OpAsmParser &parser, OperationState &result) {
-  return parseMinMaxOp(parser, result);
```
jsetoain, line 1976: Same for `parseMinMaxOp`

jsetoain (commented):
@linay-xsj Something like this :-)

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -641,6 +642,8 @@ def AIEVec_MinOp:
     Xilinx-specific intrinsic that calculates the minimum between two input vectors.
     `$result = min(`$lhs, $rhs`).
   }];
+  let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($result)";
+  let hasVerifier = 0;
```
jsetoain, line 646: Also, if we migrate all the ops that can benefit from auto-generated verifiers, we should make that the default. But that one is for the future.

jsetoain:
Solved by a different PR.

PR #484, opened 2023-05-11
aiecc.py can compile simulation

stephenneuendorffer:
Previously, aiecc.py generated a simulation directory with a Makefile.  Running simulation required running 'make'.  However the Makefile was fixed and made some assumptions about the structure of a design.

Rather than fix the Makefile, this patch modifies aiecc.py to implement the compilation.  The intention here is that with (more or less) the same command line, one can retarget a design for simulation.  Currently, there is still some minor tweaking required:

synthesis:  aiecc.py aie.mlir main.cpp -o main.exe
simulation: aiecc.py --aiesim aie.mlir main.cpp

After simulation is compiled it can be run with:
aiesimulator --pkg-dir=aie.mlir.prj/sim

Note that currently the tests doesn't quite work.


fifield:
If `aiecc.py --aiesim aie.mlir main.cpp -o output` generated a script called `output` with `aiesimulator --pkg-dir=aie.mlir.prj/sim` inside, then our unit tests could "just work". They would call `%run_on_board ./output` with `sh` as the run_on_board substitution.

stephenneuendorffer:
> If `aiecc.py --aiesim aie.mlir main.cpp -o output` generated a script called `output` with `aiesimulator --pkg-dir=aie.mlir.prj/sim` inside, then our unit tests could "just work". They would call `%run_on_board ./output` with `sh` as the run_on_board substitution for the simulation target(s).

Yeah.. I'm intrigued by that idea too.. I wasn't sure whether it was going to make things simpler, or more complex, but I think it's worth trying.  (Main issue is that aiecc.py would need to start parsing the argument list, which it doesn't do today.  There are some aspects of the way that aiecc.py handles arguments anyway, so might want to refactor other things first before tackling that.)

stephenneuendorffer:
> > If `aiecc.py --aiesim aie.mlir main.cpp -o output` generated a script called `output` with `aiesimulator --pkg-dir=aie.mlir.prj/sim` inside, then our unit tests could "just work". They would call `%run_on_board ./output` with `sh` as the run_on_board substitution for the simulation target(s).
> 
> Yeah.. I'm intrigued by that idea too.. I wasn't sure whether it was going to make things simpler, or more complex, but I think it's worth trying. (Main issue is that aiecc.py would need to start parsing the argument list, which it doesn't do today. There are some aspects of the way that aiecc.py handles arguments anyway, so might want to refactor other things first before tackling that.)

So what I settled on was that --aiesim compiles the simulation of the executable you're building.  So, we generate a script, but it's independent from the rest of the system compilation.  The awkward part is with cross-compiling: you're not allowed to use --aiesim when cross-compiling.

stephenneuendorffer:
@jackl-xilinx Would appreciate your input on the failing AIE2 tests, and also what it will take to get the shim tests working for AIE1.

PR #485, opened 2023-05-11
[aievec] Add support for function declarations in aievec-to-cpp

jsetoain:
This change enables declaring and calling external functions from MLIR in our `aievec-to-cpp` translation flow.


david-vc (commented):
Hi @jsetoain : Could you please add a test case showing what got fixed here? Thanks.

jsetoain:
Done.

PR #488, opened 2023-05-11
Tutorial 2021.2 merge

jackl-xilinx:


jackl-xilinx:
This merge actually rolls back the sim Makefile that was updated to use the static test_lib_sim and instead builds from test_library.cpp source again. We will move this back to the static lib reference again once platform and static library libxaiengine are back in sync.

AndraBisca:
The merge also adds a fix to the AIE Targets for pathfinder JSON generation: specifically, it makes several functions DeviceOp aware.

PR #492, opened 2023-05-12
[NFC] Fix formatting of AIEDialect.md and AIEDialect.html

keryell:
At the end this is to fix https://xilinx.github.io/mlir-aie/AIEDialect.html


AndraBisca:
Thank you for all of these changes!

include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -100,6 +100,7 @@ def AIE_DeviceOp: AIE_Op<"device", [AIETarget,
       %tile = aie.tile(1, 1)
       %CORE = aie.core(%tile) { ... }
     }
+    ```
```
AndraBisca, line 103: This change is also included in #489.

PR #493, opened 2023-05-13
Pathfinder packets

searsm8:
Attempt to enable Pathfinder to route Packet flows. Summary of work is detailed in docs/Packet_Routing.md

### DIFF
--- a/docs/Packet_Routing.md
+++ b/docs/Packet_Routing.md
@@ -126,4 +126,6 @@ This approach combines as many packet rules as possible without adding extra flo
 
     c.  **Suggestion: add estimated bandwidth information to packet flows.**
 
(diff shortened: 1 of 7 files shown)


keryell:
Need some merge from upstream.

hunhoffe:
Reviewing against current main: I believe this capability landed via a different route — #1576 merged aie-create-packet-flows into aie-create-pathfinder-flows, and main now has the packet group IDs, channel sharing, and mask-rule generation this PR introduced, plus a full set of create-packet-flows lit tests. Overlapping source/dest grouping was later handled in #1694.

Since the feature is covered end-to-end on main, I'm closing this as superseded/stale. If there are missing features, happy to revisit later.

PR #496, opened 2023-05-16
Cmake Files Update

AndraBisca:


runtime_lib/CMakeLists.txt
```diff
@@ -39,7 +38,19 @@ foreach(target ${AIE_RUNTIME_TARGETS})
   message("Building AIE runtime for ${target} with TOOLCHAIN=${${target}_TOOLCHAIN_FILE}")
   
   # xaiengine
-  ExternalProject_Add(xaiengine_${target}
+  # if no LibXAIE_{target}_DIR specified, build xaiengine_${target} from Vitis
```
stephenneuendorffer, line 41: Needs better comment.

runtime_lib/CMakeLists.txt
```diff
@@ -39,7 +38,19 @@ foreach(target ${AIE_RUNTIME_TARGETS})
...
+  if (DEFINED LibXAIE_${target}_DIR)
+    message("Using xaiengine from LibXAIE_${target}_DIR=${LibXAIE_${target}_DIR}.")
+    #set(xaiengine_${target} ${LibXAIE_${target}_DIR})
+    ExternalProject_Add(xaiengine_${target}
```
stephenneuendorffer, line 45: Why do you need an externa project, if this is already built somewhere?  I'd think if you have something already built, you'd need some install rules?

AndraBisca:
Thank you for the review @stephenneuendorffer. The PR is currently still in draft and not ready for review, I will make more changes to it before then!

runtime_lib/CMakeLists.txt
fifield: My preference would be to point to an existing install somewhere, no building or copying needed, like we had in the past.

stephenneuendorffer:
> Thank you for the review @stephenneuendorffer. The PR is currently still in draft and not ready for review, I will make more changes to it before then!

Thanks for doing this :)

denolf:
@AndraBisca I tried to address some of the comment above, please have a look and test in multiple build variations. Thanks!

fifield:
The latest works for me with `-DLibXAIE_x86_64_DIR` although there are some erroneous `Vitis available, libxaie will be compiled and picked up locally` messages:

```
Using xaiengine from LibXAIE_x86_64_DIR=/scratch2/jefff/embeddedsw.
-- Found LibXAIE: /scratch2/jefff/embeddedsw/install/include  
-- FOUND LibXAIE: TRUE, XILINX_XAIE_LIB_DIR: /scratch2/jefff/embeddedsw/install/lib, XILINX_XAIE_INCLUDE_DIR: /scratch2/jefff/embeddedsw/install/include
...
-- Vitis available, libxaie will be compiled and picked up locally
```

fifield (approved):
It works for the use case I was interested in. Thanks!

denolf:
merged #496 in main.

PR #500, opened 2023-05-18
Convert arith.cmpi and arith.cmpf to aievec.cmp for different input data types

linay-xsj:
The change of translation part will be in the next change set with converting arith::select to aievec.sel.

### DIFF
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -677,7 +677,7 @@ def AIEVec_CmpOp:
   Results<(outs AnyUnsignedInteger:$result)> {
   let summary = "AIE vector comparison";
   let description = [{
-    Xilinx-specific intrinsic that compares operations between two input vectors.
+    Xilinx-specific intrinsic that performs element-wise comparisonof two input vectors.
     The attribute predicate defines which type of comparison is
     performed. The following comparisons are supported:
 
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -683,10 +683,14 @@ def AIEVec_CmpOp:
 
     -   equal (mnemonic: `"eq"`)
     -   not equal (mnemonic: `"ne"`)
-    -   less than (mnemonic: `"lt"`)
-    -   less than or equal (mnemonic: `"le"`)
-    -   greater than (mnemonic: `"gt"`)
-    -   greater than or equal (mnemonic: `"ge"`)
+    -   signed less than (mnemonic: `"slt"`)
+    -   unsigned less than (mnemonic: `"ult"`)
+    -   signed less than or equal (mnemonic: `"sle"`)
+    -   unsigned less than or equal (mnemonic: `"ule"`)
+    -   signed greater than (mnemonic: `"sgt"`)
+    -   unsigned greater than (mnemonic: `"ugt"`)
+    -   signed greater than or equal (mnemonic: `"sge"`)
+    -   unsigned greater than or equal (mnemonic: `"uge"`)
   }];
 
   let assemblyFormat = "$lhs `,` $rhs ` ` `{` `pred` `=` $pred attr-dict `}` `:` type($lhs) `,` type($rhs)  `,` type($result)";
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -667,4 +667,29 @@ def AIEVec_MaxOp:
   let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($result)";
   let hasVerifier = 0;
 }
+
+def AIEVec_CmpOp:
+  AIEVec_Op<"cmp", [
+    Pure,
+    AllTypesMatch<["lhs", "rhs"]>
+  ]>,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, DefaultValuedStrAttr<StrAttr, "">:$pred)>,
+  Results<(outs AnyUnsignedInteger:$result)> {
+  let summary = "AIE vector comparison";
+  let description = [{
+    Xilinx-specific intrinsic that compares operations between two input vectors.
+    The attribute predicate defines which type of comparison is
+    performed. The following comparisons are supported:
+
+    -   equal (mnemonic: `"eq"`)
+    -   not equal (mnemonic: `"ne"`)
+    -   less than (mnemonic: `"lt"`)
+    -   less than or equal (mnemonic: `"le"`)
+    -   greater than (mnemonic: `"gt"`)
+    -   greater than or equal (mnemonic: `"ge"`)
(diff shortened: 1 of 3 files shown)


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -1119,6 +1119,121 @@ using LowerVectorMinFOpToAIEVecMinOp =
...
+    Type type =
+        mlir::IntegerType::get(srcOp.getContext(), laneSize <= 32 ? 32 : 64,
+                               mlir::IntegerType::Unsigned);
+
+    Location loc = srcOp.getLoc();
+    Value lhs = srcOp.getLhs();
+    Value rhs = srcOp.getRhs();
+    CmpTy pred = srcOp.getPredicate();
+
+    arith::CmpIPredicate ipred = convertToIntegerPredicate(pred);
```
david-vc, line 1214: Do you allow the pattern to apply if the LHS is signed and the RHS is unsigned?

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -667,4 +667,29 @@ def AIEVec_MaxOp:
...
+  Results<(outs AnyUnsignedInteger:$result)> {
+  let summary = "AIE vector comparison";
+  let description = [{
+    Xilinx-specific intrinsic that compares operations between two input vectors.
+    The attribute predicate defines which type of comparison is
+    performed. The following comparisons are supported:
+
+    -   equal (mnemonic: `"eq"`)
+    -   not equal (mnemonic: `"ne"`)
+    -   less than (mnemonic: `"lt"`)
```
david-vc, line 686: For arith.cmpi, the operands are always Signless and the predicate specifies whether the operands are interpreted to be either signed, e.g., slt, or unsigned, e.g., ult.
I think that we need to follow the same scheme, otherwise the transform will not preserve semantics.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -1119,6 +1119,121 @@ using LowerVectorMinFOpToAIEVecMinOp =
...
+                                      CmpIPredicate pred, Location loc,
+                                      Type type, Value lhs, Value rhs) {
+  switch (pred) {
+  case CmpIPredicate::eq:
+    return rewriter.create<aievec::CmpOp>(loc, type, lhs, rhs, "eq");
+  case CmpIPredicate::ne:
+    return rewriter.create<aievec::CmpOp>(loc, type, lhs, rhs, "ne");
+  case CmpIPredicate::slt:
+  case CmpIPredicate::ult:
+    return rewriter.create<aievec::CmpOp>(loc, type, lhs, rhs, "lt");
```
david-vc, line 1163: This does not preserve semantics. slt and ult cannot both be mapped to "lt".
For arith.cmpi the operands are SignLess and the interpretation of whether they both are signed or whether they both are unsigned is specified by the predicate.

david-vc (changes requested):
Hi Lina, I think that we need to differentiate between signed and unsigned comparisons. See more details in comments.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
linay-xsj: According to aie-ml intrinsics, we do not handle the comparison between unsigned and signed. In AIEVecOps.td, I also request for AllTypesMatch for lhs and rhs.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
linay-xsj: The signless type in MLIR allows values to be treated as either signed or unsigned depending on the context. If the operands are signed, it is not allowed to use unsigned operator. The reverse is also true. In our aie-ml intrinsics, they are use the same operator for unsigned and signed comparison depends on the operands' data types. I believe the current setting is good for the semantics.

david-vc: @linay-xsj : in MLIR, i32 is a 32-bit integer type with no signedness semantics. See BuiltinTypes.td; operations like comparison require signedness semantics. For arith.cmp the signedness semantics are NOT provided by the operand type, but instead they are provided by the predicate (slt vs ult). When converting an arith.cmp to an aievec.comp, we are not changing the types of the operands. Consequently the signedness semantics must be conveyed via the predicate of the aievec.cmp. Thus, "signed less than" and "unsigned less than" must use different predicates.
@jsetoain : Could you also please comment on this.

jsetoain: I agree with you, David. Coalescing `<s|u><cmp>` down to `<cmp>` loses signedness semantics. It just so happens to work here because we're lowering the comparison down to C++ from here, and the types will be "signful" in C++-land, which will allow the right function call to be invoked. But that's something that we break here, and gets coincidentally fixed in ***our*** _current_ next step. It may (and probably will) break in pretty much every other instance; e.g.: when lowering to LLVM.



> The signless type in MLIR allows values to be treated as either signed or unsigned depending on the context.

That is exactly right, and the context in this case is provided by the predicate. Removing signedness from the predicate breaks semantics. Information about signedness (whether the comparison is signed or unsigned) is lost, and we still want to keep that information at this level.

I understand we want to keep the translation simple, so we avoid "one to many" translations. But "many to one translations" do not add complexity to the translation.

jsetoain (changes requested):

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -667,4 +667,29 @@ def AIEVec_MaxOp:
   let assemblyFormat = "$lhs `,` $rhs attr-dict `:` type($result)";
   let hasVerifier = 0;
 }
+
+def AIEVec_CmpOp:
+  AIEVec_Op<"cmp", [
+    Pure,
+    AllTypesMatch<["lhs", "rhs"]>
+  ]>,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs, DefaultValuedStrAttr<StrAttr, "">:$pred)>,
```
jsetoain, line 676: We are abusing the use of attributes. If it must always be present, the predicate should be another operand, not an attribute. You probably use the same type constraints that `arith.cmp` uses for this.

david-vc (commented):
@linay-xsj , @jsetoain : 

Is there a clean way to provide a translation from arith.cmpi to C using the aie compare intrinsics without introducing a aievec.cmp? 
I'm seeing that aievec.cmp or (aievec.cmpi) is really a more restricted version of arith.cmpi with a translation to aievec.cmp. So it is duplicating a lot of functionality arith.cmpi already provides.
Maybe the answer of my question is 'no', and then I'm fine introducing a aievec.cmp. In this case, however, we should align this much more closely with airth.cmp, and for example, specify the predicate as the first operand.

david-vc (approved):

PR #503, opened 2023-05-25
ShimDMA Allocation Info

AndraBisca:
Currently, higher-level mlir-aie abstractions like the objectFifo, as well as higher-level dialects external to mlir-aie, may do shimDMA allocations as part of their lowering to physical-level mlir-aie. Because of this, there is a need to keep track of which shimDMA and which of its channels were assigned to the original operations, a feature that will be useful in writing and testing designs.

This PR introduces the ShimDMAAllocationOp and through its inputs it makes the connection between the symbolic name of a higher-level op and the channelDir, channelIndex and column of the shimDMA to which it was assigned. This allocation op is removed as part of the AIECoreToStandard.cpp.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIE.td
+++ b/include/aie/Dialect/AIE/IR/AIE.td
@@ -1277,8 +1277,36 @@ def AIE_PutCascadeOp: AIE_Op<"putCascade", [HasParent<"CoreOp">]> {
   let assemblyFormat = [{ `(` $cascadeValue `:` type($cascadeValue) `)` attr-dict }];
 }
 
+def AIE_ShimDMAAllocationInfoOp : AIE_Op<"shimDMAAllocationInfo", [HasParent<"ModuleOp">]> {
+  let summary = "Runtime allocation information for a single shim DMA";
+  let description = [{
+    Runtime allocation information for a single shim DMA. It contains attributes for the
+    sym_name of the operation which generated the shim DMA, for the DMAChannelDir and 
+    channel index, and for the column of the shim tile to which the originating operation 
+    was mapped.
 
-
+    Example:
+    ```
+      %tile00 = AIE.tile(0, 0)
+      %tile02 = AIE.tile(0, 2)
+      %connect1 = AIE.objectFifo.createObjectFifo(%tile00, {%tile02}, 2) {sym_name = "of_in_0"} : !AIE.objectFifo<memref<64xi16>>
+    ```
+    could produce the following allocation info (channel direction MM2S, channel index 1, and shim column 0):
+    ```
+      AIE.shimDMAAllocationInfo("of_in_0", MM2S, 1, 0)
+    ```
+  }];
+  let arguments = (
+    ins StrAttr:$sym_name,
+        DMAChannelDir:$channelDir,
+        I64Attr:$channelIndex,
+        I64Attr:$col
+  );
+  let results = (outs);
+  let assemblyFormat = [{
+    `(` $sym_name `,` $channelDir `,` $channelIndex `,` $col `)` attr-dict  
+  }];
+}
 
 def AIE_ObjectFifoCreateOp: AIE_Op<"objectFifo.createObjectFifo", []> {
   let summary = "Create a circular buffer or channel between two tiles";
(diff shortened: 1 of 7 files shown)


include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -1277,8 +1277,36 @@ def AIE_PutCascadeOp: AIE_Op<"putCascade", [HasParent<"CoreOp">]> {
...
-
+    Example:
+    ```
+      %tile00 = AIE.tile(0, 0)
+      %tile02 = AIE.tile(0, 2)
+      %connect1 = AIE.objectFifo.createObjectFifo(%tile00, {%tile02}, 2) {sym_name = "of_in_0"} : !AIE.objectFifo<memref<64xi16>>
+    ```
+    could produce the following allocation info (channel direction MM2S, channel index 1, and shim column 0):
+    ```
+      AIE.shimDMAAllocationInfo("of_in_0", MM2S, 1, 0)
```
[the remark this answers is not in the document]
AndraBisca: I'm open to changing this at any point! Would it be better if I added this op to AIEX instead?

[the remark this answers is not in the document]
stephenneuendorffer: > I see what you mean. However, when lowering from mlir-air for example, there are cases where the shimDMA stub (and the dmaStart op) aren't generated. In those cases we still need to keep the allocation info somewhere.

OK, to me this is the key conceptual thing I never quite wrapped my head around.  Let's just add that do the documentation. perhaps: "This op exists for cases where shimDMA configuration is performed outside of MLIR and hence there is no appropriate dmaStart operation"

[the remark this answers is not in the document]
AndraBisca: Sounds good! Will add now.

lib/Targets/AIETargets.cpp
```diff
@@ -153,6 +177,31 @@ void registerAIETranslations() {
...
       registerDialects);
 
+  TranslateFromMLIRRegistration registrationShimDMAToJSON(
+      "aie-generate-json", "Transform AIE shim DMA allocation info into JSON",
+      [](ModuleOp module, raw_ostream &output) {
+        llvm::json::Object moduleJSON;
+        for (auto shimDMA_meta : module.getOps<ShimDMAAllocationOp>()) {
+          llvm::json::Object shimJSON;
+          ShimDMAAllocationOpAdaptor shimDMAAllocationOpAdaptor(shimDMA_meta);
+          auto channelDir = shimDMAAllocationOpAdaptor.getChannelDirAttr();
```
fifield, line 187: I don't think you need the extra adaptor here. `shimDMA_meta.getChannelDirAttr();` should work.

lib/Targets/AIETargets.cpp
```diff
@@ -153,6 +177,31 @@ void registerAIETranslations() {
       },
       registerDialects);
 
+  TranslateFromMLIRRegistration registrationShimDMAToJSON(
+      "aie-generate-json", "Transform AIE shim DMA allocation info into JSON",
```
fifield, line 181: `aie-translate -aie-generate-json` needs a test.

fifield (approved):

lib/Targets/AIETargets.cpp
AndraBisca: I accidentally merged in the PR, but I will resolve this in a new fix PR.

lib/Targets/AIETargets.cpp
AndraBisca: Same as above.

fifield:
I started using this a bit and a few things struck me:
* The `ShimDMAAllocationOp` probably should have `AIE.device` as its parent, not `module`
* We should probably use the symbol of the objectfifo in the `ShimDMAAllocationOp`, not the string name of the symbol. The current design lends itself to writing string matching code to match up operations. But we should use the symbol table instead since that's what it's for.

AndraBisca:
Thank you for the feedback! I agree with all of the points you mentioned and I think it makes more sense to use the symbol with objectFifos, just like air channels. I'll start working on PRs to implement these changes.

PR #506, opened 2023-05-30
Create canFoldMulAddChainToConvOpAnalysis for FoldMulAddChainToConvOp pass in vectorizer

linay-xsj:


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -1427,10 +1427,9 @@ populateAIEVecV2TransformationPatterns(RewritePatternSet &patterns) {
   patterns.add<FoldAIEShiftAndBroadcast>(patterns.getContext());
 }
 
-static void
-populateAIEVecConvOpTransformationPatterns(RewritePatternSet &patterns,
-                                           unsigned shiftParam) {
-  patterns.add<FoldMulAddChainToConvOpPattern>(patterns.getContext(),
+static void populateAIEVecConvOpTransformationPatterns(
+    RewritePatternSet &patterns, AnalysisManager &am, unsigned shiftParam) {
+  patterns.add<FoldMulAddChainToConvOpPattern>(patterns.getContext(), am,
```
jsetoain, line 1432: The declaration of this function should be in `FoldMulAddChainToConvOp.h`, and its implementation in `FoldMulAddChainToConvOp.cpp`.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -1711,17 +1710,13 @@ configureAIEVecV2TransformationLegalizations(ConversionTarget &target) {
...
-    SmallVector<SmallVector<arith::MulIOp, 8>, 8> groupFusedOps;
-    MulDefMapTy macChainMap;
-    unsigned dupFactor = 1;
-    bool hasMulConv = false;
-    Value acc = nullptr;
-    return !canFoldMulAddChainToConvOp(op, macChainMap, groupFusedOps,
-                                       dupFactor, hasMulConv, acc);
+  target.addDynamicallyLegalOp<arith::AddIOp>([&am](arith::AddIOp op) {
+    return !am.getChildAnalysis<canFoldMulAddChainToConvOpAnalysis>(op)
+                .canFoldMulAddChainToConvOp;
```
jsetoain, line 1719: Same here, `configure<Legalizations>` should be declare in the header file and implemented in the corresponding source file.

lib/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.h
```diff
@@ -0,0 +1,330 @@
...
+      }
+      acc = convOp->getResult(0);
+    }
+
+    llvm_unreachable("the conversion should end with srs op.");
+  }
+
+  AnalysisManager &am;
+  unsigned shiftParam;
+};
```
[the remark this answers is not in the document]
jsetoain: For completion, as part of the declaration of those two functions, please add their documentation (what kind of conversions/rewrites they populate, and what kind of legality they configure).

jsetoain (changes requested):
We are much closer to what we need, but we're not quite there yet. Just a couple more changes 🙂 

Thanks!

jsetoain (approved):
Thanks, Lina! 🙂

PR #508, opened 2023-06-05
Change default device if AIE-ML compilation is demanded by user

andrej:
Currently, if no `AIE.device` op is present in MLIR, the "Canonicalize Device" pass automatically adds a **default device** `AIE.device(xcvc1902)`. **This default device is an AIE1 device.**

However, if the user wants to compile AIE-ML (AIE2) code, this device will not work for simulation, since it is a first generation AIE device. The user can currently request AIE-ML compilation if `aiecc.py` is called with `--aie-target=AIE2`.

**Thus, if the user explicitly asks for `--aie-target=AIE2`, I suggest using a more appropriate default device that supports AIE-ML.**

This pull request ...

1. ... adds an option `--aie-target` to the `AIECanonicalizeDevice` pass, which takes AIE (AIE) or AIE2 as values.
2. ... makes `aiecc.py` pass the `--aie-target` option it receives on to the `AIECanonicalizeDevice` pass.
3. ... changes the default device added in the `AIECanonicalizeDevice` to the `VE2802` instead of `VC1902` if the target architecture passed in is `AIE2`.
4. ... updates tutorial documentation to reflect this.

This pull request does not change any behavior if ...
- the target is not explicitly specified (AIE is assumed) or AIE compilation is specified
- an explicit AIE.device op is present in the user's MLIR

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPasses.td
+++ b/include/aie/Dialect/AIE/Transforms/AIEPasses.td
@@ -45,6 +45,13 @@ def AIECanonicalizeDevice : Pass<"aie-canonicalize-device", "ModuleOp"> {
   }];
 
   let constructor = "xilinx::AIE::createAIECanonicalizeDevicePass()";
+
+  let options = [
+    Option<"aieTarget", "aie-target", "xilinx::AIE::AIEArch", "xilinx::AIE::AIEArch::AIE1",
+           "Target architecture of AIE device (AIE1 or AIE2 aka AIE-ML).",
+           "::llvm::cl::values(clEnumValN(xilinx::AIE::AIEArch::AIE1, \"AIE\", \"First generation AIE.\"),\n"
+           "                   clEnumValN(xilinx::AIE::AIEArch::AIE2, \"AIE2\", \"Second generation AIE-ML.\"))">
+  ];
 }
 
 def AIECoreToStandard : Pass<"aie-standard-lowering", "ModuleOp"> {
--- a/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
@@ -39,10 +39,18 @@ struct AIECanonicalizeDevicePass
     // the new op quite yet.
     OpBuilder builder(moduleOp->getContext());
 
+    // If no device op is present, we assume the VC1902 device by default, or
+    // the VE2802 device if AIE-ML compilation is demanded by the user 
+    // (i.e. flag --aie-target=AIE2 set).
+    AIEDevice defaultDevice = AIEDevice::xcvc1902;
+    if(AIEArch::AIE2 == aieTarget) {
+      defaultDevice = AIEDevice::xcve2802;
+    }
+
     Location location = builder.getUnknownLoc();
     DeviceOp deviceOp = builder.create<DeviceOp>(
         location,
-        AIEDeviceAttr::get(builder.getContext(), AIEDevice::xcvc1902));
+        AIEDeviceAttr::get(builder.getContext(), defaultDevice));
 
     deviceOp.getRegion().takeBody(moduleOp.getBodyRegion());
     new (&moduleOp->getRegion(0)) Region(moduleOp);
--- a/tools/aiecc/aiecc/cl_arguments.py
+++ b/tools/aiecc/aiecc/cl_arguments.py
@@ -87,6 +87,7 @@ def parse_args():
     parser.add_argument('--aie-target',
             dest="aie_target",
             default="AIE",
+            choices=["AIE", "AIE2"],
             help='Target architecture of the AIE program')
     parser.add_argument('--compile-host',
             dest="compile_host",
--- a/tools/aiecc/aiecc/main.py
+++ b/tools/aiecc/aiecc/main.py
@@ -336,7 +336,7 @@ class flow_runner:
         self.file_with_addresses = os.path.join(self.tmpdirname, 'input_with_addresses.mlir')
         await self.do_call(progress.task, ['aie-opt',
                                           '--lower-affine',
-                                          '--aie-canonicalize-device',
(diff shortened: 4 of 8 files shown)


fifield:
I solved a similar problem in another branch by adding:
```
let options = [
    Option<"clDevice", "device", "std::string",
          /*default=*/"\"xcvc1902\"",
           "AIE device to target.">,
  ];
```
and,
```
diff --git a/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp b/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
index 9ef29b6a..bb6d71fb 100644
--- a/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
@@ -29,6 +29,13 @@ struct AIECanonicalizeDevicePass
 
     ModuleOp moduleOp = getOperation();
 
+    auto device = AIE::symbolizeAIEDevice(clDevice);
+    if (!device) {
+      moduleOp.emitOpError("Invalid AIE.device option");
+      signalPassFailure();
+      return;
+    }
+
     for (auto deviceOp : moduleOp.getOps<AIETarget>()) {
       (void)deviceOp;
       // We have a device..  Nothing to do.
@@ -42,7 +49,7 @@ struct AIECanonicalizeDevicePass
     DeviceOp deviceOp = builder.create<DeviceOp>(
         location,
+        AIEDeviceAttr::get(builder.getContext(), *device));
```
I'm not sure what the better approach is, or really why we want this at all long term instead of mandatory `AIE.device`

stephenneuendorffer:
Hmm.. I've actually been celebrating getting to remove the --aie-target option from aiecc, now that we have AIE.device...  (see https://github.com/Xilinx/mlir-aie/pull/484/files#diff-5e288757c4de5f1a96dbeb353c98a873482af0f20449849b77f529747a2f7e10)

I think the fundamental question is how useful is it to take mlir-aie designs and attempt to 'retarget' them to a different device or device family.  Currently there are almost no example designs for which this is the case, so my inclination is not to try to go down that path with mlir-aie and instead leave a 'device portable' abstraction to mlir-air.  I'm curious if others feel differently?  

My intention with the current implementation is to generally expect an AIE.device operation, but to allow not having one for a short period.

andrej:
Good points. If the long term goal is to force everyone to have an `AIE.device`, this change would be a step in the wrong direction.

I thought that maybe the idea of leaving out the `AIE.device` operation was to allow "portable" AIE MLIR that could be used on different devices. However, with the differences in architectures it may not be possible to have actual portable code without more abstractions.

One thing that pops in my head is that a nice additional "feature" could be to allow multiple `AIE.device` declarations and have MLIR code for different devices in one file (e.g. for the tutorials we could have an AIE and an AIE-ML block). During compilation, only the block that was targeted would be compiled.

I also see that my code did not pass the code format check. The message references a `clang-format.patch`, but I don't know where to see that. Maybe someone can help me with this but it may not be necessary if this is rejeced.

jgmelber:
I agree that this is a problem that requires a solution for the tutorials, maybe it is as simple as including another MLIR design file targeting AIE2. 

> I also see that my code did not pass the code format check. The message references a `clang-format.patch`, but I don't know where to see that. Maybe someone can help me with this but it may not be necessary if this is rejeced.

After adding files with changes, and before committing run: `git clang-format`

andrej:
> I agree that this is a problem that requires a solution for the tutorials, maybe it is as simple as including another MLIR design file targeting AIE2.

That was something I was considering as well. I can do that instead. 

Any thoughts on if I should add a `AIE.device` to the tutorial 1 code? If we want to require device ops for all files in the future anyways, it might be a good idea not to teach new people to leave it out. On the other hand, it would make the code slightly more complicated, maybe that's why it was left out for the first tutorial.

> > I also see that my code did not pass the code format check. The message references a `clang-format.patch`, but I don't know where to see that. Maybe someone can help me with this but it may not be necessary if this is rejeced.
> 
> After adding files with changes, and before committing run: `git clang-format`

Thank you!

stephenneuendorffer:
> One thing that pops in my head is that a nice additional "feature" could be to allow multiple AIE.device declarations and have MLIR code for different devices in one file (e.g. for the tutorials we could have an AIE and an AIE-ML block). During compilation, only the block that was targeted would be compiled.

My vague thinking at this point is that multiple AIE.device nodes would target multiple devices from the same design, but we haven't tried this out yet.  Your idea is intriguing, but seems like something which could be almost as simply build using multiple mlir files?

> I also see that my code did not pass the code format check. The message references a `clang-format.patch`, but I don't know where to see that. Maybe someone can help me with this but it may not be necessary if this is rejeced.

The easiest thing is to click on 'details' and then scroll down: one of the last steps displays the patch.  I tend to ignore this and just run 'git clang-format', but there can be minor differences if you don't have the same clang version as the CI (clang 12)

stephenneuendorffer:
> Any thoughts on if I should add a `AIE.device` to the tutorial 1 code? If we want to require device ops for all files in the future anyways, it might be a good idea not to teach new people to leave it out. On the other hand, it would make the code slightly more complicated, maybe that's why it was left out for the first tutorial.

Great idea! should definitely happen as we're doing more with AIE2.

andrej:
Thanks for the really quick feedback!

I think I will close this since it seems to be a step in the wrong direction. I think making `AIE.device` mandatory is a good idea too.

One last thought that I have about this: Is the goal for AIE MLIR to be portable _within_ the same architecture? I.e. should any AIE1 MLIR run on any AIE1 device, and the same for AIE-MLIR? If so, having a mandatory `AIE.device` could maybe instead be replaced with a mandatory `AIE.arch`, which would be less specific/restrictive.

> Your idea is intriguing, but seems like something which could be almost as simply build using multiple mlir files?

That is true. I suggested my idea because I thought it could enable implementing higher level abstractions that then simultaneously lower to one code block for _each_ of the architectures. Either way it's not relevant to the issue at hand, especially if you already have other uses for multiple device ops in mind.

Thanks again!

PR #510, opened 2023-06-07
Testlib cleanup for AIE2

stephenneuendorffer:


fifield:
With this can we assume it's safe to start removing mentions of `LIBXAIENGINEV2` in other places?

PR #512, opened 2023-06-08
ObjectFIFO broadcast with multiple depths

AndraBisca:
Currently, the objectFifo lowering makes assumptions about the number of AIE.buffers allocated on each tile in an objectFifo based on the maximum number of objects acquired by the code running on the tile. This allows users to only specify a single objectFifo depth for broadcasting, as the depth for each individual consumer is chosen by the lowering.

This PR extends the ObjectFifoCreateOp with the ability to take as input either a single integer depth to be applied to all tiles, or an ArrayAttr with multiple depths for each individual tile. This offers the user the possibility to bypass the simple mechanism described above and explicitly pass to the lowering a depth for each consumer of a broadcast.


include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -64,6 +64,8 @@ def AIE_ObjectFifoSubviewType :
 
 def AIE_Type : AnyTypeOf<[AIE_ObjectFifoType, AIE_ObjectFifoSubviewType]>;
 
+def AIE_ObjectFifo_Depth : AnyAttrOf<[ConfinedAttr<I64Attr, [IntMinValue<0>]>, ArrayAttr]>;
```
stephenneuendorffer, line 67: Why I64?  even I32 is alot of buffers.

include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -1319,6 +1321,10 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectFifo.createObjectFifo", []> {
     buffers (of given output type) and their locks in the Memory Module of the appropriate tile(s) after lowering, 
     based on tile-adjacency. These elements represent the conceptual depth of the objectFifo.
 
+    For the producer and for each consumer, a different size (i.e., element number) can be specified. This will
```
stephenneuendorffer, line 1324: Could we get the best of both worlds by keeping the simple analysis here and not requiring the sizes to be specified?  What's a case that the analysis gets wrong?

AndraBisca: Currently, the simple analysis I implemented always looks for the maximum acquire number for an objFifo in a core and then adds +1 to it to account for pre-fetching. Right now, even if a size is specified it may be ignored if it's larger than the number calculated before.

This update specifically concerns broadcasting. Broadcast has multiple consumers but doesn't account for any connections between them, just between the producer and each consumer. So for example, if consumers A and B receive data through broadcast from a producer P using ping pong buffering, but B also needs data from A (calculated based on data from P), higher concurrency may be achieved if instead of a ping pong buffer B had a circular buffer of size 3 (this way, P can push data faster to both consumers).

stephenneuendorffer: OK, then I would tend to view this as an optimization:  'Most of the time you shouldn't need to specify the size, because the the analysis will compute a minimum safe value.  However it's also possible to specify a larger size, which may increase performance in some cases'.  this leaves open the possibility of a more complex analysis that would take into account the situation you describe.

AndraBisca: Yes, that makes sense! I'll update the PR.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -999,14 +958,17 @@ struct AIEObjectFifoStatefulTransformPass
...
+          consumerDepth = createOp.size(consumerIndex + 1);
         builder.setInsertionPointAfter(createOp);
         AIEObjectFifoType datatype =
             createOp.getType().cast<AIEObjectFifoType>();
 
+        std::vector<Attribute> cosumerObfFifoSize = {
+            builder.getI32IntegerAttr(consumerDepth)};
         ObjectFifoCreateOp consumerFifo = createObjectFifo(
-            builder, datatype, consumerTile, consumerTile, consMaxAcquire);
+            builder, datatype, consumerTile, consumerTile, cosumerObfFifoSize);
```
keryell, line 971:
```suggestion
            builder, datatype, consumerTile, consumerTile, consumerObfFifoSize);
```

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -999,14 +958,17 @@ struct AIEObjectFifoStatefulTransformPass
...
-        int consMaxAcquire =
-            findObjectFifoSize(device, consumerTileOp, createOp);
+        if (isa<ArrayAttr>(createOp.getElemNumber()))
+          // +1 to account for 1st depth (producer)
+          consumerDepth = createOp.size(consumerIndex + 1);
         builder.setInsertionPointAfter(createOp);
         AIEObjectFifoType datatype =
             createOp.getType().cast<AIEObjectFifoType>();
 
+        std::vector<Attribute> cosumerObfFifoSize = {
```
keryell, line 968:
```suggestion
        std::vector<Attribute> consumerObfFifoSize = {
```

AndraBisca:
Based on the feedback so far I added back the simple analysis for the objectFifo size. 
Now, unless sizes are specified for every consumer tile of a broadcast, the lowering will infer the sizes as before. This is also true when there is only one consumer.

I did, however, enforce the generation of DMA regions even when a core isn't provided because an objectFifo depth was provided initially and the locks and object values can still be controlled from the host code.

fifield (approved):

PR #514, opened 2023-06-12
Implement vector reduction add, reduction min, and reduction max in auto vectorizer

linay-xsj:
This change set introduces several enhancements to the auto vectorizer, focusing on 
1. empowering the auto vectorizer with the capability to perform reduction add, reduction min, and reduction max operations across various same input and output data types, including i32, i16, i8, float, and bfloat. 
2. aiming to enhance the aie translation process by addressing initialization with constant dense splat data through the utilization of broadcast and broadcast_zero_ methods.
3. handling the aie translation for memref.store operation.
4. introducing a new operation called ExtElemOp, which allows for the extraction of an element from a vector.


lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -1281,6 +1281,165 @@ struct LowerVectorSelectOpToAIEVecSelOp
...
+        auto lUpsOp = rewriter.create<aievec::UPSOp>(loc, accType, srcVec);
+        auto rUpsOp = rewriter.create<aievec::UPSOp>(loc, accType,
+                                                     shiftBytesOp.getResult());
+        auto elemOp = rewriter.create<aievec::AddElemOp>(
+            loc, lUpsOp->getResult(0).getType(), lUpsOp->getResult(0),
+            rUpsOp->getResult(0));
+        auto curOp =
+            rewriter.create<aievec::SRSOp>(loc, vType, elemOp.getResult());
+        shiftIndex /= 2;
+        while (shiftIndex) {
```
keryell, line 1419: Some `do loop` in the neighborhood too?

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2438,6 +2490,44 @@ bool CppEmitter::hasBlockLabel(Block &block) {
...
+  for (APFloat elem : dense) {
+    denseValues.push_back(elem.convertToFloat());
+  }
+  ElTy firstV = denseValues[0];
+
+  if (llvm::all_of(denseValues, [firstV](ElTy v) { return v == firstV; })) {
+    firstValue = std::to_string(firstV);
+
+    if (firstValue == "inf") {
+      firstValue = "3.40282347e+38F";
```
keryell, line 2521: I am curious about this magical value. Is this something that could come from https://en.cppreference.com/w/cpp/types/numeric_limits or from LLVM target description?

linay-xsj: This change set is not ready for review, Ronan. I haven't requested anyone for review yet. I thought no one can see this. Once I finish it, I will add you for review.

keryell: Ah yes, this is actually a draft. Sorry for dropping by. :-)

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2438,6 +2490,63 @@ bool CppEmitter::hasBlockLabel(Block &block) {
...
+  }
+  ElTy firstV = denseValues[0];
+  if (llvm::all_of(denseValues, [firstV](ElTy v) { return v == firstV; })) {
+    firstValue = std::to_string(firstV);
+    return true;
+  }
+  return false;
+}
+
+static bool hasSameDenseValueOfFloat(DenseFPElementsAttr dense,
```
jamestclin, line 2508: For getting splat value, I think you can leverage isSplat() and getSplatValue(). See https://mlir.llvm.org/doxygen/classmlir_1_1DenseElementsAttr.html.

linay-xsj: Thanks, James.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2438,6 +2493,50 @@ bool CppEmitter::hasBlockLabel(Block &block) {
...
+  }
+  ElTy firstV = denseValues[0];
+  if (llvm::all_of(denseValues, [firstV](ElTy v) { return v == firstV; })) {
+    firstValue = std::to_string(firstV);
+    return true;
+  }
+  return false;
+}
+
+static bool hasSameDenseValueOfFloatOrBFloat16(DenseFPElementsAttr dense,
```
david-vc, line 2511: Please add comment to explain what this function is actually computing. It is not obvious from the name.
My suggestion is to refactor this into a function that just retrieves the first element from a DenseValue as a string. The caller will first test dense.isSplat() and then call the new API.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2483,9 +2582,61 @@ LogicalResult CppEmitter::emitAttribute(Location loc, Attribute attr) {
...
+          bool hasSameValue = false;
+          std::string firstValue = "";
+          if (width == 32) {
+            hasSameValue =
+                hasSameDenseValueOfFloatOrBFloat16(dense, firstValue);
+          } else if (width == 16) {
+            hasSameValue = hasSameDenseValueOfFloatOrBFloat16(
+                dense, firstValue, /*isBFloat*/ true);
+          }
+          if (hasSameValue &&
```
david-vc, line 2598: Is the case where hasSameValue == false unreachable?
Should we assert(hasSameValue) ?
Should we return failure() if !hasSameValue?

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2520,18 +2671,48 @@ LogicalResult CppEmitter::emitAttribute(Location loc, Attribute attr) {
...
           os << "()";
           return success();
         }
-        os << '{';
-        interleaveComma(dense, os, [&](const APInt &val) {
-          printInt(val, shouldMapToUnsigned(iType.getSignedness()));
-        });
-        os << '}';
+
+        if (AIEML) {
```
david-vc, line 2688: Looks like the 'else' branch handles the cases where there are multiple values.
Why are we not handling multiple values for the case AIEML=true?
Why are we not asserting that if AIEML, then there should be only a single value?

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2520,18 +2671,48 @@ LogicalResult CppEmitter::emitAttribute(Location loc, Attribute attr) {
...
+          bool hasSameValue = false;
+          std::string firstValue = "";
```
david-vc, line 2690: I find the hasSameDenseValue API quite confusing.
Perhaps instead we can do
if (dens.isSplat()) {
  // get the first value
  // broadcast it 
} else {
  // produce list { , }
}

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2438,6 +2493,50 @@ bool CppEmitter::hasBlockLabel(Block &block) {
   return blockMapper.count(&block);
 }
 
+template <typename ElTy>
+static bool hasSameDenseValue(DenseIntElementsAttr dense,
```
david-vc, line 2497: My suggestion is that the caller will call dens.isSplat() and create a separate API for retrieving the first value as a string.

test/unit_tests/aievec_tests/bf16_max_reduce/testbench.cc
```diff
@@ -0,0 +1,55 @@
...
+alignas(32) bfloat16 g_in0[IN0_SIZE];
+alignas(32) bfloat16 g_out0[OUT0_SIZE];
+alignas(32) bfloat16 g_out0Ref[OUT0_SIZE];
+
+int main(int argc, char *argv[]) {
+  // XXX Figure out how to use argv with xme_ca_udm_dbg -A
+  std::string dataDir(TO_STR(DATA_DIR));
+  srand(10);
+  std::generate(g_in0, g_in0 + IN0_SIZE,
+                [&]() { return random_bfloat16(-1, 1, 1); });
```
david-vc, line 19: Only 1 bit of mantissa? This deserves at the very least an explanation.

test/unit_tests/aievec_tests/float_min_reduce/testbench.cc
```diff
@@ -0,0 +1,57 @@
...
+
+  if (ok)
+    printf("TEST PASSED\n");
+  else
+    printf("TEST FAILED\n");
+
+  return ok ? 0 : 1;
+}
+
+// in0 and out0 are in C4 layout.
```
david-vc, line 50: Delete this comment

test/unit_tests/aievec_tests/bf16_add_reduce/testbench.cc
```diff
@@ -0,0 +1,62 @@
...
+
+  if (ok)
+    printf("TEST PASSED\n");
+  else
+    printf("TEST FAILED\n");
+
+  return ok ? 0 : 1;
+}
+
+// in0 and out0 are in C4 layout.
```
david-vc, line 48: Delete this comment. Do same for all new reduction test cases.

test/unit_tests/aievec_tests/bf16_max_reduce/testbench.cc
```diff
@@ -0,0 +1,55 @@
...
+    printf("TEST PASSED\n");
+  else
+    printf("TEST FAILED\n");
+
+  return ok ? 0 : 1;
+}
+
+// in0 and out0 are in C4 layout.
+void dut_ref(bfloat16 *in0, bfloat16 *out0) {
+  bfloat16 maxx = 0.0f;
```
david-vc, line 50: Initialize to -infinity.

test/unit_tests/aievec_tests/float_add_reduce/testbench.cc
```diff
@@ -0,0 +1,55 @@
...
+    printf("TEST PASSED\n");
+  else
+    printf("TEST FAILED\n");
+
+  return ok ? 0 : 1;
+}
+
+// in0 and out0 are in C4 layout.
+void dut_ref(float *in0, float *out0) {
+  float sum = 0.0f;
```
david-vc, line 50: Initialize to -infinity

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2483,9 +2572,56 @@ LogicalResult CppEmitter::emitAttribute(Location loc, Attribute attr) {
...
+              os << ")";
+            }
+          } else if (width == 16 && getVectorLaneSize(vType) == 16) {
+            os << "extract_v16bfloat16(";
+            if (splatValue == "0") {
+              os << "broadcast_zero_bfloat16()";
+            } else {
+              os << "broadcast_to_v32bfloat16";
+              os << "((";
+              if (failed(emitType(loc, fType)))
```
david-vc, line 2609: This code is producing incorrect numerical results.
The reason is that you cannot (bfloat16)(0x7F80) is doing (bfloat16)(32640.F).
See my comments on test case bf16_min_reduce

david-vc (changes requested):
Hi @linay-xsj , 
Thanks for addressing many issues.
However, there is another issue that needs to be addressed. See my comments on test case bf16_min_reduce.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2438,6 +2493,40 @@ bool CppEmitter::hasBlockLabel(Block &block) {
...
+// Get the first float value of a dense type value as a string.
+static std::string getSplatValueOfFloatDense(DenseFPElementsAttr dense,
+                                             bool isBFloat = false) {
+  APFloat apFloat = dense.getSplatValue<APFloat>();
+  float splatVal = apFloat.convertToFloat();
+  std::string firstValue = std::to_string(splatVal);
+
+  if (apFloat.isPosInfinity()) {
+    if (isBFloat) {
+      firstValue = std::to_string(0x1.FEp+127f);
```
david-vc, line 2513: Add comment: 
"TODO: Clean this up; emitting largest finite value in lieu of infinity; system headers do not provide a simple way to initialize a bfloat16 to infinity."

david-vc (approved):
Hi Lina:
Can you please address the comment at TranslateAIEVectToCpp.cpp:2513? Thanks.

PR #519, opened 2023-06-20
Implement target backend for AIRBIN

jnider:
AIRBIN is a stand-alone file format that contains a series of configuration instructions for AIE. This new backend outputs an ELF-compatible file that contains all of the register values needed for configuration. The file is uploaded to the control processor that executes the series of commands, instead of the host executing each command one by one.

This depends on the 'elfutils' library.
Configure elfutils with:
./configure --program-prefix="air-" --disable-debuginfod

Generate an AIRBIN by passing "--aie-generate-airbin" to aircc.py.


lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1250 @@
...
+    TILE_ADDR_ROW_SHIFT + TILE_ADDR_ROW_WIDTH;
+static constexpr auto TILE_ADDR_COL_WIDTH = 7u;
+
+static constexpr auto TILE_ADDR_ARR_SHIFT =
+    TILE_ADDR_COL_SHIFT + TILE_ADDR_COL_WIDTH;
+
+#define DMA_BD_COUNT 16
+
+#define REG_DMA_BD_BLOCK_SIZE 0x20
+#define REG_DMA_ADDR_A_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x0)
```
keryell, line 76: What about just doing C++14?
```suggestion
auto reg_dma_addr_a_bd = [](auto _idx) { return 0x1D000 + (_idx*REG_DMA_BD_BLOCK_SIZE) + 0x0; };
```
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es31-dont-use-macros-for-constants-or-functions
With C++20 you can use functions with `auto` instead of a lambda, but I guess you are stuck to C++17 in this project.

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1250 @@
...
+class TileAddress {
+public:
+  TileAddress(uint8_t column, uint8_t row, uint64_t array_offset = 0x000u)
+      : array_offset{array_offset}, column{column}, row{row} {}
+
+  // SFINAE is used here to choose the copy constructor for `TileAddress`,
+  // and this constructor for all other classes.
+  template <
+      typename Op,
+      std::enable_if_t<not std::is_same<Op, TileAddress>::value, bool> = true>
```
keryell, line 157:
```suggestion
      std::enable_if_t<!std::is_same_t<Op, TileAddress>, bool> = true>
```
`_t` is C++14, and I cannot remember the rule that say we should avoid the alias like `not`.
We can shorten a little bit this, but anyway all this will be replaced by `requires` from C++20 when we switch to.

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1250 @@
...
+                "The field must live in a 32-bit register");
+
+  static constexpr auto num_bits_used = (high_bit - low_bit) + 1u;
+  static constexpr auto unshifted_mask = (1u << num_bits_used) - 1u;
+  static_assert((low_bit != high_bit) xor (unshifted_mask == 1),
+                "1 is a valid mask iff the field is 1 bit wide");
+
+  static constexpr auto shifted_mask = unshifted_mask << low_bit;
+
+  [[nodiscard]] constexpr uint32_t operator()(uint32_t value) const {
```
keryell, line 232: Nice to think to the end-user bugs with the `[[nodiscard]]`. :-)

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1250 @@
...
+template <uint8_t high_bit, uint8_t low_bit = high_bit> class Field final {
+public:
+  static_assert(high_bit >= low_bit,
+                "The high bit should be higher than the low bit");
+  static_assert(high_bit < sizeof(uint32_t) * 8u,
+                "The field must live in a 32-bit register");
+
+  static constexpr auto num_bits_used = (high_bit - low_bit) + 1u;
+  static constexpr auto unshifted_mask = (1u << num_bits_used) - 1u;
+  static_assert((low_bit != high_bit) xor (unshifted_mask == 1),
```
keryell, line 227:
```suggestion
  static_assert((low_bit != high_bit) ^ (unshifted_mask == 1),
```
I think the rationale nowadays is that we have serious encoding nowadays and do not use digraph or trigraph anymore https://en.cppreference.com/w/cpp/language/operator_alternative

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -0,0 +1,1250 @@
...
+
+#define REG_DMA_BD_BLOCK_SIZE 0x20
+#define REG_DMA_ADDR_A_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x0)
+#define REG_DMA_ADDR_B_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x4)
+#define REG_DMA_2D_X_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x8)
+#define REG_DMA_2D_Y_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0xC)
+#define REG_DMA_PKT_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x10)
+#define REG_DMA_INT_STATE_BD(_idx)                                             \
+  (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x14)
+#define REG_DMA_CTRL_BD(_idx) (0x1D000 + ((_idx)*REG_DMA_BD_BLOCK_SIZE) + 0x18)
```
keryell, line 83: And since there is a lot of copy-paste you could even build all these functions on top of a more generic like
```
auto reg_dma_bd = [](auto idx, offset) { return 0x1D000 + (idx*REG_DMA_BD_BLOCK_SIZE) + offset; };
auto reg_dma_ctrl_bd = [](auto idx) { return reg_dma_bd(idx, 0x18)); };
```
toward cleaner embedded programming driven by the labs. :-)

lib/Targets/AIETargetAirbin.cpp
jnider: The last time I used C++ seriously was around 2005, so I really don't know any of these new features :-)
I'm happy to try all your suggestions.

keryell: For example I remember https://www.youtube.com/watch?v=ZLlGO102tVw which was interesting and I mentioned on Xilinx Slack `#cpp` along others.

jnider: done

lib/Targets/AIETargetAirbin.cpp
jnider: done

lib/Targets/AIETargetAirbin.cpp
jnider: done

lib/Targets/AIETargetAirbin.cpp
jnider: I am now using a struct to avoid the byte offsets

PR #520, opened 2023-06-21
[aievec][nfc] Reorganize conversion from vector

jsetoain:
Make the division of stages going from arbitrary Vector code to AIEVec intrinsics more explicit in the code organization.


include/aie/Dialect/AIEVec/Transforms/Passes.td
```diff
@@ -37,60 +37,4 @@ def AIEVectorize : Pass<"aie-vectorize", "ModuleOp"> {
...
-  let summary = "Transform simple aievec ops into aievec mul_conv or mac_conv ops.";
-  let options = [
-    Option<"shiftParam", "shift", "unsigned", /*default=*/"0",
-      "Shift parameter for rounding and saturation">,
-    Option<"aieTarget", "aie-target", "std::string", /*default=*/"\"aie\"",
-     "Select AIE version: \\\"aie\\\" or \\\"aieml\\\". This will determine "
-     "the vector size and available operations.">,
-  ];
-}
-
```
fifield, line 95: I'm curious about the decision to use C++ instead of tablegen for the boilerplate. I usually see code move the other direction. Is it related to this comment?
```
// TODO: Create a common base class for all AIEVec pipeline options.

fifield: Thanks for the context and explanation.

linay-xsj:
Javier, Could you please update the workspace and add reduction add, min and max transformation to vector-to-aievec in this changeset? I will not check in any code until you check in this change.

jamestclin:
A quick question, do we have a common syntax beautifier for this repo?

stephenneuendorffer:
> A quick question, do we have a common syntax beautifier for this repo?

We use 'git clang-format'

keryell:
If we are talking about code formatting, the answer is yes: https://github.com/Xilinx/mlir-aie/blob/main/.clang-format
Otherwise, if this is more about syntax highlighting, this is even more a personal taste and depends on the IDE and editors.

jamestclin:
> If we are talking about code formatting, the answer is yes: https://github.com/Xilinx/mlir-aie/blob/main/.clang-format Otherwise, if this is more about syntax highlighting, this is even more a personal taste and depends on the IDE and editors.

> > A quick question, do we have a common syntax beautifier for this repo?
> 
> We use 'git clang-format'

Thanks for the answer promptly.

PR #522, opened 2023-06-25
add `cmake/modulesXilinx` to the cmake module path

andrej:
Building and then compiling a tutorial on the current commit does not work when starting from scratch.

Here is how it breaks:

1. First, cmake does not find the `findVitis.cmake` module, because `cmake/modulesXilinx` is not in the cmake module search path. This only raises a warning, not an error -- easily overlooked.
2. Since `findVitis.cmake` is not found, the variable `VITIS_AIE_INCLUDE_PATH` is never defined. Subsequent references produce an empty string.
3. Despite this, building MLIR-AIE successfully completes (in my opinion, there should be an error).
4. Subsequently trying to compile a tutorial (e.g. tutorial 1) gives the error: `Error: Option -d appears as value of another option`, because `xchesscc_wrapper` calls unwrapped xchesscc like this:
    `$UNWRAPPED_XCHESSCC +P 4 -p me -C Release_LLVM -Y clang=$DIR/chess-clang -P $LIBDIR -d -f $@`,
    wherein `$LIBDIR` is defined as `VITIS_AIE_INCLUDE_PATH` or `VITIS_AIE2_INCLUDE_PATH`. Since those are empty strings due to the failure described above, it looks as if option -d is passed as value to option -P.

This pull request fixes this issue by adding the cmake/modulesXilinx path to the cmake module path, by resolving the issue in the first step above.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,9 @@ cmake_minimum_required(VERSION 3.21)
 
 set(TOOLCHAINFILES_PATH ${CMAKE_SOURCE_DIR}/cmake/modulesXilinx)
 
+# Allow cmake to find findVitis.cmake, needed for setting variables such as VITIS_AIE_INCLUDE_DIR
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modulesXilinx")
+
 set(AIE_RUNTIME_TARGETS "x86_64" CACHE STRING "Architectures to compile the runtime libraries for.")
 set(AIE_RUNTIME_TEST_TARGET "x86_64" CACHE STRING "Runtime architecture to test with.")
 


fifield:
If we do this then we should remove the setting of the cmake module path in the build scripts,
e.g. utils/build-mlir-aie.sh, and in the CI yaml.

andrej:
Oh! So that's where that was set. I hadn't seen that.

Curiously, it has stopped working for me. I will double-check if it's something on my end that has caused this to stop working. I always use the `build-mlir-aie.sh` script.

andrej:
Okay, I have figured out what caused this issue, and it was mostly my fault for calling the build script wrong.

If you run `./utils/build_mlir_aie.sh`, everything works as it should:

```
+ CMAKEMODULES_DIR=/home/andre/mlir-aie-fork/cmake
```

However, for some reason, I **sourced** the script instead of running it directly `source ./utils/build_mlir_aie.sh`, and that is what caused the problems:

```
++ CMAKEMODULES_DIR=/usr/cmake
```

This of course is not the intended way to call the build script, so this is user error. However, if someone (like me) mistakenly does call the build script like this, my fix from this pull request would fix the issue.

I'll leave it up to you if that's worth the "fix"!

At least it explains why suddenly it didn't work for me any more.

keryell:
I was confused by that too.
Usually in my project I use `.sh` extension only for a file to be sourced by `sh`.
If it is a shell-based executable, I do not put an extension since this is just an implementation detail of the executable.

stephenneuendorffer:
> Okay, I have figured out what caused this issue, and it was mostly my fault for calling the build script wrong.
> 
> If you run `./utils/build_mlir_aie.sh`, everything works as it should:
> 
> ```
> + CMAKEMODULES_DIR=/home/andre/mlir-aie-fork/cmake
> ```
> 
> However, for some reason, I **sourced** the script instead of running it directly `source ./utils/build_mlir_aie.sh`, and that is what caused the problems:
> 
> ```
> ++ CMAKEMODULES_DIR=/usr/cmake
> ```
> 
> This of course is not the intended way to call the build script, so this is user error. However, if someone (like me) mistakenly does call the build script like this, my fix from this pull request would fix the issue.
> 
> I'll leave it up to you if that's worth the "fix"!
> 
> At least it explains why suddenly it didn't work for me any more.

Hmm.. that's weird.  Have you root caused the difference?

andrej:
Yes, the root cause is lines 27 and 28 of the `build_mlir_aie.sh` script:

```
BASE_DIR=`realpath $(dirname $0)/..`
CMAKEMODULES_DIR=$BASE_DIR/cmake
```

When `source`ing, `$0` is not set to the path/name of the script, but of the command that launched the shell. So `BASE_DIR` does not point to the root of the repository, and so the cmake modules cannot be found.

CMakeLists.txt
```diff
@@ -17,6 +17,9 @@ cmake_minimum_required(VERSION 3.21)
 
 set(TOOLCHAINFILES_PATH ${CMAKE_SOURCE_DIR}/cmake/modulesXilinx)
```
stephenneuendorffer, line 18: Seems like this is not used anywhere.  Can we remove it?

stephenneuendorffer (changes requested):

PR #523, opened 2023-06-27
Set shiftBytes in aievec ShiftOp as an operand

linay-xsj:


lib/Dialect/AIEVec/Transforms/AIEVectorize.cpp
```diff
@@ -714,24 +714,16 @@ static aievec::SubOp generateSubOp(Operation *Op, AIEOpAttributes &opAttr,
   return subOp;
 }
 
-static aievec::ShiftOp generateShiftOp(SmallVector<Value> &sources,
-                                       VectState *state, Location loc,
-                                       int32_t shiftBytes,
+static aievec::ShiftOp generateShiftOp(Value lhs, Value rhs, VectState *state,
```
david-vc, line 717: Now that 'shiftBytes' is an operand, I recommend to change the order of the formal arguments and make shiftBytes the 3rd argument.

david-vc (approved):
Thank for for fixing this.

PR #524, opened 2023-06-27
allow aiesim.sh script to run independent of working directory

andrej:
When running `aiecc.py` with the `--aiesim` flag, a shell script named `aiesim.sh` that invokes the `aiesimulator` is generated.

This pull request adjusts the generated `aiesim.sh` script, so that it can be called from any working directory and still function. Just something to make working with it a little easier if you compile, e.g., into a separate build folder and don't want to enter that folder before running the simulation.

Previously, this script could only be called from the parent directory of the `...prj` directory where `aiesim.sh` resides.

I also added a parameter to be able to define the file name of the generated VCD file. If nothing is supplied, the same name as before, `foo` is used.


stephenneuendorffer (approved):
Good point! Thanks for this!

PR #527, opened 2023-06-30
Add "--aie-verify-connections" pass

andrej:
This new pass verifies that the connections between tiles are terminated at both ends and contain no cycles.

This pass does not transform the MLIR whatsoever; it is purely analysis. It issues errors if there are unterminated connections or cycles. This is done at the switchbox connection level.

Specifically, this pass issues the following new errors after analysis:

- Cycle errors, along with some connection along the cycle.
   ```
    error: There is a cycle in the route containing this connection.
        AIE.connect<"North" : 0, "West" : 0>
    ````
- No outgoing connection errors: If there is a switchbox configured with an incoming connection from N, E, S, W direction, but in its neighboring tile's switchbox, there is no outgoing connection from S, W, N, E direction, respectively.
    ```
    error: There is no matching outgoing connection for <"East" : 2> in tile (0, 1) for this incoming connection.
        AIE.connect<"West" : 2, "DMA" : 1>
    ```

- No incoming connection errors: Essentially the analog of above, if a switchbox is configured to send traffic out towards N, E, S, W, but its neighboring tile is not configured to accept traffic from S, W, N, E, an error is issued.
    ````
    error: There is no matching incoming connection for <"West" : 1> in tile (1, 1) for this outgoing connection.
        AIE.connect<"DMA" : 1, "East" : 1>
    ````

- No connection for DMAs: If a MM2S DMA is configured, but no switchbox connection DMA->X is configured, or if a S2MM, but no switchbox connection X->DMA is configured, an error is issued.
    ```
    error: S2MM DMA defined, but no incoming connections to the DMA are defined.
        %dma = AIE.dmaStart("S2MM", 0, ^bd, ^end)
   ```

There are tests in `test/verify-connections` that demonstrate all of the above examples.

I have added the `--aie-verify-connections` pass in `aiecc.py`.

Current limitations:
- Does not check if there is a matching configured DMA if a X->DMA `AIE.connect` is issued. (Essentially the opposite of the last error above.)
- Does not support packet based routing, only circuit switched. A warning is issued by the pss if packet based routing is used and the pass is invoked.

Please let me know what you think!


test/verify-connections/cycle-0.mlir
```diff
@@ -0,0 +1,22 @@
+// RUN: set +o pipefail; aie-opt --aie-verify-connections %s 2>&1 1>/dev/null | FileCheck %s
```
fifield, line 1: Although it isn't used much in this repo, `opt` supports a `--verify-diagnostics` option that can help check errors and warnings without pipe gymnastics. for example: https://github.com/Xilinx/mlir-aie/blob/main/test/dialect/AIE/badlock.mlir

andrej: Oh that is much nicer! Thanks for pointing this out

test/verify-connections/no-issues-0.mlir
```diff
@@ -0,0 +1,50 @@
...
+			^bd:
+				AIE.useLock(%lock12, "Acquire", 0)
+				AIE.dmaBd(<%buf12 : memref<256xi32>, 0, 256>, 0)
+				AIE.useLock(%lock12, "Release", 1)
+				AIE.nextBd ^end
+			^end:
+				AIE.end
+		}
+    }
+}
```
keryell, line 50: Add EOL at the end of each file.

keryell (commented):
Thanks for the great PR description message!
Since you made this effort, what about copy-pasting it ins a top-level comment of `lib/Dialect/AIE/Transforms/AIEVerifyConnections.cpp` for example?
Without the need of leading `//` on every line.

test/verify-connections/no-issues-0.mlir
keryell: Interestingly, this feature appeared only in clang-format 16... :-| https://releases.llvm.org/16.0.0/tools/clang/docs/ClangFormatStyleOptions.html#insertnewlineateof

test/verify-connections/cycle-0.mlir
andrej: Tests adjusted to use `--verify-diagnostics` where this was easily feasible. For the cycle detection tests, we don't know which connection the message will be associated with -- it could be any one along the cycle. So those are left as is.

test/verify-connections/no-issues-0.mlir
andrej: Done and done! Thanks for looking over this!

stephenneuendorffer:
Can we put this in --aie-find-flows? https://github.com/Xilinx/mlir-aie/pull/110

stephenneuendorffer:
@andrej  is this obsolete?

andrej:
@makslevental I just tried to push this branch into this repository, but it appears my access was revoked at the end of my internship. If you want to revive this, it might be easiest for you to add my repo as a remote, then push your copy of the branch.

PR #531, opened 2023-07-05
Implement vector reduction add, reduction min, and reduction max in auto vectorizer with different input and output data types

linay-xsj:


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -1207,24 +1348,40 @@ struct LowerVectorReductionOp
...
+        auto addElemOp = rewriter.create<aievec::AddElemOp>(
+            loc, lExtOp.getResult().getType(), lExtOp.getResult(),
+            rExtOp.getResult());
+        shiftIndex /= 2;
+        generateAIEVecOpsForReductionOp<aievec::AddElemOp>(
+            rewriter, srcOp, shiftIndex, addElemOp.getResult());
+        return success();
+      }
+      generateAIEVecOpsForReductionOp<aievec::AddElemOp>(
+          rewriter, srcOp, shiftIndex, srcOp.getVector());
```
jsetoain, line 1384: All of this pattern is actually a bunch of different patterns that act on different operand types for the same op. Instead of having a long list of `if` statements within the same pattern (with a lot of code that will never execute because of its templated parameters), it's better to have one pattern per conversion. Each "replaceOpWithNewOp" or "generateAIEVecOpsForReductionOp" should live in its own individual pattern.

jsetoain (changes requested):

jsetoain (approved):
Looks good to me. Thank you, Lina.

PR #532, opened 2023-07-05
Aie2fixes

stephenneuendorffer:


lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -335,21 +335,27 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
...
+          relLockID = lock.getLockIDValue();
           relValue = op.getLockValue();
+
         } else {
           // unreachable for current targets
           return op.emitOpError("unsupported lock action");
         }
       }
+      // if((acqEnable == disable) || (relEnable == disable))
+      //   return memOp.emitOpError("XAIEV2 requires acquire and release");
```
fifield, line 358: Remove or enable this? There are a few more instances of this comment block later in the file.

fifield (approved):

lib/Targets/AIETargetXAIEV2.cpp
stephenneuendorffer: Yeah, I think there's another bug there, was trying to figure out whether to deal with it in the same PR or not.

PR #533, opened 2023-07-06
Remove redundant clone of Xilinx/cmakeModules.git in the CI

keryell:
This is now a submodule in cmake/modulesXilinx.
But first, this PR is to check that this is really the case and does not break the CI.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -44,11 +44,6 @@ jobs:
           key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}
           max-size: 1G
 
-      - name: Get cmakeModules
-        id: clone-cmakeModules
-        run: git clone --depth 1 https://github.com/Xilinx/cmakeModules.git
-        shell: bash
-
       - name: Get LLVM
         id: clone-llvm
         run: utils/clone-llvm.sh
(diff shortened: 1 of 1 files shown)


stephenneuendorffer (approved):

stephenneuendorffer:
@keryell ready to commit?

keryell:
Perhaps this is the right time to push forwards a discussion about git hygiene I wanted to discuss during ACDC meeting but which have not happened yet.
First I am a `nobody` in this project, so I cannot commit anything anyway.
More importantly, since the teams around these projects have grown a lot recently, I would prefer we go along a more collaborative open strategy such as:
- nobody can merge her own PR except for emergency purpose;
- someone else has to merge the PR.
At least this ensures that at least 2 different persons have looked at the code (with the hypothesis the author was not in a pure write-only mode ;-) ).

stephenneuendorffer:
> First I am a `nobody` in this project, so I cannot commit anything anyway.

Added you.

stephenneuendorffer:
> I would prefer we go along a more collaborative open strategy such as:
> 
> * nobody can merge her own PR except for emergency purpose;
> * someone else has to merge the PR.
>   At least this ensures that at least 2 different persons have looked at the code (with the hypothesis the author was not in a pure write-only mode ;-) ).

I think if the team wants to commit to reviewing code on a regular basis, then we can require review, but I tend to think that people should be responsible for merging their own code.  I'd rather encourage "review early, review often" as a culture, rather than "do alot of work and then review before checkin.  This means that code is often ready for review, but not yet quite ready for committing.

keryell:
> I think if the team wants to commit to reviewing code on a regular basis, then we can require review, but I tend to think that people should be responsible for merging their own code. I'd rather encourage "review early, review often" as a culture, rather than "do alot of work and then review before checkin. This means that code is often ready for review, but not yet quite ready for committing.

This seems rather orthogonal to me, but let's talk about this next time we meet in person so it is easier to understand the subtleties.

PR #537, opened 2023-07-07
Join pattern with ObjectFifoLinkOp

AndraBisca:
This PR extends the ObjectFifoLinkOp with a list of input objectFifos that can be joined together from multiple input channels inside a sharedTile and output on a single outgoing channel.

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -431,39 +431,66 @@ xilinx::AIE::TileOp xilinx::AIE::ObjectFifoCreateOp::getProducerTileOp() {
 
 // ObjectFifoLinkOp
 LogicalResult xilinx::AIE::ObjectFifoLinkOp::verify() {
-  ObjectFifoCreateOp fifoIn = getFifoIn().getDefiningOp<ObjectFifoCreateOp>();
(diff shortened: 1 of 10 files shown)


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -431,18 +431,37 @@ xilinx::AIE::TileOp xilinx::AIE::ObjectFifoCreateOp::getProducerTileOp() {
 
 // ObjectFifoLinkOp
 LogicalResult xilinx::AIE::ObjectFifoLinkOp::verify() {
+  if (isJoin() && isDistribute())
+    return emitError("ObjectFifoLinkOp does not support 'join' and 'distribute' at the same time");
```
jgmelber, line 435: Do you think this could be possible and useful in the future?

jgmelber (approved):
Looks great!

lib/Dialect/AIE/IR/AIEDialect.cpp
AndraBisca: I think it is possible. We'll have to think a little bit more about the interactions between the memory allocs in the DMA BDs. I don't necessarily see the use of it between shimTile -> memTile -> AIEs, but between AIEs -> memTile -> AIEs it could be useful.

PR #542, opened 2023-07-12
add error handling to libXAIE calls

andrej:
If this is merged, the return values of generated libXIAE/AIE-rt calls in the generated `aie_inc.cpp` will be inspected for errors before continuing execution. An error will be printed to stderr if non-AIE_OK return values are observed; this can be disabled by putting `#define MLIR_AIE_QUIET` before the `#include aie_inc.cpp`.


stephenneuendorffer (commented):
IIRC, something similar already exists in LibXAIE, where we usually get an error message when calls fail.  What does this add over that?

andrej:
This stops execution when one of the calls fails by returning out of the function with an error code. I see two issues with the current implementation:

1.  As it stands, library calls happen back to back and continue even if a previous call fails.
2. There is no way for someone calling into our generated `aie_inc.cpp` to observe when an error occurs. As an end user of the generated code, e.g. in the tutorials, I would like to be able to stop execution / handle the error / retry when, e.g. configuring the cores fails. That code currently has now way of knowing the call failed.

I can remove printing the errors if this is redundant, but think we should keep the error returns.

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -37,6 +37,17 @@ using namespace xilinx::AIEX;
...
+// This string is output at the top of the lowered C code.
+const char xaie_c_file_header[] =
+    "#ifndef MLIR_AIE_QUIET\n"
+    "#define __mlir_aie_print(x) x\n"
+    "#else\n"
+    "#define __mlir_aie_print(x)\n"
+    "#endif\n"
+    "#define __mlir_aie_try(x) do { AieRC ret = (x); if(ret != XAIE_OK) { "
+    "__mlir_aie_print(fprintf(stderr, #x \"failed: %d\\n\", x)); return x; } "
+    "} while(0)\n";
```
keryell, line 49: You could use raw strings (see example of https://en.cppreference.com/w/cpp/language/string_literal, the equivalent of Python `"""` multi-line strings) and format normally the code instead of the current obfuscation.
It took me some time to understand the logic of the code because of the `return` I misinterpreted.
The fact that the `return` can actually be used as a kind of error exit here is super related to the way the runtime is generated on the fly. If it changes some day, it will break the code. I do not say it is bad. At least this could appear in a comment inside the macro for the next person refactoring the runtime architecture. :-)
You would probably want to print the line number, function, file, etc.

andrej: This is a good suggestion, I will make this change so the code is more readable.

What do you mean by this:
> The fact that the return can actually be used as a kind of error exit here is super related to the way the runtime is generated on the fly.

I think the return can be used as an "exit" because it is inside a macro, which behaves pretty much like cut and paste. I just did not want to litter the code with `if(XaieLibraryCall() != OK) return not_OK`, and thought a macro was a clean way of capturing this repetitive pattern. I don't think the fact that this code is generated affects how this works.

keryell: This is a broader discussion related to the specific way the runtime is designed and even the program using it.
For example if at some point MLIR-AIE mix more some CPU and AIE in a single-source way like CUDA, there might be some CPU code calling a completely independent AIE realm and returning a value would not mean anything since it would return from a random function and not `main()` which is the strong assumption made by the runtime today and by your `__mlir_aie_try` implementation.
But forget about it. Outside of this PR.
Otherwise the new code is quite clearer, thanks.
Actually you can even put comments in multi-line macros with `/* */` if you need.

stephenneuendorffer:
Hrm... it seems that this change was committed as part of a different PR (#547)

andrej:
I mentioned in the original description of #547 that it contains the changes of #542 (see last paragraph) and nobody seemed to take issue with it. Did this break something?

stephenneuendorffer:
> I mentioned in the original description of #547 that it contains the changes of #542 (see last paragraph) and nobody seemed to take issue with it. Did this break something?

no.... but since you had this in it's own PR, it's a bit nicer to commit it here and then rebase your other PR, especially given that it's unrelated. At the very least, this PR is still open when it doesn't need to be.

andrej:
Those are good points. Thank you for closing it!On Aug 4, 2023, at 6:01 PM, Stephen Neuendorffer ***@***.***> wrote:﻿

I mentioned in the original description of #547 that it contains the changes of #542 (see last paragraph) and nobody seemed to take issue with it. Did this break something?

no.... but since you had this in it's own PR, it's a bit nicer to commit it here and then rebase your other PR, especially given that it's unrelated. At the very least, this PR is still open when it doesn't need to be.

PR #550, opened 2023-07-22
fix setup_python_packages.sh script

andrej:
The most recent commit broke the script, at least for me. This simple fix just changes one line to how it was before the most recent commit


andrej:
This fix has been merged as part of #547

PR #551, opened 2023-07-24
Aie2 DMAs

stephenneuendorffer:


jgmelber (commented):
I believe a check is needed so that MemTile DMA channels 4 and 5 accessing neighboring MemTiles throws an error.

stephenneuendorffer:
> I believe a check is needed so that MemTile DMA channels 4 and 5 accessing neighboring MemTiles throws an error.

done.

jgmelber (approved):

PR #552, opened 2023-07-24
Aie2 DMA tests

stephenneuendorffer:
Jack, I've been struggling to get the aie2 simulator to work properly with ShimDMAs... I'm starting to wonder if there is some
additional configuration we need to get things to work properly?


stephenneuendorffer:
I've pulled these into a different PR.

PR #555, opened 2023-07-25
add "memory_allocator.h" target to cmake

andrej:
The "libmemory_allocator" library is successfully compiled and installed in the MLIR AIE install directory, but  the corresponding header is not added to the install 'include' directory.

This commit fixes that, thus allowing to use data movement via external buffers into the shim tiles from the host.


jackl-xilinx:
I ran into this same issue today. This is a good fix. Thanks @andrej

PR #557, opened 2023-07-26
Make object fifos into real symbols instead of referencing them by string attributes.

fifield:
The problem remains that they are also SSA values.


stephenneuendorffer:
nice!

AndraBisca (approved):

AndraBisca:
Looks really good! I will start working on fully removing the SSA values and using the symbols instead very soon. Thank you!

PR #573, opened 2023-08-10
Parallelize the CI on Release & Assert, Ubuntu 20.04 & 22.04

keryell:
- Parallelize CI for Assert and Release build.
- Use strategy in GitHub Action to run on Ubuntu 20.04 and 22.04.
  It uses also Docker with the side effect on GitHub action of
  having actually 2 CPU, instead of one per job by default.
  So it gives a parallelization factor of 2×2×2 instead of 2×2.
- Display a lot of CI information to help further development

The problem of using an official Ubuntu Docker image is that it
exposes a lot of different compared to GitHub Action Ubuntu image
which comes turbo-charged with a lot of goodies, like latest
CMake, a lot of compiler versions, etc. even on old Ubuntu
version.

So, a lot of additional software has to be installed, but it is
also more realistic about a normal user environment.

- Install lacking software to have check-out working.
- Remove sudo usage which is useless in Docker.
- Populate package list.
- Install pip.
- Update recipe to install Ninja recipe compatible with Docker.
- Install CMake with pip to benefit from latest version
- Install Clang compiler and LLVM linker required by the CI.
- Run apt package manager in the CI in non-interactive mode.
  Otherwise, on Ubuntu 20.04 the installation of tzdata asking question
  freezes libboost installation.
- Display user and group ids.
- Make safer display of environment execution context by avoiding
  special character evaluation by the shell by using environment
  variables as a communication channel.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -102,8 +102,12 @@ jobs:
           pip install psutil rich numpy pybind11
 
       - name: Install Ninja
+        # Can not use the following since it wants to use `sudo` not
+        # available in the case of Docker execution
         # https://github.com/llvm/actions/tree/main/install-ninja
-        uses: llvm/actions/install-ninja@6a57890d0e3f9f35dfc72e7e48bc5e1e527cdd6c
+        # uses: llvm/actions/install-ninja@6a57890d0e3f9f35dfc72e7e48bc5e1e527cdd6c
+        # So just use the specific implementation instead:
+        run: apt-get install -y ninja-build
 
       - name: Get Submodule Hash
         id: get-submodule-hash
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -15,6 +15,12 @@ defaults:
     # repeating elsewhere.
     shell: bash
 
+env:
+  # Run apt package manager in the CI in non-interactive mode.
+  # Otherwise, on Ubuntu 20.04 the installation of tzdata asking question
+  # freezes libboost installation.
+  DEBIAN_FRONTEND: noninteractive
+
 jobs:
   build-repo:
     name: Build and Test
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -114,6 +114,10 @@ jobs:
         # So just use the specific implementation instead:
         run: apt-get install -y ninja-build
 
+        # Because the build requires to use explicitly this compiler
+      - name: Install Clang & Clang++
+        run: apt-get install -y clang
+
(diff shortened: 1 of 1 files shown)


stephenneuendorffer (approved):

keryell:
There is an issue when printing the execution context with the "&" in the PR name being interpreted by JSON and havoc happens. o_o
GitHub Action is an interesting pile of frameworks, hacks and different sub-languages... :-(
I need to find a solution before landing this.

keryell:
I have the feeling that `ccache` does not work, even if the cache is saved at the end and reloaded. :-(

keryell:
It looks like after relaunching the workflows, CCache seems to work correctly.

keryell:
I need an admin to merge this for some reason I cannot introspect.
@stephenneuendorffer @fifield

keryell:
Probably the problem is that the life of this PR started producing just `Build and Test / Build and Test ` and evolved to produce something else, but the acceptance workflow of GitHub does still expect for `Build and Test / Build and Test` to pass.
Since it does not exist anymore sine it has been replaced by 4 other productions instead, it will never happen. And the 4 new productions do not appear as required.

jamestcl-amd:
Hello, this commit updates the `cmake/modulesXilinx` from `7915d8f` to `2c38856`. It causes the script `./utils/build-mlir-aie.sh` to have cmake error like below. Not sure if this is intended.
```
loading initial cache file /proj/xsjhdstaff1/linjames/mlir-aie/build/runtime_lib/xaiengineTmp/x86_64/tmp/xaiengine_x86_64-cache-Release.cmake
CMake Error at CMakeLists.txt:10 (if):
  if given arguments:

    "EQUAL" "2022"

  Unknown arguments specified


-- Configuring incomplete, errors occurred!

```

fifield:
> Hello, this commit updates the `cmake/modulesXilinx` from `7915d8f` to `2c38856`. It causes the script `./utils/build-mlir-aie.sh` to have cmake error like below. Not sure if this is intended.
> 
> ```
> loading initial cache file /proj/xsjhdstaff1/linjames/mlir-aie/build/runtime_lib/xaiengineTmp/x86_64/tmp/xaiengine_x86_64-cache-Release.cmake
> CMake Error at CMakeLists.txt:10 (if):
>   if given arguments:
> 
>     "EQUAL" "2022"
> 
>   Unknown arguments specified
> 
> 
> -- Configuring incomplete, errors occurred!
> ```

I just encountered the same issue. I've made a fix in https://github.com/Xilinx/mlir-aie/pull/624

keryell:
Curious this `cmake/modulesXilinx` story. Not intended at all. :-( Thank you for fixing this!

PR #574, opened 2023-08-11
[aievec] Move common utility code to library (1/N)

jsetoain:
There are several utility functions that don't have a proper place in which to live. Some of them are defined as "inline" in a header, and one of them is duplicated in two source files. This patch creates an AIEVecUtils library specifically for these, and starts the migration process by moving the duplicated function to this library.


keryell:
Thanks for this cleaning work!
This reminds me a presentation from Google mentioning they spend something like 10% of their R&D budget on code cleaning.

PR #575, opened 2023-08-11
[aievec] Fix lowering of add/mul for AIEv1

jsetoain:
This patch deals with some of the remaining vector instructions that were being left unconverted.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
@@ -94,9 +94,24 @@ extractMACOperandsFromAddOperands(Value addLhs, Value addRhs) {
     mulOp = dyn_cast<arith::MulIOp>(rhsDefOp);
     acc = addLhs;
   }
-  if (!mulOp)
-    return {};
-  return std::make_tuple(mulOp.getLhs(), mulOp.getRhs(), acc);
+  if (mulOp)
+    return std::make_tuple(mulOp.getLhs(), mulOp.getRhs(), acc);
+
+  // If the MulIOp has been already translated to aievec::MulOp:
+  aievec::SRSOp lhsSrsOp = addLhs.getDefiningOp<aievec::SRSOp>();
+  aievec::SRSOp rhsSrsOp = addRhs.getDefiningOp<aievec::SRSOp>();
+  aievec::MulOp aieMulOp = nullptr;
+  if (lhsSrsOp) {
+    aieMulOp = lhsSrsOp.getSource().getDefiningOp<aievec::MulOp>();
+    acc = addRhs;
+  }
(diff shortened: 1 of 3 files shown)


test/Conversion/VectorToAIEVec/test-arith.mlir
```diff
@@ -39,3 +39,35 @@ func.func @vecsubf(%arg0: vector<16xf32>, %arg1: vector<16xf32>) -> vector<16xf3
...
+  // CHECK: %[[RES:.*]] = aievec.srs %[[MUL]] {shift = 0 : i8} : vector<16xi80>, vector<16xi32>
+  %0 = arith.muli %arg0, %arg1 : vector<16xi32>
+  // CHECK: return %[[RES]] : vector<16xi32>
+  return %0 : vector<16xi32>
+}
+
+// CHECK-LABEL: func @vecmulf(
+// CHECK-SAME: %[[LHS:.*]]: vector<16xf32>,
+// CHECK-SAME: %[[RHS:.*]]: vector<16xf32>)
+func.func @vecmulf(%arg0: vector<16xf32>, %arg1: vector<16xf32>) -> vector<16xf32> {
```
linay-xsj, line 68: For AIEv1, it cannot deal with 16xf32 size vectors, it can only handle 256bits size vector, such as 8xf32, 8xi32, 16xi16. You can run this test case on the simulator.

jsetoain: Right! I actually saw it failing in the simulators in other tests. We really need to fix the dialect 🙂

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -737,15 +752,13 @@ struct FoldBroadcastToFMAOp : public OpConversionPattern<aievec::FMAOp> {
 };
 
 struct ConvertMulAddToAIEVecFMAOpPattern
-    : public OpConversionPattern<arith::AddIOp> {
-  using OpConversionPattern<arith::AddIOp>::OpConversionPattern;
+    : public OpConversionPattern<aievec::AddOp> {
```
jamestcl-amd, line 755: Should we move this to AIEVecOptimizations.cpp?

jsetoain: It's an option, yes, but for v1 there's an extra wrinkle to sort out first. The reason we do MUL+ADD -> MAC in the dialect conversion is because the way we convert broadcast is by merging it with FMA. If we leave aievec.mul and aievec.add with the intention of fusing them together into a FMA in the optimization layer (which is very reasonable, and my preference), then we leave vector.broadcast alive after conversion. So, the order of operations would be to first deal with the lowering of broadcast for v1, and then move everything to its rightful place.

Good eye, though, you're very quickly catching up with all the technical debt we've left behind :-)

PR #578, opened 2023-08-12
target model refactoring

stephenneuendorffer:


keryell:
On my plate, I was thinking to deploy https://editorconfig.org/#supported-properties on our projects to avoid some cross-platform pain à la dos2unix.

PR #580, opened 2023-08-14
Use EditorConfig file to pre-configure editors and IDE

keryell:
See https://editorconfig.org/

This is not to be merged before careful inspection by the project stakeholders
since I do not know exactly what matters.

This is just a start of discussion as some parameters might be controversial :-)
or to be left up to the implementer.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,33 @@
+# Configure globally editors and IDE according to this file
+# https://EditorConfig.org
+
+# This is the top-most EditorConfig file for this project, to avoid
+# being influenced by higher such files, for example if this project
+# is a submodule or subproject of an other project.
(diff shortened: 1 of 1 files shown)


keryell:
Some potential controversy: do you want to have 80 character lines like in LLVM? Is 2-character indent OK? Is space trimming at the end of the lines good for anyone?
At the end it does not prevent anyone to do something else, as this sets just the editor default, not a pre/post-commit enforcement.

keryell:
To be propagated to `mlir-air` too at some point.

SamuelBayliss:
I don't know what flow / enforcement would be triggered by this - but if it doesn't break the CI ( clang-format etc) then I'm ok with it.

PR #582, opened 2023-08-15
Pipeclean AIE2 peano flow

stephenneuendorffer:
Peano for AIE2 has matured quite a bit, but we haven't been actively using it for AIE2/phoenix.  This patch enables some simple designs to go through the flow:
1) aiecc.py now assumes the target architecture is set in the LLVMIR generated from MLIR, rather than hardcoding --march=aie

2) We generate the correct target architecture in LLVMIR based on the DeviceOp in MLIR.

3) We lower lock operations to AIE2 lock intrinsics based on the DeviceOp as well.


jgmelber:
I'd be excited to do some Peano AIE2 testing on a laptop

PR #583, opened 2023-08-16
Fix TableGen file to avoid {% Liquid templating language

keryell:
Probably I need to update the CI to compile the documentation without installing it just to check we do not break the documentation.
The problem now is that the breakage is post-commit, when the `main` branch is updated.


AndraBisca (approved):
Ah, this brings back memories. I ran into this exact issue when I first added those curly brackets and, of course, I forgot this can happen. Thank you so much for fixing it! :)

Issue #584, opened 2023-08-17
Support vector.transfer_read of scalar memref


jamestcl-amd:
- For vector canonicalization, it doesn't support vector.transfer_read from scalar memref
- Sample test at https://github.com/Xilinx/mlir-aie/blob/main/test/dialect/AIEVec/precanonicalization.mlir
- Sample code at https://github.com/Xilinx/mlir-aie/blob/1cf052849d09f0583564dca1b138b165eff9a7d1/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp#L128
- Workaround:
![image](https://github.com/Xilinx/mlir-aie/assets/142062293/0631c3c8-1ea4-4dc5-9934-89ec2baa64cf)

jsetoain:
@jamestcl-amd 

[PR#586](https://github.com/Xilinx/mlir-aie/pull/586) Deals with this issue. I tried to add you as a reviewer but looks like I'm not allowed. You should be able to review anyway.

jamestcl-amd:
Thanks. I will need Steve to give me permissions against this repo.

jsetoain:
Resolved by #586

PR #586, opened 2023-08-17
[aievec] Add support for 0-rank xfer reads

jsetoain:
This patch solves issue [#584](https://github.com/Xilinx/mlir-aie/issues/584)

### DIFF
--- a/lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
+++ b/lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
@@ -1878,6 +1878,26 @@ static LogicalResult printOperation(CppEmitter &emitter,
   return success();
 }
 
+// Print an expand shape by forwarding the value to the next op
+static LogicalResult printOperation(CppEmitter &emitter,
+                                    memref::ExpandShapeOp expandShapeOp) {
+  Value source = expandShapeOp.getSrc();
+
+  // If the memref being outputted is not already emitted,
+  // error out
+  if (!emitter.hasValueInScope(source))
+    return failure();
+
+  if (failed(emitter.emitAssignPrefix(*expandShapeOp)))
+    return failure();
+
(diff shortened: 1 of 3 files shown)


lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -1878,6 +1878,26 @@ static LogicalResult printOperation(CppEmitter &emitter,
...
+  // error out
+  if (!emitter.hasValueInScope(source))
+    return failure();
+
+  if (failed(emitter.emitAssignPrefix(*expandShapeOp)))
+    return failure();
+
+  raw_indented_ostream &os = emitter.ostream();
+
+  os << emitter.getOrCreateName(source);
```
jamestcl-amd, line 1896: For my own understanding. This op essentially is a no-op in the cpp? It only does variable assignment like "v18 = v17"?

jsetoain: Indeed, it's just forwarding the value, and [that is the semantics of the operation itself](https://mlir.llvm.org/docs/Dialects/MemRef/#memrefexpand_shape-memrefexpandshapeop): _"The operation is limited to such reassociations, where a dimension is expanded into one or multiple contiguous dimensions. Such reassociations never require additional allocs or copies."_

A better option would be to replace all uses of the op with the source value and call it a day, so you don't give extra work to the C++ compiler, but this was a more immediate solution. I still might give it a go before landing it.

jamestcl-amd: I see, that sounds reasonable to me. Hopefully, our C++ compiler can figure that out. If not, we can go for your better option later.

jamestcl-amd (approved):
Looks good to me. Thanks for fixing the issue promptly.

PR #587, opened 2023-08-18
[aievec] Improve alignment offset calculation

jsetoain:
In order to compute the alignment offset for a given transfer read operation, we impose the following constraints on the indices:

a) are constants, or
b) come from an `affine.apply` on an aligned SSA value, or
c) are an aligned SSA value

This patch tries to loosen the constraints to improve misalignment detection and computation by computing the minimum value a "well-defined" SSA value can have. Now the constraints are that the indices:

a) are constants, or
b) come from an `affine.apply` over an i.v. in an affine.for op, or
c) come from an `affine.apply` over an aligned SSA value that's not an i.v.,
c) are an aligned SSA value

This patch adds correct misalignment calculations for cases where the i.v. starts on an unaligned value (like 1), and can deal with indices computed using affine maps on a multidimensional domain, e.g.:
```
        %index =affine.apply affine_map<(d0, d1) -> (d0 * 512 + d1)>(%i, %j)
```

### DIFF
--- a/lib/Dialect/AIEVec/Utils/Utils.cpp
+++ b/lib/Dialect/AIEVec/Utils/Utils.cpp
@@ -18,56 +18,81 @@
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Debug.h"
-#include <optional>
 
 #define DEBUG_TYPE "aievec-utils"
 
 using namespace mlir;
 
 namespace xilinx::aievec {
+
+static std::optional<int64_t> getLowerBoundValue(Value idx) {
+  if (auto blkArg = dyn_cast<BlockArgument>(idx)) {
+    auto parentOp = blkArg.getOwner()->getParentOp();
+    return TypeSwitch<Operation *, std::optional<int64_t>>(parentOp)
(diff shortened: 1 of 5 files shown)


lib/Dialect/AIEVec/Utils/Utils.cpp
```diff
@@ -25,6 +25,43 @@
 using namespace mlir;
 
 namespace xilinx::aievec {
+
+static int64_t getLowerBoundValue(Value idx) {
```
david-vc, line 29: Would it be safer to change the return type of this function and std::optional<int64_t>() for cases where it is unable to infer the lower bound? In that case the caller is responsible for handling this case.

david-vc (commented):
Could you spell out in the PR message for which cases the change improves accuracy?

lib/Dialect/AIEVec/Utils/Utils.cpp
jsetoain: It's certainly cleaner, and ready for further improvement. Right now, we assume that accesses whose alignment can't be determined at compile-time, are aligned. In the future, we should at least consider if we want to treat all those cases as unaligned, and generate alignment code. It's a tricky trade-off.

david-vc (approved):
@jsetoain : thanks for explaining the limitations and being more explicit with the std :: optional.

PR #588, opened 2023-08-18
Repair and enable python test

fifield:
* Fix the python binding test for object fifo changes
* Enable test when bindings are enabled


AndraBisca (approved):
Thank you for fixing this!

PR #590, opened 2023-08-18
catch up to TOM MLIR

makslevental:
This is a bump to TOM LLVM.

Primarily it's `affine` being moved to its own namespace and the removal of `llvm::Optional` in favor of `std::optional`.


fifield:
It is problematic to update the llvm version, if that is the intent. See https://github.com/Xilinx/mlir-aie/pull/394

keryell (commented):
Thanks for the C++ modernism!

makslevental:
> It is problematic to update the llvm version, if that is the intent. See #394

@fifield what am I missing? why not actually do as you suggest and revert the change locally (instead of reverting to the commit):

```diff
diff --git a/llvm/lib/IR/AsmWriter.cpp b/llvm/lib/IR/AsmWriter.cpp
index be4a3ed79d88..8da1d4a43809 100644
--- a/llvm/lib/IR/AsmWriter.cpp
+++ b/llvm/lib/IR/AsmWriter.cpp
@@ -621,8 +621,7 @@ void TypePrinting::print(Type *Ty, raw_ostream &OS) {
   }
   case Type::TypedPointerTyID: {
     TypedPointerType *TPTy = cast<TypedPointerType>(Ty);
-    OS << "typedptr(" << *TPTy->getElementType() << ", "
-       << TPTy->getAddressSpace() << ")";
+    OS << *TPTy->getElementType() << "*";
     return;
   }
   case Type::TargetExtTyID:
diff --git a/llvm/lib/IR/Value.cpp b/llvm/lib/IR/Value.cpp
index 41260a98e3ce..bb339027eace 100644
--- a/llvm/lib/IR/Value.cpp
+++ b/llvm/lib/IR/Value.cpp
@@ -44,9 +44,9 @@ static cl::opt<unsigned> UseDerefAtPointSemantics(
 //                                Value Class
 //===----------------------------------------------------------------------===//
 static inline Type *checkType(Type *Ty) {
-  assert(Ty && "Value defined with a null type: Error!");
-  assert(!isa<TypedPointerType>(Ty->getScalarType()) &&
-         "Cannot have values with typed pointer types");
+//  assert(Ty && "Value defined with a null type: Error!");
+//  assert(!isa<TypedPointerType>(Ty->getScalarType()) &&
+//         "Cannot have values with typed pointer types");
   return Ty;
 }
 
diff --git a/mlir/lib/Target/LLVMIR/TypeToLLVM.cpp b/mlir/lib/Target/LLVMIR/TypeToLLVM.cpp
index 6d8b415ff09d..d8f1483b0c79 100644
--- a/mlir/lib/Target/LLVMIR/TypeToLLVM.cpp
+++ b/mlir/lib/Target/LLVMIR/TypeToLLVM.cpp
@@ -14,6 +14,7 @@
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/IR/DataLayout.h"
 #include "llvm/IR/DerivedTypes.h"
+#include "llvm/IR/TypedPointerType.h"
 #include "llvm/IR/Type.h"
 
 using namespace mlir;
@@ -104,9 +105,11 @@ private:
     return llvm::IntegerType::get(context, type.getWidth());
   }
 
   /// Translates the given pointer type.
   llvm::Type *translate(LLVM::LLVMPointerType type) {
-    return llvm::PointerType::get(context, type.getAddressSpace());
+    if (type.isOpaque())
+      return llvm::PointerType::get(context, type.getAddressSpace());
+    return llvm::TypedPointerType::get(translateType(type.getElementType()),
+                                  type.getAddressSpace());
   }
 
   /// Translates the given structure type, supports both identified and literal

```

then 

```
// mlir-opt -convert-func-to-llvm='use-opaque-pointers=0'
func.func private @memref_call_conv(%arg0: memref<?xf32>)
```

becomes 

```
// mlir-translate --mlir-to-llvmir
llvm.func @memref_call_conv_nested(!llvm.ptr<i32>, !llvm.ptr<i32>, i64, i64, i64) attributes {sym_visibility = "private"}
```

becomes 

```
; ModuleID = 'LLVMDialectModule'
source_filename = "LLVMDialectModule"

declare ptr @malloc(i64)

declare void @free(ptr)

declare void @memref_call_conv_nested(i32*, i32*, i64, i64, i64)

!llvm.module.flags = !{!0}
```

It's not perfect I guess (i.e., it might need further munging) but the types should be preserved, and as [this comment indicates](https://github.com/llvm/llvm-project/blob/2138c906458ed37f0f833b1c7dda1bc6c6be8236/llvm/include/llvm/IR/TypedPointerType.h#L21-L24), such a path isn't likely to go away because of GPUs...

fifield:
We can try your solution but not in the github runner because it doesn't have access to Vitis and won't show the problem. Are you able to build with Vitis and build the tests with the chess compiler?

makslevental:
> Are you able to build with Vitis and build the tests with the chess compiler?

Not unless chess ships with the standard student release (i.e., that is the only thing I have access to).

keryell:
> Not unless chess ships with the standard student release (i.e., that is the only thing I have access to).

You can register with your information when you download Vitis. It will probably be accepted and then you need to ask for a license.
Usually the real problem is the approximate 300 GiB of storage required during the install process.
But it fits on my Linux laptop.

lib/Dialect/AIEVec/Transforms/ConvertVectorToAIEVec.cpp
```diff
@@ -38,6 +38,7 @@ namespace xilinx::aievec {
 } // namespace xilinx::aievec
 
 using namespace mlir;
+using namespace mlir::affine;
```
jsetoain, line 41: I'd rather leave the `affine::` prefix everywhere, for clarity. I suspect the next step is to change the op names from `Affine<op>` to simply `<op>`, and having things like `ForOp` floating around obscures the type.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -1545,7 +1545,7 @@ struct LowerVectorExtractStridedSliceOpAIEv1Pattern
   LogicalResult
   matchAndRewrite(vector::ExtractStridedSliceOp extractOp, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto vType = extractOp.getVectorType();
+    auto vType = extractOp.getVector().getType();
```
jsetoain: The new type access method is `getSourceVectorType`. In theory, in your change, `auto` resolves to `Type` instead of `VectorType`, but it looks like `getRank` on `vType` is not failing. I guess there's some static polymorphism shenanigans at play but, for safety, let's stick with the original semantics.

jsetoain (commented):
There are a few pending changes before this compiles, and I suspect it will fail the simulator tests (which won't run without Vitis, and therefore are not being tested automatically on PRs), but we can pick it up from there once the basic tests pass.

Thanks, Maks!

fifield:
After applying this patch and getting it to build, many aievec tests are failing for me with errors like this one:
```
aie-opt: /scratch2/jefff/acdc/aie/include/aie/Dialect/AIEVec/AIEVecUtils.h:64: int32_t xilinx::aievec::getVectorSizeInBits(mlir::VectorType): Assertion `veclen >= 128 && "AIE vector size should be greater than 128 bits"' failed.
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: /scratch2/jefff/acdc/build/aie/bin/aie-opt /scratch2/jefff/acdc/aie/test/unit_tests/aievec_tests/i16xi16_gemm/gemm64_int16_unroll32.mlir -affine-super-vectorize=virtual-vector-size=32 -aieml=true --aie-vectorize
```

makslevental:
@jsetoain I restored the `affine::` prefix.

@fifield I squashed all of the assert fails

include/aie/Dialect/AIE/IR/AIE.td
```diff
@@ -1363,50 +1358,50 @@ def AIE_ShimDMAAllocationOp : AIE_Op<"shimDMAAllocation", [HasParent<"DeviceOp">
...
   }];
 }
 
 def AIE_ObjectFifoCreateOp: AIE_Op<"objectFifo", [HasParent<"DeviceOp">, Symbol]> {
   let summary = "Create a circular buffer or channel between two tiles";
   let description = [{
-    The `aie.objectFifo` operation creates a circular buffer established between a producer and one or 
-    more consumers, which are `aie.tile` operations. The`aie.objectFifo` instantiates the given number of 
-    buffers (of given output type) and their locks in the Memory Module of the appropriate tile(s) after 
-    lowering, based on tile-adjacency. These elements represent the conceptual depth of the `objectFifo` or,
```
makslevental, line 1376: I don't know how these are being erased - I tried to put them back but was unsuccessful.

include/aie/Dialect/AIEVec/AIEVecUtils.h
```diff
@@ -43,7 +43,7 @@ inline VectorType createVectorType(unsigned lanes, Type elementType) {
 
 // Return the size (in bits) of the underlying element type of the vector
 inline int32_t getElementSizeInBits(VectorType type) {
-  return type.cast<ShapedType>().getSizeInBits() / type.getNumElements();
+  return type.cast<ShapedType>().getElementTypeBitWidth();
```
makslevental, line 46: @fifield This was the source/cause of the >=128 bit assert.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -476,30 +476,31 @@ struct AIEObjectFifoStatefulTransformPass
...
       builder.setInsertionPointToEnd(device.getBody());
       MemOp newMemOp =
           builder.create<MemOp>(builder.getUnknownLoc(), objFifoTileOp);
-      producerMem = &newMemOp;
-      Region &r = producerMem->getBody();
-      r.push_back(new Block);
-      // add terminator operation to end block
-      Block &endBlock = r.back();
-      builder.setInsertionPointToStart(&endBlock);
-      builder.create<EndOp>(builder.getUnknownLoc());
```
makslevental, line 499: Some wacky things going on here that were giving me assert fails because the `MemOp *` kept being set to point at a `AIE_TileOp` which has no regions (and thus would then fail the assert in `producerMem->getBody()`).

fifield (approved):
Works for me when applied alongside https://github.com/Xilinx/mlir-aie/pull/634

I also needed to change `convert-memref-to-llvm` to `finalize-memref-to-llvm` in python/aie/compiler/aiecc/main.py

PR #598, opened 2023-08-24
[aievec] Move exp lowering pattern to vector-2-aievec

jsetoain:
The lowering pattern and legalization of math::ExpOp was in the wrong pipeline, this patch moves it to the Vec2AIEVec pipeline.


david-vc (commented):
Looks good. Thanks.

PR #599, opened 2023-08-24
Add CopyRemoval pass to the vector canonicalization before lowering to AIEVec

jamestcl-amd:
- This pass once existed in the upstream MLIR and was removed at https://reviews.llvm.org/D99172. This CopyRemoval pass was removed due to the introduction of `memref.clone` (now `bufferization.clone`). `bufferization.clone` is essentially `memref.alloc+copy` ops and is mostly created in buffer-deallocation pass. It is introduced for better memory deallocation and optimization. The canonicalization pass can optimize out some of the bufferization.clone+memref.dealloc ops. If the bufferization.clone still exists in the end of bufferization pass, we should call "-convert-bufferization-to-memref" pass to convert it back to `memref.alloc+memref.copy`.  Hence, the original CopyRemoval pass (which removes `memref.alloc/copy/dealloc`) is replaced by the canonicalization of bufferization.clone+memref.dealloc op. However, in our case, `memref.alloc()` is from the `tensor.empty()` after bufferization, `memref.copy()` is added in buffer-results-to-out-params pass, and `memref.dealloc()` is added in buffer-deallocation pass, which creates a pattern that the existing canonicalizer cannot support. As a result, we recover this CopyRemoval pass back into the canonicalization pass for Vector before lowering to AIEVec and help eliminate unnecessary memory allocation after the bufferization.
- Note that we can also try to remove copy op before the bufferization. This will require the function to be converted to destination-passing style code at tensor level, which isn't fully supported in the existing MLIR environment yet. Once the upstream MLIR has a mature support on this, removing copy op before bufferization can be another direction to pursue. 
- After adding this CopyRemoval pass, we are able to support TOSA functions end-to-end with our AIEVec based flow. TOSA end-to-end tests will be in another PR.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/CopyRemoval.cpp
+++ b/lib/Dialect/AIEVec/Transforms/CopyRemoval.cpp
@@ -17,6 +17,32 @@ using namespace MemoryEffects;
 namespace {
 
 //===----------------------------------------------------------------------===//
+// TODO: this pass once existed in the upstream MLIR and was removed at
+// https://reviews.llvm.org/D99172. This CopyRemoval pass was removed due to the
+// introduction of memref.clone (now bufferization.clone). bufferization.clone
+// is essentially memref.alloc+copy ops and is mostly created in
+// buffer-deallocation pass. It is introduced for better memory deallocation and
+// optimization. The canonicalization pass can optimize out some of the
+// bufferization.clone+memref.dealloc ops. If the bufferization.clone still
+// exists in the end of bufferization pass, we should call
+// "-convert-bufferization-to-memref" pass to convert it back to
+// memref.alloc+memref.copy.  Hence, the original CopyRemoval pass (which
+// removes memref.alloc/copy/dealloc) is replaced by the canonicalization of
+// bufferization.clone+memref.dealloc op. However, in our case, memref.alloc()
+// is from the tensor.empty() after bufferization, memref.copy() is added in
+// buffer-results-to-out-params pass, and memref.dealloc() is added in
+// buffer-deallocation pass, which creates a pattern that the existing
+// canonicalizer cannot support. As a result, we recover this CopyRemoval pass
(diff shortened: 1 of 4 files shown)


lib/Dialect/AIEVec/Transforms/CopyRemoval.cpp
```diff
@@ -0,0 +1,245 @@
...
+// buffer-results-to-out-params pass, and memref.dealloc() is added in
+// buffer-deallocation pass, which creates a pattern that the existing
+// canonicalizer cannot support. As a result, we recover this CopyRemoval pass
+// back into the canonicalization pass for Vector before lowering to AIEVec and
+// help eliminate unnecessary memory allocation after the bufferization. Note
+// that we can also try to remove copy op before the bufferization. This will
+// require the function to be converted to destination-passing style code at
+// tensor level, which isn't fully supported in the existing MLIR environment
+// yet. Once the upstream MLIR has a mature support on this, removing copy op
+// before bufferization can be another direction to pursue.
```
keryell, line 42: Great to have this design comment!

lib/Dialect/AIEVec/Transforms/CopyRemoval.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+} // end anonymous namespace
+
+//===----------------------------------------------------------------------===//
+// CopyRemovalPass construction
+//===----------------------------------------------------------------------===//
+
+std::unique_ptr<::mlir::Pass> createCopyRemovalPass() {
+  return std::make_unique<CopyRemovalPass>();
+}
```
keryell, line 245: [`61b4fa4` (#580)](https://github.com/Xilinx/mlir-aie/pull/580/commits/61b4fa449ae6abffa745d17475db3f6889204689)

jamestcl-amd: Will address this in another PR

PR #600, opened 2023-08-24
Split ccache database according to the parallel jobs

keryell:
This fixes a race condition in the ccache database writing happening at the end of each job running in parallel.

By using a unique key per job, each database is correctly written and can be used for a next CI run.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -104,7 +104,7 @@ jobs:
           echo "$execution_context_var_inputs"
           echo "::endgroup::"
 
-      - name: Configure Environment
+      - name: Configure environment
         run: echo "$GITHUB_WORKSPACE/llvm/install/bin" >> $GITHUB_PATH
 
       - name: Install git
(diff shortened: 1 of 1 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -142,15 +142,19 @@ jobs:
       - name: Install Clang & Clang++ and linker
         run: apt-get install -y clang lld
 
-      - name: Get Submodule Hash
+      - name: Get related-LLVM Submodule Hash
         id: get-submodule-hash
-        run: echo "hash=$(md5sum $(echo utils/clone-llvm.sh))" >> $GITHUB_OUTPUT
+        # Instead of using the git hash of LLVM, just use the good-enough hash
```
stephenneuendorffer, line 147: If this is now a requirement on the clone-llvm.sh script, perhaps it should be documented?

stephenneuendorffer (approved):

.github/workflows/buildAndTest.yml
keryell: It it not "now a requirement", it has always been there. :-|
It took me some time to understand why this hash was working. This is why I commented (= documented :-) ) the CI here. :-)
I have just removed some tautological code.
But I agree that if some day we remove this textual git hash from `utils/clone-llvm.sh` it would break things, so I should also comment this script.
Thinking again about it and the discussion we had at lunch about reducing the technical debt, I think I will just remove this hack and look at the hash just in the `llvm` worktree after the step running `utils/clone-llvm.sh`.

PR #601, opened 2023-08-24
TOSA end-to-end tests for element-wise op with 1-D tensors

jamestcl-amd:
- This PR requires https://github.com/Xilinx/mlir-aie/pull/599 
- Create TOSA end-to-end tests, with Vector->AIEVec->cpp and then aiesimulator with reference results
- Currently put TOSA tests at test/unit_tests/tosa_tests/*
- List of added tests for tosa element-wise op with 1-D tensors:
  - bf16xbf16_add_elem (lane=16)
  - bf16xbf16_mul_elem (lane=16)
  - bf16xbf16_sel (lane=32)
  - bf16xbf16_sub_elem (lane=16)
  - i16xi16_mul_elem (lane=32)


test/unit_tests/tosa_tests/bf16xbf16_add_elem/testbench.cc
```diff
@@ -0,0 +1,56 @@
...
+  auto cycleCount = (int)(cyclesEnd - cyclesBegin);
+  reportCycleCount(cycleCount, dataDir + "/cycle_count.txt");
+
+  writeData(g_out0, OUT0_SIZE, dataDir + "/out0.txt");
+
+  dut_ref(g_in0, g_in1, g_out0Ref);
+  writeData(g_out0Ref, OUT0_SIZE, dataDir + "/out0_ref.txt");
+
+  bool ok = true;
+  ok &= checkData(g_out0, g_out0Ref, OUT0_SIZE, 1);
```
david-vc, line 42: Remove the last argument (", 1"). 
In this case absTol, relTol, and absTolF will default to 0, which is what we want.
Since we are testing is element-wise bfloat16 addition and given the range for the input data, results should match exactly.
The same change is needed in the other test cases.

david-vc (approved):
Hi James: This is great. Can you please address the 1 comment before merging? Thanks

PR #602, opened 2023-08-25
[aievec] Add hoisting patterns for arith.extsi

jsetoain:
Hoisting cast operations as close as possible to the source of data can make later patterns more robust to typical variations in the source code.

At the moment, it hoists only `extsi` ops, but it'd be trivial to generalize to all `arith:ext` ops.

We might need to revisit this one if, in the future, this process causes unintended consequences due to the widening of data types in operations.

### DIFF
--- a/test/dialect/AIEVec/precanonicalization.mlir
+++ b/test/dialect/AIEVec/precanonicalization.mlir
@@ -66,3 +66,44 @@ func.func @rank_zero_transfer_read(%m : memref<i16>) -> vector<16xi16> {
     // CHECK: return %[[S]] : vector<16xi16>
     return %v : vector<16xi16>
 }
+
+// -----
+
+// CHECK-LABEL: func.func @extsi_hoisting_through_extract_n_bcast(
+// CHECK-SAME: %[[VEC:.*]]: vector<16xi8>
+func.func @extsi_hoisting_through_extract_n_bcast(%v : vector<16xi8>)
+                                                            -> vector<32xi32> {
+// CHECK: %[[EXV:.*]] = arith.extsi %[[VEC]] : vector<16xi8> to vector<16xi32>
+// CHECK: %[[EXS:.*]] = vector.extract %[[EXV]][7] : vector<16xi32>
+// CHECK: %[[BCAST:.*]] = vector.broadcast %[[EXS]] : i32 to vector<32xi32>
+// CHECK: return %[[BCAST]] : vector<32xi32>
+  %si8 = vector.extract %v[7] : vector<16xi8>
+  %vi8 = vector.broadcast %si8 : i8 to vector<32xi8>
+  %vi32 = arith.extsi %vi8 : vector<32xi8> to vector<32xi32>
+  return %vi32 : vector<32xi32>
+}
+
+// -----
+
+// CHECK-LABEL: func.func @extsi_hoisting_through_extract_strided_slice(
+// CHECK-SAME: %[[MEM:.*]]: memref<?xi8>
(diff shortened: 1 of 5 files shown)


david-vc (changes requested):
Hi @jsetoain : 
I'm concerned that the new patterns applies too broadly. In particuler, there is very little that is being checked about the defining op of the operand of the extsi op. Also, we are not checking whether there are any other uses of operand of the extsi. I'm concerned that the new pattern could lead to malformed IR.

test/dialect/AIEVec/precanonicalization.mlir
```diff
@@ -66,3 +66,44 @@ func.func @rank_zero_transfer_read(%m : memref<i16>) -> vector<16xi16> {
...
+// CHECK-SAME: %[[VEC:.*]]: vector<16xi8>
+func.func @extsi_hoisting_through_extract_n_bcast(%v : vector<16xi8>)
+                                                            -> vector<32xi32> {
+// CHECK: %[[EXV:.*]] = arith.extsi %[[VEC]] : vector<16xi8> to vector<16xi32>
+// CHECK: %[[EXS:.*]] = vector.extract %[[EXV]][7] : vector<16xi32>
+// CHECK: %[[BCAST:.*]] = vector.broadcast %[[EXS]] : i32 to vector<32xi32>
+// CHECK: return %[[BCAST]] : vector<32xi32>
+  %si8 = vector.extract %v[7] : vector<16xi8>
+  %vi8 = vector.broadcast %si8 : i8 to vector<32xi8>
+  %vi32 = arith.extsi %vi8 : vector<32xi8> to vector<32xi32>
```
linay-xsj, line 82: Do you have the real user test cases which generate this extsi op pattern?

jsetoain:
> Hi @jsetoain : I'm concerned that the new patterns applies too broadly. In particuler, there is very little that is being checked about the defining op of the operand of the extsi op. Also, we are not checking whether there are any other uses of operand of the extsi. I'm concerned that the new pattern could lead to malformed IR.

I've tried to make it as generic as possible so it's not a pattern that deals with "the one use case", and I believe that should be the defining principle. What we try to accomplish here, is to perform the type cast as early as possible. There are potential risks by doing this (like intermediate operations that are supported on the narrow type, but not on the wide type), but I can't think of an example that would be broken by this pattern, and work otherwise.

Regarding uses of the operand, like with our conversion patterns, if somebody else is using it then it won't be eliminated. You'd end up with two copies:

1) A wide element version that feeds to the users of the original ext
2) The original version, feeding into all the other users.

The riskiest assumption I'm making is that if the elemental type of the defining op's operand the same as it's output, it will also have to be converted. Similarly, I'm ignoring attribute types. I've thought about this, and I can't think of an exception to this rule, but we can manage those when and if it happens. I guess I should add a comment about that (I think I haven't).

test/dialect/AIEVec/precanonicalization.mlir
jsetoain: I do, but it breaks for other reasons. Before it runs end-to-end, I need to fix the issue with Vector2AIEVec letting arithmetic vector operations through unconverted, because, predictably, the legalization pattern fails for this use case. Once that's solved, and together with this, it should work. I'll add the new test case with that (longer) patch.

david-vc (commented):
Hi Javier, 
IMO, the pattern should not apply unless the defOp meets additional requirements. 
Can you consider implementing this?

jsetoain:
> Hi Javier, IMO, the pattern should not apply unless the defOp meets additional requirements. Can you consider implementing this?

Yes, of course, I was assuming we'd discuss this in our next meeting. Nothing easier than adding restrictions here 😄

PR #604, opened 2023-08-28
Put exp lookup table into run_time_lib/AIE2

linay-xsj:


keryell:
Just wondering whether it could be possible to replace the non-open-source friendly binary blob by source code, such as some `constexpr`/`constinit`/consteval` C++ computation to compute the LUT at compile time?

david-vc (changes requested):
Hi @linay-xsj 
aie_runtime_lib/AIE2/exp_lut.cpp should contain just the definitions of the global arrays.
The **definition** of getExpBf16(v16bfloat16) needs to go in aie_runtime_lib/AIE2/exp_lut.h.
This is needed because to get good QOR the function needs to be inline.
For consistency to aie_runtime_lib/AIE1 should we check in libexp_lut.o rather than libexp_lut.a?

jsetoain:
> Just wondering whether it could be possible to replace the non-open-source-friendly binary blob by source code, such as some `constexpr`/`constinit`/`consteval` C++ computation to compute the LUT at compile time?

That is an interesting proposition, but I don't think we can use `std::expf` for that. To declare a `constexpr` variable, the initializer needs to be declared as `constexpr` as well, and the standard `extf` isn't (because it's a C function). It looks like they've changed that... for C++26. You're ahead of the times, Ronan 😄

linay-xsj:
> 

Thanks for pointing this out, David.

linay-xsj:
> > Hi @linay-xsj aie_runtime_lib/AIE2/exp_lut.cpp should contain just the definitions of the global arrays.
> 
> You will also need to declare the arrays in exp_lut.h, as "extern" variables.



> > Hi @linay-xsj aie_runtime_lib/AIE2/exp_lut.cpp should contain just the definitions of the global arrays.
> 
> You will also need to declare the arrays in exp_lut.h, as "extern" variables.

Yes, we need to declare them as extern variables. Thanks for the reminder, Javier.

aie_runtime_lib/AIE2/exp_lut.cpp
```diff
@@ -0,0 +1,203 @@
+#include "aie_api/aie.hpp"
```
david-vc, line 1: Please add the same copyright notice as you have in exp_lut.h

david-vc (approved):
Looks good; please do add the copyright notice to aie_runtime_lib/AIE/exp_lut.cpp

PR #610, opened 2023-08-29
Update chess_intrinsic_wrapper.cpp

fifield:
Remove event intrinsic declarations from AIEv1 wrapper

These were added in https://github.com/Xilinx/mlir-aie/pull/591 but seem to be causing compiler crashes when targeting AIE1, so removing for now.

### DIFF
--- a/aie_runtime_lib/AIE/chess_intrinsic_wrapper.cpp
+++ b/aie_runtime_lib/AIE/chess_intrinsic_wrapper.cpp
@@ -25,5 +25,3 @@ extern "C" void llvm___aie___lock___acquire___reg(unsigned id, unsigned val) {
 extern "C" void llvm___aie___lock___release___reg(unsigned id, unsigned val) {
   release(id, val);
 }
-extern "C" void llvm___aie___event0() { event0(); }
-extern "C" void llvm___aie___event1() { event1(); }


fifield:
The crash is:
```
Internal error: while translating intrinsic `void event(uint2_t)' mismatched input type `uint2_t' DSFG type `uint2_t' is not promotable on `uint1_t' (coming from llvm)
```
So one theory is the intrinsic signature changed from AIE1 to AIE2 but somehow we are not generating the right one.

stephenneuendorffer:
Yeah I noticed that, but was working through getting the streams and cascade to work first.

aie_runtime_lib/AIE/chess_intrinsic_wrapper.cpp
```diff
@@ -25,5 +25,3 @@ extern "C" void llvm___aie___lock___acquire___reg(unsigned id, unsigned val) {
 extern "C" void llvm___aie___lock___release___reg(unsigned id, unsigned val) {
   release(id, val);
 }
-extern "C" void llvm___aie___event0() { event0(); }
-extern "C" void llvm___aie___event1() { event1(); }
```
keryell, line 29: Perhaps you can just comment them out with a comment and a `\todo` for clarity and so they can be re-instantiated even more easily?

PR #612, opened 2023-08-30
Implement inverse of a float by lookup tables

linay-xsj:


test/unit_tests/aievec_tests/bf16_inv_lut/testbench.cc
```diff
@@ -0,0 +1,55 @@
...
+  auto cycleCount = (int)(cyclesEnd - cyclesBegin);
+  reportCycleCount(cycleCount, dataDir + "/cycle_count.txt");
+
+  writeData(g_out0, OUT0_SIZE, dataDir + "/out0.txt");
+
+  dut_ref(g_in0, sum, g_out0Ref);
+  writeData(g_out0Ref, OUT0_SIZE, dataDir + "/out0_ref.txt");
+
+  bool ok = true;
+  ok &= checkData(g_out0, g_out0Ref, OUT0_SIZE, 0, 1e-2, 1e-2);
```
david-vc, line 39: Since in this test case we dividing each number in the input by 1234.56f, an absolute tolerance of 0.01 is much too large. Can we set the absolute tolerance to 0 and only specify the relative tolerance of 0.01?

aie_runtime_lib/AIE/exp_lut.cpp
```diff
@@ -214,3 +214,14 @@ alignas(aie::vector_decl_align) int16 softmax_flut_cd[512] = {
     16422, 16423, 16423, 16424, 16419, 16420, 16421, 16421, 16422, 16423, 16423,
     16424, 16425, 16425, 16426, 16427, 16427, 16428, 16429, 16429, 16425, 16425,
     16426, 16427, 16427, 16428, 16429, 16429};
+
+alignas(aie::vector_decl_align) unsigned char m_inv_lut[128] = {
```
david-vc, line 218: Now that this file contains lut tables for exp and inv, it should probably be renamed. How about lut_based_ops.cpp and lut_based_ops.h ?

david-vc: Also, can we strive towards some consistency in the naming of these global variables?

david-vc (changes requested):
I have some minor requests in the comments.

david-vc (approved):

PR #624, opened 2023-09-05
Bump cmakeModules

fifield:
The cmakeModules submodule was reverted to an earlier version in https://github.com/Xilinx/mlir-aie/pull/573. I'm guessing this was inadvertent. This PR restores the submodule to the latest commit.


keryell:
Thanks for fixing my mess!

PR #626, opened 2023-09-05
[NO MERGE] Wheels

makslevental:


.github/workflows/distro.yml
```diff
@@ -0,0 +1,324 @@
...
+    - name: Setup tmate session
+      uses: mxschmitt/action-tmate@v3
+      if: ${{ inputs.DEBUG_ENABLED && inputs.DEBUG_OS == matrix.OS && inputs.DEBUG_ARCH == matrix.ARCH }}
+      with:
+        limit-access-to-actor: true
+        detached: ${{ inputs.DEBUG_DETACHED }}
+
+    - uses: actions/checkout@v3
+
+    - uses: actions/checkout@v3
```
keryell, line 99: You should add a `name` to help the reading of the output to avoid confusion with previous step.

.gitignore
```diff
@@ -0,0 +1,8 @@
+/.idea*
+/build*
+/install*
+/sandbox*
+/llvm*
+/mlir_aie*
+.vscode
+__pycache__
```
keryell, line 8: Missing EOL in a few files.
You do not like my https://github.com/Xilinx/mlir-aie/blob/main/.editorconfig ? :-)

setup.py
```diff
@@ -0,0 +1,251 @@
...
+
+setup(
+    version=version,
+    author="",
+    name="mlir-aie",
+    include_package_data=True,
+    description=f"An MLIR-based toolchain for Xilinx Versal AIEngine-based devices.",
+    long_description=dedent(
+        """\
+        This repository contains an [MLIR-based](https://mlir.llvm.org/) toolchain for Xilinx Versal
```
keryell, line 235:
```suggestion
        This repository contains an [MLIR-based](https://mlir.llvm.org/) toolchain for AMD Versal 
```

keryell (commented):
Cool stuff!

.github/workflows/distro.yml
```diff
@@ -0,0 +1,323 @@
...
+        else
+          sudo apt-get install -y rename
+        fi
+        rename 's/cp311-cp311/py3-none/' wheelhouse/mlir_aie-*whl
+        
+        if [ x"${{ matrix.OS }}" == x"ubuntu-20.04" ] && [ x"${{ matrix.ARCH }}" == x"aarch64" ]; then
+          rename 's/x86_64/aarch64/' wheelhouse/mlir_aie-*whl
+        fi
+
+    - name: rename
```
keryell, line 248: Name conflict with previous step.

makslevental: They never execute on the same runner...

.gitignore
makslevental: If you flip to this branch you'll see there's nothing at the root - except for `.vscode` (a mistake) these are the only generated artifacts.

.github/workflows/distro.yml
keryell: Whatever, this is confusing in https://github.com/Xilinx/mlir-aie/actions/runs/6103552291/job/16564151441?pr=626 with 2 "rename".

.github/workflows/distro.yml
```diff
@@ -0,0 +1,272 @@
...
+    - name: Install boost
+      uses: MarkusJx/install-boost@v2.4.4
+      id: install-boost
+      with:
+        boost_version: 1.72.0
+        platform_version: ${{ steps.split.outputs.platform_version }}
+        arch: ${{ matrix.ARCH == 'x86_64' && 'x86' || 'aarch64' }}
+
+    - name: cibuildwheel
+      if: ${{ matrix.OS != 'ubuntu-20.04' || matrix.ARCH != 'aarch64' }}
```
keryell, line 134: While you are in a cleaning mood, you can remove probably all the `${{ }}` in `if:` https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

.github/workflows/distro.yml
```diff
@@ -0,0 +1,272 @@
...
+      working-directory: ${{ steps.setup_base.outputs.WORKSPACE_ROOT }}
+      shell: bash
```
keryell, line 136: Move to the top level?
https://github.com/Xilinx/mlir-aie/blob/main/.github/workflows/buildAndTest.yml#L16

makslevental: this can't be at top level because of 

```
    - name: rename
      if: ${{ matrix.OS == 'windows-2019' }}
      working-directory: ${{ steps.setup_base.outputs.WORKSPACE_ROOT }}
      run: |
        ls wheelhouse/mlir_aie-*whl | Rename-Item -NewName {$_ -replace 'cp311-cp311', 'py3-none' }
```

keryell: Or the opposite with `shell: ...` for the Windows case?

PR #627, opened 2023-09-06
Black/format aiecc.py

makslevental:
Upstream moved to using black for formatting months ago so e.g., four space indents like is generally considered "pythonic".


fifield:
Can it be added to the workflow checks alongside `clang-format` so that it stays formatted?

keryell (commented):
That sounds going along https://peps.python.org/pep-0008/
While you are here, could you update https://github.com/Xilinx/mlir-aie/blob/formataiecc/.editorconfig
to copy https://github.com/Xilinx/mlir-aie/blob/formataiecc/.editorconfig#L27 in another `[*.py]` section?

stephenneuendorffer (changes requested):

PR #628, opened 2023-09-07
Enable use of mlir pass manager in aiecc

fifield:
* add aievec and aiex to CAPI
* ensure passes are registered
* convert the first `aie-opt` in `aiecc` to use `PassManager` directly

### DIFF
--- a/python/aie/compiler/aiecc/main.py
+++ b/python/aie/compiler/aiecc/main.py
@@ -15,11 +15,13 @@ import stat
 import platform
 import sys
 import time
-from subprocess import PIPE, run, call
-import tempfile
+import subprocess
 import shutil
-import timeit
 import asyncio
+
+from aie.mlir.passmanager import PassManager
+from aie.mlir.ir import Module, Context, Location
+from aie.dialects import aie as aiedialect
 
 import aie.compiler.aiecc.cl_arguments
 import aie.compiler.aiecc.configure
@@ -42,7 +44,8 @@ aie_opt_passes = ['--aie-normalize-address-spaces',
                   '--cse']
 
 class flow_runner:
-  def __init__(self, opts, tmpdirname):
+  def __init__(self, mlir_module, opts, tmpdirname):
+      self.mlir_module = str(mlir_module)
       self.opts = opts
       self.tmpdirname = tmpdirname
       self.runtimes = dict()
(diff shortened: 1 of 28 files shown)


python/aie/compiler/aiecc/main.py
```diff
@@ -15,12 +15,14 @@
...
-from subprocess import PIPE, run, call
-import tempfile
+import subprocess
 import shutil
-import timeit
 import asyncio
 
+from aie.mlir.passmanager import PassManager
+from aie.mlir.ir import Module, Context, Location
+from aie.dialects import aie as aiedialect
```
stephenneuendorffer, line 24: My main concern here is in how much of a python environment dependency this implies...  For instance, this will make cross compiling much harder, since we now have a python binding dependency.  Also, there are some parts that will still need to be out of process, when we outline each core, so not everything will convert this way?

python/aie/compiler/aiecc/main.py
```diff
@@ -417,18 +429,23 @@ async def run_flow(self):
         self.progress_bar = progress_bar
         progress_bar.task = progress_bar.add_task("[green] MLIR compilation:", total=1, command="1 Worker")
 
+        aiedialect.register_dialect(ctx)
+        mlir_module = Module.parse(self.mlir_module)
```
stephenneuendorffer, line 433: This is slick... Maybe next look at some of the places where this would make life much simpler, for instance where we today use aie-translate to query the device, instead we could just parse the MLIR directly.

python/aie/compiler/aiecc/main.py
makslevental: although I tend to agree with this sentiment (if you can pass strings you probably should) cross-compiling the bindings (assuming you have the cross-compiled distro) isn't a [huge lift](https://github.com/makslevental/mlir-wheels/blob/main/.github/workflows/wheels.yml#L176); I plan to add that to bindings wheel matrix shortly.

python/aie/compiler/aiecc/main.py
```diff
@@ -83,9 +86,18 @@ async def do_call(self, task, command, force=False):
...
+      ret = subprocess.run(command, stdout=m, stderr=m, universal_newlines=True)
       return ret
 
+  def run_passes(self, pass_pipeline, mlir_module, outputfile=None):
+      if self.opts.verbose:
+        print("Running:", pass_pipeline)
+      PassManager.parse(pass_pipeline).run(mlir_module)
+      if outputfile:
+        with open(outputfile, 'w') as g:
+          g.write(str(mlir_module))
```
makslevental, line 99: @fifield 

My recommendation is to narrow the extent/scope for the `with Context() as ctx, Location.unknown()`. Currently you have nested contexts (one with in `main` and then another in `run_flow`). You could collapse them both to being just here; if instead of parsing the module in `main` and then holding onto it as a `ir.Module`, you just held the string and only constructed the `ir.Module` here (i.e., just before running the PassManager on it).

fifield: Good point. I store and parse the module as a string in `run_flow` because I need to have a new context due to the multi-threading (at least I think that's what's going on, it didn't work with a single context).

python/aie/compiler/aiecc/main.py
fifield: Yeah that's kind of the point of doing this :)

PR #630, opened 2023-09-07
need ONLY option to make cmake find numpy

denolf:
FIRST option fails to find numpy in my setup (I am running from the sandbox as indicated in the build instructions)


fifield (approved):
We have it set to ONLY elsewhere for the same reason.

keryell (approved):
I made this change in https://github.com/Xilinx/mlir-aie/commit/9e879ac343c14e92960d1fb638589561c1041e56 to have this compiling on my laptop which has a not a Xilinx Linux install. I guess there is something in Xilinx install or configuration files that breas things.
Note that in the CI it works, but this is not a Xilinx environment.
Anyway, if it works for you, this is good. :-)

denolf:
merged #630

PR #635, opened 2023-09-11
Add softmax test cases

linay-xsj:
Add two version of softmax test cases.
The change in the lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp can avoid the redundant include operations.


test/unit_tests/aievec_tests/bf16_softmax_2/dut.cc
```diff
@@ -0,0 +1,107 @@
+#include "lut_based_ops.h"
+
```
david-vc, line 2: Could you add a comment with the measured cycle count?

david-vc (approved):
Looks good. Please add the requested comment with measured cycle count.

PR #638, opened 2023-09-14
[aievec] Generalize vector passes

jsetoain:
Right now, vectorization passes hook to FuncOp, which prevents conversion to AIEVec within other top level operations, like AIE.device ops.

This patch makes all passes generic and allows for conversion within AIE.device.


lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
```diff
@@ -11,6 +11,7 @@
 // it compatible with the available vector instructions in AIE architectures
 //===----------------------------------------------------------------------===//
 
+#include "aie/Dialect/AIE/IR/AIEDialect.h"
```
jamestcl-amd, line 14: For my own understanding, why do we need to include this header?

jsetoain: We don't, I added it for one of my first attempts at this, and I forgot to take it out once it wasn't necessary. Good catch!

linay-xsj (approved):

PR #654, opened 2023-09-23
bump LLVM

makslevental:
Note this PR has 2 commits; base commit has substantive changes and second commit has what I call "be the [diff] you want to see in the world", namely I applied all of the clang-tidy suggestions to the cpp files that were touched. Feel free to push back on any of those chagnes and I will revert.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
@@ -12,7 +12,6 @@
 // to ops that can be translated to a sequence of valid AIEVec ops.
 //===----------------------------------------------------------------------===//
 
-#include <algorithm>
 #include <bitset>
 #include <optional>
 #include <tuple>
@@ -21,9 +20,7 @@
 #include "aie/Dialect/AIEVec/IR/AIEVecOps.h"
 #include "aie/Dialect/AIEVec/Pipelines/Passes.h"
 #include "aie/Dialect/AIEVec/Utils/Utils.h"
-#include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/EmitC/IR/EmitC.h"
-#include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
@@ -31,7 +28,6 @@
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/DialectConversion.h"
-#include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "mlir/Transforms/Passes.h"
 #include "llvm/ADT/SmallSet.h"
 #include "llvm/ADT/TypeSwitch.h"
@@ -71,8 +67,8 @@ extractMACOperandsFromAddOperands(Value addLhs, Value addRhs) {
     return std::make_tuple(mulOp.getLhs(), mulOp.getRhs(), acc);
 
   // If the MulIOp has been already translated to aievec::MulOp:
-  aievec::SRSOp lhsSrsOp = addLhs.getDefiningOp<aievec::SRSOp>();
-  aievec::SRSOp rhsSrsOp = addRhs.getDefiningOp<aievec::SRSOp>();
+  auto lhsSrsOp = addLhs.getDefiningOp<aievec::SRSOp>();
+  auto rhsSrsOp = addRhs.getDefiningOp<aievec::SRSOp>();
   aievec::MulOp aieMulOp = nullptr;
   if (lhsSrsOp) {
     aieMulOp = lhsSrsOp.getSource().getDefiningOp<aievec::MulOp>();
@@ -120,7 +116,7 @@ convertValueToTargetTypeAieML(ConversionPatternRewriter &rewriter, Location loc,
           loc, srcType, broadcastZeroOp->getResult(0), 0);
 
       SmallVector<Value> inputSources = {inputVal, extOp->getResult(0)};
-      aievec::ConcatOp concatOp =
+      auto concatOp =
           rewriter.create<aievec::ConcatOp>(loc, tgtType, inputSources);
 
       return concatOp.getResult();
@@ -224,8 +220,9 @@ buildAttributeListForRotationSelectOp(PatternRewriter &rewriter, VectorType vTy,
          {yoffsetsAttrName, attr0},
          {ysquareAttrName, attr0},
          {ystartAttrName, attr0}});
+  default:
+    return {};
   }
-  return {};
 }
 
 namespace xilinx {
@@ -286,8 +283,9 @@ SmallVector<NamedAttribute> buildFMAOpSplatAttrForElemTy(aievec::FMAOp fmaOp,
          {fmaOp.getZstepAttrName(), fmaOp.getZstepAttr()},
          {fmaOp.getZsquareAttrName(), fmaOp.getZsquareAttr()},
          {fmaOp.getFmsubAttrName(), fmaOp.getFmsubAttr()}});
+  default:
+    return {};
   }
-  return {};
 }
 
 } // namespace aievec
@@ -375,7 +373,8 @@ static aievec::CmpOp createCmpOpAieML(ConversionPatternRewriter &rewriter,
 template <typename DstOpTy>
 static void generateAIEVecOpsForReductionOp(ConversionPatternRewriter &rewriter,
                                             vector::ReductionOp srcOp,
-                                            int shiftIndex, Value curValue) {
+                                            unsigned shiftIndex,
+                                            Value curValue) {
   assert(shiftIndex > 0 && (shiftIndex & (shiftIndex - 1)) == 0 &&
          "shiftIndex must be power of 2");
 
@@ -387,8 +386,8 @@ static void generateAIEVecOpsForReductionOp(ConversionPatternRewriter &rewriter,
   DstOpTy curOp = nullptr;
   unsigned elWidth = scalarType.getIntOrFloatBitWidth();
 
-  for (int id = shiftIndex; id > 0; id /= 2) {
-    arith::ConstantOp constOp = rewriter.create<arith::ConstantOp>(
+  for (unsigned id = shiftIndex; id > 0; id /= 2) {
+    auto constOp = rewriter.create<arith::ConstantOp>(
         loc, rewriter.getI32IntegerAttr(id * elWidth / 8));
 
     auto shiftBytesOp = rewriter.create<aievec::ShiftOp>(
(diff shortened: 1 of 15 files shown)


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -71,8 +67,8 @@ extractMACOperandsFromAddOperands(Value addLhs, Value addRhs) {
     return std::make_tuple(mulOp.getLhs(), mulOp.getRhs(), acc);
 
   // If the MulIOp has been already translated to aievec::MulOp:
-  aievec::SRSOp lhsSrsOp = addLhs.getDefiningOp<aievec::SRSOp>();
-  aievec::SRSOp rhsSrsOp = addRhs.getDefiningOp<aievec::SRSOp>();
+  auto lhsSrsOp = addLhs.getDefiningOp<aievec::SRSOp>();
+  auto rhsSrsOp = addRhs.getDefiningOp<aievec::SRSOp>();
```
makslevental, line 71: Use auto when initializing with a template cast to avoid duplicating the type name

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -286,8 +283,9 @@ SmallVector<NamedAttribute> buildFMAOpSplatAttrForElemTy(aievec::FMAOp fmaOp,
          {fmaOp.getZstepAttrName(), fmaOp.getZstepAttr()},
          {fmaOp.getZsquareAttrName(), fmaOp.getZsquareAttr()},
          {fmaOp.getFmsubAttrName(), fmaOp.getFmsubAttr()}});
+  default:
+    return {};
```
makslevental, line 287: Uncovered code path - add default case.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -375,7 +373,8 @@ static aievec::CmpOp createCmpOpAieML(ConversionPatternRewriter &rewriter,
 template <typename DstOpTy>
 static void generateAIEVecOpsForReductionOp(ConversionPatternRewriter &rewriter,
                                             vector::ReductionOp srcOp,
-                                            int shiftIndex, Value curValue) {
+                                            unsigned shiftIndex,
+                                            Value curValue) {
```
makslevental, line 377: Narrowing from int to unsigned is implementation dependent.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -400,47 +399,12 @@ static void generateAIEVecOpsForReductionOp(ConversionPatternRewriter &rewriter,
...
-      auto userUpdOp = dyn_cast<xilinx::aievec::UPDOp>(user);
-      if (userUpdOp)
-        sizeInBits -= userUpdOp.getOffset();
-    }
-    effectiveSize = sizeInBits;
-  }
-
-  unsigned effectiveSize;
-};
-
```
makslevental, line 443: dead

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -528,8 +492,8 @@ struct ConvertMulAddToAIEVecFMAElemOpPattern
     : public OpConversionPattern<arith::AddIOp> {
   using OpConversionPattern<arith::AddIOp>::OpConversionPattern;
 
-  ConvertMulAddToAIEVecFMAElemOpPattern(MLIRContext *context,
-                                        unsigned shiftParam = 0)
+  explicit ConvertMulAddToAIEVecFMAElemOpPattern(MLIRContext *context,
```
makslevental, line 495: Single-argument constructors must be marked explicit to avoid unintentional implicit conversions

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2342,15 +2306,17 @@ struct LowerVectorToAIEVec
...
       : LowerVectorToAIEVec() {
     aieTarget = options.aieTarget;
   }
 
   // In case we want to register this pass as a standalone pass for test
   // purposes.
-  StringRef getArgument() const final { return "test-lower-vector-to-aievec"; }
-  StringRef getDescription() const final {
+  [[nodiscard]] StringRef getArgument() const final {
+    return "test-lower-vector-to-aievec";
```
makslevental, line 2317: [nodiscard](https://en.cppreference.com/w/cpp/language/attributes/nodiscard)

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2369,10 +2335,9 @@ struct LowerVectorToAIEVec
     auto op = getOperation();
     MLIRContext *context = &getContext();
     RewritePatternSet patterns(context);
-    ConversionTarget target(*context);
     AIEArch aieVersion = AIEArch::AIE;
     if (!aieTarget.empty()) {
-      std::string target = aieTarget;
+      std::string target = aieTarget.getValue();
```
makslevental, line 2340: There's an implicit downcast but better to be explicit (the actual clang-tidy reco was about 8 bytes discarded or something like that).

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2382,6 +2347,7 @@ struct LowerVectorToAIEVec
       }
     }
 
+    ConversionTarget target(*context);
```
makslevental, line 2350: Moved from above because `target` in the conditional shadows.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -391,15 +393,6 @@ static LogicalResult createLinearizedAccess(CppEmitter &emitter, Value source,
...
 }
 
-// Return true if the array accessed by this value is readonly
-static bool isReadOnly(Value read) {
-  for (auto *user : read.getUsers()) {
-    if (isa<vector::TransferWriteOp>(user))
-      return false;
-  }
-  return true;
-}
```
makslevental, line 401: replace with `std::none_of` and move to call site.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2423,8 +2417,8 @@ static LogicalResult printOperation(CppEmitter &emitter,
           for (OpResult result : op->getResults()) {
             if (failed(emitter.emitVariableDeclaration(
                     result, /*trailingSemicolon=*/true))) {
-              return WalkResult(
-                  op->emitError("unable to declare result variable for op"));
+              return {
+                  op->emitError("unable to declare result variable for op")};
```
makslevental, line 2421: Avoid repeating the return type from the declaration; use a braced initializer list instead

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2508,7 +2502,7 @@ void CppEmitter::setMemRefDimParam(Value memref, unsigned index,
                                    std::string parameter) {
   auto p = std::make_pair(memref, index);
   assert(!paramIndexMapper.count(p) && "memref dimension already set");
-  paramIndexMapper[p] = parameter;
+  paramIndexMapper[p] = std::move(parameter);
```
makslevental, line 2505: Parameter 'parameter' is passed by value and only copied once; consider moving it to avoid unnecessary copies

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2646,7 +2639,7 @@ LogicalResult CppEmitter::emitAttribute(Location loc, Attribute attr) {
       if (auto vType = dense.getType().dyn_cast<VectorType>()) {
         if (auto fType = vType.getElementType().dyn_cast<FloatType>()) {
           unsigned width = fType.getWidth();
-          std::string splatValue = "";
+          std::string splatValue;
```
makslevental, line 2642: Redundant string initialization

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -357,15 +359,15 @@ static LogicalResult createLinearizedAccess(CppEmitter &emitter, Value source,
...
+  for (int64_t dim = memRefType.getRank() - 1; dim >= 0; --dim) {
     // All the indices in the access expression must already be emitted
     if (!emitter.hasValueInScope(indices[dim]))
       return failure();
 
     // Form the access string for this dimension
     std::string cur;
     if (!paramPart.empty())
-      cur = paramPart + "*";
+      cur += paramPart + "*";
```
makslevental, line 370: String concatenation results in allocation of unnecessary temporary strings; consider using 'operator+=' or 'string::append()' instead

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2545,7 +2539,6 @@ StringRef CppEmitter::getOrCreateName(Block &block, std::string prefix) {
 bool CppEmitter::shouldMapToUnsigned(IntegerType::SignednessSemantics val) {
   switch (val) {
   case IntegerType::Signless:
-    return false;
```
makslevental, line 2548: Switch has 2 consecutive identical branches

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: Why are we switching to "unsigned" for `shiftIndex` at all? I'm not a fan of unsigned types unless they're strictly necessary, signed types are generally safer to use.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -1182,7 +1176,7 @@ static LogicalResult printFMAOrMulElemOperand(CppEmitter &emitter, T op,
 // Print lhs or rhs operand of mul_conv/mac_conv intrinsic
 template <typename T>
 static LogicalResult printFMAOrMulConvOperand(CppEmitter &emitter, T op,
-                                              int32_t M, int32_t N,
+                                              uint32_t M, uint32_t N,
```
jsetoain, line 1179: Same here, do we really need the extended range of an unsigned type?

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -2522,7 +2516,7 @@ StringRef CppEmitter::getMemRefDimParam(Value memref, unsigned index) {
 /// associated with it
 bool CppEmitter::isMemRefDimParam(Value memref, unsigned index) {
   assert([&] {
-    MemRefType type = memref.getType().dyn_cast<MemRefType>();
+    auto type = memref.getType().dyn_cast<MemRefType>();
```
jsetoain, line 2519: Since we're changing this line, can we fix the postfixed use of `dyn_cast`? I believe it's being deprecated.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -307,7 +309,7 @@ static LogicalResult parseMemRefDynamicDims(CppEmitter &emitter,
   // Step2: Iterate over all the block arguments, and make sure that the memref
   // args have a parameter associated with the dynamic sized dimension
   for (BlockArgument arg : func.getArguments()) {
-    MemRefType argType = arg.getType().dyn_cast<MemRefType>();
+    auto argType = arg.getType().dyn_cast<MemRefType>();
```
jsetoain, line 312: Here as well, `dyn_cast<....>(xxx)` is preferable to `xxx.dyn_cast<...>()`

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -326,7 +328,7 @@ static LogicalResult parseMemRefDynamicDims(CppEmitter &emitter,
 // Print the memref dims, if the memref has dynamic shape
 static LogicalResult printMemRefDims(CppEmitter &emitter, BlockArgument arg) {
   raw_indented_ostream &os = emitter.ostream();
-  MemRefType argType = arg.getType().dyn_cast<MemRefType>();
+  auto argType = arg.getType().dyn_cast<MemRefType>();
```
jsetoain, line 331: Ditto here

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -342,7 +344,7 @@ static LogicalResult printMemRefDims(CppEmitter &emitter, BlockArgument arg) {
 static LogicalResult createLinearizedAccess(CppEmitter &emitter, Value source,
                                             SmallVector<Value, 4> indices,
                                             std::string &access) {
-  MemRefType memRefType = source.getType().dyn_cast<MemRefType>();
+  auto memRefType = source.getType().dyn_cast<MemRefType>();
```
jsetoain, line 347: Ditto here

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -478,7 +471,7 @@ static LogicalResult printOperation(CppEmitter &emitter, aievec::UPDOp updOp) {
 
   raw_indented_ostream &os = emitter.ostream();
   Value result = updOp.getResult();
-  VectorType resultType = result.getType().cast<VectorType>();
+  auto resultType = result.getType().cast<VectorType>();
```
jsetoain, line 474: Ditto here

jsetoain (commented):
Thanks for the clean-up, Maks! 🙂

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
makslevental: The root cause is some of the upstream APIs return unsigned (e.g., `getPosition`) and so that carries forward and leads to narrowing conversions. I guess you're saying signed is better because it overflows rather than ... (not sure what happens for unsigned - is it UB?).

makslevental:
@jsetoain thanks for the review but just a heads up - this is now way back in queue because of https://github.com/Xilinx/mlir-aie/pull/656.

jsetoain:
> @jsetoain thanks for the review but just a heads up - this is now way back in queue because of #656.

Yes, I saw that one 🙂 There's a lot of good stuff in this patch regardless, so I thought I'd give it a review 😄

PR #658, opened 2023-09-26
Enable AIEX dialect bindings

fifield:
Enable AIEX dialect python bindings + a smoke test


python/CMakeLists.txt
```diff
@@ -36,6 +36,16 @@ declare_mlir_dialect_python_bindings(
   DIALECT_NAME AIE
 )
 
+declare_mlir_dialect_python_bindings(
+  ADD_TO_PARENT AiePythonSources.Dialects
+  ROOT_DIR "${AIE_PYTHON_ROOT_DIR}"
+  TD_FILE dialects/AiexBinding.td
```
stephenneuendorffer, line 42: Generally, I'd prefer "AIEX" to "Aiex"  (but I'd also prefer 'AIE' to 'Aie')

AndraBisca (approved):

python/CMakeLists.txt
fifield: Agreed. I've done some renaming.

PR #659, opened 2023-09-26
Add more add/sub/mul mixed precision tests

jamestcl-amd:
- Refactor the tosa-to-vector pipelines script in each test to a central place at `test/Integration/lit.local.cfg` for better maintainability. Also, make sure each .mlir test is running in a unique workdir for placing multiple .mlir test in a single directory.
- For add/sub/mul mixed precision tests, we add tests with swapped inputs
- Per the TOSA spec at https://www.mlplatform.org/tosa/tosa_spec.html#_mul, we add test coverage for i16xi16_mul_elem_i32, and i8xi8_mul_elem_i32. Our refactored mul_elem lowering pattern works on these two cases directly, and the acctype for the i8/i16 mac intrinsics we used is i32.


david-vc (approved):
Thanks for refactoring the command lines.

PR #661, opened 2023-09-28
Lower tosa.abs, math.absi and math.absf patterns to aievec dialects for i8, i16, i32, float and bfloat16 types

linay-xsj:


aie_runtime_lib/AIE2/vec_math.h
```diff
@@ -145,4 +145,39 @@ inline __attribute__((always_inline)) v16bfloat16 getErfBf16(v16bfloat16 in) {
...
+  return (v16bfloat16)absX;
+}
+
+inline __attribute__((always_inline)) v16float getAbs(v16float in) {
+  aie::vector<float, 16> x = in;
+  aie::vector<float, 16> absX = aie::abs(x);
+  return (v16float)absX;
+}
+
+inline __attribute__((always_inline)) v16int32 getAbs(v16int32 in) {
```
jamestcl-amd, line 166: Is v32int32 supported in high-level API? If so, I think we can add the support as well, thanks. We have element-wise add/sub using v32int32 in some scenarios, which might need it. Thank you.

linay-xsj: In this case, we need to convert v32int32 to v32acc32 first. If we need to support more patterns, we can add more cases in the future.

david-vc (approved):
Looks good to me.

PR #663, opened 2023-09-29
Use ubuntu-20.02 for clang-format

fifield:
clang-format-12 is no longer supported by ubuntu-latest https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -218,7 +218,7 @@ jobs:
 
   lint-repo:
     name: Check code format
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       # We'll be running clang-tidy later in this flow.
       - name: Install clang-tidy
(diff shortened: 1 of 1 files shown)


stephenneuendorffer:
Unfortunately, 20.04 and 22.04 don't support clang-12 anymore.  I'd rather run the clang-tidy task under 20.04 than move to a clang-format that isn't available on our dev machines.  I do think we should start using clang-14 on 22.04, though.

stephenneuendorffer (changes requested):
I'd rather just run this on 20.04, rather than ubuntu-latest

fifield:
> I'd rather just run this on 20.04, rather than ubuntu-latest

I agree this is a safer approach for now since most users are on 20.04. I'll update the PR. But to answer your earlier question I think we should move to clang-14 (and 22.04+) across the board.

stephenneuendorffer:
I'd agree, except that clang-14 isn't an one click install on 20.04...

PR #664, opened 2023-09-29
Add Multi-Dimensional Address Generation to Object Fifo

jgmelber:
Creating a branch within Xilinx/milir-aie/main. Comment from original PR #589

> I recently added support for the n-dimensional data layout transformations (strides/wraps) of AIE2 to the dmaBdOp, https://github.com/Xilinx/mlir-aie/pull/547. This PR adds the same functionality at the objectFifo level.
> 
> You can specify two data layout transformations; toStream and fromStream. toStream is applied at the producer side; data in the buffer referenced by the objectFifo is put in the stream by following the strides and wraps given. fromStream is the analog on the consumer side; what comes in from the stream is written into the memory addresses obtained by applying the given strides wraps.
> 
> Since these data layout transformations require a DMA to perform them, if either toStream or fromStream is used, a DMA connection between tiles is forced, even if they are adjacent and have shared memory. (This required some light refactoring of the objectFifo code.)
> 
> To be more consistent with the AIEX.dma_nd_memcpy operation, which first lists lengths and then strides, I also refactored the strides and wraps tuples in both dmaBdOp and objectFifo (this PR) to now be in order <wrap, stepsize> (== synonymously <length, stride>).


AndraBisca:
Closing because updates can be done directly to the other PR.

PR #669, opened 2023-10-04
Add test for python aiecc

fifield:


python/aie/compiler/aiecc/main.py
```diff
@@ -547,10 +547,12 @@ def run(mlir_module, args=None):
...
         sys.stderr.write('\ncompiling %s\n' % opts.filename)
 
-    if(opts.tmpdir):
+    if opts.tmpdir:
       tmpdirname = opts.tmpdir
-    else:
+    elif opts.filename:
       tmpdirname = os.path.basename(opts.filename) + ".prj"
+    else:
+      tmpdirname = 'aiecc.prj'
```
stephenneuendorffer, line 555: Should this be in a temporary directory?

stephenneuendorffer (approved):

python/aie/compiler/aiecc/main.py
fifield: yeah that's a good idea, and can be overridden anyway if needed.

PR #671, opened 2023-10-06
Dynamic size support for the auto-vectorization to CPP flow

jamestcl-amd:
- TOSA broadcastable op semantics:
   - The `Broadcastable` Trait: https://mlir.llvm.org/docs/Traits/Broadcastable/
 - Assumptions:
   - Runtime memory load/store should be memory aligned.
   - Runtime sizes should be vectorizable by the given vector lane size.
   - Runtime sizes of two operands of binary op should be equal. 
   - Runtime sizes will never be one and thus no need of implicit broadcast for the dim with runtime size = 1. 
 - Required attribute:
   - In order to get the MLIR-AIE's auto-vectorization to CPP pipeline to work, the input TOSA function should contains the `tosa.no_implicit_broadcast_of_dynamic_sizes` attribute indicating that the runtime dynamic sizes dimension will never need an implicit broadcast.
 - Change list:
   - Clean up `AIEVec/Pipelines/Passes.h` include files
   - Move `createCopyRemovalPass` to xilinx::aievec namespace
   - Add `createDynamicSizeNoImplicitBroadcastPass` to xilinx::aievec namespace
   - Add lit tests for the --dynamic-size-no-implicit-broadcast pass
   - Add tosa integration tests for dynamic size dimensions
 - Other experiments:
   - Tried the `cf.assert` + `scf.if` combo but it doens't work:
```
%dim_1 = tensor.dim %arg0, %c0 : tensor<?xbf16>
%gtone = arith.cmpi ugt, %dim_1, %c1 : index
cf.assert %gtone, "Should always > 1"
%2 = scf.if %gtone -> (tensor<?xbf16>) {
  scf.yield %arg0 : tensor<?xbf16>
} else {
  cf.assert %false, "NO BROADCAST SUPPORTED"
  scf.yield %arg1 : tensor<?xbf16>
}

// The corresponding canonicalized output:
%dim = tensor.dim %arg0, %c0 : tensor<?xbf16>
%2 = arith.cmpi ugt, %dim, %c1 : index
cf.assert %2, "Computer says yes 2"
%4 = arith.select %2, %arg0, %arg1 : tensor<?xbf16>
scf.if %2 {
} else {
  cf.assert %false, "NO BROADCAST SUPPORTED"
}
```


david-vc (approved):
Excellent work, James. Thank you.

PR #674, opened 2023-10-06
Allow packet routing to coexist with circuit routing in the same design

erwei-xilinx:


fifield:
Does this eliminate conflicts, or just make them less likely? Is it safe to run `create-pathfinder-flows` followed by `create-packet-flows` (like in aiecc)?

erwei-xilinx:
> Does this eliminate conflicts, or just make them less likely? Is it safe to run `create-pathfinder-flows` followed by `create-packet-flows` (like in aiecc)?

It just makes conflict less likely. When circuit + packet flows together use up routing resources, then the packet router doesn't know to route around the congestion. Apart from that, the aiecc routing passes can work as it is. I'm using it with IPU trace routing.

erwei-xilinx:
Merged in another PR.

PR #677, opened 2023-10-10
Convert tosa.sigmoid computation pattern to a function call getSigmoidBf16() for v16bfloat16 and v32bfloat16 types

linay-xsj:


david-vc (approved):
This looks good, but it's unfortunate that you need both isInSigmoidOperation and hasSigmoidCompmutationChain; both are lengthy and there is a lot of overlap between them.

Issue #501, opened 2023-05-23
aiecc hard codes libxaie paths


fifield:
aiecc main.py assumes that libxaie is installed as part of mlir-aie install which is wrong if the user gave the path to a preinstalled libxaie at build time (e.g. with `LibXAIE_x86_64_DIR` flag).

https://github.com/Xilinx/mlir-aie/blob/main/tools/aiecc/aiecc/main.py#L261-L263

Related to https://github.com/Xilinx/mlir-air/pull/235

stephenneuendorffer:
Fixed by #678

PR #678, opened 2023-10-11
Update runtime_lib/xaiengine cmake to always copy libxaie.

fifield:
The previous behavior was to copy libxaie into the runtime_lib install area unless the user provided cmake with the libxaie location. This commit changes cmake to always copy libxaie into the install, whether or not it was user provided.

Some pseudo code to try to explain the changes:
```python
# old way
for target in targets: # target is x86_64, aarch64
  if user provided a libxaie for target:
    # do nothing
  else if vitis was found:
    as a cmake subproject:
      if cmake_crosscompiling:
        copy libxaie source and headers from vitis
        cross compile libxaie
        install libxaie under runtime_lib/<target>
      else:
        install libxaie from vitis under runtime_lib/<target>

# new way
for target in targets: # target is x86_64, aarch64
  if user or vitis libxaie exists:
    as a cmake subproject:
      if cmake_crosscompiling:
        copy libxaie source and headers from vitis
        cross compile libxaie
        install libxaie under runtime_lib/<target>
      else:
        install user or vitis libxaie under runtime_lib/<target>
```

Resolves https://github.com/Xilinx/mlir-aie/issues/501 because we always copy libxaie to the hard coded location.

### DIFF
--- a/runtime_lib/CMakeLists.txt
+++ b/runtime_lib/CMakeLists.txt
@@ -40,19 +40,27 @@ foreach(target ${AIE_RUNTIME_TARGETS})
     message(FATAL_ERROR "Toolchain file ${${target}_TOOLCHAIN_FILE} not found! Cannot build target ${target}.")
   endif()
   message("Building AIE runtime for ${target} with TOOLCHAIN=${${target}_TOOLCHAIN_FILE}")
-  
+
   # xaiengine
-  # if no LibXAIE_{target}_DIR specified, build xaiengine_${target} from Vitis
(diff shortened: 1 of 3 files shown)


keryell:
It is sad that such a good PR description is not added as a design comment somewhere and is lost for ever.

fifield:
> It is sad that such a good PR description is not added as a design comment somewhere and is lost for ever.

Point taken, although I did just find it via the PR pointer left by the merge commit, so not totally lost.

PR #682, opened 2023-10-17
Convert tosa.ceil and tosa.floor to function calls getCeilBf16() and getFloorBf16() for bfloat16

linay-xsj:


aie_runtime_lib/AIE2/vec_math.h
```diff
@@ -229,4 +229,21 @@ getSigmoidBf16(v32bfloat16 in) {
   aie::vector<bfloat16, 32> out = msc0_acc.to_vector<bfloat16>();
   return (v32bfloat16)out;
 }
+
+inline __attribute__((always_inline)) v16bfloat16 getCeilBf16(v16bfloat16 in) {
+  aie::vector<int32, 16> in_int32 = bfloat16_to_int(
```
david-vc, line 234: This function will compute incorrect results if abs(in) > 0x1.p31
The range of bfloat16 is -1.fep127 .. +1.fep127.
For input values that are < -2^31 or > 2^31-1 we should return the input as-is.

david-vc (changes requested):
The getFloorBf16 and getCeilBf16 need to be revised to handle cases where abs(in) > 2^31.

david-vc (approved):

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2606,6 +2721,48 @@ static void configureAIEVecCommonLegalizations(ConversionTarget &target,
...
+
+    Type scalarType = srcType.getElementType();
+
+    if (!isa<FloatType>(scalarType)) {
+      return true;
+    }
+
+    unsigned laneSize = getVectorLaneSize(srcType);
+    unsigned elWidth = scalarType.getIntOrFloatBitWidth();
+    if (elWidth != 16 || (laneSize != 16 && laneSize != 32)) {
```
jamestcl-amd, line 2759: neat: we can just do `return (elWidth != 16 || (laneSize != 16 && laneSize != 32))`.

test/Integration/Dialect/TOSA/bf16_floor_v32/bf16_floor.mlir
```diff
@@ -0,0 +1,13 @@
+// REQUIRES: valid_xchess_license
+// RUN: aie-opt %s %tosa-to-linalg% | aie-opt %linalg-to-vector-v32% --convert-vector-to-aievec="aie-target=aieml" -lower-affine | aie-translate -aieml=true --aievec-to-cpp -o dut.cc
+// RUN: xchesscc_wrapper aie2 -f -g +s +w work +o work -I%S -I%aie_runtime_lib%/AIE2 -I %aietools/include -D__AIEARCH__=20 -D__AIENGINE__ -I. %S/testbench.cc dut.cc
+// RUN: mkdir -p data
+// RUN: xca_udm_dbg --aiearch aie-ml -qf -T -P %aietools/data/aie_ml/lib/ -t "%S/../profiling.tcl ./work/a.out" >& xca_udm_dbg.stdout
+// RUN: FileCheck --input-file=./xca_udm_dbg.stdout %s
+// CHECK: TEST PASSED
+// Cycle count: 919
```
jamestcl-amd, line 8: Looks like the v32 doesn't save us much of cycles?

linay-xsj: Yes, because bfloat16_to_int only support v16bfloat16, we need to extract lower and higher part of v32bfloat16 and convert them to int  separately, at last do concatenation. Only comparison and select ops can save some cycle counts.

jamestcl-amd: I see, thanks for the explanation.

PR #688, opened 2023-10-19
[chess-clang] Add proper defines for Vitis kernel code

stephenneuendorffer:
The vitis headers use some defines to switch between synthesizable and functional models.  When feeding code to chess, these always need to be defined.


david-vc (commented):
LGTM.

PR #692, opened 2023-10-23
ObjFIFO Multi-Dimensional Patch

jgmelber:
* relax some constraints
* provide additional tests


test/objectFifo-stateful-transform/nd_dma_distribute_broadcast_AIE2_bad.mlir
```diff
@@ -0,0 +1,39 @@
+//===- nd_dma_distribute_AIE2.mlir -----------------------------*- MLIR -*-===//
```
AndraBisca, line 1: Same as above.

test/objectFifo-stateful-transform/nd_dma_distribute_AIE2_bad.mlir
```diff
@@ -0,0 +1,38 @@
+//===- nd_dma_distribute_AIE2.mlir -----------------------------*- MLIR -*-===//
```
AndraBisca, line 1: The name of the file doesn't match the name written here.

test/dialect/AIE/nd-dma-wrong-rank.mlir
```diff
@@ -9,6 +9,7 @@
 //===----------------------------------------------------------------------===//
 
 // RUN: aie-opt --verify-diagnostics %s
+// XFAIL: *
```
AndraBisca, line 12: Should this test and the next one fail or catch an intended error?

AndraBisca (approved):

test/dialect/AIE/nd-dma-wrong-rank.mlir
jgmelber: They pass because the error disallowing ND DMA transformations for any memref type other than 1xi32 is disabled

fifield (approved):

PR #693, opened 2023-10-23
Convert tosa.negate to aievec.neg for v16bfloat16 and v16float types and add the test cases for i32, i16 and i8 types

linay-xsj:


test/Integration/Dialect/TOSA/i8_neg/dut.cc
```diff
@@ -0,0 +1,13 @@
+void dut(int8_t *restrict v1, int8_t *restrict v2) {
+  v64int8 v3 = broadcast_zero_s8();
+  size_t v4 = 0;
+  size_t v5 = 1024;
+  size_t v6 = 64;
+  for (size_t v7 = v4; v7 < v5; v7 += v6)
+    chess_prepare_for_pipelining chess_loop_range(16, 16) {
+      v64int8 v8 = *(v64int8 *)(v1 + v7);
+      v64int8 v9 = sub(v3, v8);
```
jamestcl-amd, line 9: From my understanding, the lowering pipeline lowers the integer case of `tosa.negate %arg0` to `zeros - %arg0`. And, we can support those operations with the existing conversion patterns right?

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2401,6 +2401,55 @@ struct ComputeFloorOpPattern : public OpConversionPattern<math::FloorOp> {
...
+    }
+
+    Type scalarType = srcType.getElementType();
+
+    if (!isa<FloatType>(scalarType)) {
+      return failure();
+    }
+    unsigned laneSize = getVectorLaneSize(srcType);
+
+    if (laneSize != 16 && laneSize != 32) {
```
jamestcl-amd, line 2424: It looks like both `bf16_neg` and `float_neg` has the vector lanes=16. Do we like support lane size = 32 here?

test/Integration/Dialect/TOSA/i8_neg/dut.cc
linay-xsj: Yes, the cycle count is the same as the case using neg intrinsic.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
linay-xsj: neg instrinsc only supports v16accfloat type, so we only support v16float and v16bfloat16 types.

david-vc: Hi @linay-xsj : So we allow lanesize=32, but we do not have a test the this case. If lanesize=32 is supported, then please also add a test case that exercises this.

david-vc (changes requested):

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
linay-xsj: It seems I made a mistake here.  It should be if (laneSize != 16), because neg instrinsc only supports v16accfloat type, so we only support v16float and v16bfloat16 types.

david-vc: OK. Please upload fix and I'll approve.

david-vc (approved):
Looks good to me.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
linay-xsj: > OK. Please upload fix and I'll approve.

Done.

PR #707, opened 2023-10-31
Revert "remove extraneous transform tests"

jsetoain:
This reverts commit 6dab630c74084f49d13322f7dd81829cf68466df.

There's nothing "extraneous" about these tests.

### DIFF
--- a/lib/Dialect/AIEVec/TransformOps/AIEVecTransformOps.cpp
+++ b/lib/Dialect/AIEVec/TransformOps/AIEVecTransformOps.cpp
@@ -50,13 +50,9 @@ static Value vectorizeTensor(OpBuilder &rewriter, Location loc, Value tensor) {
   auto opMemrefVecTy = MemRefType::get(newShape, opVecElemTy);
   auto typeCastOp =
       rewriter.create<vector::TypeCastOp>(loc, opMemrefVecTy, toMemRefOp);
-  // Missing a 'restrict' qualifier
   auto toTensorOp = rewriter.create<bufferization::ToTensorOp>(
       loc, RankedTensorType::get(newShape, opVecElemTy), typeCastOp);
-  // TODO: This won't work until we next update llvm, with the newly added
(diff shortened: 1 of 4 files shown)


makslevental:
@jsetoain sorry maybe extraneous isn't the right word but I saw that they don't test `aie` functionality (only upstream). In addition, as you can see, they now error.

jsetoain:
> @jsetoain sorry maybe extraneous isn't the right word but I saw that they don't test `aie` functionality (only upstream). In addition, as you can see, they now error. I guess it's fine if they stay but I wasn't sure the benefit against the (current) cost.

They test the `transform.structured.vectorize_contraction` transformation, which was added together with the tests themselves (f388ebe06f5d9d7262eae190c82df2a814ebb2bc). It's not upstream functionality, it's ours.

lib/Dialect/AIEVec/TransformOps/AIEVecTransformOps.cpp
```diff
@@ -239,7 +231,7 @@ DiagnosedSilenceableFailure transform::VectorizeContractionOp::applyToOne(
   for (unsigned i = 0; i < numOuterMostDims; i++)
     remOuterDims.push_back(getAffineDimExpr(i, ctx));
   unsigned numResults = indexingMaps[0].getNumResults();
-  llvm::ArrayRef<int64_t> positions = {numResults - 2, numResults - 1};
+  SmallVector<int64_t> positions = {numResults - 2, numResults - 1};
```
makslevental, line 234: Sorry this bit you :(

makslevental (approved):

lib/Dialect/AIEVec/TransformOps/AIEVecTransformOps.cpp
jsetoain: It's all good, you solved it yourself so... 😄

PR #709, opened 2023-11-01
Convert tosa.bitwise_not to aievec.bneg to compute bitwise not for integer types

linay-xsj:


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2396,9 +2396,30 @@ struct ComputeNegOpPattern : public OpConversionPattern<arith::NegFOp> {
...
+  auto cstDense = dyn_cast<DenseIntElementsAttr>(constOp.getValue());
+  if (!cstDense) {
+    return false;
+  }
+
+  if (elWidth == 32) {
+    return cstDense.getSplatValue<int32_t>() == -1;
+  } else if (elWidth == 16) {
+    return cstDense.getSplatValue<int16_t>() == -1;
+  } else {
```
david-vc, line 2414: In case that elWidth == 64 the function would return the incorrect result if any of the top 56 bits is 0.

david-vc (approved):
Please fix the 1 minor issue that I pointed out before merging. Thank you.

PR #706, opened 2023-10-31
[aie] Remove boost dependency by implementing Dijkstra using ADT

makslevental:
This PR excises the dependency on boost by using `llvm/ADT/DirectedGraph.h` types and implementing Dijkstra by hand. 

Other general cleanup is done in-place.

Currently, the only remaining issue is this implementation isn't deterministic in the ordering of the switchboxes and flows created; the routings it finds are still correct, they're just out of order  (hence the sprinkling of `CHECK-DAG`). Obviously it's preferrable that everything be deterministic so I'll keep hunting.

### TODO 

1. ~~Ablate the `Switchbox` and `Channel` constructor panoply~~
2. ~~use `shared_ptr` instead of `new Channel`~~
3. ~~fix `CHECK-DAG`~~
4. ~~add error throwing mechanism for max iters reached~~
5. ~~turn `SwitchSetting`, `PathEndPoint`, and `Flow` into structs.~~ (next refactor...)
6. ~~handle no matching channel in `addFixedConnection` correctly instead of silently.~~
7. ~~rename `AIEPathfinder.h` and `AIEPathfinder.cpp`.~~

PR #711, opened 2023-11-02
remove clang dep

makslevental:
This PR removes the dependency on clang-12 for tests and examples. I say tests and examples because there was never a real dependency (I've never built with clang-12...) because as far as I understand the dependency was only on the artifacts passed to chess rather than any compilation aie artifacts themselves.

Note this PR also drops the `DCMAKE_COMPILE_WARNING_AS_ERROR` because gcc will fail to compile MLIR with it on (lol).

**EDIT**:

Just want to be clear: this works across all platforms with all compilers currently and I don't know why this was needed (before my time). The wheels I build for us are built using gcc on Linux, clang on Mac, and MSCV on Windows:

1. Linux: [The CXX compiler identification is GNU 12.2.1](https://github.com/Xilinx/mlir-aie/actions/runs/6670407503/job/18130199379#step:7:15368);
2. Mac: [The CXX compiler identification is AppleClang 13.0.0.13000029](https://github.com/Xilinx/mlir-aie/actions/runs/6670407503/job/18130199804#step:7:269);
3. Windows: [The CXX compiler identification is MSVC 19.29.30152.0](https://github.com/Xilinx/mlir-aie/actions/runs/6670407503/job/18130200230#step:7:454).

Regarding the `utils/` scripts: there seems to be nothing to be gained from suggesting/implying/guaranteeing clang compatibility[^1] and not gcc. Yes, there are some [clang specific](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIECreatePathfindFlows.cpp#L35-L36) `#pragma`s (and [gcc too](https://github.com/Xilinx/mlir-aie/blob/57b4fbfd17c2acea42f190f1be680c565aef49f4/include/aie/Dialect/AIE/Transforms/AIEPathfinder.h#L17-L21)) but

1. They don't present any issues as indicated by passing tests here;
2. They are all (including the gcc pragmas) related to boost which [we will soon be rid of entirely](https://github.com/Xilinx/mlir-aie/pull/706).

cc @nirvedhmeshram

[^1]: And plenty to be lost (by signaling there's this added friction, of having clang, to using the tool).


cmake/toolchainFiles/toolchain_x86_64.cmake
```diff
@@ -2,17 +2,5 @@
 # Copyright (C) 2022, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 
-# specify the compiler
-set(CLANG_VER 12)
```
keryell, line 6: Great ! This one was a pain to me about looking for such a compiler in an antique boutique.

jsetoain (approved):

jsetoain (changes requested):
lgtm

jsetoain (approved):

CMakeLists.txt
```diff
@@ -134,6 +134,15 @@ include_directories(${PROJECT_BINARY_DIR}/include)
...
 add_definitions(${LLVM_DEFINITIONS})
 
+# Silence a false positive GCC -Wunused-but-set-parameter warning in constexpr
+# cases, by marking SelectedCase as used. See
+# https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85827 for details. The issue is
+# fixed in GCC 10.
+if(CMAKE_CXX_COMPILER_ID MATCHES "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "10.0")
+  check_cxx_compiler_flag("-Wno-unused-but-set-parameter" CXX_SUPPORTS_WNO_UNUSED_BUT_SET_PARAMETER)
+  append_if(CXX_SUPPORTS_WNO_UNUSED_BUT_SET_PARAMETER "-Wno-unused-but-set-parameter" CMAKE_CXX_FLAGS)
+endif()
```
makslevental, line 144: See https://github.com/llvm/llvm-project/pull/68528.

fifield (approved):

keryell (approved):
Thanks!

PR #713, opened 2023-11-02
Convert tosa.bitwise_and and tosa.bitwise_or to aievec.band and aievec.bor to compute bitwise and and bitwise or for integer types

linay-xsj:


include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -775,4 +775,37 @@ def AIEVec_BnegOp:
   let assemblyFormat = "$source attr-dict `:` type($result)";
   let hasVerifier = 0;
 }
+
+def AIEVec_BorOp:
+  AIEVec_Op<"bor", [
+    Pure,
+    AllTypesMatch<["lhs", "rhs", "result"]>
+  ]>,
+  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs)>,
```
jamestcl-amd, line 784: I'm not sure if this has been discussed before. Do we like to restrict vector element type from `AnyVector` to `VectorOf<[I8, I16, I32]>`? Will it be too restricted or over-design?

linay-xsj: These aievec ops and the corresponding intrinsics not only support integer types but also support float types. Tosa dialect only supports integer types. Thus, we set it AnyVector and check the legality in VectorToAIEVecConversions.cpp.

jamestcl-amd: I see, thanks for the explanation. We can do `VectorOf<[BF16, F32, I8, I16, I32]>` to align with the underlying intrinsics, if that's the case. Anyway, I was just thinking about what's the disadvantage of restricting the datatypes. For example, if we later want to support new datatype, this restriction will need an update.

david-vc (approved):
Looks good to me.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
jsetoain: > I see, thanks for the explanation. We can do `VectorOf<[BF16, F32, I8, I16, I32]>` to align with the underlying intrinsics, if that's the case. Anyway, I was just thinking about what's the disadvantage of restricting the datatypes. For example, if we later want to support new datatype, this restriction will need an update.

That's not a disadvantage, that's how verification works :-) On the other hand, add proper constraints not only prevents invalid code, it becomes documentation.

Right now, you're right, these instructions should all be properly constrained to the supported data types, like everywhere else in LLVM and MLIR. This is something we've been doing wrong for the start.

jsetoain (changes requested):

PR #715, opened 2023-11-03
Add IPU to TargetModel

fifield:
Add IPU device to TargetModel.

Maybe there is a better device name than `ipu` to not collide with future IPU variant?


jgmelber (approved):
Agreed, we should consider alternate naming, however, since it is a somewhat "virtual" target it may be appropriate

PR #716, opened 2023-11-03
AIE Python Bindings

AndraBisca:
This PR introduces an extension of the auto-generated python bindings we currently have in the repo. In particular, the extended operations have specialised constructors that can take in python integers as inputs. For now, only the most widely used operations in the AIE dialect have been extended. The others will be added in future PRs.

This PR relies on #715


fifield (approved):

python/dialects/_AIE_syntax.py
```diff
@@ -0,0 +1,257 @@
...
+    with Context() as ctx, Location.unknown():
+        register_dialect(ctx)
+        module = Module.create()
+        with InsertionPoint(module.body):
+            f()
+        print(module)
+
+
+# Create an aie device on specified target architecture.
+class Device(DeviceOp):
```
makslevental, line 80: This is going to break as soon as we bump due to this PR I landed upstream https://github.com/llvm/llvm-project/pull/68853.

python/dialects/_AIE_util.py
```diff
@@ -0,0 +1,169 @@
...
+
+from ..ir import *
+from ..dialects import arith
+from ..dialects import memref
+from ..dialects.func import *
+from ..dialects.scf import *
+
+
+# Create a signless arith constant of given width (default is i32).
+class integerConstant(arith.ConstantOp):
```
makslevental, line 12: Instead of having many different such specializations you're better off doing something like this https://github.com/makslevental/mlir-python-utils/blob/main/mlir/utils/dialects/ext/arith.py#L57.

makslevental (commented):
In general I think this isn't the right thing to do because

1. There are now value builders being generated automatically https://github.com/llvm/llvm-project/pull/68308 (if you check in `_AIE_ops_gen.py` you will see them just under the conventional builders);
2. Shortly they will support infix operators https://github.com/llvm/llvm-project/pull/69644;
3. My next upstream PR will be `region_op`.

fifield:
> In general I think this isn't the right thing to do because
> 
> 1. There are now value builders being generated automatically [[mlir][python] generate value builders llvm/llvm-project#68308](https://github.com/llvm/llvm-project/pull/68308) (if you check in `_AIE_ops_gen.py` you will see them just under the conventional builders);
> 2. Shortly they will support infix operators [[mlir][python] value casting llvm/llvm-project#69644](https://github.com/llvm/llvm-project/pull/69644);
> 3. My next upstream PR will be `region_op`.

We have a set of examples and tests using these bindings and we are maintaining them downstream. This is one in a set of PRs to get all that landed. Are these all things we can apply incrementally to what we have?

makslevental:
> Are these all things we can apply incrementally to what we have?

You can just choose not to use it but it's all exactly the same functionality (and more).

makslevental:
@fifield where are these examples? Can you link so I can see how they're written?

jgmelber:
@makslevental this is representative of the style in the examples: test/python/codeRegion.py

jgmelber (approved):

makslevental:
> @makslevental this is representative of the style in the examples: [test/python/codeRegion.py](https://github.com/Xilinx/mlir-aie/blob/python-bindings/test/python/codeRegion.py)

You can already rewrite this to use just regular python for loops https://github.com/llvm/llvm-project/blob/main/mlir/test/python/dialects/scf.py#L64.

python/dialects/_AIE_syntax.py
AndraBisca: Hey Maks, thank you for your detailed review! I looked through the PRs you mentioned above. Could you please help me understand why this would break exactly? In the PR above it looks like extensions that were calling super.__init__() can stay unchanged.

python/dialects/_AIE_util.py
AndraBisca: I agree. The reason why I didn't make this more complete is because I am looking forward to using your proposed method when it is available.

AndraBisca:
I understand that what I have hand-implemented in this PR will become auto-generated in the near future. I look forward to that and will be ready to modify all of these examples accordingly. However, because of internal deadlines, our priority right now is to support all the downstream work that is currently using these simple sugar bindings. We should discuss internally how this aligns with the release of your automated python bindings.

makslevental:
> I understand that what I have hand-implemented in this PR will become auto-generated in the near future. I look forward to that and will be ready to modify all of these examples accordingly. However, because of internal deadlines, our priority right now is to support all the downstream work that is currently using these simple sugar bindings. We should discuss internally how this aligns with the release of your automated python bindings, @makslevental

Just want to underscore somethings: a lot of this _is_ already autogenerated.

makslevental:
> I understand that what I have hand-implemented in this PR will become auto-generated in the near future. I look forward to that and will be ready to modify all of these examples accordingly. However, because of internal deadlines, our priority right now is to support all the downstream work that is currently using these simple sugar bindings. We should discuss internally how this aligns with the release of your automated python bindings, @makslevental

Whoops finget slipped...

Anyway of course that's reasonable. I was just trying to emphasize that some of those PRs are only a couple of days away from being merged. Anyway I'll just port this code here when they're merged.

makslevental (approved):

AndraBisca:
> > I understand that what I have hand-implemented in this PR will become auto-generated in the near future. I look forward to that and will be ready to modify all of these examples accordingly. However, because of internal deadlines, our priority right now is to support all the downstream work that is currently using these simple sugar bindings. We should discuss internally how this aligns with the release of your automated python bindings, @makslevental
> 
> Whoops finget slipped...
> 
> Anyway of course that's reasonable. I was just trying to emphasize that some of those PRs are only a couple of days away from being merged. Anyway I'll just port this code here when they're merged.

Sounds good! I will still inquire more about the deadlines, because we might be able to wait(?). I'll let you know!

python/dialects/_AIE_syntax.py
makslevental: Sorry I didn't realize that `AIEDevice` is already an `EnumAttr`. So you can get this for free by signing up [using `GEN_ENUM_BINDINGS`](https://github.com/llvm/llvm-project/blob/main/mlir/python/CMakeLists.txt#L65).

stephenneuendorffer:
> Sounds good! I will still inquire more about the deadlines, because we might be able to wait(?). I'll let you know!

I think we really need to move to using the bindings that are upstream.  There's no real value in diverging from that, as far as I can tell.  I think in the short term, we need to make sure we can get up to HEAD and then propose a time frame to migrate the python bindings.  @AndraBisca can you figure out a time frame for that?

makslevental:
> we need to make sure we can get up to HEAD 

my contribution: https://github.com/Xilinx/mlir-aie/pull/724

AndraBisca:
> > Sounds good! I will still inquire more about the deadlines, because we might be able to wait(?). I'll let you know!
> 
> I think we really need to move to using the bindings that are upstream. There's no real value in diverging from that, as far as I can tell. I think in the short term, we need to make sure we can get up to HEAD and then propose a time frame to migrate the python bindings. @AndraBisca can you figure out a time frame for that?

Yes, I can do that!

jgmelber (approved):
LGTM

PR #723, opened 2023-11-07
Make access to value of signless attributes signed by default

makslevental:
This PR implements our own version of `TypedSignlessIntegerAttrBase` such that for signless constraints like `I16Attr`, `I32Atrr`, `I64Attr,`, etc., generated getters return `int`s instead of `uint`s. Why? Because all other uses/builders of `IntegerAttr` take/return `int`s as well. See this discussion on the ["pairs to structs" PR](https://github.com/Xilinx/mlir-aie/pull/719#discussion_r1382551200) and the follow-up on [discourse](https://discourse.llvm.org/t/signless-semantics-integerattr-vs-signlessintegerattr) about the mismatch between `IntegerAttr` and `TypedSignlessIntegerAttrBase` constraints. 

In order to accomplish this goal `AIE.td` is factored into separate files for types, attributes, and ops (which is the right way to do things anyway).

This PR also updates some stray uses of `SI32Attr`, which is technically correct given our intent to use `int`s everywhere, but doesn't conform to our use of `I32Attr` everywhere else. Updated tests are all for handling this change.

With this change, there remain almost no `(int)` casts and almost no "narrowing conversions".

Only remaining question is what to do about 

https://github.com/Xilinx/mlir-aie/blob/eae36affdcc4d5b191833ab270f098059994aaae/include/aie/Dialect/AIE/IR/AIETargetModel.h#L222-L225


lib/Targets/AIETargets.cpp
```diff
@@ -164,16 +156,16 @@ void registerAIETranslations() {
...
-          if (auto tile = target_model.getMemEast(srcCoord))
-            doBuffer(tile, target_model.getMemEastBaseAddress());
+          if (auto tile = targetModel.getMemSouth(srcCoord))
+            doBuffer(tile, targetModel.getMemSouthBaseAddress());
+          if (auto tile = targetModel.getMemWest(srcCoord))
+            doBuffer(tile, targetModel.getMemWestBaseAddress());
+          if (auto tile = targetModel.getMemNorth(srcCoord))
+            doBuffer(tile, targetModel.getMemNorthBaseAddress());
+          if (auto tile = targetModel.getMemEast(srcCoord))
+            doBuffer(tile, targetModel.getMemEastBaseAddress());
```
jsetoain, line 168: These style fixes are not really part of the path, and they would be better on their own, in its own fix, with its own description.

makslevental: It's true but there's so many that it's basically impossible to do that.

include/aie/Dialect/AIE/Transforms/AIENormalizeAddressSpaces.td
```diff
@@ -12,6 +12,7 @@
 #define AIE_NORMALIZE_ADDRESS_SPACES
 
 include "aie/Dialect/AIE/IR/AIE.td"
+
```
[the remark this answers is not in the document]
makslevental: Well top top includes are directory local right? And those are usually separated. And since conventional order is most specific -> least specific you need `"llvm/..."` separated out right (otherwise sorting would put them above `"mlir/...")? Well this is just the natural extension of that.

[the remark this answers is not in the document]
jsetoain: I get the ordering, I haven't read any guidelines about separating by subproject. I've seen people separating the project headers from the system headers (<iostream>/<utility>/<list>...), but that's not in the coding guidelines either. Anyway, it's up to you, I just wanted to point out that I don't think that's necessary, and when it's just 3 includes, it looks odd.

lib/Targets/AIETargets.cpp
jsetoain: Why? You can leave them as they are, and do another patch that says "[nfc] change target model variable name to camelcase". Not saying you must do that for this patch right now, but patches are more manageable when they are single issue, rather than "<issue>(+unspecified things here and there)". The "side things" are lost from the history, which may be source of pain further down the line, so it's something to avoid.

makslevental: > Not saying you must do that for this patch right now, but patches are more manageable when they are single issue, rather than "(+unspecified things here and there)". The "side things" are lost from the history, which may be source of pain further down the line, so it's something to avoid.

Of course I agree and I would rather not have to do this (and since you prefer I didn't, I won't anymore) but here's what I'm saying

1. These kinds of lint are absolutely everywhere;
2. They are hard to find (a `[a-z]_[a-z]` regex catches stl methods and such);
3. They're not actually a priority.

So short of just keeping a list of line numbers, I don't see how to aggregate these into one NFC. So I'm taking the approach [recommended here](https://github.com/Xilinx/mlir-air/pull/286#pullrequestreview-1594655052).

PR #725, opened 2023-11-07
delete `NetlistAnalysis`

makslevental:
Per this [recommendation](https://github.com/Xilinx/mlir-aie/pull/719#discussion_r1382674722).


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1218,6 +1218,17 @@ LogicalResult xilinx::AIE::BufferOp::verify() {
...
 
+// FIXME: make address assignment for buffers explicit and move this function to
+// an interface
+uint64_t xilinx::AIE::getBufferBaseAddress(Operation *bufOp) {
+  if (auto buf = dyn_cast<BufferOp>(bufOp)) {
+    return buf.address();
+  } else if (auto buf = dyn_cast<ExternalBufferOp>(bufOp)) {
+    assert(false && "External buffer addresses are assigned at runtime.");
+  }
+  llvm_unreachable("unknown buffer type");
```
jsetoain, line 1229: Asserts are not supposed to be used to handle error cases, but to bail on a broken/corrupted state. Here it looks like we're using it as a convenient way to "return an error". If this function can only be called on a `BufferOp`, you should assert on that. If calling on an `ExternalBufferOp` is a user error, then you should find a way to return an error (e.g.: optional), and handle it properly from the caller.

makslevental: Most place this is called are `void`. Only thing I can do here is `report_fatal_error`. I don't disagree that this is wrong but lack of correct error propagation in the codebase is too big for me to tackle by myself :shrug:.

fifield:
@makslevental A gentle reminder that not all tests run in the github CI because of lack of Vitis.  I'm seeing some regressions after this one. For example the container iterated over [here](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetXAIEV2.cpp#L795) is now never populated, since the API doing so was removed.

makslevental:
> @makslevental A gentle reminder that not all tests run in the github CI because of lack of Vitis. I'm seeing some regressions after this one. For example the container iterated over [here](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetXAIEV2.cpp#L795) is now never populated, since the API doing so was removed.

Ah that's unfortunate (we should get vitis in CI...) but it's an easy fix since. I'll push it shortly.

fifield:
> 
> Ah that's unfortunate (we should get vitis in CI...) but it's an easy fix since. I'll push it shortly.

What do you need in order to run the tests yourself?  I'm also having trouble with aiecc.py....

```
  File "/scratch2/jefff/ipu/build-mlir-aie/python/aie/compiler/aiecc/main.py", line 275, in minmax
    mn = min(mn, o.col.value)
AttributeError: 'aie._mlir_libs._mlir.ir.OpView' object has no attribute 'col'
```

makslevental:
> What do you need in order to run the tests yourself?

I can, there's just a lot of friction because I do most of my dev on a macbook. I have a [docker with Vitis](https://github.com/makslevental/mlir-aie/pkgs/container/mlir-aie%2Fxilinx) that I setup for myself (and potentially others) when the premature opaque-pointers bump broke things but I just haven't had the time to add it to your CI.

> I'm also having trouble with aiecc.py....

This isn't in anywhere to be found in current `main.py` so I'm not sure exactly what you're attempting but you have there an abstract `OpView` rather than a `TileOp`.

PR #726, opened 2023-11-07
Cleanup TOSA regressions

jamestcl-amd:
- Move TOSA tests to correct directory and update the lowering pipeline strings. The list of tests are from https://github.com/Xilinx/mlir-aie/commit/47cd2e65651693bc1e0ea45cc25992514543ec0e#diff-f1aa19177fcb53b7fdb68b4f0738ea9f8f27881140fe93a3b4a8dcd5af222c17
- Fix all the `tosa.mul` tests by changing `shift = 0 : i32` to `shift = 0 : i8`.
- Remove XFAIL marks for all the passing tests that works with the latest LLVM/MLIR ver.


david-vc (approved):
Thanks for fixing this.

PR #731, opened 2023-11-08
[nfc] Sanitize headers (1/n)

jsetoain:
This patch removes the last few uses of "using namespace" in headers, removes unnecessary includes, and replaces some of them with forward declarations.

### DIFF
--- a/lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
+++ b/lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // (c) Copyright 2022 Xilinx Inc.
+// (c) Copyright 2023 Advanced Micro Devices, Inc.
 //
 //===----------------------------------------------------------------------===//
 // This file defines helpers to emit C++ code for AIE vector dialect.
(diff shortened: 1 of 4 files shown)


makslevental:
Thanks! Btw are you familiar with https://github.com/include-what-you-use/include-what-you-use ?

jsetoain:
> Thanks! Btw are you familiar with https://github.com/include-what-you-use/include-what-you-use ?

Nop! It looks intriguing, and it may have saved me some time 😂 Thanks!

keryell:
Are you familiar to LSP server `clangd`? There is for example an `=iwyu` option and some other header-related options.
I use it with Emacs `lsp-mode` but I guess it can work with other editor/IDE.

makslevental:
> Are you familiar to LSP server `clangd`? There is for example an `=iwyu` option and some other header-related options. I use it with Emacs `lsp-mode` but I guess it can work with other editor/IDE.

Jetbrain's CLion (which embeds clangd) does it too. Anyway I would like to set it up in CI (something like [this](https://github.com/marketplace/actions/include-what-you-use-github-action)) but haven't gotten around.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -27,6 +28,7 @@
 #include "mlir/Support/IndentedOstream.h"
 #include "mlir/Support/MathExtras.h"
 #include "llvm/ADT/DenseMap.h"
+#include "llvm/ADT/ScopedHashTable.h"
```
makslevental, line 31: Did this one get added by accident or really it was missing?

makslevental (approved):
LGTM (module that one new include that might be stray?)

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
jsetoain: It's used by the code (LN 178, 179, 197, 198, 209, 212). It was being included by one of the headers that did not make use of it.

PR #732, opened 2023-11-08
Add vector lanes and element type constraints for bitwise operators

linay-xsj:


jamestcl-amd (approved):
Thanks for making the changes!

PR #735, opened 2023-11-08
check for lld in shell scripts

makslevental:


.github/workflows/buildAndTest.yml
```diff
@@ -175,7 +175,7 @@ jobs:
             -DCMAKE_MODULE_PATH=`pwd`/../cmake/modulesXilinx \
             -DMLIR_DIR=../llvm/install/lib/cmake/mlir \
             -DLLVM_DIR=../llvm/install/lib/cmake/llvm \
-            -DCMAKE_LINKER=lld \
```
makslevental, line 178: This is not the right thing https://stackoverflow.com/a/25274328

.github/workflows/buildAndTest.yml
```diff
@@ -175,7 +175,7 @@ jobs:
...
+            -DLLVM_USE_LINKER=lld \
```
makslevental, line 178: [this flag sets](https://github.com/llvm/llvm-project/blob/4aae5387a874a55ee491f5dc23ce0506c5cdc678/llvm/cmake/modules/HandleLLVMOptions.cmake#L361).

```
-DCMAKE_EXE_LINKER_FLAGS_INIT="-fuse-ld=lld"
-DCMAKE_MODULE_LINKER_FLAGS_INIT="-fuse-ld=lld"
-DCMAKE_SHARED_LINKER_FLAGS_INIT="-fuse-ld=lld"
```

utils/build-llvm-cross-local.sh
```diff
@@ -57,18 +58,16 @@ cmake ../llvm \
   -DLLVM_BUILD_EXAMPLES=OFF \
   -DLLVM_BUILD_UTILS=ON \
   -DLLVM_INSTALL_UTILS=ON \
-  -DLLVM_USE_LINKER=lld \
   -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
   -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR \
-  -DLLVM_ENABLE_PROJECTS="clang;lld;mlir" \
```
makslevental, line 63: Don't need these.

utils/build-llvm-cross-local.sh
```diff
@@ -78,19 +77,19 @@ cmake ../llvm \
   -DMLIR_INCLUDE_INTEGRATION_TESTS=OFF \
   -DLINKER_SUPPORTS_COLOR_DIAGNOSTICS=OFF \
   -DLLVM_ENABLE_TERMINFO=OFF \
-  -DLLVM_DEFAULT_TARGET_TRIPLE="aarch64-linux-gnu" \
-  -DLLVM_TARGET_ARCH="AArch64" \
-  -DCMAKE_CXX_STANDARD=17 \
```
makslevental, line 83: This is already specified in [`llvm/CMakeLists.txt`](https://github.com/llvm/llvm-project/blob/3fe69bab1149e20f87ac31f8b2557baa88fcdd14/llvm/CMakeLists.txt#L70).

keryell (approved):
Thanks or the cleaning.

PR #736, opened 2023-11-08
Add dma-to-ipu pass

fifield:
This pass converts ipu.dma_memcpy_nd operations to ipu.writebd_shimtile and ipu.write32 operations.

depends on https://github.com/Xilinx/mlir-aie/pull/734

PR #737, opened 2023-11-08
Add aie-ipu-instgen target to aie-translate

fifield:
This adds translation from aie ipu operations to their binary representation.

depends on https://github.com/Xilinx/mlir-aie/pull/736

### DIFF
--- a/lib/Targets/AIETargetIPU.cpp
+++ b/lib/Targets/AIETargetIPU.cpp
@@ -1,0 +1,148 @@
+//===- AIETargetIPU.cpp -----------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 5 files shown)


jgmelber (approved):

lib/Targets/AIETargetIPU.cpp
```diff
@@ -0,0 +1,148 @@
...
+  os << "314E5A5F\n";
+  os << "635F5F31\n";
+  os << "676E696C\n";
+  os << "39354E5F\n";
+  os << "6E693131\n";
+  os << "5F727473\n";
+  os << "64726F77\n";
+  os << "00004573\n";
+  os << "07BD9630\n";
+  os << "000055FF\n";
```
keryell, line 43: We should have a random straw poll to ask whether this is so obvious it does not deserve a single line of comment...

PR #738, opened 2023-11-08
Add xclbin generation to aiecc

fifield:
This adds a cdo target to aie-translate and an xclbin target to aiecc.py.

depends on https://github.com/Xilinx/mlir-aie/pull/737


python/compiler/aiecc/main.py
```diff
@@ -256,6 +260,213 @@ async def process_core(self, core):
...
+              "type" : "char *",
+              "offset" : str(hex(offset))
+            }
+      arguments.append(arg)
+      offset += 0x8
+    design = {"ps-kernels" : {
+                "kernels" : [{"name" : kernel_name,
+                              "type" : "dpu",
+                              "extended-data" : {
+                                "functional" : "1",
```
jgmelber, line 343: An additional field is needed here inside extended-data, before functional:
```
"subtype" : "DPU",
```

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -13,6 +13,23 @@
...
+    AIEX Ipu Ops are removed.
+
+  }];
+
+  let constructor = "xilinx::AIEX::createAIEXToStandardPass()";
+  let dependentDialects = [
+    "func::FuncDialect",
+    "memref::MemRefDialect",
+    "xilinx::AIE::AIEDialect",
+    "xilinx::AIEX::AIEXDialect",
```
makslevental, line 29: After the `using namespace` cleanup, you need the `mlir::` namespace here.

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -13,6 +13,23 @@
 
 include "mlir/Pass/PassBase.td"
 
+def AIEXToStandard : Pass<"aiex-standard-lowering", "ModuleOp"> {
```
makslevental, line 16: Tests are failing because after the using namespace cleanup, you need the mlir:: namespace here before the `ModuleOp`.

makslevental, line 16: Tests are failing because after the `using namespace` cleanup, you need the mlir:: namespace here before the `ModuleOp`, like this:

```suggestion
def AIEXToStandard : Pass<"aiex-standard-lowering", "mlir::ModuleOp"> {
```

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
makslevental, line 29: Here too:

```suggestion
    "mlir::func::FuncDialect",
    "mlir::memref::MemRefDialect",
    "xilinx::AIE::AIEDialect",
    "xilinx::AIEX::AIEXDialect",
```

makslevental (changes requested):
Please use `AIEI32Attr`s everywhere that @jsetoain and I added here https://github.com/Xilinx/mlir-aie/pull/723.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -463,4 +463,155 @@ def AIE_SelectOp: AIEX_Op<"select", []>, Results<(outs Index)> {
...
+        I32Attr:$iteration_wrap,
+        I32Attr:$iteration_stepsize,
+        I32Attr:$next_bd,
+        I32Attr:$use_next_bd,
+        I32Attr:$valid_bd,
+        I32Attr:$lock_rel_val,
+        I32Attr:$lock_rel_id,
+        I32Attr:$lock_acq_enable,
+        I32Attr:$lock_acq_val,
+        I32Attr:$lock_acq_id
```
makslevental, line 607: Please use `AIEI32Attr`s everywhere that @jsetoain and I added here https://github.com/Xilinx/mlir-aie/pull/723.

fifield: That's not part of this PR so we'll cleanup in a subsequent patch.

makslevental (approved):

jgmelber (approved):

python/compiler/aiecc/cl_arguments.py
```diff
@@ -142,6 +142,41 @@ def parse_args(args=None):
...
+            default="PP_FD_PRE",
+            help='Kernel name in xclbin file')
+    parser.add_argument('--xclbin-instance-name',
+            dest="instance_name",
+            default="MLIRAIE",
+            help='Instance name in xclbin metadata')
+    parser.add_argument('--xclbin-kernel-id',
+            dest="kernel_id",
+            default="0x201",
+            help='Kernel id in xclbin file')
```
fifield, line 179: FYI there's a change here vs. what we had before. I added `xclbin-` prefix to these three arguments.  @jgmelber

jgmelber: The default kernel name should be MLIR_AIE here

jgmelber: And kernel ID is 0x901

python/compiler/aiecc/main.py
```diff
@@ -256,6 +260,214 @@ async def process_core(self, core):
       if(task):
         self.progress_bar.update(task,advance=0,visible=False)
 
+  async def emit_design_partition_json(self, output_filename, kernel_id='0x201'):
```
jgmelber, line 263: kernel_id should be 0x901

python/compiler/aiecc/cl_arguments.py
fifield: Thanks. And `MLIRAIEV1` instance name? I think I rebased wrong commits in this PR and I'm still picking through the wreckage.

jgmelber: Yes that’s right

lib/Targets/AIETargetCDO.cpp
```diff
@@ -0,0 +1,1084 @@
...
+      << "//#include \"adf/adf_api/AIEControlConfig.h\"\n\n"
+      << "#define __mlir_aie_try(x) x\n"
+      << "static XAie_DmaDimDesc *__mlir_aie_alloc_dim_desc(size_t ndims) {\n"
+      << "  XAie_DmaDimDesc *ret = NULL;\n"
+      << "  ret = (XAie_DmaDimDesc *)calloc(sizeof(XAie_DmaDimDesc), ndims);\n"
+      << "  if(NULL == ret) {\n"
+      << "    fprintf(stderr, \"Allocating DmaDimDesc failed.\\n\");\n"
+      << "  }\n"
+      << "  return ret;\n"
+      << "}\n";
```
keryell, line 74: Why not using raw string literals?
https://en.cppreference.com/w/cpp/language/string_literal

lib/Targets/AIETargetCDO.cpp
```diff
@@ -0,0 +1,1084 @@
...
+      << "#endif\n"
+      << "\n"
+      << "#ifdef __CDO__\n"
+      << "      int trace_config_stream_option = 2;\n"
+      << "#else\n"
+      << "      int trace_config_stream_option = 0;\n"
+      << "#endif\n"
+      << "    }\n"
+      << "  } initAIEControl;\n\n";
+  return str;
```
keryell, line 137: At the I wonder whether it would not be easier to just provide some `.h` files and a run-time library instead of generating all this, for example if we want to call this from another program or runtime, to mix-and-match say AIE + GPU, potentially from other vendors.

lib/Targets/AIETargetCDO.cpp
```diff
@@ -0,0 +1,1084 @@
...
+  // u8 ChNum, XAie_DmaDirection Dir, u8 Pause); AieRC
+  // XAie_DmaChannelPauseMem(XAie_DevInst *DevInst, XAie_LocType Loc, u8 ChNum
+  // XAie_DmaDirection Dir, u8 Pause); AieRC XAie_DmaChannelConfig(XAie_DevInst
+  // *DevInst, XAie_DmaDesc *DmaDesc, XAie_LocType Loc, u8 ChNum,
+  // XAie_DmaDirection Dir, u8 RepeatCount, u8 EnTokenIssue, u8 ControllerId);
+  // AieRC XAie_DmaChannelPushBdToQueue(XAie_DevInst *DevInst, XAie_LocType Loc,
+  // u8 ChNum, XAie_DmaDirection Dir, u8 BdNum); AieRC
+  // XAie_DmaChannelEnable(XAie_DevInst *DevInst, XAie_LocType Loc, u8 ChNum,
+  // XAie_DmaDirection Dir); AieRC XAie_DmaChannelDisable(XAie_DevInst *DevInst,
+  // XAie_LocType Loc, u8 ChNum, XAie_DmaDirection Dir);
```
keryell, line 284: Unfortunate formatting? Use `#if 0`/`#endif` instead?

lib/Targets/AIETargetCDO.cpp
```diff
@@ -0,0 +1,1084 @@
...
+               << tileLocStr(col, row) << ", "
+               << "/* ChNum */ " << chNum
+               << ", "
+               // TODO hack until physical dialect changes
+               << "/* dmaDir */ DMA_" << dmaDir << ");\n";
+      }
+    }
+  }
+
+  // for (auto op : module.getOps<ExternalBufferOp>()) {
```
keryell, line 665: `#if 0`?
Well, at least I cannot complains about this is purely uncommented opaque open-source software. ;-)

lib/Targets/AIETargetCDO.cpp
```diff
@@ -0,0 +1,1084 @@
...
+               << ", "
+               // TODO hack until physical dialect changes
+               << "/* dmaDir */ DMA_" << dmaDir << ", "
+               << "/* BdNum */" << bdNum << ");\n";
+        output << "XAie_DmaChannelEnable(" << deviceInstRef << ", "
+               << tileLocStr(col, row) << ", "
+               << "/* ChNum */ " << chNum
+               << ", "
+               // TODO hack until physical dialect changes
+               << "/* dmaDir */ DMA_" << dmaDir << ");\n";
```
keryell, line 470: Perhaps it is time to add a dependency on https://github.com/fmtlib/fmt while waiting for LLVM to move to C++20 with Python f-string-like `std::format`?
@makslevental what our library-dependency master thinks?

python/compiler/aiecc/main.py
```diff
@@ -256,6 +260,214 @@ async def process_core(self, core):
...
+             }}
+    print(output_filename)
+    with open(output_filename, 'w') as f:
+      f.write(json.dumps(design, indent=2))
+
+  async def emit_design_bif(self, path, output_filename):
+    s = """
+all:
+{
+  id_code = 0x14ca8093
```
keryell, line 363: ?

lib/Targets/AIETargetCDO.cpp
makslevental: for these long unparameterized ones I agree. For the parameterized ones, I would consider taking on https://github.com/fmtlib/fmt which implements both [C++20 std::format](https://en.cppreference.com/w/cpp/utility/format) and [C++23 std::print](https://en.cppreference.com/w/cpp/io/print) (so eventually we'll be able to just remove the dep easily).

lib/Targets/AIETargetCDO.cpp
makslevental: LOL I left my earlier comment before seeing this one. Yes, I support it - especially in header only mode.

makslevental: There are several other places (linker scripts?) where fmt would be a boon as well.

lib/Targets/AIETargetCDO.cpp
fifield: There's some code like you suggest checked into a new `data` area, but it would be better to delete that and the code you highlight. I think the preference is to generate the configuration bits directly and remove the dependence on the Vitis library. @jgmelber has already started some work down that path.

PR #740, opened 2023-11-09
[aievec] Add conversion from MatMulOp to LLVM intrinsic call

jsetoain:
In order to flatten the 2D vectors we generate `vector.shape_cast` ops, which currently treat 2D vectors as a struct of 1D vectors in for LLVM IR purposes. Once we add support to lower the shape casts as bitcasts, this issue will disappear.

### DIFF
--- a/include/aie/Conversion/Passes.td
+++ b/include/aie/Conversion/Passes.td
@@ -22,7 +22,9 @@ def ConvertAIEVecToLLVM : Pass<"convert-aievec-to-llvm", "mlir::ModuleOp"> {
     This pass converts AIEVec dialect ops to LLVM dialect calls to builtins.
   }];
   let constructor = "xilinx::aievec::createConvertAIEVecToLLVMPass()";
-  let dependentDialects = ["LLVM::LLVMDialect"];
+  let dependentDialects = ["LLVM::LLVMDialect",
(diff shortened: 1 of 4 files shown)


david-vc (approved):
This is very nice work. Thank you.

makslevental:
> Once LLVM IR supports multidimensional vectors, the expectation is that this will correct itself.

Interesting - hadn't heard about this. Any links to info?

stephenneuendorffer:
> > Once LLVM IR supports multidimensional vectors, the expectation is that this will correct itself.
> 
> Interesting - hadn't heard about this. Any links to info?

I think you're reading more into this comment.  The intrinsics defined in Peano are currently type-erased.  They should probably have proper types.  I think Javier is working around this until Peano can be fixed.

jsetoain:
> > > Once LLVM IR supports multidimensional vectors, the expectation is that this will correct itself.
> > 
> > 
> > Interesting - hadn't heard about this. Any links to info?
> 
> I think you're reading more into this comment. The intrinsics defined in Peano are currently type-erased. They should probably have proper types. I think Javier is working around this until Peano can be fixed.

Apologies, I chose the worst wording possible. I'll clarify the comment.

PR #744, opened 2023-11-09
Dedicated build script for Ryzen AI

jgmelber:


keryell (approved):
It starts becoming complicated with all these scripts with a lot of code bloat.
We need to think to a way to reduce the technical debt.

stephenneuendorffer:
I'm confused why there's a build-time difference here?

jgmelber:
> I'm confused why there's a build-time difference here?

Building aie-rt from GitHub with sed change to two lines in the CDO backend file. Then passing that aie-rt with cmake to the build

denolf (approved):

stephenneuendorffer:
> > I'm confused why there's a build-time difference here?
> 
> Building aie-rt from GitHub with sed change to two lines in the CDO backend file. Then passing that aie-rt with cmake to the build

That's scary...  If that's really different for different AIE targets, then needs to get pushed into aie-rt.

jgmelber:
> > > I'm confused why there's a build-time difference here?
> > 
> > 
> > Building aie-rt from GitHub with sed change to two lines in the CDO backend file. Then passing that aie-rt with cmake to the build
> 
> That's scary... If that's really different for different AIE targets, then needs to get pushed into aie-rt.

Agreed. The number of different versions of aie-rt we have needs to be addressed.

stephenneuendorffer:
> > > > I'm confused why there's a build-time difference here?
> > > 
> > > 
> > > Building aie-rt from GitHub with sed change to two lines in the CDO backend file. Then passing that aie-rt with cmake to the build
> > 
> > 
> > That's scary... If that's really different for different AIE targets, then needs to get pushed into aie-rt.
> 
> Agreed. The number of different versions of aie-rt we have needs to be addressed.

Then suggest you put a big FIXME around that :)

PR #746, opened 2023-11-09
fix cmakes

makslevental:
Not sure what the idea here is/was but having multiple reloads of everything (LLVM/MLIR/etc) resets CMake global properties like `MLIR_DIALECT_LIBS`, `MLIR_CONVERSION_LIBS`, etc..

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,11 +173,47 @@ add_subdirectory(lib)
 add_subdirectory(python)
 add_subdirectory(runtime_lib)
 add_subdirectory(aie_runtime_lib)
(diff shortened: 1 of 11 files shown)


fifield (approved):
I don't know the history either but the cleanup makes sense to me.

keryell (approved):
Great cleaning.

stephenneuendorffer:
There's a specific point here: it enables configuring and running the tests against installed binaries.

stephenneuendorffer (changes requested):
If you have a better idea how to accomplish running the tests independently, I'd love to hear it.

PR #747, opened 2023-11-10
[nfc] Generalize type constraints for AIE operands

jsetoain:
Instead of using unnecessarily specific types, add classes that allow us to express the constraints in a more widely applicable way.


david-vc (approved):
LGTM. Thanks.

PR #750, opened 2023-11-10
fix path issues

denolf:
the mlir-aie build script will find its CMakeLists.txt in to root folder, not in the llvm subfolder. I think this is a cut and paste mistake from the llvm build scripts.
Most likely other mlir-aie build scripts need fix as well.


makslevental (approved):
Yea that's definitely my fault. Sorry.

denolf:
merged #750

PR #754, opened 2023-11-13
Python bindings cleanup

makslevental:
Further cleanup of the bindings. Most of `aie/types.py` will go away after https://github.com/llvm/llvm-project/pull/71218 merges.


AndraBisca (approved):
Thank you for the great cleanup!

fifield (approved):

PR #756, opened 2023-11-14
Hello World printf-style log message example

STFleming:
This PR contains a small PoC reference design demonstrating a lightweight, printf-style log messaging utility for AIE tiles. 

It parses the format strings and their addresses host-side and passes the format strings address and parameters at runtime, avoiding string transfers. It also uses variadic templates to have minimal impact on instruction memory compared to traditional embedded printf implementations.


denolf (approved):

reference_designs/log_hello_world/kernel.cc
```diff
@@ -0,0 +1,34 @@
+#define NOCPP
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define REL_WRITE 0
+#define REL_READ 1
```
keryell, line 8: Are these used?

jgmelber:
I pushed a version with the python bindings here: https://github.com/Xilinx/mlir-aie/tree/printf 

Ideally this is inside the ipu-xrt subdirectory in reference-designs for now since these are device specific. That appears: https://github.com/Xilinx/mlir-aie/tree/examples2

jgmelber (approved):
I think this is RTM. https://github.com/Xilinx/mlir-aie/commits/examples2 moves this to the ipu-xrt subdirectory and adds python bindings to this example.

reference_designs/log_hello_world/decodelog.hpp
```diff
@@ -0,0 +1,127 @@
+#include "xrt/xrt_bo.h"
+#include <boost/filesystem.hpp>
```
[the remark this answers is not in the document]
STFleming: Good idea! Thanks.

reference_designs/log_hello_world/kernel.cc
STFleming: Nope, these are not needed. Thanks for the catch.

stephenneuendorffer (commented):
This is quite neat, but it seems like the rendering of the elf strings is a proof of concept at best: we'd need to deconflict different elfs for different cores and there's really no good reason I can see for passing in the strings as CSV.  My preference would be to cook this a bit more before it lands.

PR #758, opened 2023-11-14
add missing ipu example (sans for loop)

makslevental:
Note the `scf.for %arg3 = %c0 to %c8 step %c1` is commented out because the current (i.e., in main) implementation of `forLoop` doesn't support induction vars.

### DIFF
--- a/test/python/ipu.py
+++ b/test/python/ipu.py
@@ -66,6 +66,7 @@ def constructAndPrintInModule(f):
 # CHECK:   }
 # CHECK: }
 
+
 @constructAndPrintInModule
 def my_vector_scalar():
     N = 4096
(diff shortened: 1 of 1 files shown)


jackl-xilinx:
The inner loop for the add_one example is still missing (see aie.mlir). @makslevental,  is this functionality is not possible?

makslevental:
> The inner loop for the add_one example is still missing (see aie.mlir). @makslevental, is this functionality is not possible?

As the description states: not in main.

makslevental:
You can find a fully functional implementation on the `python-bindings-cleanup-2` branch: https://github.com/Xilinx/mlir-aie/pull/754/commits/bc7295cf101706bf1243c0b79c6b7c5f98444744 .

jackl-xilinx (approved):
Good first step to add this design but we'll need the inner loop functionality to be supported that you have in the python-bindings-cleanup-2 branch.

PR #762, opened 2023-11-16
Scripts to enable local build from github actions workflow using nektos/act

SamuelBayliss:
These script enable a containerized build locally using nektos/act. If these are widely adopted we can have many people able to replicate the CI flow exactly in their local machine ( with an environment controlled using the same mechanisms as the CI). 

These scripts contain some additions to make this go fast. 
- CCache usage backed by a persistent container
- Worktree checkout enabled by a bind-mount of LLVM
- Substitution of local pre-built development container, rather than fetching packages on each build.

### DIFF
--- a/.github/workflows/buildAndTestLocal.yml
+++ b/.github/workflows/buildAndTestLocal.yml
@@ -43,7 +43,8 @@ jobs:
 
     container:
       image:  amd/acdcbuild:1.1
-
+      volumes: 
+        - ${{github.workspace}}/llvm-central-repo:/workspace/llvm-central-repo
     steps:
       
       - name: Display environment variables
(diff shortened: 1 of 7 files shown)


makslevental:
My 2cents: the solution to mismatch between CI and local is not to make local more like CI but to make CI more like local. Going down the `act` road leads to you inevitably trying discern the differences between that environment and GHAs _when act breaks_.

keryell (commented):
Quite useful!
Probably we should also run this from... GitHub CI :-) to check it gives consistent results.

hunhoffe:
We're doing some cleanup of the PR backlog, and this one has gone stale — it's fallen behind main and would need a non-trivial rebase to land.

I'm going to close it for now to keep the queue tidy. If you're still interested, we are happy to revisit this idea in the future!

PR #765, opened 2023-11-18
clang-tidy AIEDialect.cpp

makslevental:


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -33,26 +34,22 @@ struct AIEInlinerInterface : public DialectInlinerInterface {
   // We don't have any special restrictions on what can be inlined into
   // destination regions. Always allow it.
   bool isLegalToInline(Region *dest, Region *src, bool wouldBeCloned,
-                       IRMapping &valueMapping) const final override {
```
makslevental, line 36: final and override are redundant

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -546,19 +531,18 @@ LogicalResult xilinx::AIE::ObjectFifoLinkOp::verify() {
...
+    auto fifoType = fifoOut.getElemType().cast<AIEObjectFifoType>();
+    auto elemType = fifoType.getElementType().cast<MemRefType>();
     int64_t outputSize = 1;
     for (auto dim : elemType.getShape())
       outputSize *= dim;
 
     int inputSize = 0;
     for (auto fifoIn : getInputObjectFifos()) {
-      AIEObjectFifoType fifo = fifoIn.getElemType().cast<AIEObjectFifoType>();
-      MemRefType elemType = fifo.getElementType().cast<MemRefType>();
```
makslevental, line 559: we need to stop doing this in general - LLVM highly recommends spelling out types fully unless they're long template specializations.

keryell (commented):
Thanks, it looks more modern.

PR #772, opened 2023-11-21
Fixes for --assign-lock-ids pass 

newling:
This fixes a few issues with this pass. The issues were

1) hard coded number of locks per tile at 15. 
2) in some cases, locks were still being assigned IDs greater than the number of HW locks 
3) did not signal failure (just emitted error from op, and returned) to halt pass pipeline 

This PR also implements the pass a bit more efficiently (does less work to find a free ID).

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
@@ -57,9 +57,10 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
       if (isAssigned) {
         auto lockID = lockOp.getLockID().value();
         auto iter = tileToLocks.find(tileOp);
-        if (iter == tileToLocks.end()) {
+
+        if (iter == tileToLocks.end())
           tileToLocks.insert({tileOp, {{lockID}, /* unassigned = */ {}}});
-        } else {
+        else {
           if (iter->second.assigned.find(lockID) !=
               iter->second.assigned.end()) {
             auto diag = lockOp->emitOpError("is assigned to the same lock (")
@@ -70,26 +71,23 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
           }
           iter->second.assigned.insert(lockID);
         }
-      }
-
-      else {
+      } else {
         auto iter = tileToLocks.find(tileOp);
-        if (iter == tileToLocks.end()) {
+        if (iter == tileToLocks.end())
           tileToLocks.insert({tileOp, {/* assigned = */ {}, {lockOp}}});
-        } else {
+        else
           iter->second.unassigned.push_back(lockOp);
-        }
       }
     }
 
     // IR mutation: assign locks to all unassigned lock ops.
-    for (auto &&[tileOp, locks] : tileToLocks) {
+    for (auto [tileOp, locks] : tileToLocks) {
 
       const auto locksPerTile =
           getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
 
       uint32_t nextID = 0;
-      for (auto &&lockOp : locks.unassigned) {
+      for (auto lockOp : locks.unassigned) {
         while (nextID < locksPerTile &&
                (locks.assigned.find(nextID) != locks.assigned.end())) {
           ++nextID;
--- a/lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
@@ -40,12 +40,12 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
     DeviceOp device = getOperation();
     OpBuilder rewriter = OpBuilder::atBlockEnd(device.getBody());
 
-    // Map from tiles to all their lock ops (stored by ID) which have been
-    // assigned to locks.
-    DenseMap<TileOp, DenseSet<int>> tileToAssignedLocks;
+    struct TileLocks {
+      DenseSet<int> assigned;
+      SmallVector<LockOp> unassigned;
+    };
 
-    // Map from tiles to all lock ops which have not been assigned to locks.
-    DenseMap<TileOp, SmallVector<LockOp>> tileToUnassignedLocks;
+    DenseMap<TileOp, TileLocks> locks;
 
     // Separate lock ops into assigned and unassigned ops, and store by tile.
     for (auto lockOp : device.getOps<LockOp>()) {
@@ -56,75 +56,53 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
       // Append to set of assigned locks.
       if (isAssigned) {
         auto lockID = lockOp.getLockID().value();
-        auto assignedLocksIter = tileToAssignedLocks.find(tileOp);
-        if (assignedLocksIter == tileToAssignedLocks.end()) {
-          tileToAssignedLocks.insert({tileOp, {lockID}});
+        auto assignedLocksIter = locks.find(tileOp);
+        if (assignedLocksIter == locks.end()) {
+          locks.insert({tileOp, {{lockID}, {}}});
         } else {
-          if (assignedLocksIter->second.find(lockID) !=
-              assignedLocksIter->second.end()) {
+          if (assignedLocksIter->second.assigned.find(lockID) !=
+              assignedLocksIter->second.assigned.end()) {
             auto diag = lockOp->emitOpError("is assigned to the same lock (")
                         << lockID << ") as another op.";
             diag.attachNote(tileOp.getLoc())
                 << "tile has lock ops assigned to same lock.";
             return signalPassFailure();
           }
-          assignedLocksIter->second.insert(lockID);
+          assignedLocksIter->second.assigned.insert(lockID);
         }
       }
 
       // Append to set of unassigned locks.
       else {
-        auto unassignedLocksIter = tileToUnassignedLocks.find(tileOp);
-        if (unassignedLocksIter == tileToUnassignedLocks.end()) {
-          tileToUnassignedLocks.insert({tileOp, {lockOp}});
+        auto unassignedLocksIter = locks.find(tileOp);
+        if (unassignedLocksIter == locks.end()) {
+          locks.insert({tileOp, {{}, {lockOp}}});
         } else {
-          unassignedLocksIter->second.push_back(lockOp);
+          unassignedLocksIter->second.unassigned.push_back(lockOp);
         }
       }
     }
 
     // IR mutation: assign locks to all unassigned lock ops.
-    for (auto &&[tileOp, unassignedLocks] : tileToUnassignedLocks) {
+    for (auto &&[tileOp, tileLocks] : locks) {
 
       const auto locksPerTile =
           getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
 
-      auto assignedLocksIter = tileToAssignedLocks.find(tileOp);
-
-      // No locks have been assigned to this tile, so we don't do any collision
-      // checking.
-      if (assignedLocksIter == tileToAssignedLocks.end()) {
-        if (unassignedLocks.size() >= locksPerTile) {
-          auto diag = tileOp->emitOpError("has more lock ops (")
-                      << unassignedLocks.size() << ") than locks ("
-                      << locksPerTile << ").";
-          return signalPassFailure();
-        }
-        for (auto lockOp : llvm::enumerate(unassignedLocks)) {
-          lockOp.value()->setAttr("lockID",
-                                  rewriter.getI32IntegerAttr(lockOp.index()));
-        }
-      }
-
-      // Locks have been assigned to this tile, so we check for collisions when
-      // assigning locks to lock ops.
-      else {
-        const auto &assignedLocks = assignedLocksIter->second;
-        uint32_t nextID = 0;
-        for (auto &&lockOp : unassignedLocks) {
-          while (nextID < locksPerTile &&
-                 (assignedLocks.find(nextID) != assignedLocks.end())) {
-            ++nextID;
-          }
-          if (nextID == locksPerTile) {
-            auto diag = lockOp->emitOpError("not allocated a lock.");
-            diag.attachNote(tileOp.getLoc())
-                << "tile has only " << locksPerTile << " locks available.";
-            return signalPassFailure();
-          }
-          lockOp->setAttr("lockID", rewriter.getI32IntegerAttr(nextID));
+      uint32_t nextID = 0;
+      for (auto &&lockOp : tileLocks.unassigned) {
+        while (nextID < locksPerTile &&
+               (tileLocks.assigned.find(nextID) != tileLocks.assigned.end())) {
           ++nextID;
         }
+        if (nextID == locksPerTile) {
+          auto diag = lockOp->emitOpError("not allocated a lock.");
+          diag.attachNote(tileOp.getLoc())
+              << "tile has only " << locksPerTile << " locks available.";
+          return signalPassFailure();
+        }
+        lockOp->setAttr("lockID", rewriter.getI32IntegerAttr(nextID));
+        ++nextID;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,63 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
-        tileToLastID[lock_tile].second.insert(lock.getLockIDValue());
+    // Map from tile to all its locks.
+    DenseMap<TileOp, SmallVector<LockOp>> tileToLockOps;
+    for (auto lockOp : device.getOps<LockOp>()) {
+      auto tileOp = lockOp.getTileOp();
+      auto iter = tileToLockOps.find(tileOp);
+      if (iter == tileToLockOps.end()) {
+        tileToLockOps.insert({tileOp, {lockOp}});
+      } else {
+        iter->second.push_back(lockOp);
```
stephenneuendorffer, line 50: Could be DenseSet instead of SmallVector, then you wouldn't need to do the collision detection below?

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,63 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+    for (auto &&[tileOp, locks] : tileToLockOps) {
+
+      const auto locksPerTile =
+          getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
+
+      // All lock IDs which are assigned to locks before this pass.
+      DenseSet<int> idsPreAssigned;
+
+      // All lock ops which do not have IDs before this pass.
+      SmallVector<LockOp> idsNotAssigned;
```
stephenneuendorffer, line 64: not really "ids", but "locks"

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,63 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
-            else
-              break;
+    // For each tile, ensure that all its locks have unique and valid IDs.
+    for (auto &&[tileOp, locks] : tileToLockOps) {
+
+      const auto locksPerTile =
+          getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
+
+      // All lock IDs which are assigned to locks before this pass.
+      DenseSet<int> idsPreAssigned;
```
stephenneuendorffer, line 61: assignedLocks?

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,63 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
         }
+      }
+
+      assert(idsPreAssigned.size() + idsNotAssigned.size() == locks.size());
 
-        // The lockID is assigned and is stored in the set.
-        lock->setAttr("lockID", rewriter.getI32IntegerAttr(
-                                    tileToLastID[lock_tile].first));
-        tileToLastID[lock_tile].second.insert(tileToLastID[lock_tile].first);
+      uint32_t nxtId = 0;
```
stephenneuendorffer, line 86: nextID
Gnrlly avd abbrv, pls.

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,63 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+      for (auto &&lockOp : idsNotAssigned) {
+        while (nxtId < locksPerTile &&
+               (idsPreAssigned.find(nxtId) != idsPreAssigned.end())) {
+          ++nxtId;
+        }
+        if (nxtId == locksPerTile) {
+          tileOp->emitOpError("can have a maximum of ")
```
stephenneuendorffer, line 93: Would make more sense if this was an error on the lock op, with a note pointing to the tile op.

stephenneuendorffer (changes requested):
Thanks for this!

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
newling: Updated. Now looks something like

within split at X.mlir:124 offset :24:10: error: 'AIE.lock' op was not allocated a lock.
  %l16 = AIE.lock(%tMemTile)
         ^
within split at X.mlir:124 offset :24:10: note: see current operation: %17 = "AIE.lock"(%0) : (index) -> index
within split at X.mlir:124 offset :7:15: note: tile only has 16 locks available.
  %tMemTile = AIE.tile(4,4)
              ^

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
newling: lol okay

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
stephenneuendorffer: Right... but why have a vector *and* a set... why not just have a set?

newling: I've updated the design to avoid this initial vector. It's a bit more complex, but probably more efficient (only possible downside it that previously there was only 1 DenseSet, in the inner loop, now there is 1 per tile).

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +40,91 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
-        tileToLastID[lock_tile].first = 0;
-        tileToLastID[lock_tile].second.insert(lock.getLockIDValue());
-      }
-    }
+    // Map from tiles to all their lock ops (stored by ID) which have been
+    // assigned to locks.
+    DenseMap<TileOp, DenseSet<int>> tileToAssignedLocks;
+
+    // Map from tiles to all lock ops which have not been assigned to locks.
+    DenseMap<TileOp, SmallVector<LockOp>> tileToUnassignedLocks;
```
stephenneuendorffer, line 48: could be a single map?

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +40,91 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+      }
+    }
+
+    // IR mutation: assign locks to all unassigned lock ops.
+    for (auto &&[tileOp, unassignedLocks] : tileToUnassignedLocks) {
+
+      const auto locksPerTile =
+          getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
+
+      auto assignedLocksIter = tileToAssignedLocks.find(tileOp);
```
stephenneuendorffer, line 92: Which would avoid this find()

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +40,91 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+        auto assignedLocksIter = tileToAssignedLocks.find(tileOp);
+        if (assignedLocksIter == tileToAssignedLocks.end()) {
+          tileToAssignedLocks.insert({tileOp, {lockID}});
+        } else {
+          if (assignedLocksIter->second.find(lockID) !=
+              assignedLocksIter->second.end()) {
+            auto diag = lockOp->emitOpError("is assigned to the same lock (")
+                        << lockID << ") as another op.";
+            diag.attachNote(tileOp.getLoc())
+                << "tile has lock ops assigned to same lock.";
```
stephenneuendorffer, line 68: Could just be "in this tile", I think.  (i.e. a continuation of the previous statement)

newling: So, drop the number of available locks? 

Currently it is 
```
../test/assign-lock-ids/test.mlir:24:10: error: 'AIE.lock' op not allocated a lock.
  %l15 = AIE.lock(%tMemTile)
         ^
../test/assign-lock-ids/test.mlir:24:10: note: see current operation: %16 = "AIE.lock"(%0) : (index) -> index
../test/assign-lock-ids/test.mlir:7:15: note: tile has only 16 locks available.
  %tMemTile = AIE.tile(4,4)
  ```
  
  With the change (if I understand your suggestion) it is
  
  ```
  ../test/assign-lock-ids/test.mlir:24:10: error: 'AIE.lock' op not allocated a lock.
  %l15 = AIE.lock(%tMemTile)
         ^
../test/assign-lock-ids/test.mlir:24:10: note: see current operation: %16 = "AIE.lock"(%0) : (index) -> index
../test/assign-lock-ids/test.mlir:7:15: note: in this tile
  %tMemTile = AIE.tile(4,4)
  ```
  
  My default is to add any information which might be useful, and the number of locks on this tile might be?

File: lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp (hunk 6 above)
Comment (newling) [reply lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp:48]: Yeah, that's better.

File: lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp (hunk 8 above)
Comment (stephenneuendorffer) [reply lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp:68]: Ew yeah good point.  maybe "because only 16 locks available in this tile"?

File: lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp [hunk 9]
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+        if (iter == tileToLocks.end()) {
+          tileToLocks.insert({tileOp, {{lockID}, /* unassigned = */ {}}});
+        } else {
+          if (iter->second.assigned.find(lockID) !=
+              iter->second.assigned.end()) {
+            auto diag = lockOp->emitOpError("is assigned to the same lock (")
+                        << lockID << ") as another op.";
+            diag.attachNote(tileOp.getLoc())
+                << "tile has lock ops assigned to same lock.";
+            return signalPassFailure();
```
stephenneuendorffer, line 69: return diag;

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
           }
-          tileToLastID[lock_tile].first = potential_ID;
+          iter->second.assigned.insert(lockID);
+        }
+      }
+
+      else {
```
stephenneuendorffer, line 75: clang-format should put the else with the preceeding close brace.

stephenneuendorffer (approved):
Thanks!

makslevental:
@newling is this good to go?

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
makslevental: this is a `void runOnOperation()` - you can't return `diag` here.

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+        auto iter = tileToLocks.find(tileOp);
+        if (iter == tileToLocks.end()) {
+          tileToLocks.insert({tileOp, {/* assigned = */ {}, {lockOp}}});
         } else {
-          lock->emitError() << "Exceeded the number of unique LockIDs";
-          return;
+          iter->second.unassigned.push_back(lockOp);
```
makslevental, line 80: Please remove braces for single line conditions (matching upstream).

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+                << "tile has lock ops assigned to same lock.";
+            return signalPassFailure();
           }
-          tileToLastID[lock_tile].first = potential_ID;
+          iter->second.assigned.insert(lockID);
+        }
+      }
+
+      else {
+        auto iter = tileToLocks.find(tileOp);
```
makslevental, line 76: Please spell out `auto` except in cases where it's hard to do or when there's an explicit cast on the right (like upstream).

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
         }
+      }
+    }
 
-        // The lockID is assigned and is stored in the set.
-        lock->setAttr("lockID", rewriter.getI32IntegerAttr(
-                                    tileToLastID[lock_tile].first));
-        tileToLastID[lock_tile].second.insert(tileToLastID[lock_tile].first);
+    // IR mutation: assign locks to all unassigned lock ops.
+    for (auto &&[tileOp, locks] : tileToLocks) {
```
makslevental, line 86: I believe since `tileToLocks` is `DenseMap` the range-for will give you references even with a structured binding so you don't need the `&&` (not sure).

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
```diff
@@ -40,50 +39,69 @@ struct AIEAssignLockIDsPass : AIEAssignLockIDsBase<AIEAssignLockIDsPass> {
...
+
+      const auto locksPerTile =
+          getTargetModel(tileOp).getNumLocks(tileOp.getCol(), tileOp.getRow());
+
+      uint32_t nextID = 0;
+      for (auto &&lockOp : locks.unassigned) {
```
makslevental, line 92: Same here - not sure but do you need the `&&`?

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
stephenneuendorffer: Ah, my mistake :)

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
newling: I often use the universal reference `&&` as a just-do-the-right thing. Let the compiler deduce the reference type.  

You're right that DenseMap (and as far as I know all common map type) will give references, and it would require a weird custom map class for the `auto [tileOp, locks]` case to actually incur a copy of locks. 

Also, it seems like the universal reference approach is not common upstream (251 vs 31 in llvm/mlir/lib) so I'll change this case to what you suggest.

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
newling: This will be deduced `const LockOp & lockOp`. If I had to write it explicitly I would drop the const-ref so will drop the && here too.

lib/Dialect/AIE/Transforms/AIEAssignLockIDs.cpp
newling: I didn't realize that was the upstream policy, can you please point me to a reference? 

I don't know what this type is, so personally this classifies as a "hard-to-do" case :-)

newling: image

makslevental: Ah I definitely dinged you for the worst possible instance - that definitely qualifies as "hard to spell". Anyway some of the others I would definitely appreciate you spelling out (but if your crags or whatever aren't forthcoming don't worry about it - I'll come back around later and change em using my IDE).

newling: Ok cool, thanks. Will skim through my autos before landing

newling:
@makslevental and @stephenneuendorffer I don't seem to have write permission -- I'm happy with the current state if you want to go ahead and merge this.

PR #773, opened 2023-11-21
[aievec] Add lowering pattern for vector contraction

jsetoain:


test/Conversion/VectorToAIEVec/test-contract.mlir
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// CHECK-LABEL: func.func @contracti32i32i32(
+// CHECK-SAME: %[[A:[a-zA-Z0-9]+]]: vector<4x4xi16>,
+// CHECK-SAME: %[[B:[a-zA-Z0-9]+]]: vector<4x8xi8>,
+// CHECK-SAME: %[[C:[a-zA-Z0-9]+]]: vector<4x8xi32>) -> vector<4x8xi32> {
+// CHECK:        %[[R:.*]] = aievec.matmul %[[A]], %[[B]], %[[C]] :
+// CHECK-SAME:   vector<4x4xi16>, vector<4x8xi8> into vector<4x8xi32>
+// CHECK:        return %[[R]] : vector<4x8xi32>
+func.func @contracti32i32i32(%A : vector<4x4xi16>,
```
david-vc, line 86: I see, the version first promotes the operands to i32 whereas the version at L31 promotes the i8 operand to i16.

david-vc (approved):
Nice work.

test/Conversion/VectorToAIEVec/test-contract.mlir
jsetoain: I'm dealing with two issues here. First, when you generalize `linalg.matmul` with mixed operands, it generates widening operations before the arithmetic, and when you vectorize you end up with something like what you see there. At the same time, although the `vector.contract` operation supports different precision for inputs and outputs, it doesn't support different precision between the two inputs (A & B must have the same elemental type). This is me just making sure that there's a lowering strategy for all our supported cases 🙂

PR #775, opened 2023-11-22
fix hidden pathfinder bugs

makslevental:
There are `create-flows` tests that currently fail but because the tests are written in a weird way the failures are not caught by lit: `create-flows/mmult.mlir`, `create-flows/routed_herd_3x1.mlir`, and `create-flows/routed_herd_3x2.mlir` all fail because the existing implementation of `DynamicTileAnalysis::runAnalysis` doesn't actually care if [`pathfinder->addFixedConnection` fails](https://github.com/Xilinx/mlir-aie/blob/9ce3079e66f4c01c4854c9e28e3cdac107200f78/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L78). The tests hide this because they only check for a round trip (since they use `aie-find-flows`) which effectively happens because the original source prints again (after the `switchboxOp.emitOpError`s get printed at the top).

Going back to figure out when it broke (just after I merged in https://github.com/Xilinx/mlir-aie/pull/706) I could not, no matter what I did, repro the behavior of the old implementation in terms of paths found by dijkstra. It boiled down to three things:

1. Boost's graph iterates over edges in `vertex_descriptor` order, i.e., by `id`;
2. `std::set` is a tree implementation not a heap implementation (I knew this);
3. Boost's d-ary heap (used by Boost's dijkstra) and `std::pop_heap` operate differently[^1] and therefore produce different heap orders (recall that heap sort isn't stable).

**The solution** was to steal Boost's d-ary heap and reimplement dijkstra like they do (using node coloring and etc). After this change, the tests 100% agree before and after the removal of Boost except for the failing tests which have always failed and just never been caught (prior no error was even emitted). To verify/validate that it's in fact 100% I extended/changed the tests to match on the actual emitted assembly and removed `aie-find-flows` pass, which (again) was just for some kind of round-trip like test.

**Three remaining tests** all have fixed connections that can't be added;  you can see [which connections in the commit where I commented them out to get the test suite to pass](https://github.com/Xilinx/mlir-aie/pull/775/commits/324f7fba580ed1959af8663359a030d0385c37d8). I do not have enough experience to know whether this is user error (and they shouldn't have been "connected" in the first place) or whether the [implementation of `addFixedConnection` is broken](https://github.com/Xilinx/mlir-aie/blob/a1df725251c6c2c491a399b156ee4fdd67bd56fc/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L260). 

[^1]: `std::pop_heap` doesn't actually pop the last element and expects you to pop, while Boost's heap does pop the last element for you. This means that `std::pop_heap` does heap down on an underlying container that's 1 longer than the underlying container in Boost, when Boost performs the heap down after a pop.


erwei-xilinx:
Thanks for working on this, Maks! To me this seems to suggest that when user tries to add extra routings to a previously routed design, then the previous routing's connections to tile DMA ports shall be lost... To my knowledge, in our AIR and AIE designs we haven't really tried in practice to incrementally route a previously routed design, which could be why this bug hasn't been discovered previously...

To inspect the routings, this visualization tool might be useful: https://xilinx.github.io/mlir-aie/AIERouting (Visualizing Routing).

makslevental:
> Thanks for working on this, Maks! To me this seems to suggest that when user tries to add extra routings to a previously routed design, then the previous routing's connections to tile DMA ports shall be lost... 

That's not correct; `AIEPathfinder` will respect [existing connections](https://github.com/Xilinx/mlir-aie/blob/a1df725251c6c2c491a399b156ee4fdd67bd56fc/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L415). That's not what's going wrong here; `addFixedConnection` only fails to add if no matching edge exists on the fabric:

```cpp
auto matchingCh =
    std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
      return ch.src.col == coords.col && ch.src.row == coords.row &&
             ch.bundle == port.bundle;
    });
if (matchingCh == edges.end())
  return false;
```

In particular, this mean that the connections that were failing to be added (the connections that I had to comment out) were on bundles that do not exist for those switchboxes[^1]. Now maybe they should exist and the fabric/graph initialization [isn't adding them to the graph](https://github.com/Xilinx/mlir-aie/blob/a1df725251c6c2c491a399b156ee4fdd67bd56fc/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L190-L212) but that's exactly where I need someone with more experience to judge because I don't know the architecture intimiately yet.

[^1]: Because I do know that those switchboxes do exist at those coordinates so it must be `ch.bundle == port.bundle` that's failing.

erwei-xilinx:
> > Thanks for working on this, Maks! To me this seems to suggest that when user tries to add extra routings to a previously routed design, then the previous routing's connections to tile DMA ports shall be lost...
> 
> That's not correct; `AIEPathfinder` will respect [existing connections](https://github.com/Xilinx/mlir-aie/blob/a1df725251c6c2c491a399b156ee4fdd67bd56fc/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L415). That's not what's going wrong here; `addFixedConnection` only fails to add if no matching edge exists on the fabric:
> 
> ```c++
> auto matchingCh =
>     std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
>       return ch.src.col == coords.col && ch.src.row == coords.row &&
>              ch.bundle == port.bundle;
>     });
> if (matchingCh == edges.end())
>   return false;
> ```
> 
> In particular, this means that the connections that were failing to be added (the connections that I had to comment out) were on ports/bundles that do not exist for those switchboxes[1](#user-content-fn-1-80457d376ead655917716d44644e90fc). Now maybe those ports/bundles should exist and the fabric/graph and initialization [isn't adding them to the graph](https://github.com/Xilinx/mlir-aie/blob/a1df725251c6c2c491a399b156ee4fdd67bd56fc/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp#L190-L212) but that's exactly where I need someone with more experience to judge because I don't know the architecture intimiately yet.
> 
> ## Footnotes
> 1. Because those switchboxes do exist at those coordinates so it must be `ch.bundle == port.bundle` that's failing. [↩](#user-content-fnref-1-80457d376ead655917716d44644e90fc)

```
%[[VAL_68:.*]] = AIE.switchbox(%[[VAL_16]]) {
  AIE.connect<South : 0, DMA : 0>
  AIE.connect<East : 0, DMA : 1>
}
```
Should be a legal routing, for example.

makslevental:
> Should be a legal routing for the fabric, for example.

This is the original assembly:

```mlir
    %switchbox_0_3 = AIE.switchbox(%tile_0_3) {
      // AIE.connect<South : 0, DMA : 0>
      // AIE.connect<East : 0, DMA : 1>
    }
```

So these are both incoming connections to an edge tile right? Currently (if my understanding is correct) Pathfinder doesn't allow/enable incoming connections to any edge tiles.

erwei-xilinx:
> > Should be a legal routing for the fabric, for example.
> 
> This is the original assembly:
> 
> ```mlir
>     %switchbox_0_3 = AIE.switchbox(%tile_0_3) {
>       // AIE.connect<South : 0, DMA : 0>
>       // AIE.connect<East : 0, DMA : 1>
>     }
> ```
> 
> So these are both incoming connections to an edge tile right? Currently (if my understanding is correct) Pathfinder doesn't allow/enable incoming connections to any edge tiles.

This tile (0, 3) is on the west edge, but those two connections should be legal because they come from east and south. Any incoming/outgoing edges to the west would've been illegal, IMO.

keryell:
At the end, using Boost might be too much for you, but including some textual snapshots of Boost does not seem super healthy either.

makslevental:
> At the end, using Boost might be too much for you, but including some textual snapshots of Boost does not seem super healthy either.

Neither of these is right - there should be unit tests that don't rely on implementation details of a heap so that no one can afterwards verify whether the code is correct without running an E2E test. 

> At the end, using Boost might be too much for you

I'm not some wacky crank here and I cannot reiterate enough how much of a non-starter Boost is for downstream users/projects https://github.com/nod-ai/iree-amd-aie/pull/19#issuecomment-1813216513.

makslevental:
> This tile (0, 3) is on the west edge, but those two connections should be legal because they come from east and south. Any incoming/outgoing edges to the west would've been illegal, IMO.

@erwei-xilinx how about these?

```mlir
module {
  AIE.device(xcvc1902) {
    %tile_2_0 = AIE.tile(2, 0)
    %tile_3_0 = AIE.tile(3, 0)
    %tile_6_0 = AIE.tile(6, 0)
    %tile_7_0 = AIE.tile(7, 0)

    %switchbox_2_0 = AIE.switchbox(%tile_2_0) {
      AIE.connect<North : 0, South : 2>
      AIE.connect<North : 1, South : 3>
    }
    %switchbox_3_0 = AIE.switchbox(%tile_3_0) {
      AIE.connect<North : 0, South : 2>
      AIE.connect<North : 1, South : 3>
    }
    %switchbox_6_0 = AIE.switchbox(%tile_6_0) {
      AIE.connect<North : 0, South : 2>
      AIE.connect<North : 1, South : 3>
    }
    %switchbox_7_0 = AIE.switchbox(%tile_7_0) {
      AIE.connect<North : 0, South : 2>
      AIE.connect<North : 1, South : 3>
    }
  }
}
```

Each of these are southern edge tiles that don't have incoming edges on southern ports and thus fail `addFixedConnection`.

erwei-xilinx:
> > This tile (0, 3) is on the west edge, but those two connections should be legal because they come from east and south. Any incoming/outgoing edges to the west would've been illegal, IMO.
> 
> @erwei-xilinx how about these?
> 
> ```mlir
> module {
>   AIE.device(xcvc1902) {
>     %tile_2_0 = AIE.tile(2, 0)
>     %tile_3_0 = AIE.tile(3, 0)
>     %tile_6_0 = AIE.tile(6, 0)
>     %tile_7_0 = AIE.tile(7, 0)
> 
>     %switchbox_2_0 = AIE.switchbox(%tile_2_0) {
>       AIE.connect<North : 0, South : 2>
>       AIE.connect<North : 1, South : 3>
>     }
>     %switchbox_3_0 = AIE.switchbox(%tile_3_0) {
>       AIE.connect<North : 0, South : 2>
>       AIE.connect<North : 1, South : 3>
>     }
>     %switchbox_6_0 = AIE.switchbox(%tile_6_0) {
>       AIE.connect<North : 0, South : 2>
>       AIE.connect<North : 1, South : 3>
>     }
>     %switchbox_7_0 = AIE.switchbox(%tile_7_0) {
>       AIE.connect<North : 0, South : 2>
>       AIE.connect<North : 1, South : 3>
>     }
>   }
> }
> ```
> 
> Each of these are southern edge tiles that don't have incoming edges on southern ports and thus fail `addFixedConnection`.

Those connections look legal to me too. The south edge tiles are adjacent to the shimmux so inbound/outbound connections to south should be allowed.

makslevental:
> Those connections look legal to me too. The south edge tiles are adjacent to the shimmux so inbound/outbound connections to south should be allowed.

Ah but shimmuxes aren't considered part of the fabric according to Pathfinder (those are handles later by the [conversion itself](https://github.com/Xilinx/mlir-aie/blob/00279bde2fd030d502e3118b96e45e625b94c497/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp#L101))

erwei-xilinx:
> > Those connections look legal to me too. The south edge tiles are adjacent to the shimmux so inbound/outbound connections to south should be allowed.
> 
> Ah but shimmuxes aren't considered part of the fabric according to Pathfinder (those are handles later by the [conversion itself](https://github.com/Xilinx/mlir-aie/blob/00279bde2fd030d502e3118b96e45e625b94c497/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp#L101))

I think it makes sense to have pathfinder pull those architecture information from `AIETargetModel` instead of hard-coded configuration like those... `AIETargetModel` would've known what connections are legal.

makslevental:
@erwei-xilinx I think I have fixed the issue - the specific commit is https://github.com/Xilinx/mlir-aie/pull/775/commits/f7dcb9557f268ba9662a6a123a161c0e39ac315b but the new implementation of `addFixedConnections` now looks like this:

```cpp
bool Pathfinder::addFixedConnection(ConnectOp connectOp) {
  auto sb = connectOp->getParentOfType<SwitchboxOp>();
  // TODO: keep track of capacity?
  if (sb.getTileOp().isShimNOCTile())
    return true;

  TileID sbTile = sb.getTileID();
  WireBundle sourceBundle = connectOp.getSourceBundle();
  WireBundle destBundle = connectOp.getDestBundle();

  // find the correct Channel and indicate the fixed direction
  // outgoing connection
  auto matchingCh =
      std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
        return static_cast<TileID>(ch.src) == sbTile && ch.bundle == destBundle;
      });
  if (matchingCh != edges.end())
    return matchingCh->fixedCapacity.insert(connectOp.getDestChannel())
               .second ||
           true;

  // incoming connection
  matchingCh = std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
    return static_cast<TileID>(ch.target) == sbTile &&
           ch.bundle == getConnectingBundle(sourceBundle);
  });
  if (matchingCh != edges.end())
    return matchingCh->fixedCapacity.insert(connectOp.getSourceChannel())
               .second ||
           true;

  return false;
}
```

Let me know what you think.

erwei-xilinx:
> @erwei-xilinx I think I have fixed the issue - the specific commit is [f7dcb95](https://github.com/Xilinx/mlir-aie/commit/f7dcb9557f268ba9662a6a123a161c0e39ac315b) but the new implementation of `addFixedConnections` now looks like this:
> 
> ```c++
> bool Pathfinder::addFixedConnection(ConnectOp connectOp) {
>   auto sb = connectOp->getParentOfType<SwitchboxOp>();
>   // TODO: keep track of capacity?
>   if (sb.getTileOp().isShimNOCTile())
>     return true;
> 
>   TileID sbTile = sb.getTileID();
>   WireBundle sourceBundle = connectOp.getSourceBundle();
>   WireBundle destBundle = connectOp.getDestBundle();
> 
>   // find the correct Channel and indicate the fixed direction
>   // outgoing connection
>   auto matchingCh =
>       std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
>         return static_cast<TileID>(ch.src) == sbTile && ch.bundle == destBundle;
>       });
>   if (matchingCh != edges.end())
>     return matchingCh->fixedCapacity.insert(connectOp.getDestChannel())
>                .second ||
>            true;
> 
>   // incoming connection
>   matchingCh = std::find_if(edges.begin(), edges.end(), [&](ChannelEdge &ch) {
>     return static_cast<TileID>(ch.target) == sbTile &&
>            ch.bundle == getConnectingBundle(sourceBundle);
>   });
>   if (matchingCh != edges.end())
>     return matchingCh->fixedCapacity.insert(connectOp.getSourceChannel())
>                .second ||
>            true;
> 
>   return false;
> }
> ```
> 
> Let me know what you think.

Those changes make sense to me. Thanks!

PR #757, opened 2023-11-14
Ryzen AI Examples

jgmelber:

PR #777, opened 2023-11-22
Color detect

denolf:
Adding the color detect vision pipeline. This is should only be merged after #757.


AndraBisca (approved):

jackl-xilinx (approved):
Added missing lut_inv_8b.h

makslevental:
@AndraBisca I tried to bump this but I can't get it to work; something is going wrong in the objectfifo passes; I get 

```
  File "/home/mlevental/dev_projects/mlir-aie/cmake-build-debug/python/aie/compiler/aiecc/main.py", line 848, in run_flow
    self.run_passes(
  File "/home/mlevental/dev_projects/mlir-aie/cmake-build-debug/python/aie/compiler/aiecc/main.py", line 102, in run_passes
    PassManager.parse(pass_pipeline).run(module.operation)
aie._mlir_libs._site_initialize.<locals>.MLIRError: Failure while executing pass pipeline:
error: "-":122:9: 'func.call' op operand type mismatch: expected operand type 'memref<256xui8>', but provided 'memref<64xui8>' for operand number 1
 note: "-":122:9: see current operation: "func.call"(%9, %6, %85) <{callee = @gray2rgbaLine}> : (memref<64xui8>, memref<64xui8>, i32) -> ()
```

even though the input ir to `aiecc` is legal/fine.

makslevental:
nm I got it - `OF_5to5b` was the wrong size (needs to be `ofifo_line_bytes_ty`).

AndraBisca:
Thank you @makslevental! That check makes sense to add, I actually thought I already did but it seems not.

makslevental:
> Thank you @makslevental! That check makes sense to add, I actually thought I already did but it seems not.

FYI I added it here to double check my understanding.

PR #780, opened 2023-11-26
implement llvm-cov

makslevental:
This PR implements a `generate-aie-coverage-report` build target that generates a test/code coverage report using `llvm-cov` and `llvm-profdata` (i.e., you need to compile with `clang`). This PR also implements an additional CI job that appends the coverage report to every PR (seen below).


.github/workflows/buildAndTest.yml
```diff
@@ -168,6 +168,8 @@ jobs:
           cmake .. \
             -GNinja \
             -DCMAKE_BUILD_TYPE=Debug \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_EXE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_MODULE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_SHARED_LINKER_FLAGS_INIT="-fuse-ld=lld" \
```
stephenneuendorffer, line 172: LLVM_USE_LINKER isn't sufficient?

.github/workflows/buildAndTest.yml
```diff
@@ -338,3 +345,102 @@ jobs:
...
+
+
+      - name: Update PR with coverage results
+        uses: edumserrano/find-create-or-update-comment@v2
+        with:
+          issue-number: ${{ github.event.pull_request.number }}
+          body-includes: '<!--<!doctype codecov html>-->'
+          comment-author: 'github-actions[bot]'
+          body-path: /home/runner/work/mlir-aie/mlir-aie/build_release/report/index.html
+          edit-mode: replace
```
stephenneuendorffer, line 446: Is this better than having a build artifact?

stephenneuendorffer (approved):
I dig it.

.github/workflows/buildAndTest.yml
stephenneuendorffer: In particular, would be interested in your thought on how to make this more actionable?  For instance, is there a way to link to annotated source code with llvm-cov, so we can figure out which lines aren't covered?

.github/workflows/buildAndTest.yml
makslevental: I've been ever able to get `LLVM_USE_LINKER` to behave consistently across platforms/environments/projects.

stephenneuendorffer: Interesting... my experience is that it always just worked..  has it not been working in the CI?

.github/workflows/buildAndTest.yml
makslevental: There is a [build artifact as well](https://gist.github.com/makslevental/c9462d50816727f9f3e9a5f5e0dedbd7) that could be `cat`ed 
image
but it would take more fiddling with `cut` then I'm willing to do to 🤷. 

The value of appending coverage for all source files is marginal I admit so I'm fiddling right now to see if I can narrow it to just files touched by the PR.

makslevental: > For instance, is there a way to link to annotated source code with llvm-cov, so we can figure out which lines aren't covered?

If you genereate the report locally then yes you can click each source file and get line-by-line stats. 

>  In particular, would be interested in your thought on how to make this more actionable?

Honor system: if you're adding code and you didn't add tests, this will at least be a loud reminder.

.github/workflows/buildAndTest.yml
makslevental: No clue but you can see here https://github.com/Xilinx/mlir-aie/actions/runs/6999303703/job/19038425447 where I tried both `LLVM_USE_LINKER` and `LLVM_CCACHE_BUILD` and both "didn't take".

PR #793, opened 2023-11-28
Remove extra AIETypesGen.inc header

stephenneuendorffer:
The types are already built as part of the dialect.


makslevental:
FWIW I did this intentionally, in order to be explicit, so that the teleology of all the `.td` files were clear. Especially since some do need explicit tablegen targets. But I'm not opposed.

stephenneuendorffer:
It is a bit odd that add_mlir_dialect() builds the types by default, but not any of the others...  But it seems even more odd to generate 2 different copies of this.

makslevental:
Might be better to ditch `add_mlir_dialect`? But that's bikeshedding at that point. Anyway like I said I'm not opposed.

PR #797, opened 2023-11-28
rename `objectFifo` -> `objectfifo`

makslevental:
This PR renames `objectFifo` -> `objectfifo` in order to better agree with both MLIR and Python naming conventions.


makslevental:
> I think the plan was to rename this to orderedobjectbuffer in the near term. @AndraBisca ?

I don't have a horse in this race but `orderedobjectbuffer` and `objectfifo` seem to be saying the exact same thing (a fifo is an ordered buffer...) and the former is appreciably longer. Anyway even if you guys do want to change the name this doesn't impede.

AndraBisca:
The plan to make the change to orderedobjectbuffer still stands. However, this PR is separate from that.
@makslevental the reason for the name change is an issue of consistency: we've switched to using OrderedObjectBuffer in the python bindings while in MLIR it is still objectFifo.

makslevental:
> we've switched to using OrderedObjectBuffer in the python bindings while in MLIR it is still objectFifo.

In the recent update to use upstream builders I have removed those names https://github.com/Xilinx/mlir-aie/commit/6f4bba37689deb3837b6d599703cdfaa02de2d36#diff-9949c2edaed1398afea384a2d73d8088919a0d656838952b54f15c7dfc91f126.

AndraBisca:
> > we've switched to using OrderedObjectBuffer in the python bindings while in MLIR it is still objectFifo.
> 
> In the recent update to use upstream builders I have removed those names [6f4bba3#diff-9949c2edaed1398afea384a2d73d8088919a0d656838952b54f15c7dfc91f126](https://github.com/Xilinx/mlir-aie/commit/6f4bba37689deb3837b6d599703cdfaa02de2d36#diff-9949c2edaed1398afea384a2d73d8088919a0d656838952b54f15c7dfc91f126).

I see, I did not have time to look over that PR. All the more reason then to switch to orderedobjectbuffer.

makslevental:
> I see, I did not have time to look over that PR. All the more reason then to switch to orderedobjectbuffer.

Sure no problem. Just add me to the PR when you rename and I'll fixup all the tests and python bindings and etc.

AndraBisca:
We may as well use this PR instead of another one to do the name change. Especially with the deletion of the ordered object buffer in the python bindings.

makslevental:
> Especially with the deletion of the ordered object buffer in the python bindings.

They're not entirely deleted https://github.com/Xilinx/mlir-aie/blob/main/python/dialects/aie.py#L193 - they're just not used in the tests anymore.

> We may as well use this PR instead of another one to do the name change.

I would rather send up a separate one because that one will have many more lines of C++ mixed in.

AndraBisca (approved):
I think the PR is good to go. Any subsequent name change related to the objectfifo can go in another PR dedicated to that. Thank you for your work Maks!

jbelot:
Hi,
I think there is a typo in your code for all "objectfifosubivew" => "objectfifosubview". I tried to do a PR for it but I'm not sure I have the permission to push a branch.

PR #801, opened 2023-11-29
`black` aiecc

makslevental:


python/compiler/aiecc/main.py
```diff
@@ -9,353 +9,701 @@
...
+            task,
+            [
+                "sed",
+                "-i",
+                "-E",
+                "s/memory\(argmem: read\)/argmemonly readonly/g",
+                llvmir_chesslinked,
+            ],
+        )
+        await self.do_call(
```
stephenneuendorffer, line 196: Not an improvement, I think...

stephenneuendorffer (commented):
I'd really rather do things like this through the CI, rather than wholesale?

makslevental:
> I'd really rather do things like this through the CI, rather than wholesale?

There is no Python formatting in CI. I tried to set it up and then got badgered to death (or something like that).

> Not an improvement, I think...

`black` is just the standard so there's no point in debating better/worse just like you wouldn't debate whether `clang-format` made things better or worse.

keryell:
Can anyone explain me the title of this PR? I am not an English native speaker.

makslevental:
> Can anyone explain me the title of this PR? I am not an English native speaker.

[`black`](https://github.com/psf/black) is a python formatter; you run it on a source file like so: `black source_file.py`. It also works on directories of python sources files (such as `python/compiler/aiecc`).

fifield:
>
> `black` is just the standard so there's no point in debating better/worse just like you wouldn't debate whether `clang-format` made things better or worse.

I really this to land but agree with @stephenneuendorffer that the formatting of `await self.do_call(...` here is terrible.  Maybe we can refactor somehow so that the black format is better? Or the equivalent of `// clang-format off`?

makslevental:
> I really want this to land but agree with @stephenneuendorffer that the formatting of `await self.do_call(...` here is terrible. Maybe we can refactor somehow so that the black format is better? Or the equivalent of `// clang-format off`?

This is the enormous benefit of black: you cannot change almost anything about it (only thing is line length). I do not understand the value/point/meaning of debating formatting. The only thing you want from a formatter is ironclad consistency *so that you yourself don't have to think about formatting in the least*. If those calls are too "tall" then the solution is to break out the list of args just above them into it's own named variable. I'm happy to do this if it means we can bury this debate.

stephenneuendorffer (approved):

PR #802, opened 2023-11-29
Fix for polygeist tests following llvm bump.

david-vc:
Data layout attributes need to be changed from vector<2xi32> to vector<2xi64>.


makslevental:
Hmm I thought I caught all of these.

makslevental:
Hmm I thought I caught all of these.

david-vc:
The build bot is not running these tests as they require chesscc.

makslevental:
> The build bot is not running these tests as they require chesscc.

Yes but I had tested locally. Blame Vitis! Thanks for the fix!

keryell:
I guess that these tests come from an old output from a Polygeist run?
Otherwise I am currently working on the integration of 'mlir-aie` into an LLVM mono with a modernized fork of Polygeist https://github.com/intel/llvm/tree/sycl-mlir/polygeist
I still have a lot of issues with 'mlir-aie` CMake configuration.

david-vc:
Hi @keryell : Yes. the input MLIR was produced by polygeist more than 1/2 year ago.

Issue #799, opened 2023-11-29
tutorial-2a make failed


Dinngger:
ubuntu 20.04 with vitis 2023.2
tutorial-1 make ok while tutorial-2a make failed.
![image](https://github.com/Xilinx/mlir-aie/assets/41265073/bbc237a5-d533-4c6b-bac8-7c409d405135)

cabreraam:
A dirty solution is to remove the `libcdo_driver.so` dependence by removing the `-lcdo_driver` option [here](https://github.com/Xilinx/mlir-aie/blob/abbcb9a05a37e33217db8a0771594b81e3c4b95f/python/compiler/aiecc/main.py#L520C57-L520C57).

But that doesn't solve the underlying linking issue, which I am unsure of how to proceed.

After trying tutorial 2b, I observed a couple of issues:
1. When you run `make`, there is no `sim` directory created in `air.mlir.prj`, though one gets created in tutorial 2a.
2. When running `make -C {path to}/sim on the tutorial 2a `sim` directory that gets created, I observe that there is a dependence on a boost utility. Specifically:
    ```bash
     Vitis/2023.2/aietools/bin//../include/adf/wrapper/logging.h:53:10: fatal error: boost/log/trivial.hpp: No such file or directory
     #include <boost/log/trivial.hpp>
    ```
 Does the [Getting started doc](https://github.com/Xilinx/mlir-aie/blob/abbcb9a05a37e33217db8a0771594b81e3c4b95f/docs/Building.md) need to be updated to include a dependence on the Boost libraries?

makslevental:
> I've made that change, but now I'm seeing an error ld.lld: error: /auto/software/swtree/ubuntu22.04/x86_64/Xilinx/Vitis/2023.2/aietools/lib/lnx64.o/libcdo_driver.so is incompatible with elf64-littleaarch64

That sysroot is for cross-compiling to aarch64 (zynq?). Indeed looking at the CLI args in the screencap there's a lot of that littered in there. Is that what you're trying to do @Dinngger?

>  Does the [Getting started doc](https://github.com/Xilinx/mlir-aie/blob/abbcb9a05a37e33217db8a0771594b81e3c4b95f/docs/Building.md) need to be updated to include a dependence on the Boost libraries?

[...]

PR #803, opened 2023-11-29
[tutorials] update makefile-common libstdc++ version to 12.2.0 to reflect Vitis 2023.2

cabreraam:
As found in #799, specifically [here](https://github.com/Xilinx/mlir-aie/issues/799#issuecomment-1832110083)


makslevental:
@cabreraam I just reread your response on the other issue (about how the tuts assume/default to cross-compilation). I didn't know that (gotta admit I've ever used them a tutor...). And thus I'm hesitant to merge this in simply because I don't know how this stuff is actually meant to be used; like there's a comment right there about PetaLinux 22 including the current version. So would this break on PetaLinux?

makslevental:
I gotta defer to @stephenneuendorffer on this one.

stephenneuendorffer:
Dup'd in #827

PR #804, opened 2023-11-29
Reduce parallelism in aievec tests

fifield:
This replaces single invocations of chesscc on multiple C++ input files with multiple invocations on single C++ files.

For example,
```
RUN: xchesscc ... foo.cc bar.cc baz.cc ...
```
becomes,
```
RUN: xchesscc -c foo.cc -o foo.o
RUN: xchesscc -c bar.cc -o bar.o
RUN: xchesscc foo.o bar.o baz.cc
```

The motivation is that when chesscc is given multiple C++ inputs, it will parallelize their compilation. Ordinarily this is not a problem but in the case of chesscc each C++ compilation spawns a child process with a memory footprint of ~4-5GB.  Most of the aievec tests compile 2-3 C++ files so each could potentially be using ~15GB of memory at any given time. This means that on most machines we must limit the overall job parallelism when running tests so that the aievec tests fit into memory.

With this PR each test uses at most ~5GB and the amount of job parallelism (i.e. `-j`) can be higher and is easier to control.


jsetoain (approved):
Good to know! Eager to test how many of these tests I can run at once now 😬 Thanks!

keryell:
Good!
My desktop got some OOM kill on `noodle` behind `xchesscc` recently... :-(

PR #821, opened 2023-12-02
Comprehensive linting and formatting

makslevental:
This PR implements "comprehensive" linting and formatting for both Python and C/C++. It runs

- `black` and `pylint` for Python;
- `clang-format` and `clang-tidy` for C/C++.

Notes:

- All checking/linting/formatting is only on changed **lines**.
- Annotations are posted on the changed lines themselves.
- Lint issues are warnings/suggestions (and thus applying is optional).
- **Formatting issues are errors and need to be applied (or fixed locally)**.
- Annotations/suggestions can be batch committed from the `Files changed` tab.

This PR itself serves as a test; I made "anti-changes" to `lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp` and `python/util.py` to exercise the formatters and linters. Thus (hopefully), this PR doesn't accurately represent the typical amount of noise the linters introduce.

Note: I do not intend to "own" this code. This is a shared resource and should be shared responsibility.


.clang-tidy
```diff
@@ -8,7 +8,7 @@ Checks: >
   -misc-non-private-member-variables-in-classes,
   -misc-no-recursion,
   -misc-use-anonymous-namespace,
-  readability-identifier-naming'
```
jsetoain, line 11: ?

.github/workflows/buildAndTest.yml
```diff
@@ -68,7 +58,6 @@ jobs:
           echo "MLIR_DIR=$PWD/mlir" | tee -a $GITHUB_OUTPUT
 
       - name: Ccache for C++ compilation
-        # https://github.com/hendrikmuhs/ccache-action/releases/tag/v1.2.9
```
jsetoain, line 71: Why remove the comment? I don't see how knowing from which tag this is coming can hurt.

.github/workflows/buildAndTest.yml
```diff
@@ -8,13 +8,6 @@ on:
     types: [assigned, opened, synchronize, reopened]
   workflow_dispatch:
 
-defaults:
-  run:
-    # This is already the default, except when running inside another Docker
-    # image, which is the case here. So set it up globally to avoid
-    # repeating elsewhere.
-    shell: bash
-
```
jsetoain, line 17: Out of curiosity, is this no longer true? Was it never true?

.github/workflows/buildAndTest.yml
```diff
@@ -139,133 +128,225 @@ jobs:
...
-            echo "Clang-tidy patch"
-            echo "================"
-            cat clang-tidy.patch
-            echo "================"
-          fi
+          # FORMATTED=0
+          black --exclude python/compiler/aiecc/main.py . || true
+          black -l 10000 python/compiler/aiecc/main.py || true
+          # FORMATTED=$(git status -s -uno | wc -l | xargs)
+          # echo "FORMATTED=$FORMATTED" | tee -a $GITHUB_OUTPUT
```
jsetoain, line 337: Do these comments serve a purpose?

.pylintrc
```diff
@@ -0,0 +1,58 @@
...
+# on implicit string concatenation in sequences defined over several lines.
+check-str-concat-over-line-jumps=yes
+
+[MESSAGES CONTROL]
+
+# C0116: Missing function or method docstring (missing-function-docstring)
+# C0114: Missing module docstring (missing-module-docstring)
+# E0402: Attempted relative import beyond top-level package (relative-beyond-top-level)
+# C0115: Missing class docstring (missing-class-docstring)
+disable=C0116,C0114,E0402,C0115
```
jsetoain, line 58: Neat!

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -718,7 +718,7 @@ class MatMulOpConversion
   using ConvertOpToLLVMPattern<aievec::MatMulOp>::ConvertOpToLLVMPattern;
 
   struct DecodedMatMulOp {
-    typedef enum { I32, I64, BF16 } Kind;
+        typedef enum { I32, I64, BF16 } Kind;
```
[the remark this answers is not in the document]
jsetoain: What's going on with all this null reformatting?

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -776,25 +776,25 @@
...
-        }
-      }
-    }
+          if (accBitWidth == 32) {
+            if (lhsBitWidth == 8) {
+              if (rhsBitWidth == 4) {
+                // <4x16xi8> x <16x8xi4> + <4x8xi32>
+                return {DecodedMatMulOp::Kind::I32, lhs, rhs, acc, signConf};
+              } else {
+                // <4x8xi8> x <8x8xi8> + <4x8xi32>
```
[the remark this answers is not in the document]
jsetoain: Nop!

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -776,25 +776,25 @@
...
+                return {DecodedMatMulOp::Kind::I32, lhs, rhs, acc, signConf | 8};
+              }
+            } else {
+              if (rhsBitWidth == 8) {
+                // <4x4xi16> x <4x8xi8> + <4x8xi32>
+                return {DecodedMatMulOp::Kind::I32, lhs, rhs, acc, signConf | 16};
+              } else {
+                // <4x2xi16> x <2x8xi16> + <4x8xi32>
```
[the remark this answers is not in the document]
jsetoain: No!

utils/git/clang_tidy_pr.py
```diff
@@ -0,0 +1,626 @@
...
+            filter(
+                lambda review_comment: not (
+                    review_comment["path"]
+                    == comment["path"]  # pylint: disable=cell-var-from-loop
+                    and review_comment["line"]
+                    == comment["line"]  # pylint: disable=cell-var-from-loop
+                    and review_comment["side"]
+                    == comment["side"]  # pylint: disable=cell-var-from-loop
+                    and review_comment["body"]
+                    == comment["body"]  # pylint: disable=cell-var-from-loop
```
jsetoain, line 598: This reads awful. Can't you put the `pylint` hints over each line and have something like:
```
X == A
and Y == B
and Z == C
```
It would certainly read better. In fact, since it's the same warning to disable, can't you just put it right after the `for` line and disable the warning for the whole block?

jsetoain (commented):
I'm not super familiar with CI scripts nor all that auto-linting shenanigans, I'd let somebody else take a deeper look.

.clang-tidy
makslevental: There's a previous PR where I combined LLVM's and MLIR's clang-tidy styles and I messed it up with that straight quote.

.github/workflows/buildAndTest.yml
makslevental: The hash is right there in the name? It's completely redundant (you can just go to the commit indicated by the hash and see the tag).

.github/workflows/buildAndTest.yml
makslevental: Because I removed the containerization and so it is indeed now the default.

.clang-tidy
jsetoain: Oh! I see! For a second there I thought that line had been like that for a while and I was wondering how this was working at all 🤷

utils/git/clang_tidy_pr.py
makslevental: I didn't [write this](https://github.com/platisd/clang-tidy-pr-comments/blob/121c8f88ff614d196bbdfb15b641417423ec162f/run_action.py) so I don't really want to fix it (especially since pylint gives it a 10/10 score :p).

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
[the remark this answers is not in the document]
makslevental: That's why I made clang-tidy annotations only a suggestion...

.github/workflows/buildAndTest.yml
jsetoain: Maybe it's just me, but I think having a full URL (and without hashes!) is far more human readable than an encoded artefact. At the end of the day, it must be the same thing (obvs.), but as a non-expert I did appreciate the clarification.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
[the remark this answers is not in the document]
makslevental: These are spurious changes I made to the files in order to trigger the CI, which as the PR description states, is **touched lines only**.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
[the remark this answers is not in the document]
jsetoain: Yeah, I realised that after a while.

.github/workflows/buildAndTest.yml
makslevental: As someone that's been looking at the files a lot recently, the effect of many of these comments, mixed in with the yaml meta language and inlined shell scripts which themselves support hash comments is **most of them are noise**. I'd also argue that inline like this comments (i.e., in the source itself) aren't for "non-experts", i.e. explaining conventional things, but for "experts" and they should explain weird/unconventional things.

utils/git/clang_tidy_pr.py
jsetoain: As much as I appreciate them, I don't let auto-linters have the last word. After all, the whole point of having a linter is to make the code easier to read _for people_ ;-)

In any case, if this has been copied from somewhere else, then don't give it a second thought.

.github/workflows/buildAndTest.yml
jsetoain: Or! You just need to read the comment and know it's pointing to `releases/tag/v1.2.9` Zero steps :-)

makslevental: Does reading `releases/tag/v1.2.9` magically download the notes into your head? Or are you a maintainer/contributor there and you remember those notes? So it's not zero steps.

stephenneuendorffer (approved):

PR #828, opened 2023-12-05
disable pylint - too much noise

makslevental:


keryell (commented):
Since you did the great work, you could have just left it but just add a global condition to disable it. Of course there is a risk o bit-rotting if it is never used.

makslevental:
It's in the commit history (that's my philosophy on commented out dead code vs. removing) and I do plan restore it as soon I can reduce the barrage.

PR #834, opened 2023-12-06
Add a pass to normalize address spaces before lowering to standard ops.

MaheshRavishankar:
This takes care of handling memrefs with address space '2'.

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -33,8 +33,7 @@ import aie.compiler.aiecc.configure
 import rich.progress as progress
 import re
 
-aie_opt_passes = ['--aie-normalize-address-spaces',
-                  '--canonicalize',
+aie_opt_passes = ['--canonicalize',
                   '--cse',
                   '--convert-vector-to-llvm',
                   '--expand-strided-metadata',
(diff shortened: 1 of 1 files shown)


python/compiler/aiecc/main.py
```diff
@@ -694,6 +694,7 @@ async def run_flow(self):
         if(opts.unified):
           self.file_opt_with_addresses = os.path.join(self.tmpdirname, 'input_opt_with_addresses.mlir')
           await self.do_call(progress_bar.task, ['aie-opt', '--aie-localize-locks',
+                              '--aie-normalize-address-spaces',
```
stephenneuendorffer, line 697: This is already part of aie-opt-passes?

MaheshRavishankar: Yes, but it was suggested that this has to run before conversion to standard based on an error with handle address space 2

stephenneuendorffer: Thanks, After your patch, I went back and actually processed what Jeff had indicated.

stephenneuendorffer: It doesn't make sense to me to run this pass twice... so I factored it out of aie_opt_passes

fifield (approved):

makslevental:
@stephenneuendorffer can I merge?

PR #812, opened 2023-12-01
Enable 7940hs runner

fifield:
try switching workflow to 7940hs runner

PR #844, opened 2023-12-07
new API for instruction buffer creation

newling:
A vector based API:

```
std::vector<uint32_t> AIETranslateToIPU(mlir::ModuleOp);
```

is introduced. This is more useful to IREE than the existing API

```
mlir::LogicalResult AIETranslateToIPU(mlir::ModuleOp module, llvm::raw_ostream &output);
```

because the flatbuffer (flatbuffer = the file type which IREE will store the instructions in) can be generated directly from a container of uint32_t's. 

Actually this API isn't specific to IREE, it seems like a layer of copying could be removed from the AIR example with this API: https://github.com/Xilinx/mlir-air/blob/main/examples/air_to_ipu/test.cpp#L52-L65

Final API open to suggestions.

### DIFF
--- a/lib/Targets/AIETargetIPU.cpp
+++ b/lib/Targets/AIETargetIPU.cpp
@@ -15,39 +15,45 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Tools/mlir-translate/MlirTranslateMain.h"
 
+#include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/TypeSwitch.h"
 #include "llvm/Support/Format.h"
+#include <vector>
 
 using namespace mlir;
 using namespace xilinx;
 using namespace xilinx::AIE;
 using namespace xilinx::AIEX;
 
-static void emitProlog(llvm::raw_ostream &os) {
-  os << "00000011\n";
-  os << "01000405\n";
-  os << "01000100\n";
-  os << "0B590100\n";
-  os << "000055FF\n";
-  os << "00000001\n";
-  os << "00000010\n";
-  os << "314E5A5F\n";
-  os << "635F5F31\n";
-  os << "676E696C\n";
-  os << "39354E5F\n";
-  os << "6E693131\n";
-  os << "5F727473\n";
-  os << "64726F77\n";
-  os << "00004573\n";
-  os << "07BD9630\n";
-  os << "000055FF\n";
+namespace {
+
+std::vector<uint32_t> getProlog() {
+  return {0x00000011, 0x01000405, 0x01000100, 0x0B590100, 0x000055FF,
+          0x00000001, 0x00000010, 0x314E5A5F, 0x635F5F31, 0x676E696C,
+          0x39354E5F, 0x6E693131, 0x5F727473, 0x64726F77, 0x00004573,
+          0x07BD9630, 0x000055FF};
 }
 
-static void emitSync(raw_ostream &output, IpuSyncOp op) {
-  std::vector<uint32_t> words(2, 0);
+// Example:
+// - instructions = {3,4,5}
+// - tailSize = 2
+// instructions becomes {3,4,5,0,0} and
+// a mutable reference to the tail {0,0} is returned.
+llvm::MutableArrayRef<uint32_t>
+reserveAndGetTail(std::vector<uint32_t> &instructions, uint64_t tailSize) {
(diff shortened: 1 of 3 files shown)


newling:
Is this code being tested in CI? I guess I could intentionally break it and see what happens.

makslevental:
> Is this code being tested in CI? I guess I could intentionally break it and see what happens.

I'm not sure actually - I have some sense that it's waiting on eg https://github.com/Xilinx/mlir-aie/pull/812 ? But maybe I'm wrong. Can you actually try to do that as a smoke test and then also I/we can run the "aie tools" [suite of tests](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestAieTools.yml) on it.

fifield:
> > Is this code being tested in CI? I guess I could intentionally break it and see what happens.
> 
> I'm not sure actually - I have some sense that it's waiting on eg #812 ? But maybe I'm wrong. Can you actually try to do that as a smoke test and then also I/we can run the "aie tools" [suite of tests](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestAieTools.yml) on it.
> 
> Just ping me here when you've "broken" it and I'll run those tests.

It would be good to wait for #812 because this will silently break any ipu-xrt test or example otherwise. All host code for those tests will need to be updated.

newling:
Ok, I've intentionally broken the returned the sequence of uint32_ts, let's see what happens.

newling:
> Ok, I've intentionally broken the returned the sequence of uint32_ts, let's see what happens.
It didn't break. This should have a unit test, shouldn't need to use HW or anything fancy to fail. I'll try and add one. 

@makslevental not sure I follow, you mean change something like to start additional testing? https://github.com/Xilinx/mlir-aie/pull/845

makslevental:
> @makslevental not sure I follow, you mean change something like to start additional testing? #845

that's the right thing but you can just do it right here in this PR and it'll apply to this PR.

lib/Targets/AIETargetIPU.cpp
```diff
@@ -15,71 +15,76 @@
...
-  os << "39354E5F\n";
-  os << "6E693131\n";
-  os << "5F727473\n";
-  os << "64726F77\n";
-  os << "00004573\n";
-  os << "07BD9630\n";
-  os << "000055FF\n";
+namespace {
+
+std::vector<uint32_t> getProlog() {
```
jgmelber, line 30: Just leaving this as a comment. I'd like to be able to remove this, but haven't had time to dig into it. @STFleming was also looking at removing it

newling:
> > @makslevental not sure I follow, you mean change something like to start additional testing? #845
> 
> that's the right thing but you can just do it right here in this PR and it'll apply to this PR.

Thanks. It's still very fast through CI -- if it's just running additional tests, should it maybe be on by default?

makslevental:
> Thanks. It's still very fast through CI -- if it's just running additional tests, should it maybe be on by default?

Hmm it didn't [run](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestAieTools.yml) for the other PR even though you enabled triggering with a pull request. Possibly this is because it needs CI secrets? 

> should it maybe be on by default

It should I agree but it takes too long (~2 hours) because it runs chess tests.

> So I'd like to add a unit test. Does that make sense?

I'd definitely appreciate that but since I'm not familiar with this part of the codebase yet I can't advise on how exactly.

makslevental:
> What's a CI secret, how do I get one? I can also just wait for #812 if that's likely to land in the next day.

uh it's kind of [don't ask don't tell](https://github.com/Xilinx/mlir-aie/blob/main/.github/workflows/buildAndTestAieTools.yml#L58-L64) lol. Really I just haven't figured out how to share secrets with workflows that aren't branches of the main repo. Supposedly it's [possible](https://stackoverflow.com/a/75910144).

> I'll investigate

:+1:

lib/Targets/AIETargetIPU.cpp
jgmelber: Yes just this prolog bit. I am not sure it is necessary but there's some testing to do here, it won't work if it is just deleted.

makslevental:
@newling what's the first comment? You mean the PR description?

jgmelber (approved):

newling:
> @newling what's the first comment? You mean the PR description?

I guess. The one that starts "A vector based API:"

PR #859, opened 2023-12-12
One column Matrix Multiplication for Ryzen AI

jgmelber:


jgmelber:
@andrej

reference_designs/ipu-xrt/matrix_multiplication_column/Makefile
```diff
@@ -0,0 +1,69 @@
...
+	word_size_out?=${word_size_in}
+endif
+ifeq ($(dtype_in), bf16)
+	word_size_in?=2
+	r?=4
+	s?=8
+	t?=4
+	dtype_out?=f32
+	word_size_out?=4
+endif
```
andrej, line 35: I believe lines 11 to 35 could be removed to avoid confusion as I don't see these environment variables used anywhere anymore. Unless you plan to adjust the Python script to take them as input arguments again and want to keep those defaults, which would be cool.

reference_designs/ipu-xrt/matrix_multiplication_column/README.md
```diff
@@ -0,0 +1,24 @@
...
+//
+// Copyright (C) 2022, Advanced Micro Devices, Inc.
+// 
+//===----------------------------------------------------------------------===//-->
+
+# <ins>Matrix Multiplication Column</ins>
+
+Multiple tiles in a single column perform a `matrix * matrix` multiply on int16 data type where `MxKxN` is `256x128x128`. The kernel itself computes `64x32x64 (MxKxN)` so it is invoked multiple times to complete the full matmul compute.
+
+To compile desing in Windows:
```
andrej, line 15: design

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
```diff
@@ -0,0 +1,246 @@
...
+    m = 64
+    k = 32
+    n = 64
+    r = 4
+    s = 4
+    t = 4
+    word_size_in = 2
+    word_size_out = 2
+
+    n_cores = 4
```
andrej, line 31: It could be cool to make these variables (line 19 - 31) command line arguments to the script down the line at some point to make it easy to experiment with different sizes.

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
```diff
@@ -0,0 +1,246 @@
...
+    n_in_i32s = n * word_size_in // 4
+    N_in_i32s = N * word_size_in // 4
+    k_x_N_in_i32s = k * N * word_size_in // 4
+    N_div_n_div_n_cores  = N // n // n_cores
+    n_x_n_cores_in_i32s  = n * n_cores * word_size_in // 4
+
+    # Output Matrix C: MxN
+    n_in_i32s_out = n * word_size_out // 4
+    N_in_i32s_out = N * word_size_out // 4
+    m_x_n_cores_x_N_in_i32s_out = m * n_cores * N * word_size_out // 4
```
andrej, line 59: These variables were necessary in the templating approach, but maybe less so in the Python script, since some of the variable names are almost as long as the expression itself... I'm also not sure if they are all still used

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
```diff
@@ -0,0 +1,246 @@
...
+                    2,
+                    ofifo_memRef_A_ty,
+                    [
+                        (m // r, r * k * word_size_in // 4),
+                        (k // s, s * word_size_in // 4),
+                        (r, k * word_size_in // 4),
+                        (s * word_size_in // 4, 1),
+                    ],
+                    [],
+                )
```
andrej, line 121: I like the formatting, this would be an awfully long line without it 👍🏻

reference_designs/ipu-xrt/matrix_multiplication_column/mm.cc
```diff
@@ -0,0 +1,216 @@
...
+  X(int16, i16, int16, i16, 4, 4, 4)                                           \
+  X(bfloat16, bf16, bfloat16, bf16, 4, 8, 4)                                   \
+  X(bfloat16, bf16, float, f32, 4, 8, 4)
+
+#define matmul_vectorized_c_func(ctype_in, mlir_type_in, ctype_out,            \
+                                 mlir_type_out, r, s, t)                       \
+  void matmul_##mlir_type_in##_##mlir_type_out(ctype_in *a_in, ctype_in *b_in, \
+                                               ctype_out *c_out) {             \
+    matmul_vectorized_##r##x##s##x##t##_##mlir_type_in##_##mlir_type_out<      \
+        64, 32, 64>(a_in, b_in, c_out);                                        \
```
andrej, line 191: I know we probably don't want to use templating any more, and we don't have any way to generate the C code from Python easily, but are there any alternatives to hard-coding the 64x32x64 size here? Maybe, we could at least make it an argument to the `combos()` macro, so it would be easy to generate a bunch of different tile sizes by adding lines to the same macro?

Something like:
```
#define combos(X)                                                              \
  X(int16, i16, int16, i16, 4, 4, 4, 64, 32, 64)                                           \
  X(bfloat16, bf16, bfloat16, bf16, 4, 8, 4, 64, 32, 64)                                   \
  X(bfloat16, bf16, float, f32, 4, 8, 4, 64, 32, 64)
```

(Then you could generate all necessary functions for another tile size by just adding a line below that with differing last three arguments.)

and then here (added M, N, K to function):
```
#define matmul_vectorized_c_func(ctype_in, mlir_type_in, ctype_out,            \
                                 mlir_type_out, r, s, t, M, N, K)                       \
  void matmul_##mlir_type_in##_##mlir_type_out(ctype_in *a_in, ctype_in *b_in, \
                                               ctype_out *c_out) {             \
    matmul_vectorized_##r##x##s##x##t##_##mlir_type_in##_##mlir_type_out<      \
        M, N, K>(a_in, b_in, c_out);  
```

and analogously for the scalar and zeroing functions below.

Just a thought ... hard-coding the size here just seems like a bit of a regression from how easy it was to experiment with different sizes before.

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
andrej: This would also require the C kernel to be generated with different tile sizes (or support a select set of sizes that are always generated), so maybe out of scope.

andrej (commented):
It's really cool to see this in Python, Joe. Much cleaner than with the templates. I know you didn't request a review from me but I hope you don't mind that I left some comments. Feel free to ignore them. The comment I added to the C kernel file is the only one I feel strongly about.

makslevental:
@jgmelber n00b question: is it not possible to write the kernel code (stuff in `mm.cc`) in python as well? E.g, doesn't `aievec` exist and have a lowering to cpp that can emit roughly the same kind of stuff? Or are there lots of things missing from the `aievec` dialect that you therefore (today) have to write by hand (in cpp).

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
AndraBisca: A few of these were indeed not used so I removed them. I didn't change the variable names.

AndraBisca:
Thank you for all of your comments @andrej, they are really helpful!
While I didn't resolve all of them, I think it might be best to merge this PR for now to avoid accumulating conflicts with main. @jgmelber can make any subsequent changes in another update-PR.

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
```diff
@@ -0,0 +1,261 @@
...
+# (c) Copyright 2023 AMD Inc.
+
+import aie
+from aie.ir import *
+from aie.dialects.func import *
+from aie.dialects.scf import *
+from aie.dialects.aie import *
+from aie.dialects.aiex import *
+from aie.dialects.extras import memref, arith
+from aie.util import mlir_mod_ctx
```
makslevental, line 15:
```suggestion
from aie.extras.dialects.ext import memref, arith
from aie.extras.context import mlir_mod_ctx
```

makslevental: @fifield is it safe to add `reference-designs` to the ryzenai CI flow?

fifield: > @fifield is it safe to add `reference-designs/` to the ryzenai CI flow?

yes. I thought it already was, but maybe not.

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -75,3 +75,4 @@ jobs:
 
             sed -i.bak 's/-sv --timeout 600/-sv --timeout 600 -j12 --show-unsupported/g' $PWD/../test/CMakeLists.txt
             ninja check-aie
+            ninja reference_designs
```
fifield, line 78: `ninja check-reference-designs` ?

I've only ever run the reference designs with `lit` directly. I just tried via ninja and get unresolved tests and errors...

makslevental: > `ninja check-reference-designs` ?

uh whoops

> I've only ever run the reference designs with `lit` directly.

TIL! How do you run with lit directly?

fifield: > TIL! How do you run with lit directly?

I do one of these:
```
$ pip install lit
$ cd <mlir-aie build dir>
$ lit [options] reference_deisgns/ipu-xrt
```
or
```
$ cd <mlir-aie build dir>
$ python <llvm build dir>/bin/llvm-lit [options] ...
```

makslevental:
Heads up @AndraBisca I marked [reference_designs/autocorrelation/objectFifo_version/aie.mlir](https://github.com/Xilinx/mlir-aie/blob/a48f16ec68c7dccd3bc40e6cf906b3a65db68e4a/reference_designs/autocorrelation/objectFifo_version/aie.mlir) as XFAIL because it uses old objectfifo syntax that I can't quite patch up.

AndraBisca:
> Heads up @AndraBisca I marked [reference_designs/autocorrelation/objectFifo_version/aie.mlir](https://github.com/Xilinx/mlir-aie/blob/a48f16ec68c7dccd3bc40e6cf906b3a65db68e4a/reference_designs/autocorrelation/objectFifo_version/aie.mlir) as XFAIL because it uses old objectfifo syntax that I can't quite patch up.

Thank you for the heads-up @makslevental. I'll take a look at it very soon, along with all the other reference designs and clean them up.

PR #869, opened 2023-12-15
Fixed edge detect python src and expanded default sizes to 1080p

jackl-xilinx:


fifield (approved):
edge_detect works for me @ 1920x1080 after these changes.

Can you also enable a larger size in edge_detect run.lit ?

PR #877, opened 2023-12-16
rename remaining aie ops

makslevental:
This PR normalizes the remainder of the op names to match upstream convention (snake_case).


stephenneuendorffer (commented):
This is a somewhat painful change, but it's probably better to do it sooner rather than later.

makslevental:
> This is a somewhat painful change, but it's probably better to do it sooner rather than later.

FYI this change "simplifies" the python bindings in that we will no longer need to wrap miscased things there to achieve proper casing.

keryell:
And there is the elephant in the room: `AIE` → `aie` :-)

makslevental:
> And there is the elephant in the room: `AIE` → `aie` :-)

I would definitely like to make this change as well!

PR #878, opened 2023-12-18
add mlir-python-extras as dep

makslevental:
This PR factors out the dependency on [mlir-python-extras](https://github.com/makslevental/mlir-python-extras) as true dependency rather than copy-paste dependency. That means `region_op` is now being gotten from there (but soon will be moved fully into upstream https://github.com/llvm/llvm-project/pull/75673). In addition it means we now have access to [tensor and memref indexing/slicing](https://github.com/makslevental/mlir-python-extras/blob/main/tests/test_tensor.py#L29).


makslevental:
TODO: move to submodule

AndraBisca (approved):
Looks very good to me! I am also glad to hear that there are plans in the very near future to move the region_op functionality to upstream! :)

makslevental:
> Looks very good to me! I am also glad to hear that there are plans in the very near future to move the region_op functionality to upstream! :)

Slowly but surely I'm upstreaming everyting in mlir-python-extras :)

PR #879, opened 2023-12-20
Clean up AIE attrs

makslevental:
This PR normalizes/cleans up some things about AIE Attrs.

1. `DMABDOp::AB` because an actual enum `DMABDBuffer` instead of a `ConfinedAttr[0,1]`
2. `DimTuple` becomes the more semantically legible `BDDimLayout` with struct parsing `<wrap = 2, step = 128>` instead of tuple parsing. `DimTupleArray` and `DimTupleArrayArray` are renamed correspondingly.


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -808,9 +808,8 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
     ins AnyMemRef:$buffer,
         AIEI32Attr:$offset,
         AIEI32Attr:$len,
-        // TODO(max): this should be an enum
-        ConfinedAttr<AIEI32Attr, [IntMinValue<0>, IntMaxValue<1>]>:$AB, // 0: A, 1: B
-        OptionalAttr<AIE_DimTupleArrayAttr>:$dimensions
+        DefaultValuedAttr<DMABDBuffer, "xilinx::AIE::DMABDBuffer::A">:$AB,
```
stephenneuendorffer, line 811: It turns out that the A/B option is an AIE1 only feature... We've also never really leveraged it.  Since this requires touching every test, it seems like we can take the opportunity to make this optional in the syntax so we wouldn't need to touch it again?  I'm curious what others think?

makslevental: This does effectively make it optional (and sets to `A`). Fancier than this (i.e., actually moving to `OptionalAttr`) and you will have to change logic. I'm for it of course but one hurdle at a time please.

PR #881, opened 2023-12-20
remove compiler specific warning pragmas

makslevental:


include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
```diff
@@ -79,10 +79,6 @@ using SwitchboxNode = struct SwitchboxNode : SwitchboxNodeBase, Switchbox {
   int id;
 };
 
-// warning: 'xilinx::AIE::ChannelEdge::src' will be initialized after
-// SwitchboxNode &src; [-Wreorder]
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wreorder"
```
stephenneuendorffer, line 85: Personally, I think handling the diagnositcs properly is preferable to having noisy warnings..  That said, I don't understand what this is actually warning about...

CMakeLists.txt
```diff
@@ -163,6 +163,7 @@ cmake_dependent_option(AIECC_LINK_WITH_XCHESSCC
 
 
 add_flag_if_supported("-Werror=sign-compare" WERROR_SIGN_COMPARE)
+add_flag_if_supported("-Werror=switch-enum" WERROR_SWITCH_ENUM)
```
stephenneuendorffer, line 166: That's much better!

include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
makslevental: Down below there's some juggling of llvm/ADT requirements and our own (how a `ChannelEdge` is created most often). I'm the one that put this here and I'm taking it out because I prefer the noisy warning (which is at least precise) to the noisy warning about ~unsupported diagnostics when compiling using clang.

CMakeLists.txt
makslevental: It doesn't work - upstream has a failing case https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/Twine.h#L443-L451. I plan to submit a patch but who knows what they'll say.

PR #889, opened 2023-12-22
Generate cdo using python

makslevental:


fifield:
I have an alternate branch where I rewrote AIETargetCDO.cpp to call the libxaie api directly, so that `aie-translate` generates the cdo's and I could delete the generated-source directory, calls to clang, etc. It's already passing the e2e tests. Is that the intent here, but "in python"?

makslevental:
> I have an alternate branch where I rewrote AIETargetCDO.cpp to call the libxaie api directly, so that `aie-translate` generates the cdo's and I could delete the generated-source directory, calls to clang, etc. It's already passing the e2e tests. Is that the intent here, but "in python"?

That's indeed the intent - I'm not married to ctypes. If you PR that as a branch on this repo I'll happily patch up whatever's missing.

PR #908, opened 2024-01-03
Adding the full LLVM hash to ./utils/clone-llvm.sh and then slicing it for the WHEEL_VERSION

eddierichter-amd:
Without this change, just running `./utils/clone-llvm.sh` fails because `git fetch --depth=1` requires a full commit hash.

### DIFF
--- a/utils/clone-llvm.sh
+++ b/utils/clone-llvm.sh
@@ -13,9 +13,9 @@
 ##===----------------------------------------------------------------------===##
 
 # The LLVM commit to use.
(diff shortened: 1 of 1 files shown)


makslevental (approved):
LGTM

makslevental:
But I'll also point out that you could just save yourself the hassle entirely and download a pre-backed distro at exactly that commit https://github.com/Xilinx/mlir-aie/releases/tag/mlir-distro ;p

PR #910, opened 2024-01-04
generate cdo directly by binding libcdo_driver

makslevental:
This PR implements a "direct" CDO generation utility - ie direct bindings and calls to `libcdo_driver.so` (instead of emitting C++, compiling+linking against `libcdo_driver.so`, and then running).

It works by building `libxaienginecdo` _and statically linking it_. The point of statically linking is to minimize the number of floating libs that need to be in the environment in order for things to work. In addition `libcdo_driver.so` itself (which `libxaienginecdo` depends on) is "re-exported" to `${CMAKE_INSTALL_PREFIX}/lib` using cmake's `install(IMPORTED_RUNTIME_ARTIFACTS ...)`. As far as I can there are no other (non-ubiquitous) runtime dependencies. 

The relevant source file is [lib/Targets/AIETargetCDODirect.cpp](https://github.com/Xilinx/mlir-aie/pull/910/files#diff-3d10bbae3356dcf7f7c77717bf8e5ef326a44af53151d51d70ec5b3bb9c0bb42)

Currently there's only a smoke test (through python bindings) but I plan to basically dump all currently generated bins (generating using the other way) and diff against what's producing with this path.

cc @MaheshRavishankar


lib/Targets/CMakeLists.txt
```diff
@@ -19,6 +21,28 @@ add_mlir_library(AIETargets
...
+  add_aiert_headers(xaienginecdo
+    ${XAIE_SOURCE}
+    ${AIE_BINARY_DIR}/include
+    ${CMAKE_INSTALL_PREFIX}/include/xaiengine)
+  add_aiert_library(xaienginecdo ${XAIE_SOURCE} STATIC)
+  target_compile_options(xaienginecdo PRIVATE "-Wimplicit-function-declaration")
+  target_compile_definitions(xaienginecdo PRIVATE -D__AIECDO__)
+  install(TARGETS xaienginecdo DESTINATION ${CMAKE_INSTALL_PREFIX}/lib EXPORT xaienginecdo)
+  # just to make cmake happy since AIETargets will need to re-export
+  install(EXPORT xaienginecdo DESTINATION ${CMAKE_INSTALL_PREFIX}/cmake)
```
stephenneuendorffer, line 41: Why compile this again?

makslevental: Compile it as a static archive so that it's statically linked in.

stephenneuendorffer: If you want a static library, then we can probably just change the version that's already compiled to be a static library.  regardless, I don't think it makes sense to compile this in a completely different directory than where the source is.

makslevental: Sure I agree but the way the build is setup right now, with `ExternalProject`, you don't get a cmake target to link against.

include/aie/Targets/cdo_driver.h
```diff
@@ -0,0 +1,52 @@
+// 67d7842dbbe25473c3c32b93c0da8047785f30d78e8a024de1b57352245f9689
+/* cdo_driver.h  */
+#ifndef _CDO_DRIVER_H_
+#define _CDO_DRIVER_H_
```
stephenneuendorffer, line 4: It seems weird to have these Decls in our code.  Where did this come from?

CMakeLists.txt
```diff
@@ -154,6 +154,12 @@ cmake_dependent_option(AIE_ENABLE_PYTHON_PASSES
 cmake_dependent_option(AIE_ENABLE_XRT_PYTHON_BINDINGS
   "Enables building of python bindings to XRT runtime." ON "XRT_COREUTIL" OFF)
 
+find_library(CDO_DRIVER_LIBRARY cdo_driver
+  HINTS "${VITIS_AIETOOLS_DIR}/lib/lnx64.o"
+)
+cmake_dependent_option(AIE_ENABLE_GENERATE_CDO_DIRECT
```
stephenneuendorffer, line 160: Also depends on libxaie?

makslevental: What's libxaie? After a successful compile of `aie-translate` this is all I have 

```
mlevental@mlevental-CORSAIR-ONE-PRO-a200:~/dev_projects/mlir-aie/cmake-build-debug$ ldd ./bin/aie-translate 
        linux-vdso.so.1 (0x00007ffe471bf000)
        libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f548f735000)
        libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007f548cd31000)
        libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007f548f703000)
        libcdo_driver.so => /media/mlevental/9b03e9bc-780d-4e14-9b46-ec8089db51a6/Xilinx/Vitis/2023.2/aietools/lib/lnx64.o/libcdo_driver.so (0x00007f548f6fd000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f548ca00000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f548cc4a000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f548c9e0000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f548c600000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f548f775000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f548f6f6000)
```

include/aie/Targets/cdo_driver.h
makslevental: No clue - I wasn't sure if they were (or weren't) used before I started but now I can see it's safe to delete all except `byte_ordering`.

PR #911, opened 2024-01-04
Fixup MLIR target export in AIEConfig.cmake

stephenneuendorffer:
MLIR_CONFIG_INCLUDE_EXPORTS_FILE was not generated for a build directory, which caused errors when this was being used.


cmake/modules/CMakeLists.txt
```diff
@@ -87,10 +88,6 @@ install(
   EXPORT AIETargets
   DESTINATION ${AIE_INSTALL_PACKAGE_DIR}
   COMPONENT aie-cmake-exports)
-install(
```
makslevental, line 90: Transcribing offline discussion: this is necessary to support downstream users because of how `add_mlir_library` works.

makslevental (changes requested):
Let me take a closer look at the bug itself.

hunhoffe:
Reviewing against current main: none of these three changes are present anymore and the patch no longer applies cleanly (context shifted after #960).
  
I don't think the underlying issue is still present, although the build tree still has empty include("") lines, but those are silent no-ops in CMake, so I think that's cosmetic.

Closing as stale/no longer applicable.

PR #912, opened 2024-01-05
Reimplement IREE target code as a C++ tool

spaceotter:
The goal here is to provide some of the functions that were ported from aiecc.py to the IREE plugin as a standalone tool which will be called from the IREE plugin instead of the current arrangement. This will provide an abstracted interface so that IREE doesn't have to keep getting updated as the build process is reworked.


tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,862 @@
...
+  std::string errorMessage;
+
+  for (auto tileOp : tileOps) {
+    int col = tileOp.colIndex();
+    int row = tileOp.rowIndex();
+    auto coreOp = tileOp.getCoreOp();
+    if (!coreOp) {
+      continue;
+    }
+    auto fileAttr = coreOp->getAttrOfType<StringAttr>("elf_file");
```
stephenneuendorffer, line 142: Move inside the if?

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,862 @@
...
+  if (aie_target == "AIE2") {
+    Args.push_back("-D__AIEARCH__=20");
+  } else {
+    Args.push_back("-D__AIEARCH__=10");
+  }
+}
+
+// Generate the elf files for the core
+LogicalResult generateCoreElfFiles(ModuleOp moduleOp, sfs::path &objFile, Toolkit &TK) {
+  auto deviceOps = moduleOp.getOps<xilinx::AIE::DeviceOp>();
```
stephenneuendorffer, line 125: Probably a 'using namespace xilinx::AIE' is reasonable in this code.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,862 @@
...
+
+struct Toolkit {
+  std::string TargetArch;
+  sfs::path PeanoDir;
+  sfs::path InstallDir;
+  sfs::path AIEToolsDir;
+  sfs::path TempDir;
+};
+
+int runTool(StringRef Program, ArrayRef<StringRef> Args,
```
stephenneuendorffer, line 86: It seems like the rest of the code is significantly complicated by having this take StringRefs...

tools/aiecc/aiecc.cpp
makslevental: Also you can just `getElfFile()`.

And for the record (not personally @ you @spaceotter), this same thing was elsewhere in the codebase and I removed it recently. I would like to never see this again - we define attrs as `ins` params in order to exactly avoid string matching.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,862 @@
...
+#include "configure.h"
+#include "llvm/Support/JSON.h"
+
+using namespace llvm;
+using namespace mlir;
+namespace sfs = std::filesystem;
+
+cl::opt<std::string> FileName(cl::Positional, cl::desc("<input mlir>"), cl::Required);
+cl::opt<std::string> TmpDir("tmpdir", cl::desc("Directory used for temporary file storage"));
+cl::opt<std::string> SysRoot("sysroot", cl::desc("sysroot for cross-compilation"));
```
makslevental, line 39: All of these need to be put under a `cl::opt::Category` so they pollute the args list. [Random example from upstream](https://github.com/llvm/llvm-project/blob/ea668144d9c6d93fbe2ea7827bd8e8cbaf4c50dd/llvm/tools/llvm-reduce/ReducerWorkItem.cpp#L51-L54).

tools/aiecc/aiecc.cpp
spaceotter: Mahesh did it this way - it did occur to me that getElfFile might exist, then IntelliSense couldn't find it so I kept going...

makslevental: It indeed didn't exist until I noticed this kind of thing was happening (reliance on an arbitrarily inserted attribute) and so I added it. Like I said not your fault just that I would like to avoid anything like this in the future.

spaceotter:
This produces an xclbin now. I will address your comments soon. A bunch of cleanup is needed.

tools/aiecc/aiecc.cpp
spaceotter: That's what is accepted by the LLVM helper function I'm using.

erieaton-amd: I changed it to std::strings.

fifield:
> The goal here is to provide some of the functions that were ported from aiecc.py to the IREE plugin as a standalone tool which will be called from the IREE plugin instead of the current arrangement. This will provide an abstracted interface so that IREE doesn't have to keep getting updated as the build process is reworked.

So we end up maintaining two standalone tools?  Why not call aiecc.py?

stephenneuendorffer:
> > The goal here is to provide some of the functions that were ported from aiecc.py to the IREE plugin as a standalone tool which will be called from the IREE plugin instead of the current arrangement. This will provide an abstracted interface so that IREE doesn't have to keep getting updated as the build process is reworked.
> 
> So we end up maintaining two standalone tools? Why not call aiecc.py?

Because python 'can't' exist in an onnx execution provider.  I think the better path to reduce the duplicated code paths is for aiecc.py to get simpler and to call into c++ bits more.

makslevental:
> we shouldn't just call into c++ bits more for xclbin creation, we should reimplement all of compiler/aiecc/main.py in a c++ library. Then aiecc.py or aiecc.exe or anyone else just uses the C api and we can have a python free build option.

I already did some of this 

https://github.com/Xilinx/mlir-aie/blob/main/lib/CAPI/Translation.cpp

including binding xrt (though not through a C API).

 https://github.com/Xilinx/mlir-aie/blob/main/python/XRTModule.cpp

tools/aie2xclbin/aie2xclbin.cpp
```diff
@@ -0,0 +1,162 @@
...
+// (c) Copyright 2024 Xilinx Inc.
+//
+//===---------------------------------------------------------------------===//
+
+#include "XCLBinGen.h"
+#include "aie/Dialect/AIE/Transforms/AIEPasses.h"
+#include "aie/Dialect/AIEX/Transforms/AIEXPasses.h"
+#include "aie/InitialAllDialect.h"
+#include "aie/Targets/AIETargets.h"
+#include "configure.h"
```
makslevental, line 16: nit: can you shuffle this to the top

tools/aie2xclbin/aie2xclbin.cpp
```diff
@@ -0,0 +1,162 @@
...
+#include "llvm/Support/Program.h"
+#include "llvm/Support/SourceMgr.h"
+#include <iostream>
+#include <regex>
+#include <stdlib.h>
+#include <string>
+
+using namespace llvm;
+using namespace mlir;
+namespace sfs = std::filesystem;
```
makslevental, line 43: so what's the verdict on `std::filesystem` vs `llvm/Support/Filesystem`?

makslevental (commented):
I think overall this is fine - I'm basically fine approving but can you add some tests? There are a lot of paths (obviously) so I'm not expecting full coverage but just _something_.

tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -0,0 +1,763 @@
...
+    SmallVector<std::string> flags;
+    flags.push_back("--work-dir-path");
+    // CDO tool doesn't join paths correctly :/
+    std::string workDir = TK.TempDir + "/";
+    flags.push_back(workDir);
+    if (runTool(cdoBinary, flags, TK.Verbose,
+                ArrayRef<StringRef>{libraryPath})) {
+      return moduleOp.emitOpError("failed to execute cdo_main binary");
+    }
+  }
```
makslevental, line 411: so you can pretty safely now replace this with just calls to [AIETranslateToCDODirect](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetCDODirect.cpp#L723).

tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -0,0 +1,763 @@
...
+
+  copy->erase();
+  return success();
+}
```
makslevental, line 415: so you can pretty safely now replace this with just calls to [AIETranslateToCDODirect](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetCDODirect.cpp#L723).

tools/aie2xclbin/aie2xclbin.cpp
spaceotter: I need to remove all usage of std::filesystem.

tools/aie2xclbin/aie2xclbin.cpp
spaceotter: This is how clang-format sorts the includes, I think if I change the order it will fail the linter.

makslevental: clang-format sorts blocks of includes separated by a newline so in fact i was going to ask for 

```cpp
#include "configure.h"
#include "XCLBinGen.h"

#include "aie/Dialect/AIE/Transforms/AIEPasses.h"
#include "aie/Dialect/AIEX/Transforms/AIEXPasses.h"
#include "aie/InitialAllDialect.h"
#include "aie/Targets/AIETargets.h"

#include "mlir/Dialect/Affine/Passes.h"
#include "mlir/Dialect/Arith/IR/Arith.h"
#include "mlir/Dialect/MemRef/IR/MemRef.h"
#include "mlir/IR/BuiltinOps.h"
#include "mlir/IR/MLIRContext.h"
#include "mlir/IR/OwningOpRef.h"
#include "mlir/Parser/Parser.h"
#include "mlir/Pass/PassManager.h"
#include "mlir/Target/LLVMIR/Dialect/Builtin/BuiltinToLLVMIRTranslation.h"
#include "mlir/Target/LLVMIR/Dialect/LLVMIR/LLVMToLLVMIRTranslation.h"
#include "mlir/Target/LLVMIR/Export.h"
#include "mlir/Tools/mlir-translate/Translation.h"

#include "llvm/ADT/SmallString.h"
#include "llvm/Support/CommandLine.h"
#include "llvm/Support/FileSystem.h"
#include "llvm/Support/JSON.h"
#include "llvm/Support/Path.h"
#include "llvm/Support/Program.h"
#include "llvm/Support/SourceMgr.h"

#include <iostream>
#include <regex>
#include <stdlib.h>
#include <string>
```

but didn't want to be too nitpicky :)

tools/aie2xclbin/aie2xclbin.cpp
spaceotter: The problem is that on windows the std::filesystem uses a wide char type internally, which breaks interop with most string stuff. There was a discussion about this on discord and we decided to just avoid std::filesystem.

keryell: That makes sense to use the LLVM solution then. But this should appear in some comment explaining why because it is not obvious to the reader.

PR #914, opened 2024-01-06
disable minis runner

makslevental:
@fifield the minis runner is out of disk space. disabling for now.


fifield:
> @fifield the minis runner is out of disk space. disabling for now.

It seems to have been the pip cache of the github user account that filled the partition. I ran `pip cache purge`  which deleted 273G.

PR #916, opened 2024-01-07
remove `dut.cc` from Integration/TOSA tests

makslevental:
Not sure why these are checked in - they are never touched by the tests as they're in the source directory and they are generated by the tests themselves. Note, there's no runner that tests these (since the minis is down) but I tested locally.


jsetoain:
These have been pushed as a reference of the output the tests should generate. You can see them as documentation of sorts. They are never tested, and consistency is neither checked nor enforced (i.e.: if somebody changes the test and the `dut.cc` is not updated accordingly, that won't trigger any error; as far as we know, this could have happened already), if something goes wrong, the only sure way to know what's happening is to actually generate `dut.cc` on the spot, so I truly see no point in having them there. Also, I'm not entirely sure what would happen if the generation of the `dut.cc` fails, but I suspect that in some circumstances it might compile the checked-in `dut.cc` and pass the test.

I've raised this issue several times since I noticed them, asking permission to remove them and/or trying to block the continued uploading of these files, and I've been overruled every time.

makslevental:
> These have been pushed as a reference of the output the tests should generate. You can see them as documentation of sorts.

But they're not canonical and shouldn't be? In this use (as a translation medium between mlir-aie and chess), they function as (effectively) IR. The MLIR is the canonical representation of the kernel.

> but I suspect that in some circumstances it might compile the checked-in `dut.cc` and pass the test.

No that's definitely not the case because that would require referencing them by `%S/dut.cc` in the `// RUN: ...` directives. But that doesn't mean it's not very confusing for anyone stumbling through that directory :)

> so I truly see no point in having them there... I've raised this issue several times since I noticed them, asking permission to remove them and/or trying to block the continued uploading of these files, and I've been overruled every time.

Two is a sufficient quorum in my book :).

PR #917, opened 2024-01-07
Unified `DMAOp`

makslevental:
This PR introduces `aie.dma` which is a single region (single block) op that encompasses both `aie.dma_start` and then dma instructions (such as `use_lock` and `dma_bd`). It looks like this:

```mlir
      aie.dma(S2MM, 0) {
        aie.use_lock(%objFifo_in0_cons_prod_lock, AcquireGreaterEqual, 1)
        aie.dma_bd(%objFifo_in0_cons_buff_0 : memref<16xi32>, 0, 16)
        aie.use_lock(%objFifo_in0_cons_cons_lock, Release, 1)
        aie.use_lock(%objFifo_in0_cons_prod_lock, AcquireGreaterEqual, 1)
        aie.dma_bd(%objFifo_in0_cons_buff_1 : memref<16xi32>, 0, 16)
        aie.use_lock(%objFifo_in0_cons_cons_lock, Release, 1)
      }
```

This PR also uses the previous PR https://github.com/Xilinx/mlir-aie/pull/913 to lower/translate these ops to the CDO backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,15 @@
 # MLIR-based AI Engine toolchain
 
-![GitHub Workflow Status](https://img.shields.io/github/actions/workflow/status/Xilinx/mlir-aie/buildAndTest.yml?branch=main)
+[![Build and Test](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTest.yml/badge.svg)](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTest.yml)
+
+[![Build and Test across Python versions](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestPythons.yml/badge.svg)](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestPythons.yml)
+
+[![Build and Test with AIE tools](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestAieTools.yml/badge.svg)](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestAieTools.yml)
+
+[![Build and Test with AIE tools on Ryzen AI](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestRyzenAI.yml/badge.svg)](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestRyzenAI.yml)
+
+[![Compile across platforms](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestMulti.yml/badge.svg)](https://github.com/Xilinx/mlir-aie/actions/workflows/buildAndTestMulti.yml)
+
 ![GitHub Pull Requests](https://img.shields.io/github/issues-pr-raw/Xilinx/mlir-aie)
 
 ![](https://mlir.llvm.org//mlir-logo.png)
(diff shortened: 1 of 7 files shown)


AndraBisca:
I really like this proposal. If my understanding is correct, with this op you don't need to chain BDs on different DMA channels using a dma_start op (like [here](https://github.com/Xilinx/mlir-aie/blob/main/test/unit_tests/aie/19_shim_dma_with_core_routed/aie.mlir#L77)). And there actually wouldn't be any ordering of the dma_start ops inside the aie.mem / memtile / shim ops. This would greatly simplify some of the code in the objFifo pass that generates these DMA regions.

makslevental:
> I really like this proposal. If my understanding is correct, with this op you don't need to chain BDs on different DMA channels using a dma_start op (like [here](https://github.com/Xilinx/mlir-aie/blob/main/test/unit_tests/aie/19_shim_dma_with_core_routed/aie.mlir#L77)). And there actually wouldn't be any ordering of the dma_start ops inside the aie.mem / memtile / shim ops. This would greatly simplify some of the code in the objFifo pass that generates these DMA regions.

Yes this is all correct but just to be clear: you also _can't chain bds_ with this op. The specific point is to _only_ handle the common case (as far as I've seen) when 2 channels don't share bds at all.

AndraBisca:
> > I really like this proposal. If my understanding is correct, with this op you don't need to chain BDs on different DMA channels using a dma_start op (like [here](https://github.com/Xilinx/mlir-aie/blob/main/test/unit_tests/aie/19_shim_dma_with_core_routed/aie.mlir#L77)). And there actually wouldn't be any ordering of the dma_start ops inside the aie.mem / memtile / shim ops. This would greatly simplify some of the code in the objFifo pass that generates these DMA regions.
> 
> Yes this is all correct but just to be clear: you also _can't chain bds_ with this op. The specific point is to _only_ handle the common case (as far as I've seen) when 2 channels don't share bds at all.

So then something like this wouldn't be possible? Where the aie buffers are shared between incoming and outgoing DMA channels?

makslevental:
> So then something like [this](https://github.com/Xilinx/mlir-aie/blob/main/test/objectFifo-stateful-transform/link_test_AIE2.mlir#L79) wouldn't be possible? Where the aie buffers are shared between incoming and outgoing DMA channels?

Sorry I should've been clearer - you can share buffers, you just can't share buffer descriptor instructions. Quite literally I'm saying you can't use the mechanism of basic blocks that are jumped two by two separate DMA "starts". You can still share buffers and etc, you just have to emit the `aie.dma_bd` op in both places (both of these `aie.dma {...}` ops).

AndraBisca:
> > So then something like [this](https://github.com/Xilinx/mlir-aie/blob/main/test/objectFifo-stateful-transform/link_test_AIE2.mlir#L79) wouldn't be possible? Where the aie buffers are shared between incoming and outgoing DMA channels?
> 
> Sorry I should've been clearer - you can share buffers, you just can't share buffer descriptor instructions. Quite literally I'm saying you can't use the mechanism of basic blocks that are jumped to by two separate DMA "starts". You can still share buffers and etc, you just have to emit the `aie.dma_bd` op in both places (both of these `aie.dma {...}` ops).

I understand now. I'm not actually sure that we have any examples where we used that functionality so far..

Issue #927, opened 2024-01-11
current CDO generation _always_ loops


makslevental:
In an example like [this](https://github.com/Xilinx/mlir-aie/blob/1a53802b77d39098df04f2a31bb8d905b42f25f8/test/Targets/AIEGenerateCDODirect/02_LM_SHIM_DDR_FillRate.mlir#L23-L28)

```mlir
    %mem_0_2 = aie.mem(%tile_0_2) {
      %0 = aie.dma_start(MM2S, 1, ^bb1, ^bb2)
    ^bb1:  // pred: ^bb0
      aie.use_lock(%lock_0_2, Acquire, 0)
      aie.dma_bd(%buf71_0 : memref<7168xi32>, 0, 7168)
      aie.use_lock(%lock_0_2, Release, 1)
      aie.next_bd ^bb2
    ^bb2:  // 2 preds: ^bb0, ^bb1
      aie.end
    }
```

the expected behaviour is that the bd is executed only once (nevermind why). And thus, `XAie_DmaSetNextBd` should **not** be called/executed. Unfortunately that's not the case; because `^bb1` has a successor [this conditional](https://github.com/Xilinx/mlir-aie/blob/main/lib/Targets/AIETargetCDO.cpp#L427-L433) is executed (note the TODO): 

```cpp
if (block.getNumSuccessors() > 0) {
  Block *nextBlock = block.getSuccessors()[0]; // should have only one
                                               // successor block
  int nextBdNum = blockMap[nextBlock];
  output << "XAie_DmaSetNextBd(" << tileDMAInstRefStr(col, row, bdNum)
         << ", "
         << " /* nextbd */ " << nextBdNum << ", "
         << " /* enableNextBd */ 1);\n"; // TODO Check if br ^end: to
                                         // disable this?
}
```

**and because STL semantics means map access creates key-val pairs** the map access `blockMap[nextBlock];` doesn't fail/error and sets `nextBdNum` to 0 (and `XAie_DmaSetNextBd` will be executed).

This affects both `AIETargetsCDO` and `AIETargetsCDODirect` because I transcribed much of `AIETargetsCDO` into `AIETargetsCDODirect`. Will send a patch for both but documenting here as a reminder.

PR #928, opened 2024-01-11
fix cdo gen always emits bd loops

makslevental:
fix https://github.com/Xilinx/mlir-aie/issues/927


lib/Targets/AIETargetCDO.cpp
```diff
@@ -399,6 +399,7 @@ mlir::LogicalResult AIETranslateToCDO(ModuleOp m, raw_ostream &output) {
         packetID = op.getPacketID();
       }
 
+      assert(blockMap.contains(&block));
       int bdNum = blockMap[&block];
       if (foundBd) {
```
fifield, line 404: It looks like `foundBd` is the same as `blockMap.contains(&block)`, and both the assert and the lookup should move inside the if. That might explain the test failure.

PR #933, opened 2024-01-13
fix docs

makslevental:


docs/_layouts/default.html
jackl-xilinx, line 1: Why are we removing the ADFPasses page? I would think these options are helpful to people running the commands by hand.

makslevental: Because there no such passes - /home/mlevental/dev_projects/mlir-aie/include/aie/Dialect/ADF/ADF.td

PR #938, opened 2024-01-15
Cleanup wrappers for python bindings

AndraBisca:
This PR aims to bring more consistency into the MLIR-AIE python bindings codebase.

In particular, the `AIEOrderedObjectBuffer` naming has been deprecated and replaced by the original `objectfifo`. Additionally, the `object_fifo` wrapper class now has two new functions, `acquire` and `release` that can directly be called on an object_fifo instance in order to make the API feel more like the Python experience. There is also now a wrapper for `object_fifo_link` that can take Python variables as inputs.

As an example the code below,

```
        objectfifo(
            "of0",
            S,
            [M],
            2,
            TypeAttr.get(ObjectFifoType.get(T.memref(256, T.i32()))),
            [],
            [],
        )
        objectfifo(
            "of1",
            M,
            [N],
            2,
            TypeAttr.get(ObjectFifoType.get(T.memref(8, 8, T.i32()))),
            [],
            [],
        )
        objectfifo_link(["of0"], ["of1"])

        @core(N, "test.o")
        def core_body():
            for _ in range_(10):
                elem0 = acquire(
                    ObjectFifoPort.Consume, "of1", 1, T.memref(8, 8, T.i32())
                ).acquired_elem()
                res = Call("test_func", [elem0], [T.i32()])
                objectfifo_release(ObjectFifoPort.Consume, "of1", 1)
                yield_([])
```

is replaced by

```
      of0 = object_fifo("of0", S, M, 2, T.memref(256, T.i32()))
      of1 = object_fifo("of1", M, N, 2, T.memref(8, 8, T.i32()))
      object_fifo_link(of0, of1)

      @core(N, "test.o")
      def core_body():
          for _ in range_(10):
              elem0 = of1.acquire(1)
              res = call("test_func", [elem0], [T.i32()])
              of1.release(1)
              yield_([])
```

This PR also changed the MLIR `ObjectFifoAcquireOp` and `ObjectFifoReleaseOp` to have an `OptionalAttr<ObjectFifoPort>` input (instead of the original `ObjectFifoPort`) as the port can be inferred from the tile on which the code execution these operations is running.

Finally, the python wrappers have been renamed following the same lowercase naming convention as the auto-generated python bindings.


python/dialects/aie.py
```diff
@@ -139,101 +139,82 @@ def __init__(self, tile, link_with=None):
...
-            destBundle=dest_port,
-            destChannel=dest_channel,
+            fifoIns=fifoInRefs,
+            fifoOuts=fifoOutRefs,
         )
 
 
 # Create a packet flow between source and destination tile ports.
-class PacketFlow(PacketFlowOp):
+class packet_flow(PacketFlowOp):
```
makslevental, line 216: This is not a good idea - lowercased names (with no `Op`) are reserved for "value builders" https://github.com/llvm/llvm-project/pull/68308 so this class will shadow an already generated function (you can grep in `_aie_ops_gen.py`).

AndraBisca: I see, thank you Maks! I will adjust the name to something else that isn't the same as the value builders. I would still like to keep it lowercase since every other function call in the .py examples is lowercase.

jgmelber (approved):

PR #941, opened 2024-01-16
adding instructions to install mlir-aie tools from whls

denolf:
WIP on using wheels on windows to avoid building tools


denolf:
Obsolete.

PR #954, opened 2024-01-24
ResNet reference design on Ryzen AI

singagan:
The reference design shows how a dataflow ResNet pipeline can be constructed using a bottleneck block.


makslevental (changes requested):
None of these examples have corresponding lit tests therefore none of them will ever be tested in CI. That is a recipe for code rot and a very disappointing experience for the unprepared, novice, user. Please add `run.lit` files for each test (see elsewhere in this same `ipu-xrt` directory).

reference_designs/ipu-xrt/resnet/baseline/conv3x3_cifar_vector/make.log
makslevental, line 1: Please also remove build artifact files such as this one and the `.ipynb_checkpoints` directory below.

makslevental (changes requested):

PR #956, opened 2024-01-24
Allow f32 type in AIEX::IpuDmaMemcpyNdOp::verify()

nirvedhmeshram:
f32 type work with the current flow so no reason to fail them in the verifier.

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -66,8 +66,8 @@ LogicalResult AIEX::BroadcastPacketOp::verify() {
 
 LogicalResult AIEX::IpuDmaMemcpyNdOp::verify() {
   MemRefType buffer = getMemref().getType();
-  if (!buffer.getElementType().isInteger(32))
-    return emitOpError("must be used with memref type i32.");
+  if (buffer.getElementTypeBitWidth() != 32)
+    return emitOpError("must be used with memref type with element width 32.");
   if (!llvm::all_of(getMixedStrides(), [](OpFoldResult s) {
         return getConstantIntValue(s).has_value();
       }))
--- a/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
@@ -246,7 +246,10 @@ struct DmaToIpuPattern : OpConversionPattern<IpuDmaMemcpyNdOp> {
     MemRefType my_memref = op.getMemref().getType();
     auto shape = my_memref.getShape();
     size_t R = shape.size();
-    size_t S = my_memref.getElementType().getIntOrFloatBitWidth() / 8;
+    size_t el_bit_width = my_memref.getElementTypeBitWidth();
+    assert(el_bit_width % 8 == 0 &&
+           "Expected Memref element bitwidth to be multiple of 8.");
+    size_t S = el_bit_width / 8;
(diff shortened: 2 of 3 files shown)


makslevental (approved):
LGTM

jgmelber:
LGTM. What is the case for using f32?

jgmelber (approved):

nirvedhmeshram:
> LGTM. What is the case for using f32?

Do you mean test case within mlir-aie or more generally? I dont think there is a test case in the repo showing it but upstream models (coming from onxx/pytorch) have f32 types in them forvarious operators that use mlir-aie through iree-amd-aie.

jgmelber:
> > LGTM. What is the case for using f32?
> 
> Do you mean test case within mlir-aie or more generally? I dont think there is a test case in the repo showing it but upstream models (coming from onxx/pytorch) have f32 types in them for various operators that use mlir-aie through iree-amd-aie.

Sounds good if it makes it easier to integrate. The OP matches the hardware where all lengths and strides occur at 4B boundaries so these have to be cast to I32 or F32.

makslevental:
> Sounds good if it makes it easier to integrate. The OP matches the hardware where all lengths and strides occur at 4B boundaries so these have to be cast to I32 or F32.

Just to be a little pedantic - the stream width is 32b (4B) right? There's no notion of data type (signedness, floating point, etc.) right? So this op really should only be doing a `getMemref().getType().getElementTypeBitWidth() == 32` check right? @nirvedhmeshram can you actually make it so.

nirvedhmeshram:
> > Sounds good if it makes it easier to integrate. The OP matches the hardware where all lengths and strides occur at 4B boundaries so these have to be cast to I32 or F32.
> 
> Just to be a little pedantic - the stream width is 32b (4B) right? There's no notion of data type (signedness, floating point, etc.) right? So this op really should only be doing a `getMemref().getType().getElementTypeBitWidth() == 32` check right? @nirvedhmeshram can you actually make it so.

Done. Btw @erieaton-amd is working on enabling bf16 through the flow which will need to relax this assumption further and add logic to pack the 4B boundary with two elements, @jgmelber @makslevental please let us know your thoughts and pointers for that.

makslevental:
> Done. Btw @erieaton-amd is working on enabling bf16 through the flow which will need to relax this assumption further and add logic to pack the 4B boundary with two elements, @jgmelber @makslevental please let us know your thoughts and pointers for that.

Not an expert but I feel like the packing should happen _before_ this op rather than _in_ this op? Not sure.

makslevental:
@nirvedhmeshram can you update the corresponding lowering here to also be purely in terms of bit width https://github.com/Xilinx/mlir-aie/blob/76427507e39e3d74d9a1ffdf092f47c195d38bea/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp#L249.

nirvedhmeshram:
> Not an expert but I feel like the packing should happen _before_ this op rather than _in_ this op? Not sure.
It feels wrong to pack two bf16



> @nirvedhmeshram can you update the corresponding lowering here to also be purely in terms of bit width
> 
> https://github.com/Xilinx/mlir-aie/blob/76427507e39e3d74d9a1ffdf092f47c195d38bea/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp#L249
> 
> .

Done, so @makslevental it will be awkward to pack 2 bf16's in a f32 or 4 i8's into i32 before the ipuDmaMemcpyNdOp is created so I was thinking in the DmaToIpuPattern we can teach it to handle non 4B elements as a next step?

PR #957, opened 2024-01-25
find/handle locks and buffers everywhere

makslevental:
This PR extends buffer address and lock id assignment:

- both passes now look for buffers/locks everywhere in a device instead of just in the region (i.e. `device.walk<WalkOrder::PreOrder>`). The point/value is to declutter the main region of buffers/locks that only play in one tile (e.g. managing data passthrough in a memtile).

### DIFF
--- a/test/ipu-xrt/add_256_using_dma_op_no_double_buffering/aie.mlir
+++ b/test/ipu-xrt/add_256_using_dma_op_no_double_buffering/aie.mlir
@@ -15,11 +15,6 @@ module {
     %buffer_0_2 = aie.buffer(%tile_0_2) : memref<16xi32>
     %buffer_0_2_1 = aie.buffer(%tile_0_2) : memref<16xi32>
 
-    %lock_0_1_0 = aie.lock(%tile_0_1, 0) {init = 1 : i32}
-    %lock_0_1_1 = aie.lock(%tile_0_1, 1) {init = 0 : i32}
-    %lock_0_1_2 = aie.lock(%tile_0_1, 2) {init = 1 : i32}
-    %lock_0_1_3 = aie.lock(%tile_0_1, 3) {init = 0 : i32}
-    
     %lock_0_2_0 = aie.lock(%tile_0_2, 0) {init = 1 : i32}
     %lock_0_2_1 = aie.lock(%tile_0_2, 1) {init = 0 : i32}
     %lock_0_2_2 = aie.lock(%tile_0_2, 2) {init = 1 : i32}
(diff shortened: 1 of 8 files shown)


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -275,158 +275,11 @@ static LogicalResult generateCDO(MLIRContext *context, ModuleOp moduleOp,
         "failed to run passes to prepare of XCLBin generation");
   }
 
-#ifdef AIE_ENABLE_GENERATE_CDO_DIRECT
   if (failed(AIE::AIETranslateToCDODirect(copy, TK.TempDir,
                                           byte_ordering::Little_Endian, false,
                                           false, false))) {
     return moduleOp.emitOpError("failed to emit CDO");
   }
```
makslevental, line 282: cc @spaceotter

test/ipu-xrt/add_256_using_dma_op_no_double_buffering/aie.mlir
```diff
@@ -50,10 +45,14 @@ module {
...
-    %buffer_0_1_0 = aie.buffer(%tile_0_1) : memref<16xi32>
-
     %memtile_dma_0_1 = aie.memtile_dma(%tile_0_1) {
+      %buffer_0_1 = aie.buffer(%tile_0_1) : memref<16xi32>
+      %buffer_0_1_0 = aie.buffer(%tile_0_1) : memref<16xi32>
+      %lock_0_1_0 = aie.lock(%tile_0_1, 0) {init = 1 : i32}
+      %lock_0_1_1 = aie.lock(%tile_0_1, 1) {init = 0 : i32}
+      %lock_0_1_2 = aie.lock(%tile_0_1, 2) {init = 1 : i32}
+      %lock_0_1_3 = aie.lock(%tile_0_1, 3) {init = 0 : i32}
+
```
stephenneuendorffer, line 55: It's not clear to me that this is a net-win.. OK, the scoping for these ops gets a little bit tighter, but this is somewhat of an unusual case... most designs won't be able to leverage this.  At the same time, we expand the set of things that we need to verify....

stephenneuendorffer (commented):
I'd rather see the removal of the cdo backend as a separate PR, since it's most of what this is actually touching.

test/ipu-xrt/add_256_using_dma_op_no_double_buffering/aie.mlir
makslevental: Kicking a hornett's nest: in reality these shouldn't be ops that float around _at all_ and should be closer to attributes of a tile (which itself should be a region rather than an SSA handle that also floats around).

> but this is somewhat of an unusual case

Every single memtile and shimtile doesn't "share" its buffers. Every single memtile connected only through/by/via streaming doesn't share its locks. Any core tile that uses private scratch space doesn't "share" buffers. To wit: I didn't invent this pattern, I only noticed it in the multitude of examples.

makslevental:
> I'd rather see the removal of the cdo backend as a separate PR, since it's most of what this is actually touching.

Done. Top (last) commit is now the relevant commit (until https://github.com/Xilinx/mlir-aie/pull/958 merges).

test/ipu-xrt/add_256_using_dma_op_no_double_buffering/aie.mlir
stephenneuendorffer: Memtile DMAs can access neighboring memtiles.

makslevental: > Memtile DMAs can access neighboring memtiles.

Fair enough, I stand corrected, but it's still the case that this enables a simplification of a very common pattern.

PR #958, opened 2024-01-26
remove old cdo generation path

makslevental:
Factored out from https://github.com/Xilinx/mlir-aie/pull/957.


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -39,7 +37,6 @@
 #include "llvm/Support/ToolOutputFile.h"
 
 #include <regex>
-#include <stdlib.h>
 
 #ifdef AIE_ENABLE_GENERATE_CDO_DIRECT
```
stephenneuendorffer, line 41: Seems like this is extraneous at this point?

makslevental: It's not for the same reason I mentioned on the airbin PR (which is also guarded the same way because of a dep on libelf): someone (me) might be doing dev on various unrelated (and unencumbered) parts on a platform (mac) that doesn't have libcdo.

stephenneuendorffer: I guess It might be a less surprising failure mode in such cases for the executable to not exist, rather than for it to do nothing?  (i.e.  CMake should just not compile the file if libcdo isn't found)

makslevental: Ya sure that makes sense. I'll guard the directory instead of the code.

PR #960, opened 2024-01-26
Fixup export of xaienginecdo_static for downstream users

fifield:
This fixed problems I was having building mlir-air with `AIE_ENABLE_GENERATE_CDO_DIRECT` enabled.

Without this I get errors like:
```
CMake Error at CMakeLists.txt:44 (find_package):
  Found package configuration file:

    /mlir-aie/install/lib/cmake/aie/AIEConfig.cmake

  but it set AIE_FOUND to FALSE so package "AIE" is considered to be NOT
  FOUND.  Reason given by package:

  The following imported targets are referenced, but are missing:
  xaienginecdo_static
```


makslevental (approved):
Ah very good - now I understand how `install(EXPORT` really works. Need to bookmark this PR next time I need it. Thanks.

PR #963, opened 2024-01-29
Allow user to override `aie2xclbin` install location

spaceotter:


stephenneuendorffer (approved):

stephenneuendorffer (commented):
This seems a little odd... we could for instance do something like:
https://unix.stackexchange.com/questions/317211/absolute-path-to-currently-executing-program

Although, we're also working to move me_basic.o into peano, so it's maybe still a short-term workaround.

erieaton-amd:
I originally did it the way you suggested, and there was some problem I forgot.

PR #965, opened 2024-01-29
[aievec] Add multidim vector transfer normalization

jsetoain:
Flatten 2D vector transfer read/write to lower 2D vector arithmetic to LLVM IR. The `vector.shape_cast` ops will cancel each other out with the `shape_cast` injected by the arithmetic ops, and the `memref.collapse_shape` ops will disappear as no-ops during lowering.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
@@ -20,6 +20,7 @@
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
+#include "mlir/Dialect/Utils/ReshapeOpsUtils.h"
 #include "mlir/IR/PatternMatch.h"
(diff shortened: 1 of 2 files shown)


lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
```diff
@@ -300,14 +444,21 @@ static void configureAIEMLCanonicalizeLegalizations(ConversionTarget &target) {
...
+        return cast<VectorType>(op.getVector().getType()).getRank() < 2;
       });
 }
 
 static void
 populateAIEMLCanonicalizeConversionPatterns(RewritePatternSet &patterns) {
   patterns.add<SplitUnalignedTransferReadPattern>(patterns.getContext(), 1024,
                                                   256);
+  patterns.add<FlattenMultDimTransferReadPattern,
+               FlattenMultDimTransferWritePattern>(patterns.getContext());
```
makslevental, line 461: Somehow this broke some of my E2E stuff but in the strangest way - on linux it seems to be no problem but on mac there's an issue...

makslevental: ugh false alarm - some stale link/compilation units somewhere

PR #966, opened 2024-01-30
[aievec] Fix lowering of aievec.matmul to LLVM Dialect

jsetoain:


jsetoain:
@stephenneuendorffer My understanding is that this is how you interface between the different types of vector types and the intrinsics handling matrix-multiply instructions, i.e.: through bitcast ops to force the type required by the intrinsic. The problem is that bitcast won't accept source & destination casts that don't have the same bitsize. I'm living this here until we can figure it out.

PR #973, opened 2024-02-02
[WIP] ConfigureCascadeOp for AIE2

AndraBisca:
This PR adds a new operation to the AIE dialect, `ConfigureCascadeOp`, that takes as input a tile, an input direction, and an output direction (`ChannelDir` has been added to the list of AIE attributes to represent a cascade direction). The op is used like in the example below:

```
module @test {
  aie.device(xcve2802) {
    %t13 = aie.tile(1, 3)
    %t23 = aie.tile(2, 3)
    aie.configure_cascade(%t13, West, East)
    aie.configure_cascade(%t23, North, South)
  }
}
```

Additionally, this PR adds a new target, `mlir_aie_configure_cascade`, to AIETargetXAIEV2.cpp that generates libXAIE calls to `XAie_CoreConfigAccumulatorControl` for each `ConfigureCascadeOp`. This new target should be called from the host code at the beginning of execution to properly setup the registers for each tile.


AndraBisca:
Opening new PR as this one was branched from an un-merged branch.

Issue #517, opened 2023-06-15
AIE2 cascade needs a switchbox-like connection


stephenneuendorffer:
The cascade is programmable in AIE2:

/*
*
* This API configures the core accumulator control register to specify the
* direction of cascade stream.
*
* @param       DevInst: Device Instance
* @param       Loc: Location of the aie tile.
* @param       InDir: Input direction. Valid values: NORTH, WEST
* @param       OutDir: Output direction. Valid values: SOUTH, EAST
*
* @return      XAIE_OK on success, Error code on failure.
*
* @note                None.
*
******************************************************************************/
AieRC XAie_CoreConfigAccumulatorControl(XAie_DevInst *DevInst,
               XAie_LocType Loc, StrmSwPortType InDir, StrmSwPortType OutDir)

hunhoffe:
Closing as implemented. AIE2 programmable cascade configuration is now fully supported through the compiler:
- aie.cascade_flow(%src, %dest) provides the high-level switchbox-like connection between tiles.
- aie.configure_cascade(%tile, inputDir, outputDir) exposes per-tile direction control with a CascadeDir enum (North/West input, South/East output) — mapping directly to the XAie_CoreConfigAccumulatorControl() call named in this issue.
- The --aie-lower-cascade-flows pass lowers cascade_flow into per-tile configure_cascade ops based on relative placement, and the placer treats cascade flows as a first-class constraint (so logical_tile designs get cascadable positions assigned automatically).
- XAIE codegen emits XAie_CoreConfigAccumulatorControl(...) (see lib/Targets/AIETargetXAIEV2.cpp and lib/Targets/AIERT.cpp).

It's tested end-to-end: dialect verifiers, the lowering pass, XAIE codegen (test/Targets/AIEGenerateXAIE/aie2_configure_cascade.mlir), placement, Python bindings, and on-hardware NPU-XRT tests (test/npu-xrt/cascade_flows/ and matrix_multiplication_using_cascade/).

If there are still missing support for cascade, please do feel free to open a new issue!

PR #974, opened 2024-02-02
CascadeFlowOp and ConfigureCascadeOp

AndraBisca:
This PR adds two new operations to the AIE dialect, `CascadeFlowOp` and `ConfigureCascadeOp`. 

`CascadeFlowOp` represents a cascade stream flow between the two input tiles. The op is used like in the example below:

```
module @test {
  aie.device(xcve2802) {
    %t13 = aie.tile(1, 3)
    %t23 = aie.tile(2, 3)
    aie.cascade_flow(%t13, %t23)
  }
}
```

Each `CascadeFlowOp` is lowered into a `ConfigureCascadeOp` which takes as input a tile, an input cascade direction and an output cascade direction. The input and output directions of the cascade for that tile are based on the relative positions of the tiles in the `CascadeFlowOp`. Unless specified otherwise, the default input and output directions for the cascade are `North` and `South` respectively. The lowering is implemented in `AIELowerCascadeFlows.cpp` and is called using `aie-opt --aie-lower-cascade-flows <path to mlir code>`. As an example, the code above lowers to:

```
module @test {
  aie.device(xcve2802) {
    %tile_1_3 = aie.tile(1, 3)
    %tile_2_3 = aie.tile(2, 3)
    aie.configure_cascade(%tile_1_3, North, East)
    aie.configure_cascade(%tile_2_3, West, South)
  }
}
```

Additionally, this PR adds a new configuration function, `mlir_aie_configure_cascade`, to `AIETargetXAIEV2.cpp` that generates libXAIE calls to `XAie_CoreConfigAccumulatorControl` for each `ConfigureCascadeOp`. This new target should be called from the host code at the beginning of execution to properly setup the registers for each tile. Similar configuration functions have been added to `AIETargetCDODirect.cpp` and `AIETargetAirbin.cpp`.


AndraBisca:
Related to #517.

makslevental:
Is this not supported on IPU? Ie shouldn't CDODirect also be extended to support?

AndraBisca:
> Is this not supported on IPU? Ie shouldn't CDODirect also be extended to support?

Yes, it is. I'm working on modifying CDODirect.

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -697,6 +705,18 @@ struct AIEControl {
             WIRE_BUNDLE_TO_STRM_SW_PORT_TYPE.at(connectOp.getDestBundle()),
             connectOp.destIndex());
     }
+
+    // Cascade configuration
+    for (auto configOp : targetOp.getOps<ConfigureCascadeOp>()) {
+      TileOp tile = dyn_cast<TileOp>(configOp.getTile().getDefiningOp());
```
makslevental, line 711: Please don't use `dyn_cast`  (especially unchecked `dyn_cast`) and instead use `cast`: https://llvm.org/docs/ProgrammersManual.html#the-isa-cast-and-dyn-cast-templates.

In fact if you could please change the 3 others in this file (that I forgot to come back and change), that would be great.

include/aie/Dialect/AIE/IR/AIEAttrs.td
```diff
@@ -144,4 +144,15 @@ def BDDimLayoutArrayArrayAttr : ArrayOfAttr<
     /*eltName*/BDDimLayoutArrayAttr.cppClassName
 >;
 
+def CascadeDir: I32EnumAttr<"CascadeDir", "Directions for cascade",
+  [
+    I32EnumAttrCase<"South", 0>,
+    I32EnumAttrCase<"West", 1>,
+    I32EnumAttrCase<"North", 2>,
+    I32EnumAttrCase<"East", 3>
```
makslevental, line 152: Doesn't it make more sense to move the cases out of `WireBundle` and reuse them here [like this](https://github.com/llvm/llvm-project/blob/e6598b053dc71845423132ac0803b1ecd0b71d45/mlir/test/lib/Dialect/Test/TestEnumDefs.td#L18). This would 

1. Prevent renumbering/reassociating enum values (confusing)
2. Save from having to define  `CASCADE_DIR_TO_STRM_SW_PORT_TYPE` in CDODirect (would be able to reuse `WIRE_BUNDLE_TO_STRM_SW_PORT_TYPE`)

makslevental (changes requested):

include/aie/Dialect/AIE/IR/AIEAttrs.td
AndraBisca: I see what you mean, I'll do that! Thank you! :)

keryell:
Would it be possible to make clearer in the operation what is the input and what is the output. Perhaps if runing an LSP server knowing the dialect the IDE could display it as when the mouse is above it, but that does not seem super clear in other context, such as reading slides or random code samples. :-)

keryell:
> > Perhaps if runing an LSP server knowing the dialect
> 
> We don't have such functionality.

This is why I was suggesting to have an MLIR AIE syntax clearer so we can only rely on plain sight.

AndraBisca:
> > > Perhaps if runing an LSP server knowing the dialect
> > 
> > 
> > We don't have such functionality.
> 
> This is why I was suggesting to have an MLIR AIE syntax clearer so we can only rely on plain sight.

So something more like: `aie.configure_cascade(tile = %t13, inputDir = West, outputDir = East)`?
Also, if we're going to push the python bindings instead of MLIR ops, then the syntax can also be adjusted there (if needed).

makslevental (approved):

test/ipu-xrt/cascade_flows/run.lit
```diff
@@ -0,0 +1,12 @@
+// (c) Copyright 2024 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// REQUIRES: ryzen_ai
```
fifield, line 4: also REQUIRES `chess`

AndraBisca: Thank you Jeff!

denolf (approved):
Looking great!

PR #975, opened 2024-02-03
Updates for amdxdna driver

fifield:
fixes for updated amdxdna driver

### DIFF
--- a/python/XRTModule.cpp
+++ b/python/XRTModule.cpp
@@ -31,7 +31,7 @@ public:
            int deviceIndex)
       : xclBin(std::make_unique<xrt::xclbin>(xclBinPath)),
         device(std::make_unique<xrt::device>(deviceIndex)) {
-    assert(device->get_info<xrt::info::device::name>() == "Phoenix" &&
+    assert(device->get_info<xrt::info::device::name>() == "RyzenAI-Phoenix" &&
            "only Phoenix supported by xrt python bindings");
     device->register_xclbin(*xclBin);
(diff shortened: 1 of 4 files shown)


makslevental:
@fifield in the future can you please tag me on changes to this code (or any code) that I've written/am actively using. If I hadn't looked through closed PRs, I wouldn't have noticed this, would've pulled and would've been hyper-confused about why it doesn't work since I haven't yet upgraded the driver. And in truth, these strings need to be promoted to params/env variables/whatever so that _I don't have to upgrade_.

fifield:
> @fifield in the future can you please tag me on changes to this code (or any code) that I've written/am actively using. If I hadn't looked through closed PRs, I wouldn't have noticed this, would've pulled and would've been hyper-confused about why it doesn't work since I haven't yet upgraded the driver.
> 

Sorry for any confusion, I'll be more thoughtful in the future.

> And in truth, now that we have multiple configurations in the wild, the right thing to do would have been to promote these strings to params/env variables/whatever so that _I don't have to upgrade_.

There is only one configuration in the wild, [the open source one](https://github.com/amd/xdna-driver). Our internal dev boxes and the github runner have all been switched over. You should too.

makslevental:
> There is only one configuration in the wild

I don't understand what this means - I have a box with the previous version of the drivers - I am using that box. I am "in the wild" and thus there are at least two configurations in the wild.

> You should too.

`SHOULD` and `MUST` are two very different things.

keryell:
I am using a laptop and this is already obsolete if you are using the latest public driver.
So there are at least 3 configurations in the wild. ;-)

PR #988, opened 2024-02-07
Add chess support to aie2xclbin

spaceotter:
This copies the pile of hacks from aiecc.py to use the Vitis chess compiler with aie2xclbin (thereby allowing IREE to use chess).

### DIFF
--- a/tools/aie2xclbin/XCLBinGen.cpp
+++ b/tools/aie2xclbin/XCLBinGen.cpp
@@ -198,55 +198,113 @@ static LogicalResult generateCoreElfFiles(ModuleOp moduleOp,
     } else {
       elfFileName = std::string("core_") + std::to_string(col) + "_" +
                     std::to_string(row) + ".elf";
-      std::cout << "assigned core elf name " << elfFileName << std::endl;
       coreOp.setElfFile(elfFileName);
     }
 
-    SmallString<64> ldscript_path(TK.TempDir);
-    sys::path::append(ldscript_path, elfFileName + ".ld");
-    {
-      auto ldscript_output = openOutputFile(ldscript_path, &errorMessage);
-      if (!ldscript_output) {
-        return coreOp.emitOpError(errorMessage);
-      }
-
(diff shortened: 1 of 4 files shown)


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -501,6 +502,184 @@ static LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
   return success();
 }
 
+static std::string chesshack(const std::string &input) {
```
stephenneuendorffer, line 505: Can you factor this out into a library that can be called from aiecc.py?

spaceotter: Just this one function or the whole thing?

stephenneuendorffer: We shouldn't have the same code duplicated between aiecc.py and aie2xclbin.  This function seems like a good granularity to reduce that duplication?  I think the same goes for much of the other code you've factored out.

tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -197,55 +198,113 @@ static LogicalResult generateCoreElfFiles(ModuleOp moduleOp,
...
+                                     "-d",
+                                     "+l",
+                                     std::string(bcfPath),
+                                     "-o",
+                                     std::string(elfFile),
+                                     "-f",
+                                     std::string(objFile)};
+      for (const auto &inc : extractedIncludes) {
+        flags.push_back(inc);
+      }
```
makslevental, line 258: nit: can you remove braces for all single line for/ifs (ie follow upstream conventions...)

stephenneuendorffer (approved):

tools/aie2xclbin/XCLBinGen.cpp
makslevental: @erieaton-amd thanks :)

PR #1019, opened 2024-02-13
buidHostLin.md improvements

andrej:
I made these edits after following the instructions on a completely fresh install of Ubuntu 22.04.

1. Reformat.
2. Rewrite and give rationale for steps in a couple places for better user experience.
3. Include steps that were previously only referenced by links. I think it makes more sense to include them so that a user can simply follow a linear list of steps.
4. Add a couple missing prerequisites.
5. Rewrite code so that it should mostly just work when users blindly copy and paste it in. This should be fairly close to a script that just does all of this.

### DIFF
--- a/docs/buildHostLin.md
+++ b/docs/buildHostLin.md
@@ -1,85 +1,260 @@
-# Linux Build Instructions: </h1>
-- Rely on Ubuntu 22.04 LTS for Vitis tool install and to build and run our MLIR-AIE tools
+# Linux Build Instructions
+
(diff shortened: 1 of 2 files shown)


jgmelber (approved):

jgmelber:
LGTM. If possible propagate any improvements for clarity to the Windows 11 getting started guide.

andrej:
I've refactored the Windows instructions a little bit and added a troubleshooting section to the Linux to cover some issues I had.

Please merge if it looks okay, as I do not have the rights to merge

PR #1020, opened 2024-02-13
Make aie2xclbin relocatable

### LINKED ISSUE #1012: aie2xclbin is not relocatable
The install directory is hardcoded at configure time.  This means that the code cannot be moved to a different install location.

erieaton-amd:
Infer install location from aie2xclbin's location by default. Remove compile time location.

Fixes #1012


stephenneuendorffer (approved):
Most AMD/Xilinx tools rely on an environment variable for this kind of homing.  Perhaps something we should figure out a better way once things settle down.  Thanks!

PR #1024, opened 2024-02-15
compile cdo_driver locally

stephenneuendorffer:
Build will probably fail because there's a small fixup in the bootgen repo necessary


runtime_lib/CMakeLists.txt
```diff
@@ -128,148 +128,21 @@ foreach(target ${AIE_RUNTIME_TARGETS})
...
+add_aiert_headers(xaienginecdo_static
+        ${XAIE_SOURCE_DIR}
+        ${AIE_BINARY_DIR}/include
+        ${CMAKE_INSTALL_PREFIX}/include/xaiengine
+)
+add_aiert_library(xaienginecdo_static ${XAIE_SOURCE_DIR} STATIC)
 
-  target_link_libraries(xaienginecdo_static PUBLIC cdo_driver)
-  target_compile_definitions(xaienginecdo_static PRIVATE -D__AIECDO__)
+target_link_libraries(xaienginecdo_static PRIVATE cdo_driver)
```
makslevental, line 141: need a `target_include_directory(xaienginecdo_static INTERFACE <SOMEPLACE_INSIDE_BOOTGEN>/cdo_driver)` here I believe to get tests to pass.

stephenneuendorffer: My hope was to avoid having cdo_driver.h exposed to users..  did I miss some place?

makslevental:
Regarding all the updated tests in `test/Targets/AIEGenerateCDODirect/*` (excluding `AIEGenerateCDODirect/dma_op_test`) - just FYI, after removing the old CDO generation path, these now basically function as smoke tests ie they don't `CHECK` anything. So they'll catch some catastrophic failure but not much else. I left them in after removing the old generation path because there were no semantic tests of the old path so I figured it's better to have at least smoke tests :shrug:. But at this point it wouldn't be unreasonable to DCE all of them (or replace them with useful semantic tests).

Note: `AIEGenerateCDODirect/dma_op_test` does actually test something useful (fidelity between `aie.dma_start` and `aie.dma` lowerings through CDODirect) so I'm not talking about removing those.

lib/Targets/AIETargets.cpp
```diff
@@ -31,6 +31,10 @@
 #include "llvm/Support/FileSystem.h"
 #include "llvm/Support/JSON.h"
 
+extern "C" {
+#include "cdo_driver.h"
```
makslevental, line 35: @stephenneuendorffer this is the spot you missed I believe (see warnings below in file view...)

PR #1027, opened 2024-02-16
[wip] workaround maximum 5 kernel args

makslevental:


jgmelber:
Nice workaround. I am wondering if there's a performance impact going from 1 write_bd_extend_shim_tile_opt to multiple write32s for each dma_memcpy_nd.

makslevental:
> Nice workaround. I am wondering if there's a performance impact going from 1 write_bd_extend_shim_tile_opt to multiple write32s for each dma_memcpy_nd.

:shrug: but looking at the implementation of the interpreter loop in dpufw maybe? You're doing `|args|`*k word unpacking now. If they had something like a `ExecWriteK32Opt(instr_ptr, &pc, partBaseAddr, word, k)` with an impl like `for (i=&word; i <= word+k; ++i) *(bd) = iptr[i];` then that would mitigate (but I'm about to start asking for fw features...).

PR #1037, opened 2024-02-20
add tracing to matrix multiply reference design

andrej:
- refactor reference_designs/ipu-xrt/matrix_multiplication/aie2.py slightly to avoid * imports
- add tracing code which is only included if `enable_tracing=TRUE` in aie2.py and `ENABLE_TRACE=true` in `test.cpp`

Note I had to use `AIEX._generated_ipu_write32` instead of `AIEX.ipu_write32`. The latter wrapper is too restrictive to really write arbitrary values.

Successfully tested but it does appear to lead to verification fails after the first run (back-to-back subsequent runs) if the device is not reset in between. It appears that if too much tracing information is generated, it "spills over" into the results of the next run? Not sure.


jgmelber (approved):

jgmelber (changes requested):
Please set tracing default to OFF for the CI

andrej:
I'm just realizing the way I checked out `utils/parse_trace.py` from the `resnet` branch, all commit history for that file.

It also shows up as if I authored it.

Is there a way that I can include this file while properly attributing it to whoever wrote it (including commit history)?

jackl-xilinx:
Looks like this fails in the same way as my trace implementation on my setup so it's likely my setup issue. I'll update my env and try my trace addition again. As for the parse script, we haven't finalized it's inclusion into main but it really should be merged in sooner rather than later and not be held up from the readiness of the resnet branch (that was intended to be merged in earlier as well but has been delayed). I'll work to separate the trace code out and add it in as part of adding trace to matmul.

andrej:
> Looks like this fails in the same way as my trace implementation on my setup so it's likely my setup issue. I'll update my env and try my trace addition again. 

Are you running verification in the host code? If I do verification for the 2048x2048x2048 size, the host code takes about a minute to compute the result, which may look like hanging.

> As for the parse script, we haven't finalized it's inclusion into main but it really should be merged in sooner rather than later and not be held up from the readiness of the resnet branch (that was intended to be merged in earlier as well but has been delayed). I'll work to separate the trace code out and add it in as part of adding trace to matmul.

Cool. I'll make sure to remove that file from this branch

jackl-xilinx:
Update and clarification. I see the same back to back error that you see when running the smaller size 128x128x128 in my case but I see it in your 256x256x256. I also continue to see the 2048x2048x2048 hang in my setup (with verification but waited >5min). So it may still be something in my env that's causing it but the back to back is a different situation I feel.

andrej:
@jgmelber I addressed your comments. I upgraded to the most recent wheel of MLIR-AIE and this compiles and runs with that version.

jackl-xilinx:
So I reran this design on our xco phoenix minisforum machines (which has a later xdna driver installed than my local miniforum linux box) and the back-to-back run error and 2048x2048x2048 hang issue goes away for my trace additions. I also verified that the back-to-back error goes away for this trace PR on the updated linux machine, so it does seem to be an issue with the older driver. Comparing our two trace changes, they are nearly identical so let's go ahead and prep this one for merging. I believe in our meeting, we discussed possibly using the import * in order for AIE and AIEX keywords not to be needed in these example designs just to make the examples cleaner (even though there can be issues as additional packages are included). @jgmelber Can you confirm this? If that's true, please make those final modifications and I think this PR will be ready to merge.

reference_designs/ipu-xrt/matrix_multiplication/test.cpp
```diff
@@ -39,7 +39,11 @@ constexpr int A_SIZE = (A_VOLUME * sizeof(A_DATATYPE));
 constexpr int B_SIZE = (B_VOLUME * sizeof(B_DATATYPE));
 constexpr int C_SIZE = (C_VOLUME * sizeof(C_DATATYPE));
 
-constexpr bool VERIFY = true;
+constexpr bool VERIFY = false;
```
jgmelber, line 42: We want VERIFY = true for the CI

jackl-xilinx:
Hmm ... it looks like I didn't notice VERIFY=false was set and thought designs were passing. Now, when I run MKN=256 and 2048, it seems to fail verification on my setup but passes if MKN=128. Can you verify if this is true on your end as well? Something seems to break when tracing is enabled for 256 and 2048.

andrej:
I removed the namespace prefixes. Sorry for just making that change without asking, I didn't think it would be controversial.

I also added the `VERIFY=true`.

> Hmm ... it looks like I didn't notice VERIFY=false was set and thought designs were passing. Now, when I run MKN=256 and 2048, it seems to fail verification on my setup but passes if MKN=128. Can you verify if this is true on your end as well? Something seems to break when tracing is enabled for 256 and 2048.

I can't reproduce this error. 256 passes verification for me. But I have seen failed verifications when I switch sizes and then re-run. I think it has something to do with the issue I described at the top, where maybe we're getting some old output from the device? Have you tried resetting the device between runs?

andrej:
I can confirm verification errors happen on my end if I run at some size, e.g. 256 (verification passes on that run), then change the size to 128, recompile, and re-run (verification errors). 

Running `sudo modprobe -r amdxdna && sudo modprobe amdxdna`, then re running makes verification pass again.

So it seems that part of what runs on the NPU is not updated, or some of the "old" data is still streaming out when we run with the new size...

@jackl-xilinx Can you confirm if running above commands between runs fixes the issue for you?

jackl-xilinx:
@andrej I think I figured out the issue. It seems to have to do with the event port running that's monitored not being configured. I was about to merge in the change but you had added some adjustments. Let me pull your changes, confirm that the fix works and them commit the fix.

jackl-xilinx:
@andrej Strange. It was working just a second ago and then I pulled your latest changes and now it's not working for some reason. Need a little more time to figure out what changed.

andrej:
@jackl-xilinx Sorry! I can revert my changes if needed. All I did was enable `verify=True`, remove the `AIE`, `AIEX`, and `SCF` prefixes, and rebase on the most recent `main`. Functionally, none of that should have changed anything!

Have you tried the reset procedure?

jackl-xilinx:
@andrej I tried that change but i'm getting some strange inconsistent verification errors that I wasn't getting earlier today and the files should not have changed as you said. I'm trying to figure out what might be affecting this in my environment. the modprobe command didn't solve it issue for me but I suspect something else is happening right now. I'll let you know if there is an update to this.

jackl-xilinx:
@andrej Ok, it looks like there's some inconsistency in the behavior and the modprobe command doesn't quite fix it for me. But we'll push out fixing that for now since I'm not able to converge on a consistent solution. Can you give me permissions to push a commit to your branch/ fork? Thanks.

jackl-xilinx:
@andrej I think this URL gives some info on that. 
https://docs.github.com/en/get-started/learning-about-github/access-permissions-on-github 
Since I couldn't push directly, I tried created a branch from your branch and create a PR to it but I also cannot merge it in without write permission.

jackl-xilinx:
@andrej Nvm. Looks like ssh solved my access problems. :)

jackl-xilinx (approved):

jgmelber (approved):

PR #1047, opened 2024-02-23
chess_wrapper: remove hard coded paths

denolf:
avoid hard coded paths after cmake configure

### DIFF
--- a/tools/chess-clang/xchesscc_wrapper
+++ b/tools/chess-clang/xchesscc_wrapper
@@ -7,10 +7,11 @@
 # (c) Copyright 2021 Xilinx Inc.
 # (c) Copyright 2022-2023 Advanced Micro Devices, Inc.
 
-TARGET_AIE_LIBDIR=${VITIS_AIE_INCLUDE_DIR:-"@VITIS_AIE_INCLUDE_DIR@"}
-TARGET_AIE2_LIBDIR=${VITIS_AIE2_INCLUDE_DIR:-"@VITIS_AIE2_INCLUDE_DIR@"}
+VITIS_ROOT=`realpath $(dirname $(which vitis))/../`
+AIETOOLS=$VITIS_ROOT/aietools
 
-AIETOOLS=${VITIS_AIETOOLS_DIR:-"@VITIS_AIETOOLS_DIR@"}
+TARGET_AIE_LIBDIR=$AIETOOLS/data/versal_prod/lib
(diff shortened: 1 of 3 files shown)


jgmelber (approved):

tools/chess-clang/xchesscc_wrapper
```diff
@@ -7,10 +7,11 @@
 # (c) Copyright 2021 Xilinx Inc.
 # (c) Copyright 2022-2023 Advanced Micro Devices, Inc.
 
-TARGET_AIE_LIBDIR=${VITIS_AIE_INCLUDE_DIR:-"@VITIS_AIE_INCLUDE_DIR@"}
-TARGET_AIE2_LIBDIR=${VITIS_AIE2_INCLUDE_DIR:-"@VITIS_AIE2_INCLUDE_DIR@"}
+VITIS_ROOT=`realpath $(dirname $(which vitis))/../`
```
stephenneuendorffer, line 10: I think we need an 'architected' way to describe how tools find eachother.  Is it path, or by specific environment variables.  should discuss with IREE folks here, since different bits of the flow have different opinions on this.  would be good to develop some consistency.

stephenneuendorffer (approved):

tools/chess-clang/xchesscc_wrapper
makslevental: you can easily support both (that's the intent of the removed lines with the `:-`)

denolf: our rule of thumb used to be (and should still be in my opinion) to depend on as few as possible environment variables, ideally none, hence the proposed change.

Counting on cmake (@VITIS_AIE_INCLUDE_DIR@) makes the path hard coded for what cmake finds during its configure time: FindVitis.cmake also tries to find vitis, similar to the $(which vitis).

So while the previous version supported both options, they were both potentially unreliable.

Maybe we should do 'which xchesscc' instead of which vitis, because in this wrapper, we need xchesscc and not vitis.

makslevental: Depending on `which` doesn't mean not supporting env variables -  it means supporting exactly one env variable: `PATH`, which IIRC people do use to fiddle with `which` and has caused problems wrt which clang ends up in the path.

Anyway I'm not advocating in either direction - ie I have no strong opinion (since I personally don't go through `xchess_wrapper`) - I was just pointing out that both `which` and env variables is possible.

denolf:
merged #1047

PR #1053, opened 2024-02-26
[aievec] Generalize flattening of multi-dim load/store

jsetoain:


lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
```diff
@@ -386,12 +389,23 @@ struct FlattenMultDimTransferWritePattern
...
+
+    auto inBoundsArrayAttrOpt = adaptor.getInBounds();
+    if (inBoundsArrayAttrOpt) {
+      SmallVector<bool> inBounds = llvm::to_vector(
+          inBoundsArrayAttrOpt.value().getAsValueRange<BoolAttr>());
+      SmallVector<bool> newInBounds({false});
+      newInBounds[0] = std::all_of(inBounds.begin(), inBounds.end(),
+                                   [](bool v) { return v; });
+      newOp.getProperties().setInBounds(rewriter.getBoolArrayAttr(newInBounds));
+    }
```
jamestcl-amd, line 408: Can we add a test to cover the changes here? If I understand correctly, this can lead to `in_bounds = [false]` in some cases, which we cannot support yet? Thanks!

jsetoain: Indeed, this _will_ lead to `in_bounds = [false]` when the access is not in bounds. But it's not that we don't support it, we don't allow it: the fist pass in the next pipeline (`lower-vector-to-aievec`) sets all the accesses in bounds, indiscriminately, without verification. That's fundamentally wrong, but also we know we don't generate out of bounds accesses. The test I'm adding with the improvements to `vector.contract` to `aievec.matmul` will include coverage for that bit of code, fear not 🙂

PR #1059, opened 2024-02-27
Add vectorization to aie2xclbin path (targeting llvm-aie) 

newling:
This PR allows a complete lowering from IREE for vectorized bfloat16 matmul. 

I would like to add a test with aie2xclbin, but it is proving challenging (I haven't managed to run all the mlir-aie tests successfully) ...


tools/aie2xclbin/CMakeLists.txt
```diff
@@ -17,22 +17,30 @@ llvm_map_components_to_libnames(llvm_libs support)
...
 target_link_libraries(aie2xclbin
   ${dialect_libs}
+  ${conversion_libs}
+  ${extension_libs}
   MLIRParser
   MLIRSCFToControlFlow
   MLIRAffineToStandard
+  MLIRAIEVecDialect
+  MLIRAIEVecToLLVM
+  MLIRAIEVecTransformOps
```
jsetoain, line 35: You might not need this particular one (`MLIRAIEVecTransformOps`), that's the `transform.structured` extension I added to vectorize matmul-like contractions in `linalg.generic` ops.

newling: Good point. I sprayed everything I could at this cmake file to get everything compiling and linking. I need to remove the unnecessary parts (like MLIRAIEVecTransformOps)

jsetoain:
FYI: I've tested changing the lhs/rhs load alignment from 2 to 4, and that's enough to get it to compile. This might be a much narrower issue than I first thought, and it "smells" like it's an issue with the compiler.

tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -576,18 +586,77 @@ static std::string chesshack(const std::string &input) {
...
+struct RemoveAlignment2FromLLVMLoadPass
+    : public PassWrapper<RemoveAlignment2FromLLVMLoadPass,
+                         OperationPass<ModuleOp>> {
+  void runOnOperation() override {
+    getOperation().walk([](Operation *op) {
+      if (auto loadOp = dyn_cast<LLVM::LoadOp>(op)) {
+        auto alignmentAttr = loadOp.getAlignmentAttr();
+        if (alignmentAttr && alignmentAttr.getValue() == 2)
+          loadOp.setAlignment(std::optional<uint64_t>());
+      }
```
jsetoain, line 618: Eventually, we might have to do the same for `LLVM::StoreOp`. No change this now, with a little bit of luck it will be fixed for good and we won't need it any more, but we should make a note in case this lives longer than we wish and the "misaligned" stores start triggering the issue again.

jsetoain (approved):
Looks good to me.

tools/aie2xclbin/XCLBinGen.cpp
newling: Good point, will keep in mind. I'll land this PR as is as soon as it passes CI.

PR #1062, opened 2024-02-27
[aievec] Make accum. materialization optional for matmul

jsetoain:
When lowering to LLVM IR, we don't need to make transfers between the accumulator and regular bank registers explicit.


jsetoain:
@david-vc

PR #1067, opened 2024-02-28
DCE AIETargetCDODirect

makslevental:
The removed code here seemingly has no effect nor do I understand its intent since on IPU these configs would be overwritten by direct programming by ipu instructions.


jgmelber (approved):
Tests pass. Nice cleanup

PR #1068, opened 2024-02-28
re-enable windows (fix elf dep)

makslevental:
Lucky for us bootgen has `elf.h`


stephenneuendorffer (approved):

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -746,26 +745,31 @@ LogicalResult generateCDOBinariesSeparately(AIEControl &ctl,
...
+              .str(),
+          [&ctl, &targetOp, &workDirPath, &aieSim] {
+            return ctl.addAieElfsToCDO(targetOp, workDirPath, aieSim);
+          })))
     return failure();
 
   if (failed(generateCDOBinary(
-          workDirPath.str() + ps + "aie_cdo_init.bin",
+          (llvm::Twine(workDirPath) + std::to_string(ps) + "aie_cdo_init.bin")
+              .str(),
```
makslevental, line 765: I suspect this is sloppy but I am too tired...

makslevental:
I make no guarantees that this works - note the tests on those windows jobs were disabled (for a reason that I can't remember).

PR #1069, opened 2024-02-28
Remove dead code: iree-amd-aie with werror fails 

newling:


makslevental:
My bad...

jgmelber:
It seems like the tests are failing after this commit. Should this have been merged into main?

makslevental:
> It seems like the tests are failing after this commit. Should this have been merged into main?

This commit removed dead code - it could not have affected tests. The tests are failing on the minis runner for the same reason they're always failing (the driver is flaky).

jgmelber:
> > It seems like the tests are failing after this commit. Should this have been merged into main?
> 
> This commit removed dead code - it could not have affected tests. The tests are failing on the minis runner for the same reason they're always failing (the driver is flaky).

Yeah, that needs to be fixed. We need to be able to reliably test and keep all components from main working

makslevental:
> Yeah, that needs to be fixed.

I have asked for more runners so that one single runner isn't the single point of failure for 3 projects  _numerous times_..

PR #1073, opened 2024-02-28
Fix wheels Revert "Stephen's "compile cdo_driver locally""

jgmelber:
Reverts Xilinx/mlir-aie#1060


makslevental (changes requested):
Why would you revert when we're literally fixing this right now https://github.com/Xilinx/mlir-aie/pull/1068

jgmelber:
> Why would you revert when we're literally fixing this right now #1068

It wasn't clear this was being addressed

PR #1077, opened 2024-02-29
Pin wheels version in quick setup

jgmelber:
Initially manually pin wheels to a particular verified version in the quick setup script. Future work to use actions to update the script after automatic verification.


jgmelber:
This PR also updates the required python version to 3.10. @STFleming

jackl-xilinx (approved):

PR #1081, opened 2024-03-01
Fix for unused variable errors

jtuyls:
I got unused variable errors after https://github.com/Xilinx/mlir-aie/pull/1079. Not sure why this isn't catched in CI?

cc @makslevental @jsetoain

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.cpp
+++ b/lib/Dialect/AIEVec/Transforms/FoldMulAddChainToConvOp.cpp
@@ -245,7 +245,7 @@ struct LongestConvMACChainAnalysis {
       opBwdSlices.insert(mulOpOperand);
 
       LLVM_DEBUG(llvm::dbgs() << "opBwdSlices = [\n");
-      for (auto op : opBwdSlices) {
+      for ([[maybe_unused]] auto op : opBwdSlices) {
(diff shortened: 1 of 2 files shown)


makslevental:
I don't know what's going on here because these aren't unused (unless your compiler can't figure out the control flow?), if they were our CI should have caught before I landed [_and also IREE's should have_](https://github.com/Xilinx/mlir-aie/pull/1069), ie the reason I turned this on is because one of the suspected causes of the breakage yesterday was a fix of an unused for IREE folks which have the same `werror`. @newling any chance you know if maybe the files touched here aren't compiled by IREE (and thus escaping your guy's `Wunused`?

Anyway I'm happy to merge to unblock but obv just adding `[[maybe_unused]]`s defeats the purpose of the check...

AndraBisca:
I have the same issue on my personal build, using the scripts in utils/.

makslevental:
> I have the same issue in my personal build, using the scripts in utils/. It's considering them unused because the values are only checked in an assert().

I thought it might be that but we have CI that tests with assertions on right? https://github.com/Xilinx/mlir-aie/blob/main/.github/workflows/buildAndTest.yml#L119

makslevental (approved):
Anyway I'm approving to unblock and making a mental note to debug next time someone gets bitten.

PR #1082, opened 2024-03-01
Fix ipu_sync for multicore in matrix vector ref design

jtuyls:
@jgmelber In the multicore matrix vector reference design, I believe every shim connection requires its own sync to make sure all output data is properly returned?

### DIFF
--- a/reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
+++ b/reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
@@ -20,7 +20,7 @@ def my_matmul():
     word_size_in = 2
     word_size_out = 4
 
-    n_cores = 3
+    n_cores = 1
 
     A_sz_in_i32s = M * K * word_size_in // 4
     B_sz_in_i32s = K * word_size_in // 4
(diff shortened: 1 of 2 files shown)


jgmelber (approved):
Thanks for fixing the sync! I would like to keep the matrix size as 288x288 (matches the weights size for Llama2 tinystories15M weights

jtuyls:
> Thanks for fixing the sync! I would like to keep the matrix size as 288x288 (matches the weights size for Llama2 tinystories15M weights

Ok, but then 

> Thanks for fixing the sync! I would like to keep the matrix size as 288x288 (matches the weights size for Llama2 tinystories15M weights

Ok! Is updating to 3 cores ok, or should it be 1?

jtuyls:
> > > Thanks for fixing the sync! I would like to keep the matrix size as 288x288 (matches the weights size for Llama2 tinystories15M weights
> > 
> > 
> > Ok! Is updating to 3 cores ok, or should it be 1?
> 
> I'd recommend 1, 2 or 4, it divides the work evenly

288 / 32 = 9, which is not divisible by 2/4 and I believe this failed. I set the number of cores to 1 now.

jtuyls:
@jgmelber CI has passed with a latest formatting fix. If it looks good, I think we can merge?

jgmelber:
> @jgmelber CI has passed with a latest formatting fix. If it looks good, I think we can merge?

Yes, LGTM

PR #1083, opened 2024-03-01
Simplify verification in matmul

jgmelber:
Enable slightly larger tests in CI by simplifying the verification methodology.

### DIFF
--- a/reference_designs/ipu-xrt/matrix_multiplication_array/test.cpp
+++ b/reference_designs/ipu-xrt/matrix_multiplication_array/test.cpp
@@ -24,9 +24,9 @@
 #include "xrt/xrt_device.h"
 #include "xrt/xrt_kernel.h"
 
-constexpr int M = 256;
-constexpr int K = 256;
-constexpr int N = 256;
+constexpr int M = 512;
+constexpr int K = 512;
+constexpr int N = 512;
 
 constexpr int A_VOLUME = M * K;
 constexpr int B_VOLUME = N * K;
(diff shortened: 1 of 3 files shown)


reference_designs/ipu-xrt/matrix_multiplication_array/test.cpp
```diff
@@ -229,17 +228,22 @@ int main(int argc, const char *argv[]) {
   if (VERIFY) {
     std::vector<C_DATATYPE> output_ref0;
     for (uint32_t i = 0; i < C_VOLUME; i++)
-      output_ref0.push_back(0);
-    matmul(AVec, BVec, output_ref0);
+      output_ref0.push_back(K);
+    // matmul(AVec, BVec, output_ref0);
```
denolf, line 232: why can this line be commented out when output_ref0 is used in the verification check?

reference_designs/ipu-xrt/matrix_multiplication_array/test.cpp
```diff
@@ -229,17 +228,22 @@ int main(int argc, const char *argv[]) {
   if (VERIFY) {
     std::vector<C_DATATYPE> output_ref0;
     for (uint32_t i = 0; i < C_VOLUME; i++)
-      output_ref0.push_back(0);
-    matmul(AVec, BVec, output_ref0);
+      output_ref0.push_back(K);
```
denolf, line 231: Are we initializing with K?

denolf (approved):

reference_designs/ipu-xrt/matrix_multiplication_array/test.cpp
jgmelber: Instead of running the GEMM on the CPU, comparing the output to a single value

PR #1085, opened 2024-03-02
[xllvm] Create dialect for external LLVM IR intrinsics

jsetoain:
This dialect works as an extension to the LLVM Dialect to support LLVM-based compilers other than the one mlir-aie is compiled against.


jsetoain:
James @newling , this change might affect your work on IREE, but I think it's necessary and the sooner we get it the better.

lib/Dialect/XLLVM/XLLVMOps.cpp
```diff
@@ -0,0 +1,68 @@
...
+llvm::CallInst *createExternalLLVMIntrinsicCall(
+    llvm::IRBuilderBase &builder, LLVM::ModuleTranslation &moduleTranslation,
+    Operation *intrOp, llvm::StringRef intrinsicName) {
+  // We support 0 or 1 results
+  assert(intrOp->getNumResults() <= 1 &&
+         "external multi-result intrinsics not supported");
+  llvm::Type *resTy = nullptr;
+  if (intrOp->getNumResults())
+    resTy = moduleTranslation.convertType(*(intrOp->getResultTypes().begin()));
+  auto operands = moduleTranslation.lookupValues(intrOp->getOperands());
```
makslevental, line 55: Interesting - I didn't know you could extend llvm translation like this. Cool.

lib/Target/LLVMIR/Dialect/XLLVM/XLLVMToLLVMIRTranslation.cpp
```diff
@@ -40,23 +38,22 @@ class AIEVecDialectLLVMIRTranslationInterface
...
+  registry.addExtension(+[](MLIRContext *ctx, xllvm::XLLVMDialect *dialect) {
+    dialect->addInterfaces<XLLVMDialectLLVMIRTranslationInterface>();
   });
 }
 
-void xilinx::aievec::registerAIEVecDialectTranslation(MLIRContext &context) {
+void xilinx::xllvm::registerXLLVMDialectTranslation(MLIRContext &context) {
   DialectRegistry registry;
-  registerAIEVecDialectTranslation(registry);
+  registerXLLVMDialectTranslation(registry);
```
makslevental, line 57: :+1:

lib/Dialect/XLLVM/XLLVMOps.cpp
jsetoain: The "magic" is happening [during registration of the -mlir-to-llvmir pass](https://github.com/Xilinx/mlir-aie/blob/ce3d8d9215a4c24969363d869f722316620f060e/tools/aie-translate/aie-translate.cpp#L25-L44). A bit hackish, but it's contained and gets the job done.

newling:
> James @newling , this change might affect your work on IREE, but I think it's necessary and the sooner we get it the better.

That's fine, I can check on my end if you land this before my changes land

include/aie/Dialect/XLLVM/IR/XLLVMDialect.td
```diff
@@ -0,0 +1,42 @@
+//===- XLLVM.td - XLLVM Dialect definition ----------------*- tablegen -*-====//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+// Defines External LLVM (XLLVM) dialect.
```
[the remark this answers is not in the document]
jsetoain: Uhm... this shouldn't exist. It's not a copy, it's a rename. Let me fix that...

PR #1089, opened 2024-03-04
[wip] ipu sync + shimtile dma

makslevental:


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -891,7 +891,7 @@ def AIE_DMAStartOp: AIE_Op<"dma_start", [
   );
   let successors = (successor AnySuccessor:$dest, AnySuccessor:$chain);
   let assemblyFormat = [{
-    `(` $channel_dir `,` $channel_index `,` $dest `,` $chain (`,` `repeat_count` `=` $repeat_count^)? `)` attr-dict
```
AndraBisca, line 894: This change means that the MLIR syntax will change from
`aie.dma_start("MM2S", 0, ^bd0, ^end, repeat_count=2)`
to 
`aie.dma_start("MM2S", 0, ^bd0, ^end) {repeat_count=2}`
right?

makslevental: There's a reason I'm doing (that's important) and I will discuss before I land

AndraBisca: Sounds good! :)

PR #1093, opened 2024-03-05
Faster Matrix Multiplication Verification

andrej:
- New `matmul` reference computation that should be faster. Verified correct for 512x512x512. When compiled, it uses SIMD instructions on my machine:
image
- Proper floating point comparison adapted from a source online
- Remove some duplicated code. This is just a start, there's still a lot of duplicated stuff, which has already started to diverge.

### DIFF
--- a/reference_designs/ipu-xrt/matrix_multiplication.h
+++ b/reference_designs/ipu-xrt/matrix_multiplication.h
@@ -1,0 +1,254 @@
+//===- matrix_multiplication.h ----------------------------000---*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2024, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// This file contains common helper functions for the matrix multiplication
+// host code, such as verifying and printing matrices.
+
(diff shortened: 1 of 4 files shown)


jgmelber (commented):
Great changes. Can you also fix _column?

andrej:
I'm sorry ... I think I was a little to prematurely excited and made this PR. I saw the SIMD instructions and thought this must be an improvement.

Turns out if I time it it's actually  really not faster than the naive reference ... 

I'll look into it some more. I'll also address your comments regarding restructuring the directories. That's a good idea.

For now, I'm making this a draft again. Hey, at least it passes verification!

jgmelber:
> I'm sorry ... I think I was a little to prematurely excited and made this PR. I saw the SIMD instructions and thought this must be an improvement.
> 
> 
> 
> Turns out if I time it it's actually  really not faster than the naive reference ... 
> 
> 
> 
> I'll look into it some more. I'll also address your comments regarding restructuring the directories. That's a good idea.
> 
> 
> 
> For now, I'm making this a draft again. Hey, at least it passes verification!

Maybe for the small size it's not faster. Have you compared times on larger GEMMs?

andrej:
Yeah, I just did 2048x2048x2048. 1m19s naive vs like 1m13s with this. Must be that either it's not utilizing the SIMD fully, or the compiler is smart enough to SIMD-ize the naive implementation too and that's just the best you can do on the CPU. If it's the former, I'll look into how to optimize it. If it's the latter, I'll see if there's a library that's easy enough to plug-and-play into it to do it on the GPU maybe. Feels like there should be something, no need to reinvent the wheel.

PR #1095, opened 2024-03-06
Bump required cmake (3.21 -> 3.23) to get OpenSSL::applink fix

newling:
Without this change I get:

```
CMake Error at tools/bootgen/CMakeLists.txt:147 (add_executable):
  Target "bootgen" links to target "OpenSSL::applink" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
  ```
  when I run 
  
  ```
  ./utils/build-mlir-aie.sh llvm/build
  ```
  
 I don't know what bootgen, or "applink" is.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 #  -DAIE_RUNTIME_TARGETS: list of targets (x86_64,aarch64) to build runtime libs for, default: x86_64; cross compilation for aarch64 against default Vitis Sysroot
 #  -DAIE_RUNTIME_TEST_TARGET: runtime test target (x86_64 or aarch64) used for running unit test and tutorials, default x86_64
 
-cmake_minimum_required(VERSION 3.21)
+cmake_minimum_required(VERSION 3.23)
 
 set(TOOLCHAINFILES_PATH ${CMAKE_SOURCE_DIR}/cmake/modulesXilinx)
 


makslevental:
No shot this is getting merged lol - these are both components/pieces necessary for assembling xclbins (and it was all introduced as a direct dep in this PR https://github.com/Xilinx/mlir-aie/pull/1060 and then fixed up in this PR https://github.com/Xilinx/mlir-aie/pull/1068). @erieaton-amd can relay all of the details.

makslevental (changes requested):
.

makslevental:
I guess there is something fishy if the tests pass? `OpenSSL::applink` is a component of openssl that is required by bootgen because [...reasons...](https://github.com/Xilinx/bootgen/blob/65ed60ae22dbd540a2a480a40732b2c5c1d5eab7/main.cpp#L39)? When we were fighting this last week, that was one of the failures on windows build.

newling:
How about I make it a dependency (in cmake) only for windows builds?

I guess I could find out why my linux setup doesn't like this when clearly the linux CI doesn't mind it. I am pretty sure I've apt installed all the openssl things available.

makslevental:
> How about I make it a dependency (in cmake) only for windows builds?
> 
> I guess I could find out why my linux setup doesn't like this when clearly the linux CI doesn't mind it. I am pretty sure I've apt installed all the openssl things available.

you can surround things in CMake with `if(WIN32)` or something like that.

> I am pretty sure I've apt installed all the openssl things available.

You sure you installed `libssl-dev` and not `openssl`? You need the former.

newling:
> > How about I make it a dependency (in cmake) only for windows builds?
> > I guess I could find out why my linux setup doesn't like this when clearly the linux CI doesn't mind it. I am pretty sure I've apt installed all the openssl things available.
> 
> you can surround things in CMake with `if(WIN32)` or something like that.
> 
> > I am pretty sure I've apt installed all the openssl things available.
> 
> You sure you installed `libssl-dev` and not `openssl`? You need the former.

Yes, libssl-dev, sorry I mistyped.. `libssl-dev is already the newest version (3.0.2-0ubuntu1.12). 0 upgraded, 0 newly installed, 0 to remove and 11 not upgraded`.

Ok cool, I'll play around in CMake if I can't resolve this locally.

makslevental:
Late breaking developments in the case: https://github.com/Xilinx/mlir-aie/pull/1100#issuecomment-1981910376

makslevental:
Just aggregating info: https://github.com/openssl/openssl/issues/17767#issuecomment-1050607089

@newling can you double check what version of CMake you're using when you build?

newling:
I am using
$ which cmake # /usr/bin/cmake
$ cmake --version # cmake version 3.22.1

newling:
lol yes, bingo indeed. Updating cmake is an easy fix. A cmake conditional might save someone in the future. If I have some time I will do this. 

Thank you for investigating and connecting dots !

makslevental:
> lol yes, bingo indeed. Updating cmake is an easy fix. A cmake conditional might save someone in the future. If I have some time I will do this.

Fantastic idea - just bump the minimum CMake version here https://github.com/Xilinx/mlir-aie/blob/main/CMakeLists.txt#L16 to `3.23`.

makslevental:
Did upgrading CMake not work? I would prefer to gate like that if that's the root cause (otherwise sure we can land this and kick the can down the road...)

newling:
@makslevental  I missed your suggestion before adding the win32 conditional approach. Increasing the minimum required cmake works fine (as it better long term).

makslevental (approved):

PR #1101, opened 2024-03-06
Revert "Pin wheels version in quick setup"

jgmelber:
Reverts Xilinx/mlir-aie#1077. Now that building the wheels is tested on every commit/PR


jgmelber:
This pulls the latest wheels even if they don't work

PR #1104, opened 2024-03-07
[wip] use transaction API

makslevental:


lib/Targets/AIERTX.cpp
```diff
@@ -0,0 +1,457 @@
...
+#include "xaiengine/xaiegbl.h"
+#include "xaiengine/xaiegbl_defs.h"
+}
+
+using namespace mlir;
+
+#define DEBUG_TYPE "aie-aiertx"
+
+llvm::raw_ostream &operator<<(llvm::raw_ostream &os, const XAie_LocType &loc) {
+  os << "XAie_LocType(col: " << std::to_string(loc.Col)
```
keryell, line 28: I am curious about the rationale of all these `std::to_string`.

lib/Targets/AIERTX.cpp
```diff
@@ -0,0 +1,457 @@
...
+  TRY_XAIE_API_FATAL_ERROR(XAie_StartTransaction, &devInst,
+                           XAIE_TRANSACTION_DISABLE_AUTO_FLUSH);
+}
+
+void AIERTXControl::exportSerializedTransaction() {
+  XAie_TxnInst *txnInst = XAie_ExportTransactionInstance(&devInst);
+  std::ios_base::fmtflags f(std::cout.flags());
+  for (size_t i = 0; i < txnInst->NumCmds; ++i) {
+    std::cout.flags(f);
+    std::cout << "Txn OpCode: " << std::hex
```
keryell, line 446: Since you are using LLVM, you could use the Pythonic `llvm::formatv` from
```c++
#include "llvm/Support/FormatVariadic.h"
```
instead of old `<<` spaghetti, while waiting for C++20 or compile with C++20/23...

makslevental: this is just for sanity checking - i'm going to ditch this soon (but thanks for the pointer to `format`.

lib/Targets/AIERTX.cpp
makslevental: These are all `uint8_t` which `llvm::raw_ostream` doesn't like for some reason (or something like that) - basically just `<<` didn't do the right thing and I didn't think too hard about it :shrug:

makslevental: Yea that was it - lots of wingdings

fifield:
Most of this has been merged as part of other PRs, closing.

Issue #1094, opened 2024-03-05
`isShimNOCTile` is wrong for 5x4 designs.


makslevental:
Physically I believe the (0,0) tile is the Shim PL Tile; if you set the partition start column to the 0th column, then `aie-rt` identifies that tile as the [shim PL tile](https://github.com/stephenneuendorffer/aie-rt/blob/4ce26c8cd2a635ec27055f4ddd689f954ed30660/driver/src/device/xaie_device_aieml.c#L59-L58). This is an issue for 5x4 designs because our [isShimNOCTile](https://github.com/Xilinx/mlir-aie/blob/a63eb5dac38c89257c94174d0cb214332083d597/include/aie/Dialect/AIE/IR/AIETargetModel.h#L468) does not take into account "virtualization" i.e. partition start column. Not sure what the fix is without extending the Target model to capture partition start.

jgmelber:
Yes, currently we consider the IPU target a virtualized 4x4 device to limit the complexity. The device model would need to be extended, perhaps with a special case for 5 column designs if we decide that this is something we want to support.

AndraBisca:
@jgmelber is this still an issue?

AndraBisca:
Closed due to inactivity. Please feel free to re-open.

PR #1106, opened 2024-03-07
[wip] adjust target model to support 5x4

makslevental:
This PR parameterizes `AIETargetModel` with a `bool virtualized`, which is then queried when checking `isShimNOCTile(col, row) || isShimPLTile(col, row)`. Note, I have only updated `IPUTargetModel` to the correct result because that is the only one I am familiar with (the rest fallback to `virtualized=true`, i.e., the current behavior).

Note, this PR touched a lot of lines only because the models were `static` prior and had to be updated to `shared_ptr`s.

fixes https://github.com/Xilinx/mlir-aie/issues/1094

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIETargetModel.h
+++ b/include/aie/Dialect/AIE/IR/AIETargetModel.h
@@ -53,8 +53,12 @@ using TileID = struct TileID {
 };
 
 class AIETargetModel {
+  bool virtualized;
+
 public:
-  AIETargetModel() = default;
+  AIETargetModel(bool virtualized = true) : virtualized(virtualized) {}
+
+  bool isVirtualized() const { return virtualized; }
 
   virtual ~AIETargetModel();
 
@@ -451,8 +455,6 @@ public:
 };
 
 class IPUTargetModel : public AIE2TargetModel {
-  llvm::SmallDenseSet<unsigned, 16> nocColumns = {0, 1, 2, 3};
-
 public:
   using AIE2TargetModel::AIE2TargetModel;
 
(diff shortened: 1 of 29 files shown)


jgmelber:
Do you have a new test for 5 columns?

makslevental:
> Do you have a new test for 5 columns?

In another branch yes

lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
```diff
@@ -39,8 +39,8 @@ struct AIECanonicalizeDevicePass
 
     Location location = builder.getUnknownLoc();
     auto deviceOp = builder.create<DeviceOp>(
-        location,
-        AIEDeviceAttr::get(builder.getContext(), AIEDevice::xcvc1902));
+        location, AIEDeviceAttr::get(builder.getContext(), AIEDevice::xcvc1902),
+        /*virtualized*/ builder.getBoolAttr(true));
```
stephenneuendorffer, line 43: This seems wrong, because the vc1902 is not virtualized.

lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
```diff
@@ -27,9 +27,9 @@ struct AIECanonicalizeDevicePass
 
     ModuleOp moduleOp = getOperation();
 
-    for (auto deviceOp : moduleOp.getOps<AIETarget>()) {
+    for (auto deviceOp : moduleOp.getOps<DeviceOp>()) {
```
stephenneuendorffer, line 30: Why did you need this?

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -53,8 +53,12 @@ using TileID = struct TileID {
 };
 
 class AIETargetModel {
+  bool virtualized;
+
 public:
-  AIETargetModel() = default;
+  AIETargetModel(bool virtualized = true) : virtualized(virtualized) {}
```
stephenneuendorffer, line 59: This seems insufficient, since we actually have many potential 'virtual' targets (for instance, of different column widths.

stephenneuendorffer (changes requested):

lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
makslevental: I removed that interface - there was only one op that implemented it `AIEDeviceOp`

stephenneuendorffer:
Obsolete with #1443

PR #1108, opened 2024-03-07
[wip] Simple bank-aware malloc scheme 

AndraBisca:
This PR is currently a work-in-progress, bringing back an experimental proposal for a very simple, bank-aware memory alloc scheme doing round-robin allocations across available memory banks. The PR is currently intended to provide an easy way for the functionality to be tested further, but will not be ready for review before it is properly discussed with the team.


makslevental:
Hey very cool - I was planning on doing something like this (but with ILP). Will take a close look!

AndraBisca:
It's supposed to be a very simple scheme to ensure that allocation is bank-aware. Right now, the allocation in mlir-aie just looks at a memory module as a single, big region of memory. And this one also worked well to ensure that the first objects in different objectFifos on the same tile don't land in the same bank, if possible. We should discuss this soon! :)

PR #1109, opened 2024-03-07
[AMD-AIE] Build microkernels for AMD-AIE

Abhishek-Varma:
-- This commit builds the first microkernel which would be needed
   by iree-amd-aie during the linking/execution of the vmfb.

Signed-off-by: Abhishek Varma <abhvarma@amd.com>


Abhishek-Varma:
Hi @MaheshRavishankar @erwei-xilinx @fifield @stephenneuendorffer 

I was able to figure out the issue and have been able to create the first step towards the microkernel workflow.

With the current PR we have a way to build microkernels while building MLIR-AIE and then subsequently be able to use it by pointing to the directory via `--iree-amdaie-path-to-ukernels`.

This PR gets us an i32 vanilla matmul microkernel8 built with MLIR-AIE and installed in and as `mlir-aie/install/aie_ukernel_lib/mm.o`.

Have tested it with latest iree-amd-aie (and changes from https://github.com/nod-ai/iree-amd-aie/pull/203) - it works!

MaheshRavishankar:
cc @makslevental @erieaton-amd and @jsetoain . Could you loop in folks that could help review this.

erwei-xilinx:
cc @andrej @jgmelber @jackl-xilinx

erwei-xilinx (approved):

aie_runtime_lib/CMakeLists.txt
```diff
@@ -52,7 +52,29 @@ function(add_aie_runtime_libs arch)
...
+    endif()
+  else()
+      message(STATUS "Tool 'xchesscc' not found. Skipping compilation of microkernel .o files.")
+  endif()
+
+endfunction()
+
 add_subdirectory(AIE)
 add_subdirectory(AIE2)
+add_subdirectory(Ukernels)
```
denolf, line 78: I am not sure if aie kernels belong in a subfolder of the aie_runtime_lib
@jackl-xilinx @jgmelber should we start a top-level subfolder collecting all the kernels to avoid too much duplication?

denolf (changes requested):

aie_runtime_lib/CMakeLists.txt
Abhishek-Varma: Hi @denolf - let me know if I should indeed bring `Ukernels` under it's own umbrella -> `aie_ukernel_lib` (I'm actually doing that while trying to install the binaries to `mlir-aie/install`).

This PR is a blocker for the initial demo of `i32` vanilla matmul (`matmul_scalar`) via iree-amd-aie - the `mm.cc` that I've added here is a bit different from other kernels - `"offset"` has been added to the APIs and accordingly the microkernel `matmul_scalar` has been modified.

Regarding all microkernels, I don't think it'd be a nice idea to plumb them down rightaway as part of this PR and can be handled perhaps in a subsequent set of PRs - the compilation and installation is being done only for the particular `mm.cc` (the one with "offsets").

Let me know your thoughts and I can address them soonest.

denolf: Hi @Abhishek-Varma,

Let's indeed start by having the kernel for this PR in a "aie_kernels" subfolder, so that you can make progress.

As you point out, there are more aie hardcoded kernels on the mlir-aie repo that then can be moved in a subsequent PR.

Thanks for addressing this.

denolf (approved):

Abhishek-Varma:
Since I don't have merging rights - can someone please help merge this in?

PR #1111, opened 2024-03-07
offset, len, and stride in words not bytes (or 32b)

makslevental:
The PR changes `offset` and `len` on `dma_bd` to be denominated in **multiple of the element type width**, e.g., `memref<32xi16>` with `offset=4` and `len=8` will send **elements** 4 through 12 inclusive (3rd through 11th). 

The arithmetic to convert between the element type width and 32b alignments is now done in the op definition (~cf. `getLenIn32bWords` and `getOffsetIn32bWords`~ cf. `getLenInBytes` and `getOffsetInBytes` [the spec is in bits but libxaie is in bytes :shrug:]). 

Note, this PR does **not** update [AIEDmaToIpu::IpuDmaMemcpyNdOp](https://github.com/Xilinx/mlir-aie/blob/174398da1e140bcd9b4527aa0697ae3ec8c50e1d/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp#L249) because if I understand correctly, it's already treated in this way there (though I could be wrong - I am not familiar with that op).

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -745,44 +745,66 @@ def AIE_DMABDPACKETOp: AIE_Op<"dma_bd_packet", []> {
   }];
 }
 
-def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
-  let summary = "Declare a dma block descriptor op";
+def AIE_DMABDOp: AIE_Op<"dma_bd", [
+    ParentOneOf<["MemOp", "MemTileDMAOp", "ShimDMAOp", "DMAOp"]>,
+  ]> {
+  let summary = "Declare a dma buffer descriptor op";
   let description = [{
-    This operation describes a block descriptor for DMA operations. In particular, it specifies
-    what buffer addresss to use, the transfer length, and the buffer type (A or B).
+    This operation describes a buffer descriptor for DMA operations. In particular, it specifies
+    what buffer to use, and optionally:
 
-    This operation must be used in an MLIR block that lives inside a MemOp's region.
-    The block descriptor specifies what lock to use and the buffer configuration.
+        1. the offset into the buffer;
+        2. the transfer length;
+        3. the sizes and strides for n-d tensor addressing (described below);
+        4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
+
+    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bits; e.g., transferring the whole of
+    `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 1024`.
+
+    `dma_bd` ops must appear in their own bbs (basic blocks) and such bbs can (optionally) include `use_lock`
+    operations (specifying an "acquire" and a "release" lock; see the `use_lock` operation) and subsequent bds in
+    a "chain" of bds (using `next_bd` as a "jump" to the next bb which contains a `dma_bd`).
 
     Example:
     ```
       // this defines a BD that uses lock %lck0 and buffer %buf0
       ^bd5:
         aie.use_lock(%lck, "Acquire", 0)
-        aie.dma_bd(<$buf0 : memref<512xi32>, 0, 512>, 1)
+        // transfer the first 32 elements of the memref
+        aie.dma_bd(<$buf0 : memref<128xi32>, 0, 32)
         aie.use_lock(%lck, "Release", 1)
-        br ^bd6 // point to the next Block, which is also a different Block Descriptor
+        aie.next_bd ^bd6 // point to the next bb, which describes the next buffer descriptor
+      ^bd6:
+        aie.use_lock(%lck, "Acquire", 1)
+        // transfer the last 32 elements of the memref
+        aie.dma_bd(<$buf1 : memref<128xi32>, 96, 32)
+        aie.use_lock(%lck, "Release", 0)
+        aie.next_bd ^end
 
       ...
 
       // this defines a BD that does not use any lock
       ^bd8:
-        aie.dma_bd(<$buf1 : memref<64xi32>, 0, 64>, 0)
+        aie.dma_bd(<$buf2 : memref<64xi32>, 0, 64)
     ```
-    A DMA channel in a Memory Module can process one block descriptor after another by chaining them.
-    There are 16 block descriptors per Memory Module. They are shared by four DMA channels.
+
+    ## Background/context:
+
+    A DMA channel in a Memory Module can process one buffer descriptor after another by chaining them.
+    There are 16 buffer descriptors per Core memory module and 48 buffer descriptors per Memtile memory module.
+    They are shared by four DMA channels (or 12).
 
     ## DMA Data Layout Transformations on AIE-ML Devices
 
     AIE-ML devices can apply data layout transformations at the buffer
     descriptor level. These transformation are described by strides and sizes in up to three dimensions (four
     dimensions on memtiles). Strides and sizes can be supplied to the `dma_bd`
-    through an optional argument, an array of tuples `<size, stride>`.
+    through an optional argument, an array of "tuple-like" attributes `bd_dim_layout<size, stride>`.
 
-    The first element of this array gives the _highest-dimension_ stride and
-    size, the last element of the array gives the lowest-dimension.
+    The first element of this array gives the outer-most dimension's stride and
+    size, the last element of the array gives the inner-most dimension's stride and size.
 
-    Strides are always expressed in units of `i32`s; this is an architectural
+    Sizes and strides are always expressed in units of 32 bits; this is an architectural
     requirement, as data is moved by the DMA at this fundamental size.
 
     We can model the access pattern strides and sizes generate by a series of
@@ -824,7 +846,9 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
 
   let arguments = (
     ins AnyMemRef:$buffer,
-        OptionalAttr<AIEI32Attr>:$offset,
+        // in words (multiples of 32b) not bytes
+        DefaultValuedAttr<AIEI32Attr, "0">:$offset,
+        // in words (multiples of 32b) not bytes
         OptionalAttr<AIEI32Attr>:$len,
         OptionalAttr<BDDimLayoutArrayAttr>:$dimensions,
         OptionalAttr<AIEI32Attr>:$bd_id,
@@ -840,8 +864,16 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
 
   let extraClassDeclaration = [{
     BufferOp getBufferOp();
-    int getOffsetValue() { return getOffset().value_or(0); }
-    int getLenValue() { return getLen().value_or(getBuffer().getType().getNumElements()); }
+    int32_t getBufferElementTypeWidthInBytes() {
+      return getBuffer().getType().getElementTypeBitWidth() / 8;
+    }
+    int32_t getLenIn32bWords() {
+      if (std::optional<int32_t> len = getLen())
+        return len.value() * getBufferElementTypeWidthInBytes() / 4;
+      else
+        return getBuffer().getType().getNumElements() * getBufferElementTypeWidthInBytes() / 4;
+    }
+    int32_t getOffsetIn32bWords() { return getOffset() * getBufferElementTypeWidthInBytes() / 4; }
   }];
 
   let hasVerifier = 1;
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -758,7 +758,7 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
         3. the sizes and strides for n-d tensor addressing (described below);
         4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
 
-    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bit words; e.g., transferring the whole of
+    `offset`, `len`, `size`s and `stride`s are all denominated in element width; e.g., transferring the whole of
     `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 256`.
 
     `dma_bd` ops must appear in their own BBs (basic blocks) and such BBs can (optionally) include `use_lock`
@@ -846,9 +846,9 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
 
   let arguments = (
     ins AnyMemRef:$buffer,
-        // in words (multiples of 32b) not bytes
+        // in words (multiples of element width) not bytes
         DefaultValuedOptionalAttr<AIEI32Attr, "0">:$offset,
-        // in words (multiples of 32b) not bytes
+        // in words (multiples of element width) not bytes
         OptionalAttr<AIEI32Attr>:$len,
         OptionalAttr<BDDimLayoutArrayAttr>:$dimensions,
         OptionalAttr<AIEI32Attr>:$bd_id,
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -867,13 +867,13 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
     int32_t getBufferElementTypeWidthInBytes() {
       return getBuffer().getType().getElementTypeBitWidth() / 8;
     }
-    int32_t getLenIn32bWords() {
+    int32_t getLenInBytes() {
       if (std::optional<int32_t> len = getLen(); len.has_value())
-        return len.value() * getBufferElementTypeWidthInBytes() / 4;
+        return len.value() * getBufferElementTypeWidthInBytes();
       else
-        return getBuffer().getType().getNumElements() * getBufferElementTypeWidthInBytes() / 4;
+        return getBuffer().getType().getNumElements() * getBufferElementTypeWidthInBytes();
     }
-    int32_t getOffsetIn32bWords() { return getOffset() * getBufferElementTypeWidthInBytes() / 4; }
+    int32_t getOffsetInBytes() { return getOffset() * getBufferElementTypeWidthInBytes(); }
   }];
 
   let hasVerifier = 1;
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -758,12 +758,12 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
         3. the sizes and strides for n-d tensor addressing (described below);
         4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
 
-    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bits; e.g., transferring the whole of
-    `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 1024`.
+    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bit words; e.g., transferring the whole of
+    `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 256`.
 
-    `dma_bd` ops must appear in their own bbs (basic blocks) and such bbs can (optionally) include `use_lock`
-    operations (specifying an "acquire" and a "release" lock; see the `use_lock` operation) and subsequent bds in
-    a "chain" of bds (using `next_bd` as a "jump" to the next bb which contains a `dma_bd`).
+    `dma_bd` ops must appear in their own BBs (basic blocks) and such BBs can (optionally) include `use_lock`
+    operations (specifying an "acquire" and a "release" lock; see the `use_lock` operation) and subsequent BDs in
+    a "chain" of BDs (using `next_bd` as a "jump" to the next BB which contains a `dma_bd`).
 
     Example:
     ```
@@ -868,7 +868,7 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
       return getBuffer().getType().getElementTypeBitWidth() / 8;
     }
     int32_t getLenIn32bWords() {
-      if (std::optional<int32_t> len = getLen())
+      if (std::optional<int32_t> len = getLen(); len.has_value())
         return len.value() * getBufferElementTypeWidthInBytes() / 4;
       else
         return getBuffer().getType().getNumElements() * getBufferElementTypeWidthInBytes() / 4;
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -847,7 +847,7 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
   let arguments = (
     ins AnyMemRef:$buffer,
         // in words (multiples of 32b) not bytes
-        DefaultValuedAttr<AIEI32Attr, "0">:$offset,
+        DefaultValuedOptionalAttr<AIEI32Attr, "0">:$offset,
         // in words (multiples of 32b) not bytes
         OptionalAttr<AIEI32Attr>:$len,
         OptionalAttr<BDDimLayoutArrayAttr>:$dimensions,
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -845,18 +845,14 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
     ## Important gotcha regarding strides
 
     All strides are expressed in multiples of the element width (just like `len` and `offset`)
-    **with the caveat that the inner-most dimension's stride must be a multiple of 32b**.
-    I.e., `(stride_0 * element_width) % 32 == 0`. For example, if you are striding a `memref<512xi16>`,
-    `stride_0` cannot be 1 (or 3, or 5, ...). The reason for this requirement is stream access to DMA memory is 32b
-    aligned (and thus, for the `memref<512xi16>` example, a stride of 1 would require 16b, half-word, access, which isn't possible).
-
+    **with the caveat that the inner-most dimension's stride must be 1**.
   }];
 
   let arguments = (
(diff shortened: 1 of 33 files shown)


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -741,44 +741,66 @@ def AIE_DMABDPACKETOp: AIE_Op<"dma_bd_packet", []> {
...
+        2. the transfer length;
+        3. the sizes and strides for n-d tensor addressing (described below);
+        4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
+
+    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bits; e.g., transferring the whole of
+    `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 1024`.
+
+    `dma_bd` ops must appear in their own bbs (basic blocks) and such bbs can (optionally) include `use_lock`
+    operations (specifying an "acquire" and a "release" lock; see the `use_lock` operation) and subsequent bds in
+    a "chain" of bds (using `next_bd` as a "jump" to the next bb which contains a `dma_bd`).
```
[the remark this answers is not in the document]
makslevental: hazy morning... thanks for the spotting

jgmelber:
Adding @AndraBisca here because this likely impacts the object FIFO join/distribute

AndraBisca:
> Adding @AndraBisca here because this likely impacts the object FIFO join/distribute

Yes that will impact some of the calculations for the join and distribute, as well as the generation of AIE_DMABdOps in the objectFifo lowering. I think mainly [this ](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L208)function needs to be updated to reflect the change in this PR. I'll take a look tomorrow and confirm.

lib/Targets/AIETargetAirbin.cpp
```diff
@@ -81,7 +81,7 @@ static constexpr auto ME_SS_SLAVE_SLOT_BASE = 0x3F200u;
 /*
   Tile DMA
 */
-static constexpr auto ME_DMA_BD_COUNT = 16;
+static constexpr auto ME_DMA_BD_COUNT = 48;
```
jgmelber, line 84: This should still be 16 since this target only considers AIE not AIE-ML. Better yet should be using the Device Models.

jgmelber:
Adding a note on [AIEDmaToIpu::IpuDmaMemcpyNdOp](https://github.com/Xilinx/mlir-aie/blob/174398da1e140bcd9b4527aa0697ae3ec8c50e1d/lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp#L249), the hardware wraps/strides/offsets are all 4B aligned and the memref type is restricted to i32 or f32 for now. A future PR could incorporate these improvements to support more data types.

jgmelber (approved):
Good improvements! I'd wait on @AndraBisca's check to merge. It's worth adding a note that this removes the skeleton AB-mode support found in AIE (not AIE-ML) hardware.

makslevental:
> I think mainly [this ](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L208)function needs to be updated to reflect the change in this PR. I'll take a look tomorrow and confirm.

I already updated `AIEObjectFifoStatefulTransform` in the latest commit (actually the only place that needed to be patched was [here](https://github.com/Xilinx/mlir-aie/blob/57a29a806f43494191850c2e6339fd21fd54fd5c/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L752)) but I'm happy to wait for a signoff from the resident expert :)

lib/Targets/AIETargetAirbin.cpp
makslevental: I'm just gonna change it back - that thing is coupled to a bunch of `static_assert`s and statically sized arrays. cc @jnider for when you come back to airbin...

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -820,7 +842,9 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
 
   let arguments = (
     ins AnyMemRef:$buffer,
-        OptionalAttr<AIEI32Attr>:$offset,
+        // in words (multiples of 32b) not bytes
+        DefaultValuedOptionalAttr<AIEI32Attr, "0">:$offset,
```
stephenneuendorffer, line 846: This doesn't make sense to me...shouldn't it me in multiples of the element type?

makslevental: comment is wrong but impl is right (ie what you said)

makslevental: But actually I realize I lied and didn't patch the n-d addressing too.

makslevental: Top commit patches size and stride as well; libxaie says 

```
* @note		None. The stepsize and wrap parameters operate at 32 bit
*		granularity. 
```

but I take that to mean just stride (what they're calling stepsize) right? Because what would it even mean for size ("wrap") to be at the 32b granularity (it's the loop bound...). Sanity check here would be much appreciated @stephenneuendorffer .

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1545,107 @@ LogicalResult DMABDOp::verify() {
...
-    MemRefType buffer = getBuffer().getType();
-    // We are not restrictive about the type of the memref used as the input
-    // to the DMABD when used with multi-dimensional strides/wraps. Since the
-    // BD will use the memref as a base address and copy from it in 32 bit
-    // chunks, while assuming the layout of the memref is contiguous. We
-    // assume the user/compiler understands and accounts for this.
-    uint64_t memrefSize = 1; // in bytes
-    uint64_t maxIdx = 0;
-    for (int64_t memrefDim : buffer.getShape())
-      memrefSize *= 4 * memrefDim;
```
makslevental, line 1566: Dunno what this is about but it's wrong for at least 2 reasons (assumes 4 bytes per element, and then for some reason computes a "byte volume"???).

stephenneuendorffer: That seems odd... it seems like there's too many '4's getting multiplied?

AndraBisca:
> > I think mainly [this ](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L208)function needs to be updated to reflect the change in this PR. I'll take a look tomorrow and confirm.
> 
> I already updated `AIEObjectFifoStatefulTransform` in the latest commit (actually the only place that needed to be patched was [here](https://github.com/Xilinx/mlir-aie/blob/57a29a806f43494191850c2e6339fd21fd54fd5c/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L752)) but I'm happy to wait for a signoff from the resident expert :)

Looks good to me! Thank you for updating Maks! :)

AndraBisca (approved):
Looks good to me! It's so much easier to read the offset and transfer length like this imo, and the extra checks in the AIEDialect are good. Thank you for working on this Maks!

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -820,7 +842,9 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
 
   let arguments = (
     ins AnyMemRef:$buffer,
-        OptionalAttr<AIEI32Attr>:$offset,
+        // in words (multiples of element width) not bytes
```
AndraBisca, line 845: This changed from `multiples of 32b` (which I found straightforward) to `multiples of element width` (which I find a bit confusing). What was the reason for the change? Do we always assume that the element width is 32b?

makslevental: No - the arithmetic is done so that if you write `aie.dma_bd(%buf33_0 : memref<256xi16>, 0, 32)` then 32*2B = 64 bytes will transferred.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -741,44 +741,66 @@ def AIE_DMABDPACKETOp: AIE_Op<"dma_bd_packet", []> {
...
+    what buffer to use, and optionally:
 
-    This operation must be used in an MLIR block that lives inside a MemOp's region.
-    The block descriptor specifies what lock to use and the buffer configuration.
+        1. the offset into the buffer;
+        2. the transfer length;
+        3. the sizes and strides for n-d tensor addressing (described below);
+        4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
+
+    `offset`, `len`, `size`s and `stride`s are all denominated in element width; e.g., transferring the whole of
```
denolf, line 757: is it in 32b multiples or is it in element with multiples (where element width changes depending on the type)?

denolf (changes requested):

include/aie/Dialect/AIE/IR/AIEOps.td
AndraBisca: So the length is always expressing the number of elements. Ok!

include/aie/Dialect/AIE/IR/AIEOps.td
denolf: or on line 758, does len need to be equal to 512 in both cases?

reference_designs/ipu-xrt/matrix_multiplication/aie2.py
jgmelber, line 1: Same comment on word size. Also noticing for all of these, the length is now wrong for the 1st dimension. It is still word_size_in // 4, but should just be "s" if the length is in multiples of data type

reference_designs/ipu-xrt/matrix_multiplication_column/aie2.py
jgmelber, line 1: Same comments as others about the 1D length and word_size // 2

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
jgmelber, line 1: Same comments as others about the 1D length and word_size // 2

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: it's "element width multiples" - line 758 is correct. this PR allows/enables the user to (almost) only think in terms of element width, forgetting 32b alignment except in one place - the final stride dim, which does have to be a multiple of 32b. I will update this doc string to say mention that (the last stride dim).

makslevental: still waking up...

lib/Dialect/AIE/IR/AIEDialect.cpp
makslevental: n-d memory! `bytes^n`!

reference_designs/ipu-xrt/matrix_multiplication_array/aie2.py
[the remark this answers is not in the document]
makslevental: that patch was done as a "go to bed see if it works in the morning" since I don't understand those designs - it mostly worked! I will polish now.

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
```diff
@@ -103,9 +103,9 @@ def device_body():
                     2,
                     memRef_A_ty,
                     [
-                        (k_in_i32s, 1),
-                        (m, k_in_i32s),
-                        (1, 1),
+                        (k_in_i32s, 2),
```
denolf, line 106: Why are we using k_in_i32s here? Does more of this PR need to be absorbed in the objFIFO?

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
```diff
@@ -103,9 +103,9 @@ def device_body():
...
+                        (m, k_in_i32s * 2),
```
denolf, line 107: Similar question, why is the *2 needed here

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
denolf: I was hoping to see here k and not k_in_i32s.

denolf (approved):
LGTM, thanks for the improvements.

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
```diff
@@ -103,8 +103,9 @@ def device_body():
                     2,
                     memRef_A_ty,
                     [
-                        (k_in_i32s, 1),
-                        (m, k_in_i32s),
+                        (1, 1),
+                        (1, 1),
```
jgmelber, line 107: (k, 1)

jgmelber (approved):

reference_designs/ipu-xrt/matrix_vector_multiplication/aie2.py
makslevental: i'm still figuring this one out... not landing yet

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -816,11 +837,22 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
         for(int k = 0; k < 8 /*size_0*/; k++)
           // access/store element at/to index (i * 16 /*stride_2*/ + j * 1 /*stride_1*/ + k * 2 /*stride_0*/)
     ```
+
+    ## Important gotcha regarding strides
+
+    All strides are expressed in multiples of the element width (just like `len` and `offset`)
+    **with the caveat that the inner-most dimension's stride must be a multiple of 32b**.
```
stephenneuendorffer, line 844: I know what you're trying to say, but this seems like its saying "Exists some k s.t. stride = k * 32b"   How about we just say "Addresses must be 32bit aligned"?

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -816,11 +837,22 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
...
+    I.e., `(stride_0 * element_width) % 32 == 0`. For example, if you are striding a `memref<512xi16>`,
+    `stride_0` cannot be 1 (or 3, or 5, ...). The reason for this requirement is stream access to DMA memory is 32b
```
stephenneuendorffer, line 846: rather than saying "cannot", I would say "stride_0" must be a multiple of 2.  if the element type is i8, then similarly the stride parameter must be a multiple of 4.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -816,11 +837,22 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
...
+    aligned (and thus, for the `memref<512xi16>` example, a stride of 1 would require 16b, half-word, access, which isn't possible).
+
   }];
 
   let arguments = (
     ins AnyMemRef:$buffer,
-        OptionalAttr<AIEI32Attr>:$offset,
+        // in words (multiples of element width) not bytes
```
stephenneuendorffer, line 853: 'words' is too ambiguous, IMO.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1547,93 @@ LogicalResult DMABDOp::verify() {
   if (!isa<BufferOp, ExternalBufferOp>(getBuffer().getDefiningOp()))
     return emitOpError(
         "BDs only support BufferOp or ExternalBufferOp operands.");
-  if (auto memOp = getOperation()->getParentOfType<MemOp>()) {
+
+  if (getLenInBytes() % 4)
+    return emitOpError("transfer length must be multiple of 4B (32b)");
```
stephenneuendorffer, line 1552: nit: the parameter is just a number so it's a multiple of 4.  Or, don't abbreviate Bytes to B.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1547,93 @@ LogicalResult DMABDOp::verify() {
...
+
+  if (getOperation()->getParentOfType<MemOp>()) {
+    auto memOp = getOperation()->getParentOfType<MemOp>();
     if (auto bufferOp = getBufferOp();
         bufferOp.getTileOp().colIndex() != memOp.colIndex() ||
         bufferOp.getTileOp().rowIndex() != memOp.rowIndex())
-      return emitOpError("can only access a buffer in the same tile.");
+      return emitOpError(
+          "Core tile DMAs can only access a buffer in the same tile.");
+    if (std::optional<int32_t> bdId = getBdId(); bdId.has_value() && *bdId > 15)
```
stephenneuendorffer, line 1561: Factor out magic constant 15, preferably to the TargetModel

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1547,93 @@ LogicalResult DMABDOp::verify() {
...
+      return emitOpError("Core tile DMAs have at most 16 buffer descriptors");
+    if (std::optional<int32_t> nextBdId = getNextBdId();
+        nextBdId.has_value() && *nextBdId > 15)
```
stephenneuendorffer, line 1564: And these should probably have negative tests.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1547,93 @@ LogicalResult DMABDOp::verify() {
...
+    auto memOp = getOperation()->getParentOfType<MemTileDMAOp>();
+    if (auto bufferOp = getBufferOp();
+        bufferOp.getTileOp().rowIndex() != memOp.rowIndex() ||
+        std::abs(bufferOp.getTileOp().colIndex() - memOp.colIndex()) > 1)
+      return emitOpError("Mem tile DMAs can only access adjacent mem tiles");
+    // TODO(max): check even channel -> <24; odd channel -> > 24
+    if (std::optional<int32_t> bdId = getBdId(); bdId.has_value() && *bdId > 47)
+      return emitOpError("Memtile DMAs have at most 48 buffer descriptors");
+    if (std::optional<int32_t> nextBdId = getNextBdId();
+        nextBdId.has_value() && *nextBdId > 47)
```
stephenneuendorffer, line 1586: Again, TargetModel for the magic numbers.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1545,60 +1547,93 @@ LogicalResult DMABDOp::verify() {
...
 
+    MemRefType buffer = getBuffer().getType();
+    int64_t maxIdx = 0;
     for (BDDimLayoutAttr dim : dims) {
       maxIdx += dim.getStride() * (dim.getSize() - 1);
       if (0 == dim.getStride())
         return emitOpError()
                << "Invalid step size; must be a positive integer.";
-      if (dim.getStride() > memrefSize)
+      if (dim.getStride() > buffer.getNumElements())
```
stephenneuendorffer, line 1621: These changes seem suspicious to me, since they seem to preclude having memref view transformations?   It's not clear to me why this is better, I guess.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -748,8 +736,9 @@ struct AIEObjectFifoStatefulTransformPass
         succ = builder.createBlock(endBlock);
 
       builder.setInsertionPointToStart(curr);
+      int offset = 0;
       if (isDistribute || isJoin)
-        offset = extraOffset * bytes;
```
stephenneuendorffer, line 752: Was this a functional error?  Do we have a test for the fix?

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -273,48 +273,63 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
       baseAddr += BASE_ADDR_A_INCR;
   }
 
   std::optional<llvm::ArrayRef<BDDimLayoutAttr>> dims = bdOp.getDimensions();
-  int lenInBytes = bdOp.getLenValue() * bytes;
-  int basePlusOffset = baseAddr + bdOp.getOffsetValue();
+  int lenInBytes = bdOp.getLenInBytes();
+  int basePlusOffsetInBytes = baseAddr + bdOp.getOffsetInBytes();
+  if (basePlusOffsetInBytes % 4)
+    return bdOp.emitOpError("bd address must be 4B (32b) aligned");
```
stephenneuendorffer, line 291: This seems like it should be in the verifier?

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -273,48 +273,63 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
+    dmaTileBdTensor.NumDim =
+        targetModel.isMemTile(tileLoc.Col, tileLoc.Row) ? 4 : 3;
     dmaTileBdTensor.Dim = static_cast<XAie_DmaDimDesc *>(
-        calloc(dims->size(), sizeof(XAie_DmaDimDesc)));
+        calloc(dmaTileBdTensor.NumDim, sizeof(XAie_DmaDimDesc)));
     if (!dmaTileBdTensor.Dim)
       return bdOp.emitError("couldn't allocate array of XAie_DmaDimDesc");
-    // TODO(max): rethink this?
+    // libxaie requires stride in multiples of 32b
+    double elementWidthIn32bWords =
```
stephenneuendorffer, line 304: double???

test/Targets/AIEGenerateXAIE/aie2_nd_DMA.mlir
```diff
@@ -15,11 +15,11 @@
 // CHECK: if(NULL == dma_tile_2_1_bd_0_tensor.Dim){
 // CHECK:   return 1;
 // CHECK: }
-// CHECK: dma_tile_2_1_bd_0_tensor.Dim[3].AieMlDimDesc = { /* StepSize */ 1, /* Size */ 2};
-// CHECK: dma_tile_2_1_bd_0_tensor.Dim[2].AieMlDimDesc = { /* StepSize */ 2, /* Size */ 3};
-// CHECK: dma_tile_2_1_bd_0_tensor.Dim[1].AieMlDimDesc = { /* StepSize */ 4, /* Size */ 2};
-// CHECK: dma_tile_2_1_bd_0_tensor.Dim[0].AieMlDimDesc = { /* StepSize */ 1, /* Size */ 1};
-// CHECK: __mlir_aie_try(XAie_DmaSetMultiDimAddr(&(dma_tile21_bd0), &dma_tile_2_1_bd_0_tensor, 0x82000,  /* len */ 128 * 4));
```
stephenneuendorffer, line 22: I think there's a somewhat significant loss of readability here...

stephenneuendorffer (commented):
This is a tricky change :)  Thanks for the effort!

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
makslevental: It wasn't an error - it's just no longer correct when `offset` is in terms of # of element rather than #number of bytes

lib/Targets/AIETargetCDODirect.cpp
makslevental: Base address is only know after the `aie-assign-buffer-addresses` pass but sure.

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -273,48 +273,63 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
+                                     elementWidthIn32bWords);
+      }
+      stride = stride > 0 ? stride : 1;
       // Assume AIE-ML architecture; we assert this above
-      // TODO(max): no we don't
-      dmaTileBdTensor.Dim[j].AieMlDimDesc = {dims.value()[i].getStride(),
-                                             dims.value()[i].getSize()};
+      dmaTileBdTensor.Dim[j].AieMlDimDesc = {stride, size};
+    }
+    for (size_t i = dims->size(); i < dmaTileBdTensor.NumDim; i++) {
```
makslevental, line 326: Assign defaults in the unassigned inner dims

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: this is wrong now.

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: This is wrong now - last stride must be 1.

lib/Dialect/AIE/IR/AIEDialect.cpp
makslevental: there's actually a verifyTraint/interface thing that does this check against the target model but dma_bd doesn't implement. Let me try to implement that interface instead of this.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -835,13 +863,21 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
   let hasVerifier = 1;
 
   let assemblyFormat = [{
-    `(` $buffer `:` type($buffer) (`,` $offset^)? (`,` $len^)? (`,` $dimensions^)? `)` attr-dict
```
AndraBisca, line 838: I think moving offset and len to the attribute dictionary should be part of another PR, centred around syntax.

AndraBisca (changes requested):

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: Sure

AndraBisca (approved):

PR #1112, opened 2024-03-08
Add LSP protocol server for AIE dialects

keryell:
This allows semantical features such as type information, auto-completion, etc. to IDE and editors using LSP.
See https://mlir.llvm.org/docs/Tools/MLIRLSP for use case.

To be used for example with Emacs:
```elisp
;; Enable MLIR mode
(require 'mlir-mode)
;; Prepare MLIR LSP mode
(load-file ".../llvm/mlir/utils/emacs/mlir-lsp-client.el")
(lsp-mlir-setup)
;; Think about customizing lsp-mlir-server-executable to use the server
;; with its own dialects.
; '(lsp-mlir-server-executable ".../mlir-aie/build/bin/aie-lsp-server")
;; Use lsp-mode in mlir-mode
(add-hook 'mlir-mode-hook 'lsp)
```


keryell:
I need to add some tests.

keryell:
@MaheshRavishankar as you are an Emacs user. :-)

fifield (changes requested):
I tried it in vscode and it worked. The new files need copyright/license.

keryell:
> I tried it in vscode and it worked. The new files need copyright/license.

Thanks for trying on your system.
Good catch for the copyright/license.

fifield (approved):

PR #1114, opened 2024-03-08
Add IpuSyncOp generation to AIEDmaToIpu

AndraBisca:
Designs that use ObjectFifos can rely on the ShimDMAAllocation ops generated by the ObjectFifo lowering to produce the corresponding IpuSync ops.


lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -419,6 +419,46 @@ struct DmaWaitToNpuPattern : OpConversionPattern<NpuDmaWaitOp> {
...
+          auto dir = builder.getI32IntegerAttr(0);
+          auto chan = builder.getI32IntegerAttr(infoOp->getChannelIndex());
+          auto col_num = builder.getI32IntegerAttr(1);
+          auto row_num = builder.getI32IntegerAttr(1);
+          builder.setInsertionPoint(returnOp);
+          builder.create<AIEX::NpuSyncOp>(dma->getLoc(), col, row, dir, chan,
+                                          col_num, row_num);
+        }
+      }
+    }
```
fifield, line 458: I don't think we want to unconditionally add a sync to every outgoing memcpy. For example, what if we are collecting N output tiles at the shim and only need to sync at the end? The N-1 extraneous syncs will have a performance penalty vs. the single (manually inserted) sync at the end.

programming_examples/basic/dma_transpose/aie2.py
```diff
@@ -59,7 +59,6 @@ def sequence(A, B, C):
                 npu_dma_memcpy_nd(
                     metadata="in", bd_id=1, mem=A, sizes=[1, K, M, 1], strides=[1, 1, K]
                 )
-                npu_sync(column=0, row=0, direction=0, channel=0)
```
fifield, line 62: I'd prefer that the tests keep the sync explicit, but using `aiex.npu.dma_wait` instead of `aiex.npu.sync`

fifield (changes requested):

programming_examples/basic/dma_transpose/aie2.py
fifield: > I'd prefer that the tests keep the sync explicit, but using `aiex.npu.dma_wait` instead of `aiex.npu.sync`

Should we close this PR in favor of https://github.com/Xilinx/mlir-aie/pull/1791 ?  Or is there still a desire to insert the sync/wait automatically?

PR #1119, opened 2024-03-11
Update quick_setup.sh to fix vanishing mlir_aie wheel

jgmelber:


jgmelber:
@makslevental do you have any idea why the other wheel would disappear?

PR #1123, opened 2024-03-13
Passthrough microbenchmark

andrej:


andrej:
Still a work in progress. I have yet to try @singagan's suggestions on how to get performance up.

PR #1129, opened 2024-03-13
Fixes for aie2xclbin on Windows

erieaton-amd:
* Fix runtime errors when using aie2xclbin
* Remove libc dependency

### DIFF
--- a/tools/aie2xclbin/XCLBinGen.cpp
+++ b/tools/aie2xclbin/XCLBinGen.cpp
@@ -299,17 +299,13 @@ static LogicalResult generateCoreElfFiles(ModuleOp moduleOp,
         std::string targetLower = StringRef(TK.TargetArch).lower();
         SmallVector<std::string, 10> flags;
         flags.push_back("-O2");
-        std::string targetFlag = "--target=" + targetLower + "-none-elf";
+        std::string targetFlag = "--target=" + targetLower;
         flags.push_back(targetFlag);
         flags.emplace_back(objFile);
         SmallString<64> meBasicPath(TK.InstallDir);
         sys::path::append(meBasicPath, "aie_runtime_lib", TK.TargetArch,
                           "me_basic.o");
         flags.emplace_back(meBasicPath);
-        SmallString<64> libcPath(TK.PeanoDir);
-        sys::path::append(libcPath, "lib", targetLower + "-none-unknown-elf",
-                          "libc.a");
-        flags.emplace_back(libcPath);
         flags.push_back("-Wl,--gc-sections");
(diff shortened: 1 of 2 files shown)


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -299,17 +299,13 @@ static LogicalResult generateCoreElfFiles(ModuleOp moduleOp,
         std::string targetLower = StringRef(TK.TargetArch).lower();
         SmallVector<std::string, 10> flags;
         flags.push_back("-O2");
-        std::string targetFlag = "--target=" + targetLower + "-none-elf";
+        std::string targetFlag = "--target=" + targetLower;
```
stephenneuendorffer, line 302: Should still have the 'none-elf' added here or it won't pick up the clang-rt IIRC.

stephenneuendorffer (approved):

tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -299,17 +299,13 @@ static LogicalResult generateCoreElfFiles(ModuleOp moduleOp,
...
         flags.push_back(targetFlag);
         flags.emplace_back(objFile);
         SmallString<64> meBasicPath(TK.InstallDir);
         sys::path::append(meBasicPath, "aie_runtime_lib", TK.TargetArch,
                           "me_basic.o");
         flags.emplace_back(meBasicPath);
-        SmallString<64> libcPath(TK.PeanoDir);
```
stephenneuendorffer, line 309: perhaps ifdef windows would be appropriate here, so we can still have full functionality on Linux?

tools/aie2xclbin/XCLBinGen.cpp
erieaton-amd: It looks in the wrong place for the builtins if I leave this as is:
`ld.lld: error: cannot open C:\s\llvm-aie\install\lib\clang\17\lib\libclang_rt.builtins-aie2.a: no such file or directory
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)`
The correct location is: `C:\s\llvm-aie\install\lib\clang\17\lib\aie2\libclang_rt.builtins.a`

stephenneuendorffer: The correct location should include the full triple "aie2-unknown-none-elf".  So, you need to have the full target triple as the target when you're building compiler-rt (and libc).  This path seems to indicate that that didn't happen...

erieaton-amd: The directions I got from Konstantin don't use the full triple, iirc.

PR #1131, opened 2024-03-15
Adapt LIT test configuration to new RyzenAI device name

keryell:
Starting with Linux 6.8 the format of the device returned by `xbutil examine` is like `[0000:66:00.1] : RyzenAI-npu1`.


makslevental:
> Starting with Linux 6.8 the format of the device returned by `xbutil examine` is like `[0000:66:00.1] : RyzenAI-npu1`.

Those strings (both the name and the id or whatever) appear in several places, and as I [mentioned last time it changed](https://github.com/Xilinx/mlir-aie/pull/975), I think the procedure should be to deprecate (as you've done in the lit config) rather than BC.

keryell:
I do not have the big picture. Just trying to run the tests on my work laptop.
It looks like nowadays the naming for RyzenAI NPU is along `npu1`/`npu2`/`npu4` according to the Linux driver from https://github.com/amd/xdna-driver/issues/3#issuecomment-1992751996

keryell:
> I think the procedure should be to deprecate (as you've done in the lit config) rather than BC.

@makslevental what do you mean by "BC"?

makslevental:
> > I think the procedure should be to deprecate (as you've done in the lit config) rather than BC.
> 
> @makslevental what do you mean by "BC"?

BC which is what I experienced after the last PR since I hadn't upgraded the driver/kernel combo.

PR #1137, opened 2024-03-19
Change `offset`, `len` syntax in `dma_bd`

makslevental:
Picked from https://github.com/Xilinx/mlir-aie/pull/1111

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -745,44 +745,66 @@ def AIE_DMABDPACKETOp: AIE_Op<"dma_bd_packet", []> {
   }];
 }
 
-def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
-  let summary = "Declare a dma block descriptor op";
+def AIE_DMABDOp: AIE_Op<"dma_bd", [
+    ParentOneOf<["MemOp", "MemTileDMAOp", "ShimDMAOp", "DMAOp"]>,
+  ]> {
+  let summary = "Declare a dma buffer descriptor op";
   let description = [{
-    This operation describes a block descriptor for DMA operations. In particular, it specifies
-    what buffer addresss to use, the transfer length, and the buffer type (A or B).
+    This operation describes a buffer descriptor for DMA operations. In particular, it specifies
+    what buffer to use, and optionally:
 
-    This operation must be used in an MLIR block that lives inside a MemOp's region.
-    The block descriptor specifies what lock to use and the buffer configuration.
+        1. the offset into the buffer;
+        2. the transfer length;
+        3. the sizes and strides for n-d tensor addressing (described below);
+        4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
+
+    `offset`, `len`, `size`s and `stride`s are all denominated in 32 bits; e.g., transferring the whole of
+    `memref<512xi32>` means `len == 512`, while transferring the whole of `memref<512xi16>` means `len == 1024`.
+
+    `dma_bd` ops must appear in their own bbs (basic blocks) and such bbs can (optionally) include `use_lock`
+    operations (specifying an "acquire" and a "release" lock; see the `use_lock` operation) and subsequent bds in
+    a "chain" of bds (using `next_bd` as a "jump" to the next bb which contains a `dma_bd`).
 
     Example:
     ```
       // this defines a BD that uses lock %lck0 and buffer %buf0
       ^bd5:
         aie.use_lock(%lck, "Acquire", 0)
-        aie.dma_bd(<$buf0 : memref<512xi32>, 0, 512>, 1)
+        // transfer the first 32 elements of the memref
+        aie.dma_bd(<$buf0 : memref<128xi32>, 0, 32)
         aie.use_lock(%lck, "Release", 1)
-        br ^bd6 // point to the next Block, which is also a different Block Descriptor
+        aie.next_bd ^bd6 // point to the next bb, which describes the next buffer descriptor
+      ^bd6:
+        aie.use_lock(%lck, "Acquire", 1)
+        // transfer the last 32 elements of the memref
+        aie.dma_bd(<$buf1 : memref<128xi32>, 96, 32)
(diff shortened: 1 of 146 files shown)


include/aie/Dialect/AIE/IR/AIEOps.td
andrej, line 1: I think GitHub does not let me add comments to lines you didn't change, but line 833 needs to be changed to use the new syntax for dims:


```
aie.dma_bd(%buf : memref<128xi32>, 0, 128, [<8, 16>, <2, 1>, <8, 2>])
```

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -835,13 +863,21 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
   let hasVerifier = 1;
 
   let assemblyFormat = [{
-    `(` $buffer `:` type($buffer) (`,` $offset^)? (`,` $len^)? (`,` $dimensions^)? `)` attr-dict
+    `(` $buffer `:` type($buffer) (`,` `dims` `=` $dims^)? `)` attr-dict
```
andrej, line 866: Do we still need `dims` in assemblyFormat? You also add "dims" as an attribute on line 894 below, does that make it part of `attr-dict`?

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -273,48 +273,57 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
+      uint16_t size;
+      uint32_t stride;
+      if (j > 0) {
+        stride = static_cast<uint32_t>(dims.value()[i].getStride() *
+                                       elementWidthIn32bWords);
+        size = dims.value()[i].getSize();
+      } else {
+        stride = dims.value()[i].getStride();
+        size = static_cast<uint16_t>(dims.value()[i].getSize() *
+                                     elementWidthIn32bWords);
```
andrej, line 317: It appears to me from this line we are making the strides be multiples of the memref data type, where previously it was always in 4-byte increments. (Correct me if wrong.) 

I'm assuming you all discussed this and agreed to make this change? The reason I bring this up is because we had a lively discussion about this back when I had my original PR for this feature, see [here](https://github.com/Xilinx/mlir-aie/pull/547#discussion_r1268836986). 

I suggested:

> So maybe, then, I should issue an error if multi-dim features are used on < i32 arrays. For > i32 arrays, we could have the dimensions be in units of elemental size. So for an i32 array, one unit of stride is 4 bytes, for an i64 arrays, one unit of stride is 8 bytes. This would be consistent with the "length" argument (but I believe the "offset" argument to the dmaBdOp is in bytes ...)

@fifield argued to not do this:

> In my opinion the wrap and step should expose the hardware registers directly without trying to do any conversion (e.g. no bytes -> words conversion). This is the cleanest and doesn't require casting to use multi-dimensional BDs in the general case (non-i32 data).

Just want to make sure you are aware of these previous discussions.

andrej:
Joe pointed me to this, I hope it's cool that I left three quick comments. Looks good

lib/Targets/AIETargetCDODirect.cpp
makslevental: This part of the PR already landed https://github.com/Xilinx/mlir-aie/pull/1111.

> It appears to me from this line we are making the strides be multiples of the memref data type, where previously it was always in 4-byte increments. (Correct me if wrong.)

You got it.

> I had my original PR for this feature, see https://github.com/Xilinx/mlir-aie/pull/547#discussion_r1268836986.

I wasn't aware such a discussion occurred (would've been nice to know - guess I should've checked...). 

Regarding prior opinions/arguments/claims, specifically

> **It doesn't matter if it's intuitive**. That's what the hardware does and at this level of aie dialect we should expose the hardware to the user.

I don't speak for everyone I'm sure but I think one of the cardinal sins of software is exposing unintuitive interfaces when you have the power to do otherwise. I mean like what's even the point of building an "abstraction" if you're just going to forward arguments/values untouched? Might as well just not build it at all and tell people to use the lower-level interface :shrug:.

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: This part of the PR, which is only about moving `offset` and `len` out into the `attr-dict` hasn't been decided on yet (and I haven't rebased either after the first part landed). But I'll try to remember that if it gets green-lit.

include/aie/Dialect/AIE/IR/AIEOps.td
makslevental: All attributes are always in the attribute dictionary (on the C++ side) but the reason I put dims here is because if you put them into `attr-dict` (i.e., tell the parser to parse them from there) then you have to annotate each size and stride with their type (`size: i32`, etc) where here you don't need to. You can do some extra work to actually get `i32`s automatically "inferred" even in the `attr-dict` (and I tried) but I just don't want to do all of that.

lib/Targets/AIETargetCDODirect.cpp
andrej: Okay I will leave it to you whether this warrants further discussion. I think Jeff's argument was that our MLIR dialect is not supposed to be a high level abstraction but instead give easy low level access to the hardware.

Why does it show up as a diff here if the PR is already merged though?

makslevental: > Why does it show up as a diff here if the PR is already merged though?

Because I haven't rebased (so its "base" is some previous commit in main before the merge).

fifield: >  I think Jeff's argument was that our MLIR dialect is not supposed to be a high level abstraction but instead give easy low level access to the hardware (i.e. allow use of data layout transformations that aren't multiples of the word size without the need for casts).

Yes that's right. I was mostly arguing against casts iirc, and not against intuitive abstractions but for dialect constructs representing the hardware, even if unintuitive. The intuitive interfaces then legalize to the low level (unintuitive) dialect, which is less intended for humans to produce/consume.

hunhoffe:
Closing as stale

PR #1140, opened 2024-03-19
Vck5000 IRON Support

eddierichter-amd:
Adding dynamic data movement support from IRON examples to VCK5000 AIE1


reference_designs/IRON-examples/matrix_add_one/run_vck5000.lit
```diff
@@ -0,0 +1,10 @@
+// (c) Copyright 2023 Advanced Micro Devices, Inc.
```
keryell, line 1: Happy new years everywhere?

reference_designs/IRON-examples/lit.local.cfg
```diff
@@ -0,0 +1,12 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023 AMD Inc.
+
+config.suffixes = ['.lit']
+config.unsupported = []
+
+if not config.has_libxaie:
+    config.unsupported = ['IRON-examples']
```
keryell, line 12: Perhaps it is a good opportunity to introduce some new variables here so instead of copy-pasting
`%python aiecc.py %VitisSysrootFlag% --host-target=%aieHostTargetTriplet% %link_against_hsa% ./aie.mlir -I%host_runtime_lib%/test_lib/include %extraAieCcFlags% %S/test_vck5000.cpp -o test.elf -L%host_runtime_lib%/test_lib/lib -ltest_lib`
all over the place we can just have
`%aiecc %iron_test`
or something like that?

eddierichter-amd: Good idea! We do need to add some logic of course to differentiate between the IPU and VCK5000 build instructions, but can encode that in %iron_test. I don't know if I want to put it in this PR as it is already incorporating a bunch of changes but if you want to take that in a separate branch that would be great :)

keryell: No emergency.
At least we could start thinking about how many different classes of tests we have first.

PR #1143, opened 2024-03-21
Simple bank-aware malloc scheme

AndraBisca:
This PR introduces a `SimpleBankAwareAllocation` scheme to `AIEAssignBuffers.cpp` that does round-robin allocations across available memory banks. This new memory allocation scheme will be enabled by default while the previous scheme, now called `BasicAllocationScheme`, can be enabled via the `--basic-alloc-scheme` flag in aiecc and `--basic-alloc` in the lowering pass using:
```
aiecc.py --basic-alloc-scheme
and
aie-opt --aie-assign-buffer-addresses="basic-alloc"
```

This PR also restructures `AIEAssignBuffers.cpp` such that future allocation schemes may be easier to add.

This is an improved version of #1108.


denolf (approved):
LGTM

PR #1148, opened 2024-03-22
[EXAMPLE] An element-wise add example

jamesroxbypb:
Here's eltwise add, vectorized and scalar versions

### DIFF
--- a/reference_designs/ipu-xrt/eltwise_add/aie2.py
+++ b/reference_designs/ipu-xrt/eltwise_add/aie2.py
@@ -15,10 +15,9 @@ from aie.extras.context import mlir_mod_ctx
 
 def my_eltwise_add():
 
-
     word_size_in = 2
     N = 65536
-    N_in_bytes = N * word_size_in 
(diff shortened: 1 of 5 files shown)


reference_designs/ipu-xrt/eltwise_add/aie2.py
```diff
@@ -0,0 +1,152 @@
...
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023 AMD Inc.
+
+import sys
+
+from aie.dialects.aie import *
+from aie.dialects.aiex import *
+from aie.dialects.scf import *
+from aie.extras.context import mlir_mod_ctx
```
keryell, line 13: I think it would be clearer to have something more understandable for beginners with less aggressive `import`.
For example just
```python
from aie.dialects import *
```
that would replace
```python
@core(cores[i], "add.o")
```
by
```python
@aie.core(cores[i], "add.o")
```
and so on.

PR #1149, opened 2024-03-22
Extend vectorization workaround to 8-bit types

newling:
With this change, (i8, i8) -> (i32) with vectorization seems to work e-2-e (from IREE)


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -622,8 +622,12 @@ struct RemoveAlignment2FromLLVMLoadPass
...
-        if (alignmentAttr && alignmentAttr.getValue() == 2)
-          loadOp.setAlignment(std::optional<uint64_t>());
+        if (alignmentAttr) {
+          int alignmentVal = alignmentAttr.getValue().getSExtValue();
+          if (alignmentVal == 2 || alignmentVal == 1) {
+            loadOp.setAlignment(std::optional<uint64_t>());
+          }
+        }
       }
     });
```
jsetoain, line 632: Do we really need this at all? With `memref.assume_alignment` on the buffer, `clang -O1` will propagate the right alignment to load/store operations.

newling: I guess it's possible that the memref.assume_alignment has been removed, since the last time I looked closely at the IR.

newling: At least [opt](https://github.com/Xilinx/mlir-aie/blob/5ef30765f7ee831d1e4d8876ceac88c63c4bd31b/tools/aie2xclbin/XCLBinGen.cpp#L799) and [llc](https://github.com/Xilinx/mlir-aie/blob/5ef30765f7ee831d1e4d8876ceac88c63c4bd31b/tools/aie2xclbin/XCLBinGen.cpp#L805) are -O2

jsetoain: Very odd, because we've verified that that's enough to propagate the alignment. `memref.assume_alignment` must be missing in these test cases. We can go back to it once I'm back from annual leave (or maybe jump on a quick call next week). In the meantime, it doesn't break anything and it's easy to revert once we figure it out. It's a +1 from me.

jsetoain (approved):

tools/aie2xclbin/XCLBinGen.cpp
newling: Hmm weird. I'll take a look to see of the assume_alignment has been removed, it'll be nice to remove pass completely if possible.

newling: image

That is the diff between the input.ll files. i.e. the outputs of  translateModuleToLLVMIR : https://github.com/Xilinx/mlir-aie/blob/d2334859d271ce43e55b4fc9c432182fb6dec239/tools/aie2xclbin/XCLBinGen.cpp#L692

On the left is with this workaround pass enabled (align 64 :-)) . On the right is without it (align 2 :-().

Here is some of the IR on the right:

```
  %41 = and i64 ptrtoint (ptr @buf2 to i64), 31
  %42 = icmp eq i64 %41, 0
  call void @llvm.assume(i1 %42)
  %43 = mul i64 %38, 512
  %44 = mul i64 %32, 32
  %45 = add i64 %43, %44
  %46 = getelementptr bfloat, ptr @buf2, i64 %45
  %47 = load <32 x bfloat>, ptr %46, align 2
```

It looks like the information determine that %47 could be aligned to 64 is there -- the llvm.assume shows that @buf2 is 64-bit aligned AFAICT. 

So I'm not too sure that the memref.assume_alignment is missing somewhere. 

When I run `opt` as follows:

```
../peano/install/bin/opt -O2 -S rhs.ll
```
 
 on the full IR, the loads remain of the form `  %101 = load <32 x bfloat>, ptr %100, align 2`. 
 
 Is the expectation that running `opt -O2` will replace loads with `align 2` with loads with `align 64` in the case where the operand to the load can be determined through `llvm.assume` operations to have 64-bit alignment? 
 
 (I suspect I've used x-bit alignment when I should have used x-byte alignment, but argument doesn't change).

jsetoain: > Is the expectation that running opt -O2 will replace loads with align 2 with loads with align 64 in the case where the operand to the load can be determined through llvm.assume operations to have 64-bit alignment?

Yes, it's the expectation and the behaviour I've observed... something odd is going on, I'll have to take a closer look.

PR #1151, opened 2024-03-23
Sort CMake configuration options and update actions/checkout

keryell:
This allows easier comparisons between the various CI configurations.
Use latest GitHub actions/checkout@v4
This fixes some deprecation warnings about old NodeJS.
Also fix some trailing space issues by side effect.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -102,28 +102,32 @@ jobs:
           mkdir build_assert
           cd build_assert
 
+          # Please keep the options sorted so it is easier to compare
           cmake .. \
             -GNinja \
-            -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-            -DCMAKE_EXE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_MODULE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_SHARED_LINKER_FLAGS_INIT="-fuse-ld=lld" \
-            -DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON \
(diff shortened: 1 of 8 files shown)


keryell:
@makslevental @fifield It is unclear to me what is happening here.
I this my fault or just some external planet misalignment?

makslevental:
https://github.com/Xilinx/mlir-aie/actions/runs/8398472772/job/23003372175?pr=1151#step:5:2301

> Cloning into '/home/runner/work/mlir-aie/mlir-aie'...
fatal: Remote branch **sort-cmake-options not found in upstream origin**
Error: Process completed with exit code 128.

I have said this many many many times: there is no point in people that are part of the org (acdc) sending PRs from forks. 

Beyond that I would appreciate it if you actually reverted the changes to `mlirDistro.yml` and `mlirAIEDistro.yml` (i.e., the two actions that triggered this fail) - they build the distro releases that downstream users depend on (i.e., **IREE**) and changing these yml triggers these rebuilds.

keryell:
> I have said this many many many times: there is no point in people that are part of the org (acdc) sending PRs from forks.

Sorry, Alzheimer is hitting me at my age. :-(
But otherwise, repository hygiene and preparing for community open-source inclusion come to my mind.

> Beyond that I would appreciate it if you actually reverted the changes to `mlirDistro.yml` and `mlirAIEDistro.yml` (i.e., the two actions that triggered this fail) - they build the distro releases that downstream users depend on (i.e., **IREE**) and changing the ymls triggers rebuilds.

Done. Interestingly this was just fixing some NodeJS deprecation.

> Finally, the orders of the CMake options weren't arbitrary - they were grouped by (CMake, LLVM, MLIR, AIE). I'm not opinionated about it but to me it was more important to distinguish between generic options (CMake), upstream (MLIR/LLVM) and our options.

I can understand but it is unclear this was fully the case and lexicographic sort is not that bad since the hierarchical prefixes enforce some locality even if it not perfect for some concepts crossing prefixes.
What does the team think about it?

PR #1152, opened 2024-03-25
Weight expand reference design

jamesroxbypb:
This is another reference example for expanding 4 bits weights with a bfloat16 scale factor shared across 32 elements.  This is suitable for accelerating vector/matrix multiplication by allowing us to read compressed weights from memory and then expand them to the compute data type (in this case bfloat16)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 #  -DAIE_RUNTIME_TARGETS: list of targets (x86_64,aarch64) to build runtime libs for, default: x86_64; cross compilation for aarch64 against default Vitis Sysroot
 #  -DAIE_RUNTIME_TEST_TARGET: runtime test target (x86_64 or aarch64) used for running unit test and tutorials, default x86_64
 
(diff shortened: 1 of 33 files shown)


CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 #  -DAIE_RUNTIME_TARGETS: list of targets (x86_64,aarch64) to build runtime libs for, default: x86_64; cross compilation for aarch64 against default Vitis Sysroot
 #  -DAIE_RUNTIME_TEST_TARGET: runtime test target (x86_64 or aarch64) used for running unit test and tutorials, default x86_64
 
-cmake_minimum_required(VERSION 3.21)
+cmake_minimum_required(VERSION 3.23)
```
makslevental, line 16: You need to rebase on top of [main](https://github.com/Xilinx/mlir-aie/blob/main/CMakeLists.txt#L16).

jgmelber:
Closed via #1158

Issue #1159, opened 2024-04-02
aie2xclbin failing with multi-column designs


erwei-xilinx:
A multi-column AIE design expects this aie_partition.json file to work:

```
      {
        "aie_partition": {
          "name": "QoS",
          "operations_per_cycle": "2048",
          "inference_fingerprint": "23423",
          "pre_post_fingerprint": "12345",
          "partition": {
            "column_width": 4,
            "start_columns": [
              1
            ]
          },
          "PDIs": [
            {
              "uuid": "00000000-0000-0000-0000-000000008025",
              "file_name": "./design.pdi",
              "cdo_groups": [
                {
                  "name": "DPU",
                  "type": "PRIMARY",
                  "pdi_id": "0x01",
                  "dpu_kernel_ids": [
                    "0x901"
                  ],
                  "pre_cdo_groups": [
                    "0xC1"
                  ]
                }
              ]
            }
          ]
        }
      }
    
```

The current json file only works for single-column designs: https://github.com/Xilinx/mlir-aie/blob/d2334859d271ce43e55b4fc9c432182fb6dec239/tools/aie2xclbin/XCLBinGen.cpp#L445

erwei-xilinx:
CC: @erieaton-amd , @newling

PR #1161, opened 2024-04-02
Modify multi-col aie_partition.json

erwei-xilinx:
Fixes issue https://github.com/Xilinx/mlir-aie/issues/1159

### DIFF
--- a/tools/aie2xclbin/XCLBinGen.cpp
+++ b/tools/aie2xclbin/XCLBinGen.cpp
@@ -443,7 +443,7 @@ static LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
       return moduleOp.emitOpError(errorMessage);
 
     std::string aie_partition_json_data = R"(
-      {
+      { 
         "aie_partition": {
           "name": "QoS",
           "operations_per_cycle": "2048",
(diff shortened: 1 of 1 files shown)


stephenneuendorffer (approved):

jgmelber:
I'm not sure this is a good change and limits the number of independent applications/processes that can use a device if we always build 4 column designs

jgmelber:
> I'm not sure this is a good change and limits the number of independent applications/processes that can use a device if we always build 4 column designs

Just to add, a desirable approach to application implementation could use multiple kernels within an XCLBIN or multiple XCLBINs for different functions (ie. GEMMs, Activations, various sizes).

erwei-xilinx:
> Just to add, a desirable approach to application implementation could use multiple kernels within an XCLBIN or multiple XCLBINs for different functions (ie. GEMMs, Activations, various sizes).

I agree. This PR is intended to do a short term fix where IREE-AMD-AIE will fail with any design using more than one AIE column, therefore blocking IREE dev progress. A long-term approach would be what you described here.

jgmelber:
> > Just to add, a desirable approach to application implementation could use multiple kernels within an XCLBIN or multiple XCLBINs for different functions (ie. GEMMs, Activations, various sizes).
> 
> I agree. This PR is intended to do a short term fix where IREE-AMD-AIE will fail with any design using more than one AIE column, therefore blocking IREE dev progress. A long-term approach would be what you described here.

What is the failure? If the design uses a tile in every column, then the metadata will enable 4 columns

erwei-xilinx:
> What is the failure? If the design uses a tile in every column, then the metadata will enable 4 columns

I created an issue on this: https://github.com/Xilinx/mlir-aie/issues/1159 Here is a gist summarizing how to reproduce the issue: https://gist.github.com/erwei-xilinx/f3d7ead1e9faad31fbf0d4ff7f82b18c

PR #1163, opened 2024-04-02
Supply more unique UUIDs for PDIs in aie2xclbin

erieaton-amd:
Windows driver expects these values to actually be unique, it causes a problem if you try to create multiple hardware contexts for multiple xclbins.

### DIFF
--- a/tools/aie2xclbin/XCLBinGen.cpp
+++ b/tools/aie2xclbin/XCLBinGen.cpp
@@ -441,7 +441,7 @@ static LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
         openOutputFile(aiePartitionJsonFile, &errorMessage);
     if (!aiePartitionJsonOut)
       return moduleOp.emitOpError(errorMessage);
-
+    uint64_t time = std::chrono::high_resolution_clock::now().time_since_epoch().count();
     std::string aie_partition_json_data = R"(
       {
         "aie_partition": {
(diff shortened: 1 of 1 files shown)


tools/aie2xclbin/XCLBinGen.cpp
```diff
@@ -441,7 +441,7 @@ static LogicalResult generateXCLBin(MLIRContext *context, ModuleOp moduleOp,
         openOutputFile(aiePartitionJsonFile, &errorMessage);
     if (!aiePartitionJsonOut)
       return moduleOp.emitOpError(errorMessage);
-
+    uint64_t time = std::chrono::high_resolution_clock::now().time_since_epoch().count();
```
stephenneuendorffer, line 444: Would hashing the design be more reasonable?

fifield: In aiecc.py we use a [random int](https://github.com/Xilinx/mlir-aie/blob/main/python/compiler/aiecc/main.py#L199) with a small range, which is not super collision resistant.  Maybe hash the time plus the design?  So that parallel builds (same time) or rebuilds of the same design (e.g. after a fix) don't collide?

kumardeepakamd:
> Windows driver expects these values to actually be unique, it causes a problem if you try to create multiple hardware contexts for multiple xclbins.

kumardeepakamd:
Should we not actually generate a uuid / guid by making a call such as  RPC_STATUS status = UuidCreate(&uuid);

stephenneuendorffer:
Obsoleted by #1513

PR #1164, opened 2024-04-03
[AMD-AIE] Fix ObjectFifoStatefulTransform to handle nested loops

Abhishek-Varma:
-- Currently ObjectFifoStatefulTransform pivots on wrong loop in
   case of nested loops and due to this the cloning operation
   in the `duplicateBlock` wrongly clones an ill-formed loop nest.
-- This commit adds a fix to ObjectFifoStatefulTransform pass to handle
   nested loops.

Signed-off-by: Jorn Tuyls <jorn.tuyls@amd.com>

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
@@ -848,6 +848,35 @@ struct AIEObjectFifoStatefulTransformPass
                       std::vector<std::vector<int>> &dependencies, Value base,
                       int64_t step, bool inLoop) {
     std::vector<Operation *> duplicatedOperations; // operations in current
+    // Recursive function to replace operands, uses recursion to handle nested
+    // loop structures.
+    std::function<void(Operation *, unsigned &, unsigned)> replaceOpsNested =
+        [&](Operation *op, unsigned &opIndex,
+            unsigned numDuplications) -> void {
+      if (auto loopOp = dyn_cast<scf::ForOp>(op)) {
+        Block *body = loopOp.getBody();
+        auto withoutTerminator = --body->end();
+        // NOTE(jornt): This only handles the cases where the nested scf::for is
+        // located at the start of the body. This should be the most common
+        // case, but is not fully generic.
+        if (auto nestedLoop = dyn_cast<scf::ForOp>(body->begin())) {
+          opIndex++;
+          auto clone = nestedLoop->clone();
+          replaceOperands(builder, clone, opIndex, base, step, inLoop,
+                          numDuplications, dependencies, duplicatedOperations);
+          replaceOpsNested(nestedLoop, opIndex, numDuplications);
+        } else {
+          for (auto loopBodyOp = body->begin(); loopBodyOp != withoutTerminator;
+               ++loopBodyOp) {
+            opIndex++;
+            replaceOperands(builder, &*loopBodyOp, opIndex, base, step, inLoop,
+                            numDuplications, dependencies,
+                            duplicatedOperations);
+          }
+        }
+      }
+    };
+
     // duplication iteration
     for (int i = 0; i < numDuplications; i++) {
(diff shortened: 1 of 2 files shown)


Abhishek-Varma:
CC: @jtuyls @MaheshRavishankar @erwei-xilinx

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -848,6 +848,35 @@ struct AIEObjectFifoStatefulTransformPass
...
+    // Recursive function to replace operands, uses recursion to handle nested
+    // loop structures.
+    std::function<void(Operation *, unsigned &, unsigned)> replaceOpsNested =
+        [&](Operation *op, unsigned &opIndex,
+            unsigned numDuplications) -> void {
+      if (auto loopOp = dyn_cast<scf::ForOp>(op)) {
+        Block *body = loopOp.getBody();
+        auto withoutTerminator = --body->end();
+        // NOTE(jornt): This only handles the cases where the nested scf::for is
+        // located at the start of the the body. This should be the most common
```
AndraBisca, line 860: Duplicated "the".

test/objectFifo-stateful-transform/nested_loop_test.mlir
```diff
@@ -0,0 +1,145 @@
+//===- nested_loop_test.mlir -----------------------------------------*- MLIR -*-===//
```
AndraBisca, line 1: It would be easier to understand the new functionality with a simpler test case, maybe with less object FIFOs (we're specifically looking at the ones for tile_1_2) and without all the toStream and fromStream attributes. It's a bit hard to read the test in its current state and I don't think all the attributes are needed to showcase the loop unrolling. Could you please make that change?

AndraBisca (changes requested):

test/objectFifo-stateful-transform/nested_loop_test.mlir
Abhishek-Varma: Hi @AndraBisca - I've made the changes. Could you re-review?

Also, I'm using ` ./utils/build-mlir-aie.sh llvm/build` to build MLIR-AIE and using the same command for running the lit test.

Strangely it's passing locally - any pointers?

AndraBisca: Thank you for making the changes! It's already so much easier to read and reason on! :) 

Did you also update your environment variables with `./utils/env_setup.sh`?

I will pull your branch locally and try it out.

Abhishek-Varma: Hi @AndraBisca - yes, I did do that.

Also, the run command I use is `./build/bin/aie-opt test/objectFifo-stateful-transform/nested_loop_test.mlir --aie-objectFifo-stateful-transform` - which works locally for me. So, I'm not really depending on `aie-opt` to be added as environment variable anyway.

Please let me know if you are able to reproduce the error - I'll try following the CI workflow to see if I can replicate it at my end. My hunch is that this might be because of Assert/Release enchilada.

AndraBisca: I have pulled the branch and was unable to reproduce the error, on my end the output is as expected in the FileCheck.

Abhishek-Varma: Oh okay. Thanks for trying @AndraBisca !

@makslevental @fifield would you happen to know the reason behind this issue and a way to reproduce the error that the CI is reporting locally?

I'm following the [installation instructions here](https://github.com/Xilinx/mlir-aie/blob/main/docs/Building.md#building-on-x86) and invoking the .sh scripts

jtuyls: @AndraBisca I was able to replicate the issue and create a fix. Could you have another look?

AndraBisca (approved):

PR #1167, opened 2024-04-03
test: build on hawkpoint

stephenneuendorffer:


.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -37,12 +37,13 @@ concurrency:
...
   DEBIAN_FRONTEND: noninteractive
   XILINXD_LICENSE_FILE: /opt/xilinx/Xilinx.lic
+  VITIS: /opt/xilinx/Vitis/2023.2
 
 jobs:
   build-repo:
     name: Build and Test with AIE tools on Ryzen AI
 
-    runs-on: amd7940hs
+    runs-on: amd8700g
```
fifield, line 46: Should we have `runs-on: [amd7940hs, amd8700g]` here?

utils/run_on_ipu.sh
```diff
@@ -3,6 +3,24 @@
...
+FIRMWARE_DIR=/lib/firmware/amdipu/1502
+# The AIE device model
+NPU_DEVICE=Phoenix
+if [[ -d /lib/firmware/amdnpu/1502_00 ]]; then
+  # New AIE driver naming and firmware location starting with Linux 6.8
+  linux_driver_version="6.8"
+  FIRMWARE_DIR=/lib/firmware/amdnpu/1502_00
+  NPU_DEVICE=NPU1
+fi
+
```
fifield, line 23: Is this still needed?

jgmelber: Not for now since we are falling back to the old (commonly used) version of the driver

.github/workflows/buildAndTestRyzenAI.yml
jgmelber: I think so

fifield: Sorry that was a swing and a miss by me. That syntax means match ALL the labels.  We could set it to `self-hosted`, but maybe we need a different label `npu` or `ryzenai` or whatever to encompass both

fifield: I created label `ryzenai` for both runner types

stephenneuendorffer: My sense is that unless you're really running the same OS image (e.g. a container), that having multiple runners on the same name is a bad idea.  Github just doesn't provide enough management/debug of runners to understand which particular instance of a runner one is running on.

stephenneuendorffer:
hawkpoint is running the wheels instead

fifield:
> hawkpoint is running the wheels instead

is there a disadvantage to not having a bigger pool that can do all the jobs?

PR #1175, opened 2024-04-04
Specify aie compiler for tests

fifield:
These fail if chess is not the default compiler


eddierichter-amd (approved):
Ahh good catch! Looks good to me.

PR #1208, opened 2024-04-10
Extracted more helper functions into test_utils and separated test.cpp and test.py versions

jackl-xilinx:


jackl-xilinx:
So I reviewed the helper/ convenience functions in xrtUtils and the additional ones that would also be helpful and consolidated them into test_utils.h/.cpp instead of xrtUtils.h/.cpp because theses functions are not strictly xrt related. I've updated vision kernels to match this (there's more changes needed to support it). 

Another choice I made was that in test_utils, I've also put it in its own namespace to avoid possible future conflicts. 

Finally, section-3 test.cpp/test.py have been simplified with run loop/timer and trace stuff moved to section-4. 

There may be some debate about how "simple" the new test.cpp/.py is. In particular, some feel that we should have more details about the xrt calls in the top level host code. But that seems like a lot of code that we can wrap up and prevent type-o's from creeping in. We can still refer to the xrt calls in test_utils and explain it but no need to have it in every test.cpp. I've also moved camelCase function names to underscore to make it more consistent with other helper functions.

jackl-xilinx:
> Is everyone aware that we have extremely simple in-tree Python bindings to XRT https://github.com/Xilinx/mlir-aie/blob/main/python/XRTModule.cpp#L146-L174 since ~December?
> 
> ```python
> from aie.xrt import XCLBin
> xclbin = XCLBin(xclbin_path, "MLIR_AIE")
> 
> xclbin.load_ipu_instructions(ipu_insts)
> [c] = xclbin.mmap_buffers([(K,)], np.int32)
> wrap_C = np.asarray(c)
> C = np.zeros((K,), dtype=np.int32)
> np.copyto(wrap_C, C, casting="no")
> 
> xclbin.sync_buffers_to_device()
> xclbin.run()
> print("Running kernel")
> xclbin.wait(30)
> xclbin.sync_buffers_from_device()
> ```

Sorry, Max. I was not aware. This looks really good for the python host code. I'll bring it up as something we can use. Thanks for pointing it out.

PR #1216, opened 2024-04-10
[ASPLOS][WIP] Passthrough kernel in basic examples

denolf:


AndraBisca (approved):
This looks good! I can point to the passthrough_kernel from the programming guide since it contains a very simple design with L3 to L1 and back data movement.

PR #1220, opened 2024-04-11
[AIEX] Add symbol-based IpuDmaWaitOp

jtuyls:
Adds a new AIEX `IpuDmaWaitOp`, which uses a global symbol to retrieve the necessary physical synchronization info like `column`, `channel`, `direction`. This new op is converted into `IpuSyncOp` inside the `DmaToIpu` pass. `IpuDmaWaitOp` should be easier to use than `IpuSyncOp` and abstracts away physical details as the latter requires users to already be aware of physical information, which might actually not be available yet. For example, a lot of reference designs use `IpuSyncOp` with channel set to `0` for output synchronization, however, in principal channel `1` could also be used and in general the exact channel has not been determined at the stage these reference designs intercept. 

See for example the syncs here: https://github.com/Xilinx/mlir-aie/blob/a74813ababb3d9631dc5483a94a10e6f2ffe401d/programming_examples/basic/matrix_multiplication/matrix_vector/aie2.py#L207, which use channel `0`, however, the operations to be synchronized don't reference a channel `0` yet and the exact channel hasn't been determined yet.

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -523,8 +523,7 @@ def AIE_IpuDmaWaitOp: AIEX_Op<"ipu.dma_wait", []> {
   let summary = "Blocking operation to wait for a DMA to complete execution.";
   let description = [{
     The IpuDmaWaitOp blocks until the DMA referenced through `symbol` completes execution
-    and issues a task-complete-token. NOTE(jornt): do we want to support synchronization
-    through polling as well?
+    and issues a task-complete-token.
 
     Example:
     ```mlir
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -478,7 +478,8 @@ def AIE_IpuDmaMemcpyNdOp: AIEX_Op<"ipu.dma_memcpy_nd", [
     symbol and specifies a descriptor `id` to be used, which will become the `bd_id` to be used
     when lowered further. The `issue_token` attribute specifies whether the execution of this
     operation should issue a token which can be received and read for synchronization purposes.
-    This `issue_token` attribute is set to `false` by default for `MM2S` for backward compatibility and **is always set to true for** `S2MM` channels.
+    This `issue_token` attribute is set to `false` by default for `MM2S` for backward compatibility
+    and **is always set to true for** `S2MM` channels.
   }];
 
   let arguments = (
@@ -515,6 +516,36 @@ def AIE_IpuDmaMemcpyNdOp: AIEX_Op<"ipu.dma_memcpy_nd", [
     std::array<unsigned, 3> $cppClass::getArrayAttrMaxRanks() { return {4, 4, 3}; }
   }];
 
+  let hasVerifier = 1;
+}
+
+def AIE_IpuDmaWaitOp: AIEX_Op<"ipu.dma_wait", []> {
+  let summary = "Blocking operation to wait for a DMA to complete execution.";
+  let description = [{
+    The IpuDmaWaitOp blocks until the DMA referenced through `symbol` completes execution
+    and issues a task-complete-token. NOTE(jornt): do we want to support synchronization
+    through polling as well?
+
+    Example:
+    ```mlir
+      ...
+      aie.objectfifo @out0(%tile_0_1, {%tile_0_0}, 4 : i32) : !aie.objectfifo<memref<32x32xi32>>
+      ...
+      aiex.ipu.dma_memcpy_nd(0, 0, %arg2[1, 1, 0, 0][1, 1, 32, 32][1, 1, 64]) {id = 0 : i64, issue_token = true, metadata = @out0} : memref<32x64xi32>
+      ...
+      aiex.ipu.dma_wait { symbol = @out0 }
+    ```
+    Here, we have an objectfifo with symbol name `out0`, which is then referenced in the
+    `ipu.dma_memcpy_nd` operation as the target for the respective DMA operation. Afterwards,
+    an `ipu.dma_wait` operation references the same symbol to block until the respective DMA
+    has executed all of its tasks.
+  }];
+  let arguments = (
+    ins FlatSymbolRefAttr:$symbol
+  );
+  let assemblyFormat = [{
+    attr-dict
+  }];
   let hasVerifier = 1;
 }
 
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -105,6 +105,16 @@ LogicalResult AIEX::IpuDmaMemcpyNdOp::verify() {
   return success();
 }
 
+LogicalResult AIEX::IpuDmaWaitOp::verify() {
+  auto dev = (*this)->getParentOfType<AIE::DeviceOp>();
+  if (!dev)
+    return emitOpError("couldn't find parent of type DeviceOp");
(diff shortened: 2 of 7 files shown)


include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -518,6 +519,36 @@ def AIE_IpuDmaMemcpyNdOp: AIEX_Op<"ipu.dma_memcpy_nd", [
   let hasVerifier = 1;
 }
 
+def AIE_IpuDmaWaitOp: AIEX_Op<"ipu.dma_wait", []> {
+  let summary = "Blocking operation to wait for a DMA to complete execution.";
+  let description = [{
+    The IpuDmaWaitOp blocks until the DMA referenced through `symbol` completes execution
+    and issues a task-complete-token. NOTE(jornt): do we want to support synchronization
```
makslevental, line 526: drive-by comments (will review thoroughly soon): I believe this note should be removed?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -105,6 +105,16 @@ LogicalResult AIEX::IpuDmaMemcpyNdOp::verify() {
   return success();
 }
 
+LogicalResult AIEX::IpuDmaWaitOp::verify() {
+  auto dev = (*this)->getParentOfType<AIE::DeviceOp>();
+  if (!dev)
+    return emitOpError("couldn't find parent of type DeviceOp");
+  auto sym = dev.lookupSymbol(getSymbol());
```
makslevental, line 112: same

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
```diff
@@ -335,6 +335,37 @@ struct DmaToIpuPattern : OpConversionPattern<IpuDmaMemcpyNdOp> {
...
+    if (!shimDmaAllocOp) {
+      op.emitOpError("couldn't find shim_dma_allocation op");
+      return failure();
+    }
+    auto channelDir = shimDmaAllocOp->getChannelDir();
+    int channel = shimDmaAllocOp->getChannelIndex();
+    int direction = (int)(channelDir == AIE::DMAChannelDir::MM2S);
+    int column = shimDmaAllocOp->getCol();
+
+    // Create with `column_num == 1` and `row_num == 1` to check for a single
```
makslevental, line 361: I guess this is reasonable (always setting both of these to 1) but I think the signal here is that actually `IpuSyncOp` just shouldn't expose these attributes (and then in the write32s set these to 1).

makslevental (commented):
Generally looks good - will come back (soon) for a closer look after it has settled in my brain...

include/aie/Dialect/AIEX/IR/AIEX.td
jtuyls: Yeah, I just wanted to bring this up. Is there any polling support right now or any interest in adding that?

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
jtuyls: Maybe it's ok in `IpuSyncOp`. I have to look at the details again, but you could use this as a `wait_all` I think? For example increase `col_num` to wait for multiple TCTs from different columns? A `DmaWaitAllOp` could be added to accommodate that if needed.

include/aie/Dialect/AIEX/IR/AIEX.td
makslevental: I don't know 100% but just yesterday I was discussing this with @benvanik - there is indeed interest and I will likely figure the how/when/where/why soon.

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
jtuyls: Ideally, `ipu_sync` wouldn't be used anymore directly as `dma_wait` should be more accessible and then `column_num` and `row_num` would only be set in a couple of places. We could also make these arguments optional and default to 1.

fifield: There's a similar `column_num` field in the `writebd_shimtile` instruction. None of these map to bits in hardware registers. Instead they are a firmware feature to write to multiple tiles with one instruction. I don't know of anyone using the them with mlir-aie.

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
```diff
@@ -335,6 +335,37 @@ struct DmaToIpuPattern : OpConversionPattern<IpuDmaMemcpyNdOp> {
...
+struct DmaWaitToIpuPattern : OpConversionPattern<IpuDmaWaitOp> {
+  using OpConversionPattern::OpConversionPattern;
+
+  DmaWaitToIpuPattern(MLIRContext *context, PatternBenefit benefit = 1)
+      : OpConversionPattern(context, benefit) {}
+
+  LogicalResult
+  matchAndRewrite(IpuDmaWaitOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto dev = op->getParentOfType<AIE::DeviceOp>();
```
makslevental, line 350: auto

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
```diff
@@ -335,6 +335,37 @@ struct DmaToIpuPattern : OpConversionPattern<IpuDmaMemcpyNdOp> {
...
+    auto shimDmaAllocOp = getAllocOpForSymbol(dev, op.getSymbol());
```
makslevental, line 351: auto

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
```diff
@@ -335,6 +335,37 @@ struct DmaToIpuPattern : OpConversionPattern<IpuDmaMemcpyNdOp> {
...
+    if (!shimDmaAllocOp) {
+      op.emitOpError("couldn't find shim_dma_allocation op");
+      return failure();
+    }
+    auto channelDir = shimDmaAllocOp->getChannelDir();
```
makslevental, line 356: auto

makslevental (approved):
LGTM modulo the few remaining `auto`s - please fix and then feel free to merge.

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
jtuyls: Changed it, but I think this would be an acceptable case in upstream as well as the type is clear from `<AIE::DeviceOp>`

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
jtuyls: Done

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
jtuyls: Done

lib/Dialect/AIEX/Transforms/AIEDmaToIpu.cpp
makslevental: Eh one reason I prefer the full type is because with these templated members sometimes you actually get a pointer rather a value so it doesn't actually match what's inside `<...>`.

PR #1221, opened 2024-04-11
Finish section 2c

AndraBisca:


programming_guide/section-2/section-2c/README.md
```diff
@@ -37,36 +37,36 @@ A data layout transformation is presented as a list of pairs, where each pair re
 ```
 Transformations can be expressed in up to three dimensions on each compute and Shim tile, and in up to four dimensions on Mem tiles. The first element of this array gives the outer-most dimension's stride and size, while the last element of the array gives the inner-most dimension's stride and size. All strides are expressed in <u>multiples of the element width</u>.
 
-Data layout transformations can be viewed as a way to specify to the hardware which location in the data to access next and it is possible to model the access pattern using a series of nested loops. For example, the transformation above can be expressed as:
+> **NOTE:**  The inner-most dimension's stride must be 1 by design.
```
AndraBisca, line 40: @jgmelber This line is part of the AIEOps.td description of `AIE_DMABDOp` but we have examples that do not have a stride of 1 for the inner-most dimension. Does this only apply to DMAs on specific tiles?

jgmelber: I think this restriction is relaxed but maybe only for 4B data types? @makslevental

programming_guide/section-2/section-2c/README.md
```diff
@@ -95,18 +96,25 @@ of0 = object_fifo
...
+        (
+            i * 16  # stride_1 
+            + j * 2 # stride_0
+        )
+```
+and further represented as in the image below,
+
+<img height="300" src="./../../assets/DataLayoutTransformation.svg">
+
+Other examples containing data layout transformations are available in the [programming_examples](../../../programming_examples/). A few notable ones are [matrix_vector_multiplication](../../../programming_examples/basic/matrix_multiplication/matrix_vector/) and [matrix_multiplication_single_column](../../../programming_examples/basic/matrix_multiplication/single_column/).
```
jgmelber, line 120: We deprecated the single_column design. Maybe highlight single_core or whole_array.

programming_guide/section-2/section-2c/README.md
```diff
@@ -10,11 +10,11 @@
...
 
-DMA blocks contain buffer descriptor operations that summarize what data is being moved, from what offset, how much of it, and in what layout. These buffer descriptors are the `AIE_DMABDOp` operations in MLIR and have their own auto-generated python binding (available under `/mlir-aie/install/python/aie/dialects/_aie_ops_gen.py` when the repository is built):
+DMA blocks contain buffer descriptor operations that summarize what data is being moved, from what offset, how much of it, and in what layout. These buffer descriptors are the `AIE_DMABDOp` operations in MLIR and have their own auto-generated python binding (available under `/mlir-aie/install/python/aie/dialects/_aie_ops_gen.py` after the repository is built):
```
jgmelber, line 17: Users may have different install directories depending on if they build the tools or use wheels or simply choose to. Can you be less specific about the exact install directory ie. <MLIR_AIE_INSTALL_PATH>

jgmelber (commented):
Looks good but I left a few comments.

programming_guide/section-2/section-2c/README.md
makslevental: Correct https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/IR/AIEDialect.cpp#L1610

AndraBisca: Thank you! I'll just clarify that in text.

PR #1226, opened 2024-04-12
Conv2d with ReLU Design

singagan:


jackl-xilinx (commented):
mlutils.py and xrtutils.py should be added to python/utils/ml.py and python/utils/xrt.py so you shouldn't need those files here.

PR #1229, opened 2024-04-12
[AIEX] Add IpuDmaWaitOp to standard lowering pass and fix device checks

jtuyls:
Makes some changes to get `IpuDmaWaitOp` through aiecc lowering. Adds the op to the `aiex-standard-lowering` pass and removes the device existence check from the verifier as `aie-standard-lowering` moves aiex ops outside of a device at some point (temporarily I think).

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -107,9 +107,9 @@ LogicalResult AIEX::IpuDmaMemcpyNdOp::verify() {
 
 LogicalResult AIEX::IpuDmaWaitOp::verify() {
   AIE::DeviceOp dev = (*this)->getParentOfType<AIE::DeviceOp>();
-  if (!dev)
-    return emitOpError("couldn't find parent of type DeviceOp");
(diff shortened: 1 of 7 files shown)


makslevental:
> removes the device existence check from the verifier

yea it "outlines" the func op containing https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp#L580

jtuyls:
> > removes the device existence check from the verifier as aie-standard-lowering moves aiex ops outside of a device at some point (temporarily I think).
> 
> yea it "outlines" the func op containing them https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp#L580.

I had a quick look and I don't really understand why it's doing that because in the end all these `aiex` ops in the function are removed and this function becomes empty, for example:

```
llvm.func @sequence(%arg0: !llvm.ptr, %arg1: !llvm.ptr, %arg2: !llvm.ptr) {
  llvm.return
}
```

makslevental:
> I had a quick look and I don't really understand why it's doing that 

:shrug:

Issue #1224, opened 2024-04-11
ObjectFifo: "aie.use_lock" op can only access a lock in the same tile


andrej:
The following code errors:
```
module {
  aie.device(ipu) {
    
    %tile_0_2 = aie.tile(0, 2)
    %tile_1_2 = aie.tile(1, 2)
    %tile_2_2 = aie.tile(2, 2)

    aie.objectfifo @a(%tile_2_2, {%tile_1_2}, 2 : i32) : !aie.objectfifo<memref<1024xbf16>>
    aie.objectfifo @b(%tile_1_2 toStream [<size = 32, stride = 1>], {%tile_0_2}, 2 : i32) : !aie.objectfifo<memref<32x32xbf16>>

    aie.objectfifo.link [@a] -> [@b]()

  }
}
```

with

```
error: unknown: 'aie.use_lock' op can only access a lock in the same tile
 note: unknown: see current operation: "aie.use_lock"(%10) <{acq_en = true, action = 2 : i32, value = 1 : i32}> : (index) -> ()
```

The error disappears when you remove the `toStream` from objectfifo b.

andrej:
@AndraBisca

andrej:
This one fails even without `toStream`:

```
module {
  aie.device(ipu) {
    %tile_0_2 = aie.tile(0, 2)
    %tile_2_2 = aie.tile(2, 2)
    %tile_3_2 = aie.tile(3, 2)

    aie.objectfifo @a(%tile_3_2, {%tile_2_2}, 2 : i32) : !aie.objectfifo<memref<1024xbf16>>
    aie.objectfifo @b(%tile_2_2, {%tile_0_2}, 2 : i32) : !aie.objectfifo<memref<1024xbf16>>

    aie.objectfifo.link [@a] -> [@b]()

  }
}
```

(Same error)

AndraBisca:
I'll double check this in more detail tomorrow but I think it's because you are using LinkOp with a shared tile that is a compute tile not a Mem tile. There's missing checks at that level that should verify whether the shared tile is not sharing memory with the compute tile, which will result in errors as the shared tile's DMA can't access the locks in the compute tile. It should disappear if you use tile (1, 1) in the first example instead of (1, 2). But I'll add actual checks to give you a more meaningful error ASAP

andrej:
Oh yeah, looks like the error does disappear when I use tile (1, 2). Thank you.

[...]

PR #1230, opened 2024-04-12
Always use DMAs when shared tile of link is compute tile

AndraBisca:
The Object FIFO lowering does not currently support LinkOps with a shared tile that is not a Mem tile. This PR adds that support by ensuring that when a compute tile is used as a shared tile the lowering will enforce going through DMAs  instead of shared memory, similar to how it currently does with Mem tiles. 

Additionally, checks have been put in place to ensure that processes running on the shared compute tile do not try to access the objectfifos which are part of the Link as this will **currently** interfere with the lock logic generated by the lowering.

This PR addresses Issue #1224


AndraBisca:
@andrej Could you please take a look at this PR and let me know if it is helpful to your designs? Or do you need to have more control over the choice of going through shared memory?

jgmelber (approved):

andrej:
Thank you Andra. I will give this a try ASAP

PR #1232, opened 2024-04-12
Add missing exports

stephenneuendorffer:
This enables running the tests on an installed area.


keryell:
Thanks. I had probably never tried to install the LSP server before using it.

PR #1233, opened 2024-04-12
Pjr move designs

jamesroxbypb:
Moving ReLU and adding a lit file


jamesroxbypb:
This is a simple move of 1 design from the basic to the ml directory, but also tests the bfloat16 support and adds a lit test for running on the device which should work

denolf (approved):

PR #1236, opened 2024-04-12
Convolution 2D with int8 support

singagan:


jackl-xilinx (commented):
In aie2.py, we're changing the order where mlir_mod_ctx() is not in the function def itself but at the end. See https://github.com/Xilinx/mlir-aie/blob/asplos/programming_examples/basic/passthrough_kernel/aie2.py#L103 for an example. Also, that example shows how we now call the trace wrapper (https://github.com/Xilinx/mlir-aie/blob/asplos/programming_examples/basic/passthrough_kernel/aie2.py#L75) so you can integrate that as well.

PR #1241, opened 2024-04-13
Correct vck5000 vector scalar lit file

eddierichter-amd:


programming_examples/basic/vector_scalar/test_vck5000.cpp
```diff
@@ -0,0 +1,134 @@
+//===- test.cpp -------------------------------------------------*- C++ -*-===//
+//
+// Copyright (C) 2020-2022, Xilinx Inc.
+// Copyright (C) 2022, Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
```
fifield, line 5: mlir-aie generally uses Apache license
```
// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
```

eddierichter-amd: Thanks for the catch will fix that right away. Was a little eager to get my fix in because it was failing tests but should have asked for a proper review.

PR #1242, opened 2024-04-15
End-to-End Bottleneck Design 

singagan:


AndraBisca:
I converted the PR to draft to avoid the notification spam from Github while you take care of the format.

PR #1270, opened 2024-04-16
Moved test_utils.py to utils/test.py

jackl-xilinx:


python/utils/ml.py
fifield, line 1: needs license/copyright

python/utils/test.py
```diff
@@ -0,0 +1,165 @@
+# test_utils.py -*- Python -*-
+#
+# Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
+# SPDX-License-Identifier: MIT
```
fifield, line 4: Should be Apache, not MIT

python/utils/xrt.py
fifield, line 1: needs license/copyright

python/utils/xrt.py
```diff
@@ -0,0 +1,154 @@
+# from npu.runtime
```
fifield, line 1: needs license/copyright

fifield (changes requested):

jgmelber:
@fifield can you check if this is ready to merge?

fifield (approved):

PR #1272, opened 2024-04-16
[ASPLOS] bring back run_on_ipu

denolf:
more run.lit in need of having run_on_ipu again


makslevental:
Not sure why this was changed (not tracking this branch) but `utils/run_on_ipu.sh` already has that env variable:

https://github.com/Xilinx/mlir-aie/blob/c1ec412bfd1082e8a1e266fc72a95f1fd4a7222b/utils/run_on_ipu.sh#L7

PR #1273, opened 2024-04-17
ResNet-50 with Offloaded Conv2_x Residual Blocks

singagan:


programming_examples/ml/resnet/layers_conv2_x/mwe.py
fifield, line 1: needs copyright

programming_examples/ml/resnet/layers_conv2_x/mwe.py
```diff
@@ -0,0 +1,545 @@
+import torch
+import torch.nn as nn
+import sys
+
+sys.path.append("../../utils")
```
fifield, line 5: You can now `import aie.utils.blah as whatever` after #1267

programming_examples/ml/resnet/layers_conv2_x/test.py
fifield, line 1: needs copyright

fifield (changes requested):
If you are going to use torch in tests it needs to be guarded in the `lit.cfg.py` with something like,
```python
try:
    import torch
    config.available_features.add("torch")
except:
    print("torch not found")
    pass
```
and a corresponding `REQUIRES: torch` in the lit test.

PR #1277, opened 2024-04-17
[asplos][wip] Update section-1 programming guide

denolf:


jackl-xilinx (approved):
I added my updates as well which has more extensive rewording of the module context.

PR #1289, opened 2024-04-17
Remove obsolete programming_examples

fifield:
It was said in hallway conversation that these are redundant, especially `programming_examples/basic/dynamic_dma_config_*` because there are now other examples for vck5000.


eddierichter-amd (approved):
Makes sense to me. Thanks for cleaning these up!

jgmelber (approved):

PR #1292, opened 2024-04-17
[ELTWISE] Add stdfloat header

jamesroxbypb:
Add stdfloat header


denolf:
@jamesroxbypb the dog is correct, #include <stdfloat> is already included in the [test_utils.h (lines 23-25)](https://github.com/Xilinx/mlir-aie/blob/asplos/runtime_lib/test_lib/test_utils.h#L23-L25)

denolf (changes requested):
no need to include <stdfloat>, it is already done in https://github.com/Xilinx/mlir-aie/blob/asplos/runtime_lib/test_lib/test_utils.h#L23-L25

denolf (approved):

jgmelber (approved):

PR #1296, opened 2024-04-18
[asplos] Passthrough dmas doc

AndraBisca:


jgmelber (approved):
Can you update from IPU to NPU?

AndraBisca:
> Can you update from IPU to NPU?

Done. There's a few other READMEs in the programming_examples that use IPU that we'll need to correct as well.

PR #1315, opened 2024-04-19
[AIE] Add Pure trait to TileOp

jtuyls:
Add the Pure trait to `AIE_TileOp` to allow cse to eliminate the operation.


makslevental:
I don't think this is a great idea (though it might be a good idea). Currently if you want to emit designs that are shifted you have to emit a "dummy" tile in col 0 and then your real design on whatever column. I haven't completely tracked down why this is but it has something to do with https://github.com/Xilinx/mlir-aie/pull/1106. 

Anyway maybe no one ever wants to do what I just described (emit a shifted design) but I found it useful for testing/debugging/etc. And if you land this then there will be no way to do that. 

An (obvious?) alternative is just a pass that prunes tiles with no uses.

jtuyls:
> Currently if you want to emit designs that are shifted you have to emit a "dummy" tile in col 0 and then your real design on whatever column. I haven't completely tracked down why this is but it has something to do with #1106.

Thanks, I wasn't aware of this. This means that if I want a design to start from column 1, in a partition which also includes column 0, I have to create a dummy tile on column 0? And if not done, column 1 would be used as column 0 in the design? That's pretty quirky.. I would expect that designs are created relative to column 0, so one starting from column 1 would include a column 0 as well.. Now, if a dummy tile is taking care of this all, I think it should not be too hard to avoid the need for that dummy tile and I would be happy to have a look at it. Is there an example that I could have a look at that demonstrates this?

> An (obvious?) alternative is just a pass that prunes tiles with no uses.

Yeah, I have a pass like this, but after a while you end up with a bunch of these passes for different ops, which you end up running back-to-back all the time. I think it should be part of cse instead if possible.

makslevental:
> Thanks, I wasn't aware of this. This means that if I want a design to start from column 1, in a partition which also includes column 0, I have to create a dummy tile on column 0? And if not done, column 1 would be used as column 0 in the design? That's pretty quirky.. I would expect that designs are created relative to column 0, so one starting from column 1 would include a column 0 as well.. 

Maybe we mean different things by "start" so let me preface with two things

1. By default (i.e., for everyone else but me) all designs are "virtualized" (not a word I made up but one I'll use) which means they are actually [launched *somewhere*](https://github.com/Xilinx/mlir-aie/blob/6354de5b430eb38b3d6c5762505633aa21fbf68b/python/compiler/aiecc/main.py#L197).
2. The physical col 0 isn't accessible (without https://github.com/Xilinx/mlir-aie/pull/1106)

That means by default, if you specify a single column design that only includes eg col 2, it will get placed somewhere (probably most often at col 1). If you want that design to actually be shifted 2 columns from the `start_col` then you need to put a dummy tile in col 0 (that does nothing). 

Hopefully that makes it clearer but if not we can hop on a call.

>Now, if a dummy tile is taking care of this all, I think it should not be too hard to avoid the need for that dummy tile and I would be happy to have a look at it. Is there an example that I could have a look at that demonstrates this?

It's not so simple because the parts are scattered between aiecc and CDO generation. The above linked PR fixes but @stephenneuendorffer doesn't agree with the implementation and is working on an alternative approach.

> Yeah, I have a pass like this, but after a while you end up with a bunch of these passes for different ops, which you end up running back-to-back all the time. I think it should be part of cse instead if possible.

Yea that makes sense. I'm fine merging this - I don't think anyone is emitting dummy tiles anyway in the way that I am.

jtuyls:
> That means by default, if you specify a single column design that only includes eg col 2, it will get placed somewhere (probably most often at col 1). If you want that design to actually be shifted 2 columns from the `start_col` then you need to put a dummy tile in col 0 (which will then be placed probably on col 1 and your col 2 will be placed at col 3).
> 
> Hopefully that makes it clearer but if not we can hop on a call.

Thanks for the clarification. I think we're talking about the same beast. I called this virtualized design a partition, which will be launched somewhere on the AIE array.

> It's not so simple because the parts are scattered between aiecc, target model, and CDO generation. The above linked PR fixes the dummy tile issue (in addition to enabling 5x4) but @stephenneuendorffer doesn't agree with the implementation and is working on an alternative approach.

I see, I need to have another more thorough look at that PR. But it seems like that dummy tile issue will be solved then in any case?

makslevental:
> I see, I need to have another more thorough look at that PR. But it seems like that dummy tile issue will be solved then in any case?

Yes. Anyway I'll approve this and we can come back if it becomes an issue.

test/dialect/AIE/tileop_cse.mlir
```diff
@@ -0,0 +1,26 @@
...
+
+// RUN: aie-opt --split-input-file --pass-pipeline="builtin.module(cse)" %s | FileCheck %s
+
+// CHECK: %[[TILE1:.*]] = aie.tile(1, 1)
+// CHECK-NOT: %[[TILE2:.*]] = aie.tile(1, 1)
+// CHECK: %[[CORE1:.*]] = aie.core(%[[TILE1]])
+// CHECK: %[[CORE2:.*]] = aie.core(%[[TILE1]])
+module {
+  %tile_1 = aie.tile(1, 1)
+  %tile_2 = aie.tile(1, 1)
```
makslevental, line 19: It's funny - this should be in a verifier somewhere (prevent duplicate tile definition).

makslevental (approved):

test/dialect/AIE/tileop_cse.mlir
jtuyls: Well, don't enable it globally.. I am creating cores with new tiles every time and then I use cse to get rid of the duplicates. Maybe it would be a good check within a `DeviceOp`.

stephenneuendorffer: I actually like this... will have to think a bit about the consequences.... @jorn would like to see more about how you're using this.

jtuyls: @stephenneuendorffer you tagged the wrong person.

I end up with duplicate tiles because I created duplicate cores (from peeling) and then eventually fuse sibling cores on the same tiles.

stephenneuendorffer: But, then don't you end up with a semantic question of what having 2 aie.core ops in the same tile means?  What meaning are you applying to that, or how are you doing the merging?  Got a repo we can look at?

jtuyls: There is still order at the point where I merge them. The meaning is that the core will first execute the operations in the first core block and then the operations in the latter one, so I just merge them into a single core block when they are siblings. https://github.com/nod-ai/iree-amd-aie/blob/c1915c6d54f463fc5bfe4e4612b4246cd90fdff3/compiler/plugins/target/AMD-AIE/iree-amd-aie/Transforms/AMDAIEFuseAieRegions.cpp#L98

PR #1345, opened 2024-04-20
[ASPLOS] Update top-level README IRON

jgmelber:


denolf (changes requested):
Can we rewrite this some more to help new users to pick the correct set of guides according to their background? Thanks.

denolf (approved):
Thanks @jgmelber

PR #1351, opened 2024-04-20
[asplos][PG section 2.b2] fix def skip connection

denolf:
Fixed definition of skip connection.

@abisca we might need some more explanation when the depth of an object fifo is given as an array. 

when a single number is used to specify the depth, like 3, it is the total depth between A and B

when an array is used one could write that as [0,3], [1,2] [2,1], [3,0] and it is not totally clear what the difference is ...


programming_guide/section-2/section-2b/02_Broadcast/README.md
```diff
@@ -29,7 +29,7 @@ The `depth` input of an Object FIFO can also be specified as an array of integer
...
 
-The main advantage of this feature comes to light during a situation like the one showcased in the example below, which we refer to as a broadcast with a <u>skip-connection</u>. In the example below two Object FIFOs are created: `of0` is a broadcast from producer tile A to consumer tiles B and C, while `of1` is a 1-to-1 data movement from producer tile B to consumer tile C. We refer to `of1` as a skip-connection because it is a dependency between the two consumer tiles of the same broadcast connection.
+The main advantage of this feature comes to light during a situation like the one showcased in the example below, which we refer to as a broadcast with a <u>skip-connection</u>. In the example below two Object FIFOs are created: `of0` is a broadcast from producer tile A to consumer tiles B and C, while `of1` is a 1-to-1 data movement from producer tile B to consumer tile C. We refer to `of0` as a skip-connection because since it skips over B in the A &rarr; B &rarr; C chain when connecting A &rarr; C.
```
AndraBisca, line 32: Thank you for adding more details @denolf ! I think either 'because' or 'since' should be removed here: ...skip-connection because / since it skips over B in the A &rarr; B &rarr; C chain when connecting A &rarr; C.

AndraBisca (approved):

PR #1368, opened 2024-04-22
Resnet with python binding

singagan:


jgmelber:
> Is the python version tested in `run.lit`?

I see it. Likely "running" as "unsupported"

PR #1370, opened 2024-04-22
Add torch to lit cfg for programming_examples

jgmelber:


programming_examples/lit.cfg.py
```diff
@@ -165,6 +165,14 @@
     opencv_flags = ""
 config.substitutions.append(("%opencv_flags", opencv_flags))
 
+try:
+     import torch
+
+     config.available_features.add("torch")
+ except:
```
makslevental, line 172: i recommend `except ImportError`. also there's an indent error here (you'll get an immediate parse error...)

fifield (approved):

PR #1378, opened 2024-04-23
[asplos] don't run resnet test

fifield:
this test breaks the runners


jgmelber:
Sorry. It looked like this was working, I guess I was wrong

jgmelber (approved):

PR #1383, opened 2024-04-23
fix log_hello_world check

fifield:
Fix the check to allow any column because in theory the test can run on any column. I observed this after another test caused the first column to become inaccessible.


jgmelber (changes requested):
We are deleting this test

PR #1384, opened 2024-04-23
fix makefiles

denolf:
Fix makefiles for reduce_ examples.

We can clean up the object file naming consistency later.


jackl-xilinx:
I have this fixed locally where instead of making %.cc.o : %.cc, I just changed the places where it looks for %.cc.o in aie2.py, etc. and make it look just for %.o. I haven't created a PR for it yet but is that a better way to do it than this fix?

PR #1386, opened 2024-04-23
2-D Array Transpose Using ObjectFifo 

HashimSharifAMD:


programming_examples/basic/vector_scalar_mul/Makefile
```diff
@@ -43,7 +43,7 @@ build/final_trace_${data_size}.xclbin: build/aie_trace.mlir build/scale.o
 	cd ${@D} && aiecc.py --aie-generate-cdo --no-compile-host --xclbin-name=${@F} \
 				--aie-generate-ipu --ipu-insts-name=insts_${data_size}.txt $(<:%=../%)
 
-${targetname}_${data_size}.exe: test.cpp
+${targetname}.exe: test.cpp
```
denolf, line 46: We want to keep the ${data_size} in the exe name to avoid deadlock when people change the data_size but did not do a make clean

HashimSharifAMD: Reverted back to this in a later commit in this PR

jgmelber (approved):
LGTM. This would be cool to extend with a few versions: transpose with the memtile, etc.

jgmelber:
Can you add a link to this in: https://github.com/Xilinx/mlir-aie/blob/asplos/programming_guide/section-5/README.md

PR #1390, opened 2024-04-23
Typos and Clarifications to ASPLOS Tutorial

andrej:
Joe asked me to give the ASPLOS tutorial a pass. Overall I thought this tutorial was well structured and gives a good introduction. 

If I could give one high-level general input for future iterations, it would be to pay particular attention to being concise, avoiding run-on sentences, and avoiding passive voice. Also, constructs like "in order to do X" or "it is the case that Y" can usually be left off without losing any meaning. The writing in some places may be easier to follow when following that advice. That being said, I still was able to follow everything nicely. Just thought I would voice this since I was asked to provide feedback!

As for the changes I am suggesting here, I think they should be uncontroversial, mostly typos, typography/formatting nitpicks, and I clarified (I hope?) a couple sentences. 

The reason I replaced `<u>...</u>` with `**...**` is that GitHub does not render the underline, so the emphasis is lost.

The most major change is that I added an introductory paragraph to section-2g. That section was very well written, but I found it seemed quite disconnected from the sections leading up to it, so I tried to tie it in a bit with some "glue sentences".  That section also didn't mention where to put the NPU operations, so I added a sentence about the "sequence" function. I also tied the discussion of data layout transformations back to their earlier discussion in section-2c.

I will open an issue for some of the things that were unclear to me that I didn't feel confident I could change by myself.


programming_guide/section-2/section-2d/README.md
```diff
@@ -18,7 +18,7 @@ ShimTile = tile(0, 0)
 MemTile = tile(0, 1)
 ComputeTile = tile(0, 2)
 ```
-For our scale out design we will keep a using a single Shim tile and a single Mem tile but we will increase the number of compute tiles to three. We can do so cleanly and efficiently in the following way:
+For our scale out design we will keep a using a single Shim tile and a single Mem tile, but we will increase the number of compute tiles to three. We can do so cleanly and efficiently in the following way:
```
AndraBisca, line 21: Just noticed this: ..keep a using a single > keep using a single

programming_guide/section-2/section-2g/README.md
```diff
@@ -10,6 +10,10 @@
 
 # <ins>Section 2g - Runtime Data Movement</ins>
 
+In the preceding sections, we looked at how we can describe data movement between tiles *within* the AIE-array. However, to do anything useful, we need to get data from outside the array, i.e. from the "host", into the AIE-array and back. On NPU devices, we can achieve this with the operations described in this section. 
+
+The operations that will be described in this section must be placed in a separate `sequence` function. This arguments to this function describe buffers that will be available on the host side; the body of the function describes how those buffers are moved into the AIE-array. [Section 3](../../../programming_examples/) contains an example.
```
AndraBisca, line 15: This arguments to this function... > The arguments to this function

AndraBisca (approved):
Thank you so much for taking the time to make a PR with all of these great improvements and typo fixes. I only read through the changes in Section 2 so far and I only noticed two additional typos. This is really helpful, thank you again Andre! :)

andrej:
Of course! Writing all of this must have been a ton of work.

denolf (approved):
Thanks for all these fixes!

jgmelber (approved):

PR #1387, opened 2024-04-23
[asplos] enable lit testing of Makefile

denolf:
example on how to run lit testing on the Makefiles for passthrough_kernel

PR #1397, opened 2024-04-24
[program examples] lit test on Makefile

denolf:
repeat of #1387

example of makefile lit testing


fifield:
* shouldn't need to make clean first
* should be using `%run_on_npu` on the `make run` line

PR #1407, opened 2024-04-25
Cleaning up VCK5000 makefiles and removing test_vck5000.cpp and run_vck5000.lit for tests hat don't have the appropriate kernels

eddierichter-amd:


fifield (changes requested):
It seems to have broken the ryzenai ci, I'll take a look

fifield:
> It seems to have broken the ryzenai ci, I'll take a look

There are additional errors if peano is the default compiler.  Maybe we can add that to the internal CI.

fifield (approved):

jgmelber (approved):

PR #1429, opened 2024-04-26
Adding calls to use the basic malloc scheme for tests that are failing with the new one

eddierichter-amd:


AndraBisca (approved):
Ah I must've missed those somehow. Thank you for adding them! :)

PR #1431, opened 2024-04-26
Compile time optimization 

newling:
This is for a large matmul in the SD3 model compiling e2e. IREE -> mlir-air -> mlir-aie -> llvm-aie -> 0100110..

```
 run_matmul_test --name_prefix sd3_mm8 --lhs_rhs_type bf16 --acc_type f32 --m 8192 --k 2432 --n 9728 --expect-compile-failure 0 --compile-only 1
 ```
 
Before
```
real    10m39.222s
user    9m27.321s
sys     0m2.217s
```

After
```
real    0m7.197s
user    0m3.882s
sys     0m0.613s
```


erwei-xilinx (approved):
LGTM!

PR #1434, opened 2024-04-29
remove duplicate pass in aiecc

Yu-Zhewen:


AndraBisca:
I'm not sure I see how these passes are duplicates. Could you please elaborate further on how the changes tie into each other?

Yu-Zhewen:
I've noticed that both `aie-lower-broadcast-packet` and `aie-create-packet-flows` are being invoked twice in our current implementation—initially within `INPUT_WITH_SWITCHBOXES_PIPELINE` and secondly in `process_host_cgen`.

The current side effect is the packet routing happens way before circuit routing, which I believe is an undesirable behaviour since we may want to merge these two routing passes together in the furture?

After the change, both passes are only invoked once.

Yu-Zhewen:
oh yes, updated now. Thanks!

AndraBisca (approved):
Looks good to me now!

PR #1437, opened 2024-04-30
[aievec] Move trunc/ext to common lowering

jsetoain:
Currently, trunc/ext ops get lowered in a separate pass with two simple patterns. This is mostly because other patterns check their users for these ops, which makes all the patterns involved dependent on each other and their order of application.

This patch addresses this issue and moves these two last patterns together with all the other vector-to-aievec patterns.


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -46,6 +46,49 @@ using namespace xilinx::aievec;
 // Utility functions
 //===----------------------------------------------------------------------===//
 
+static bool isNarrowingOp(Operation *op) {
+  if (isa<arith::TruncFOp>(op) || isa<arith::TruncIOp>(op))
+    return true;
+
+  if (auto srsOp = dyn_cast<aievec::SRSOp>(op)) {
+    auto srsOpSrcOp = srsOp.getSource().getDefiningOp();
+    if (isa<aievec::UPSOp>(srsOpSrcOp) || isa<aievec::CastOp>(srsOpSrcOp))
```
david-vc, line 55: Suppose that we have UPS from int8 => int32 followed by SRS from int32 => int16.
Would it be OK to return true in this case?

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -46,6 +46,49 @@ using namespace xilinx::aievec;
...
+    return extUIOp.getIn();
+  if (auto extFOp = src.getDefiningOp<arith::ExtFOp>())
+    return extFOp.getIn();
+  if (auto srsOp = src.getDefiningOp<aievec::SRSOp>()) {
+    // Conversion through AIE intrinsics takes two steps:
+    //     1) Load to accumulator: aievec.ups
+    //     2) Move from accumulator: aievec.srs
+    auto srsSource = srsOp.getSource();
+    if (srsSource)
+      if (auto upsOp = srsSource.getDefiningOp<aievec::UPSOp>())
```
david-vc, line 77: Suppose that we have UPS from int16 => int32 followed by SRS from int32 to int8.
IMO, we should return sct::optional<Value>{} in this case.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2757,6 +2799,98 @@ static void configureAIEVecCommonLegalizations(ConversionTarget &target,
     target.addIllegalOp<vector::TransferReadOp>();
   }
   target.addIllegalOp<vector::ExtractStridedSliceOp>();
+
+  // ****************************NEW****************************
```
david-vc, line 2803: Line 2803 is probably unintended for this PR.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2757,6 +2799,98 @@ static void configureAIEVecCommonLegalizations(ConversionTarget &target,
...
+  target.addDynamicallyLegalOp<arith::ExtFOp>([](arith::ExtFOp extfOp) {
```
david-vc, line 2804: So basically this is saying that extf ops are legal, unless that they are from v16 bfloat16  to v16 f32.

david-vc (commented):
Looks good to me, but I did have a few questions in the comments.

jamestcl-amd (commented):
The test for this conversion pattern is missing from https://github.com/Xilinx/mlir-aie/tree/main/test/Conversion/VectorToAIEVec... It might be beneficial to create it this time.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2637,14 +2676,17 @@ static void populateAIEVecV1ConversionPatterns(RewritePatternSet &patterns,
...
-                                                  1024, 256, 1024);
-  }
   // clang-format off
   // TODO: Reorder these alphabetically
+  if (backend == TargetBackend::CPP) {
+    patterns.add<
+        LowerVectorTransferReadToAIEUPD
+      >(patterns.getContext(), 128, 1024, 256, 1024);
+    patterns.add<
+        LowerVectorAddIOpToAIEVecAddElemOp
```
jamestcl-amd, line 2686: We should move `LowerVectorSubIOpToAIEVecSubElemOp` to here as well.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: I don't think we should because that's not a narrowing operation (int8 -> int16), but in this particular case it's looking for an `SRS` followed by either a `UPS` or a `aievec.cast`, which are the two possible conversion for `trunc` ops, depending on the operand types.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: Indeed, that would be the thing to do since that's a narrowing operation, not a widening one. But we're not looking for `UPS` -> `SRS` sequences here, only `SRS` -> `UPS`.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: Aren't those applicable for LLVMIR conversion as well?

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: I honestly don't know what the logic is here, I haven't changed the semantics of what we do with these ops, just where it happens and what assumptions they're making about order and other patterns.

It'd probably be a good idea, at some point, to take a good look at all these casts (and movements between accumulator/regular register files) and see what needs to be done. There's probably a sensible way to not do any of this, and simplify the conversion.

PR #1440, opened 2024-04-30
[Initial] Enable more XRT buffers

jgmelber:
Extends the generated metadata `kernels.json` to enable up to 6 buffer objects for XRT. This is hardcoded where it was previously 3, this should be improved to match the MLIR source file.


nirvedhmeshram:
what happens when the mlir source file has less buffers than the kernel.json? Not sure if mlir-aie has e2e cpp tests showing how the runtime xrt calls will change with this change?

jgmelber:
> what happens when the mlir source file has less buffers than the kernel.json? Not sure if mlir-aie has e2e cpp tests showing how the runtime xrt calls will change with this change?

You can always use less if you want to. We have a some designs with fewer memref args in the `func.func @sequence` block and they work fine. The XRT cpp/py side can also have less than 6 and work just fine. In the CI tests just now, many have 2 or 3 buffer objects and passed using this new 6 buffer object `kernels.json` metadata

nirvedhmeshram (approved):
LGTM!

PR #1443, opened 2024-05-01
Extend target model to properly support different phoenix models

stephenneuendorffer:


jgmelber:
> There's a lot I don't get here (e.g., why the constructor `NPUmodel2col(2)` both has `2` in the name but also takes `2` as an arg and also why there's no `NPUmodel5col(5)` i.e., the original intent of my patch) but mostly I'm just wondering what happens for V70 which has 16 columns? Do we need new device models `[V70model{i}col({i}) for i in range(16)]`?

V70 is xcve2802

jgmelber:
This is a bit more complicated but being explicit with how many columns a design is targeting for NPU does make more sense. @makslevental makes a good point about larger devices that are virtualized by XRT, needing to support 5, 6, 7, ... column models gets messy

stephenneuendorffer:
> There's a lot I don't get here (e.g., why the constructor `NPUmodel2col(2)` both has `2` in the name but also takes `2` as an arg and also why there's no `NPUmodel5col(5)` i.e., the original intent of my patch) but mostly I'm just wondering what happens for V70 which has 16 columns? Do we need new device models `[V70model{i}col({i}) for i in range(16)]`?

On larger devices we have today, the shim tiles are not regular, so I expect that having an arbitrary number of columns isn't that interesting. For instance, on VCK190, 4x8 with 2 shim tiles is a convenient tile size.  Or maybe 4x4 with 1 shim tile.   For larger devices I expect that powers of 2 are pretty interesting, but "13 columns" is probably less interesting?    So I guess for the moment, "I don't know", but I hope this is a step in a reasonable direction.

stephenneuendorffer:
> why there's no `NPUmodel5col(5)` i.e., the original intent of my patch

Because there's no way of implementing 5 columns with 5 shimDMAs.  If you want the whole device, then thats AIEDevice.npu1 and you have to take into account the assymmetry.

jgmelber:
> > There's a lot I don't get here (e.g., why the constructor `NPUmodel2col(2)` both has `2` in the name but also takes `2` as an arg and also why there's no `NPUmodel5col(5)` i.e., the original intent of my patch) but mostly I'm just wondering what happens for V70 which has 16 columns? Do we need new device models `[V70model{i}col({i}) for i in range(16)]`?
> 
> On larger devices we have today, the shim tiles are not regular, so I expect that having an arbitrary number of columns isn't that interesting. For instance, on VCK190, 4x8 with 2 shim tiles is a convenient tile size. Or maybe 4x4 with 1 shim tile. For larger devices I expect that powers of 2 are pretty interesting, but "13 columns" is probably less interesting? So I guess for the moment, "I don't know", but I hope this is a step in a reasonable direction.

If this discussion is around virtualizing segments in larger discrete devices, the mlir-air dialect has concepts `segment` for that. And yes, for xcvc1902 (vck190/vck5000) the 8x8 with two shims under the leftmost columns is the most regular segment shape. I have not analyzed the xcvd2802 (v70) yet but a similar shapes likely fall out.

jgmelber (changes requested):
Most of the programming examples are single core or single column. I don't think by default they should be targeting device `npu1_4col`

stephenneuendorffer:
> Most of the programming examples are single core or single column. I don't think by default they should be targeting device `npu1_4col`

True, I switch them to 4col because that is the closes to what we had before.

jgmelber:
> > Most of the programming examples are single core or single column. I don't think by default they should be targeting device `npu1_4col`
> 
> 
> 
> True, I switch them to 4col because that is the closes to what we had before.

IIRC for the 'aiecc.py' flow it would determine the number of columns in the metadata based on the design so most of these would be 'npu1_1col'.

jgmelber (approved):

PR #1446, opened 2024-05-01
[aievec] to-llvm flow for aievec.broadcast_scalar op

jamestcl-amd:
This PR add the support for `aievec.broadcast_scalar` op going through the to-llvm flow.

Changes:
- Add aievec-to-llvm conversion pattern/tests for the `aievec.broadcast_scalar` op.
- Add op verifier for the `aievec.broadcast_scalar` op.
- Add target external llvm translation tests.


lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -1257,16 +1266,17 @@ class BroadcastScalarOpConversion
       unsigned srcBitWidth = srcType.getIntOrFloatBitWidth();
 
       if (srcBitWidth < 32) {
-        src = rewriter.create<LLVM::SExtOp>(loc, rewriter.getI32Type(),
-                                            adaptor.getSource());
+        src = rewriter.create<LLVM::SExtOp>(loc, rewriter.getI32Type(), src);
       } else if (srcBitWidth > 32) {
-        src = rewriter.create<LLVM::TruncOp>(loc, rewriter.getI32Type(),
-                                             adaptor.getSource());
+        src = rewriter.create<LLVM::TruncOp>(loc, rewriter.getI32Type(), src);
```
david-vc, line 1271: What are the valid use cases where the source of the broadcast has bitwidth > 32 and truncating it to 32 bits be preserve semantics?

david-vc (approved):
LGTM.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
jamestcl-amd: Nice catch. We have limited the source scalar to be in `BF16, F32, I32, I16, I8` types. I will remove this line.

PR #1449, opened 2024-05-02
Add aie-visualize tool

stephenneuendorffer:
This tool generates a simple visualization of a design.


keryell:
Yes, C99 VLA are not even in plain C++ and they have been made optional in C11 because there were not well implemented in Microsoft compiler, if I remember correctly.

PR #1452, opened 2024-05-03
Add missing pipeline to python/util.py

AndraBisca:


include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -27,7 +27,7 @@ def AIEAssignBufferAddresses : Pass<"aie-assign-buffer-addresses", "DeviceOp"> {
   let constructor = "xilinx::AIE::createAIEAssignBufferAddressesPass()";
 
   let options = [
-    Option<"clBasicAlloc", "basic-alloc", "bool", /*default=*/"false",
+    Option<"clBasicAlloc", "basicAlloc", "bool", /*default=*/"false",
```
makslevental, line 30: That's not going to work - the `Pipeline` class turns `_` into `-` https://github.com/Xilinx/mlir-aie/blob/main/python/extras/runtime/passes.py#L126

AndraBisca: Yeah, I noticed it does that, which is why I removed the `-`. Or did I misunderstand your comment?

PR #1455, opened 2024-05-03
bump llvm

fifield:
update llvm to 1cde124
account for [expand_shape changes](https://github.com/llvm/llvm-project/commit/97069a86193a617a9e4cf742a29db6116b2bf449)

### DIFF
--- a/test/dialect/AIEVec/precanonicalization.mlir
+++ b/test/dialect/AIEVec/precanonicalization.mlir
@@ -58,7 +58,7 @@ func.func @rank_zero_transfer_read(%m : memref<i16>) -> vector<16xi16> {
     %c0_i16 = arith.constant 0 : i16
     // CHECK-DAG: %[[C0idx:.*]] = arith.constant 0 : index
     // CHECK-DAG: %[[C0i16:.*]] = arith.constant 0 : i16
-    // CHECK-DAG: %[[EXPMEM:.*]] = memref.expand_shape %[[MEM]] [] : memref<i16> into memref<1xi16>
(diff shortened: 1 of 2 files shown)


jsetoain:
Yes, I was going to say. Between the version I originally used to replicate the other issues and the latest one, another change broke another test 😬

jsetoain (approved):

jsetoain:
> Yes, I was going to say. Between the version I originally used to replicate the other issues and the latest one, another change broke another test 😬

@makslevental It looks like we're not running the integration tests any more (the ones activated with `-DAIE_INCLUDE_INTEGRATION_TESTS=ON`), that's why CI didn't catch the issues.

fifield:
> > Yes, I was going to say. Between the version I originally used to replicate the other issues and the latest one, another change broke another test 😬
> 
> @makslevental It looks like we're not running the integration tests any more (the ones activated with `-DAIE_INCLUDE_INTEGRATION_TESTS=ON`), that's why CI didn't catch the issues.

It looks like I wasn't running those tests either. I get a few failures after turning them on.

jsetoain:
> > > Yes, I was going to say. Between the version I originally used to replicate the other issues and the latest one, another change broke another test 😬
> > 
> > 
> > @makslevental It looks like we're not running the integration tests any more (the ones activated with `-DAIE_INCLUDE_INTEGRATION_TESTS=ON`), that's why CI didn't catch the issues.
> 
> It looks like I wasn't running those tests either. I get a few failures after turning them on.

The same happened to me due to recent dev. env shenanigans. Anyway, the patch I sneaked right before this one fixes those issues, but it looks like after rebase a test failed. I suspect it was one of those spurious failures we're getting later because the previous run finished successfully, might be worth looking into it.

PR #1460, opened 2024-05-07
Remove platform dependent linking in aie2xclbin

erieaton-amd:
This shouldn't be needed anymore since we figured out how to generate the libc.a on Windows.


stephenneuendorffer (approved):
Yes! Yes! A thousand times Yes!

newling (approved):

PR #1467, opened 2024-05-08
[CI] parallelize tests

stephenneuendorffer:
The tests run in about 12 minutes..  The programming examples take almost as long.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -41,7 +41,7 @@ env:
 
 jobs:
   build-tests:
-    name: Build and Test with AIE tools on Ryzen AI
+    name: Run Tests
 
     runs-on: amd7940hs
 
@@ -174,7 +174,7 @@ jobs:
           popd
 
   build-examples:
-    name: Build and Run Examples with AIE tools on Ryzen AI
+    name: Run Examples
 
     runs-on: amd8700g
 
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -38,6 +38,20 @@ concurrency:
 env:
   DEBIAN_FRONTEND: noninteractive
   XILINXD_LICENSE_FILE: /opt/xilinx/Xilinx.lic
+  VITIS: /opt/Xilinx/Vitis/2023.2
+  CMAKE_ARGS: |
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -DCMAKE_C_COMPILER_LAUNCHER=ccache \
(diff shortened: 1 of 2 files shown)


.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -170,6 +170,138 @@ jobs:
...
+          echo "::endgroup::"
+          echo "::group::matrix context"
+          echo "$execution_context_var_matrix"
+          echo "::endgroup::"
+          echo "::group::needs context"
+          echo "$execution_context_var_needs"
+          echo "::endgroup::"
+          echo "::group::inputs context"
+          echo "$execution_context_var_inputs"
+          echo "::endgroup::"
```
makslevental, line 240: my 2c: this isn't useful and unecessarily clutters these scripts - you can already get more information than even this by rerunning and "Enable debug logging":

Screenshot from 2024-05-08 07-51-35

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -170,6 +170,138 @@ jobs:
...
+            -DXRT_ROOT=/opt/xilinx/xrt \
+            -DAIE_ENABLE_PYTHON_PASSES=OFF \
+            -DAIE_ENABLE_XRT_PYTHON_BINDINGS=ON \
+            -DAIE_INCLUDE_INTEGRATION_TESTS=OFF
+
+          ninja install
+
+          # hack: timed out tests aren't fails...
+          LIT_PATH=$(python -c "import lit; from pathlib import Path; print(Path(lit.__file__).parent)")
+          sed -i.bak 's/TIMEOUT = ResultCode("TIMEOUT", "Timed Out", True)/TIMEOUT = ResultCode("TIMEOUT", "Timed Out", False)/g' $LIT_PATH/Test.py
```
fifield, line 302: I'd love to drop this timeouts are not errors hack. The tests should be stable enough that this isn't a concern. We were getting problematic false passes when merging the asplos branch a few weeks ago. One PR's was getting timeouts on a bad test, wedging some of the columns, but claiming pass. Subsequent checks on unrelated PRs would get allocation errors because of the broken columns and it took some time to figure out what was happening.  An error at the point of failure would have been better.

.github/workflows/buildAndTestRyzenAI.yml
stephenneuendorffer: Great!

.github/workflows/buildAndTestRyzenAI.yml
stephenneuendorffer: dropped.

keryell:
On my Phoenix laptop with 64GB I use ` -DLLVM_PARALLEL_LINK_JOBS=8`  and `LIT_OPTIONS` with `--workers=6`.
This might be pessimistic but it works.

PR #1469, opened 2024-05-08
[TEST] Update run_on_npu.sh for 6.8.8

jgmelber:
Remove XRT_HACK_UNSECURE_LOADING_XCLBIN


fifield:
We might want to leave this in for a while for backward compatibility, but I'm glad to see it's working without.

jgmelber:
> We might want to leave this in for a while for backward compatibility, but I'm glad to see it's working without.

I'm in no rush.

PR #1474, opened 2024-05-10
Move aie-rt submodule to third_party

stephenneuendorffer:


fifield:
It's possible these chess crashes https://github.com/Xilinx/mlir-aie/actions/runs/9026613518/job/24804197213?pr=1474 are because you ran out of memory on the runner (aie2 is 64GB). There was a `-j12` was removed by #1467, which was there to prevent out-of-memory when doing many chess jobs in parallel.

stephenneuendorffer:
> It's possible these chess crashes https://github.com/Xilinx/mlir-aie/actions/runs/9026613518/job/24804197213?pr=1474 are because you ran out of memory on the runner (aie2 is 64GB). There was a `-j12` removed by #1467, which was there to prevent out-of-memory when doing many chess jobs in parallel.

I see... probably need a better way to limit that, but I'll add it back in the meantime.

keryell:
On my Phoenix laptop with 64GB, besides the implicit `-j16` I use `-DLLVM_PARALLEL_LINK_JOBS=8`  and `LIT_OPTIONS` with `--workers=6`.
This might be pessimistic but it works.

PR #1476, opened 2024-05-10
limit number of simultaneous tests

stephenneuendorffer:
Without this, we appear to get out of memory problems on the runners.


fifield (approved):
Memory can be an issue whenever many chess jobs are going in parallel. For example in the integration tests, unit_tests, and programming example tests. I figure about ~5GB per job, but if you're unlucky it can peak a bit higher.

stephenneuendorffer:
We do default in most cases to running chess with 4 processes.  It's unclear to me how much parallelism it can actually leverage, but we could also turn that down.  The right way to do this is to limit memory more directly.

PR #1486, opened 2024-05-15
Update buildAndTestAieToolsHsaBuildOnly.yml

fifield:
update paths after https://github.com/Xilinx/mlir-aie/pull/1474

### DIFF
--- a/.github/workflows/buildAndTestAieToolsHsaBuildOnly.yml
+++ b/.github/workflows/buildAndTestAieToolsHsaBuildOnly.yml
@@ -100,12 +100,12 @@ jobs:
             git submodule update --init --recursive
 
             # Buliding libxaie
-            pushd runtime_lib/xaiengine/aie-rt/driver/src
+            pushd third_party/aie-rt/driver/src
             make -f Makefile.Linux CFLAGS="-D__AIEAMDAIR__"
             popd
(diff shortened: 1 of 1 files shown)


stephenneuendorffer (approved):
This is not running as a precommit check?  Should it be?

fifield:
> This is not running as a precommit check? Should it be?

It is not a precommit check, but it should be. The two problems are 1) it needs secrets for the vitis license so can't run on PRs from forks. 2) It needs a better runner so it doesn't take so long (and to have a local vitis install). So the compromise is to run it on commits to main and it sometimes breaks. But hopefully it catches breakage before the vck5000 runner which is non-public and checked less frequently by fewer people.

stephenneuendorffer:
> > This is not running as a precommit check? Should it be?
> 
> It is not a precommit check, but it should be. The two problems are 1) it needs secrets for the vitis license so can't run on PRs from forks. 2) It needs a better runner so it doesn't take so long (and to have a local vitis install). So the compromise is to run it on commits to main and it sometimes breaks. But hopefully it catches breakage before the vck5000 runner which is non-public and checked less frequently by fewer people.

Could we put the VCK5000 runner on the public VLAN?  We have the machines with V70's, but could easily drop a VCK5000 in, I think.

PR #1487, opened 2024-05-15
Improve type constraints for AIEVec_MulElemOp

muradq-amd:
This PR includes adding constrains to aievec::mul_elem on the number of lanes and types of operands and results.

Changes:
* Add constraints to the LHS and RHS operands and result (AIE2MulElemLHS, AIE2MulElemRHS, and AIE2MulElemACC)
* Add PredOpTrait constrains to allow only supported type/number of lanes combinations.

### DIFF
--- a/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
+++ b/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
@@ -910,14 +910,39 @@ public:
                 {VectorType::get({16}, rewriter.getBF16Type())}));
       } else if (resultVectorSize == 1024) {
         // v32bfloat16 -> v32accfloat
-        // Implement this scenario in emulation. The CPP example is below:
+        // The CPP example of the implementation is below:
         //   INTRINSIC(v32accfloat) ups_to_v32accfloat(v32bfloat16 a) {
         //     v16accfloat x0 = ups_to_v16accfloat(extract_v16bfloat16(a, 0));
         //     v16accfloat x1 = ups_to_v16accfloat(extract_v16bfloat16(a, 1));
         //     return concat(x0, x1);
         //   }
-        // TODO: implement this after adding 512->256 vector extraction
-        // intrinsic
+        auto indexZeroCst = rewriter.create<LLVM::ConstantOp>(
+            loc, rewriter.getI32Type(), rewriter.getI32IntegerAttr(0));
+        auto indexOneCst = rewriter.create<LLVM::ConstantOp>(
+            loc, rewriter.getI32Type(), rewriter.getI32IntegerAttr(1));
+        auto extractUps = [&](Value source, Value index) -> Value {
+          auto extOp = rewriter.create<xllvm::ExtI256I512IntrOp>(
+              loc, VectorType::get({8}, rewriter.getI32Type()),
+              forceCastOperandsToSignature(
+                  rewriter, loc, {source, index},
+                  {VectorType::get({16}, rewriter.getI32Type()),
+                   rewriter.getI32Type()}));
+          return rewriter.create<xllvm::Vector16BF16ToV16AccFloatIntrOp>(
+              loc, VectorType::get({8}, rewriter.getI64Type()),
+              forceCastOperandsToSignature(
+                  rewriter, loc, {extOp},
+                  {VectorType::get({16}, rewriter.getBF16Type())}));
+        };
+        auto resLo = extractUps(adaptor.getSource(), indexZeroCst);
+        auto resHi = extractUps(adaptor.getSource(), indexOneCst);
+        // Concat the two 512-bit vector to a 1024-bit vector.
+        // Note that given sources a0 and a1, the result is [a1; a0].
+        upsIntrOp = rewriter.create<xllvm::ConcatI1024I512IntrOp>(
+            loc, VectorType::get({32}, rewriter.getI32Type()),
+            forceCastOperandsToSignature(
+                rewriter, loc, {resLo, resHi},
+                {VectorType::get({16}, rewriter.getI32Type()),
+                 VectorType::get({16}, rewriter.getI32Type())}));
       }
     }
 
@@ -926,9 +951,14 @@ public:
       return failure();
     }
 
-    // create bitcast for result
-    rewriter.replaceOpWithNewOp<LLVM::BitcastOp>(op, op.getResult().getType(),
-                                                 upsIntrOp);
+    // create bitcast for result if needed
+    if (op.getResult().getType() != upsIntrOp.getType()) {
+      rewriter.replaceOpWithNewOp<LLVM::BitcastOp>(op, op.getResult().getType(),
+                                                   upsIntrOp);
+    } else {
+      rewriter.replaceOp(op, upsIntrOp);
+    }
+
     return success();
   }
 };
@@ -950,7 +980,7 @@ public:
     int resultVectorSize = resultBitWidth * resultLanes;
 
     // Integer types
-    Operation *srsIntrOp = nullptr;
+    Value srsIntrOp = nullptr;
     if (llvm::isa<IntegerType>(resultScaTy)) {
       // create constant for sign
       auto signCst = rewriter.create<LLVM::ConstantOp>(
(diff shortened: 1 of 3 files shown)


include/aie/Dialect/AIEVec/IR/AIEVecTypeConstraints.td
```diff
@@ -144,4 +144,64 @@ class IsValidAIE2MatMulShapeAndType<string lhs, string rhs, string acc> :
...
+
+def AIE2MulElemRHS :
+  AnyTypeOf<[VectorOfShapeAndType<[32], I8>,
+             VectorOfShapeAndType<[32], I16>,
+             VectorOfShapeAndType<[16], I32>,
+             VectorOfShapeAndType<[16], BF16>,
+             VectorOfShapeAndType<[16], F32>],
+            "a vector compatible with a rhs operand of element-wise multiply and "
+            # "accumulate",
+            "::mlir::VectorType">;
```
jsetoain, line 165: This is overkill. Lhs & Rhs types have to match, you don't need to spell out every single valid type for each of the operands.

include/aie/Dialect/AIEVec/IR/AIEVecTypeConstraints.td
```diff
@@ -144,4 +144,64 @@ class IsValidAIE2MatMulShapeAndType<string lhs, string rhs, string acc> :
...
+                                   acc, VectorOfShapeAndType<[16], I32>>,
+                  VectorTypesMatch<lhs, VectorOfShapeAndType<[16], BF16>,
+                                   rhs, VectorOfShapeAndType<[16], BF16>,
+                                   acc, VectorOfShapeAndType<[16], BF16>>,
+                  VectorTypesMatch<lhs, VectorOfShapeAndType<[16], BF16>,
+                                   rhs, VectorOfShapeAndType<[16], BF16>,
+                                   acc, VectorOfShapeAndType<[16], F32>>,
+                  VectorTypesMatch<lhs, VectorOfShapeAndType<[16], F32>,
+                                   rhs, VectorOfShapeAndType<[16], F32>,
+                                   acc, VectorOfShapeAndType<[16], F32>>]>>;
```
jsetoain, line 205: This constraint alone already makes all the checks necessary for the supported types, but it's unnecessarily complex for such a regular operation.

jsetoain (changes requested):
I understand you've followed the constraint scheme for `aievec.matmul`, but that's is a very bad template for this use case because that one was a very irregular type combination, while this one is pretty regular.

You need to check that lhs/rhs match, and that either the element type of lhs/rhs matches acc, or acc is the "wide type" (i32 for integrals, f32 for floats).

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -867,4 +848,37 @@ def AIEVec_MatMulOp:
...
+         lhs                | rhs                | Accumulator
+        :------------------:|:------------------:|:-----------------:
+         `vector<32xi8>`    | `vector<32xi8>`    | `vector<32xi8>`
+         `vector<32xi8>`    | `vector<32xi8>`    | `vector<32xi32>`
+         `vector<32xi16>`   | `vector<32xi16>`   | `vector<32xi16>`
+         `vector<32xi16>`   | `vector<32xi16>`   | `vector<32xi32>`
+         `vector<16xi32>`   | `vector<16xi32>`   | `vector<16xi32>`
+         `vector<16xbf16>`  | `vector<16xbf16>`  | `vector<16xbf16>`
+         `vector<16xbf16>`  | `vector<16xbf16>`  | `vector<16xf32>`
+         `vector<16xf32>`   | `vector<16xf32>`   | `vector<16xf32>`
```
jamestcl-amd, line 876: For `aievec.mul_elem` the accumulator is `acc32` (i32) for `i8/i16` and `acc64` (i64) for `i32`. Your Accumulator column here is more like a result type for a `arith.mulf/i` op from what you see in the e2e unit tests. To handle different result types, we introduce either `aievec.cast` or `aievec.srs` to the result of `aievec.mul_elem`. See https://github.com/Xilinx/mlir-aie/blob/aba1887b9750dea8db14fbe16417068f83a6b94c/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp#L702-L712

include/aie/Dialect/AIEVec/IR/AIEVecTypeConstraints.td
jamestcl-amd: The op verifier checks that the lanes and datatypes are the same for operands. See https://github.com/Xilinx/mlir-aie/blob/aba1887b9750dea8db14fbe16417068f83a6b94c/lib/Dialect/AIEVec/IR/AIEVecOps.cpp#L872-L894

This can probably be replaced by `SameTypeOperands` and `SameOperandsAndResultShape` traits....

jamestcl-amd: However, I'm not sure how to better constraint the fact that the `acc` is i32 for i8/i16 operands, `i64` for i32 operands, and `f32` for bf16/f32 operands. Maybe just constraints it in the op verifier in cpp code?

jsetoain: > The op verifier checks that the lanes and datatypes are the same for operands. See
> 
> https://github.com/Xilinx/mlir-aie/blob/aba1887b9750dea8db14fbe16417068f83a6b94c/lib/Dialect/AIEVec/IR/AIEVecOps.cpp#L872-L894
> 

Notice that the C++ verifier will be replaced by the automatically generated one.

> This can probably be replaced by `SameTypeOperands` and `SameOperandsAndResultShape` traits....

Indeed.

jsetoain: > However, I'm not sure how to better constraint the fact that the `acc` is i32 for i8/i16 operands, `i64` for i32 operands, and `f32` for bf16/f32 operands. Maybe just constraints it in the op verifier in cpp code?

For that, you can either define a new predicate ('IsValidAccumulatorTypeFor`) that verifies the accumulator type validity using simpler pre-existing predicates, or define a C++ function that makes that check and invoke it from the predicate.

muradq-amd:
Thanks @jamestcl-amd, @jsetoain for the comments.
I just revised the code changes and added simplified constraints on the types and shapes of aievec::mul_elem's operands/results. Here is the list of supported type/shape combinations:                                   

| lhs  | rhs |acc |
| ------------- | ------------- |------------- |
| <32xi8>       | <32xi8>  | <32xi8>  |
| <32xi8>      | <32xi8>   | <32xi32> |
| <64xi8>      | <64xi8>   | <32xi32> |
| <32xi16>     | <32xi16>  |<32xi16>  |
| <32xi16>     | <32xi16> | <32xi32> |
| <16xi32>     | <16xi32>  | <16xi32> |
| <16xi32>     | <16xi32>  | <16xi64> |
| <16xbf16>    | <16xbf16> | <16xbf16> |
| <16xbf16>    | <16xbf16>  | <16xf32> |
| <16xf32>    | <16xf32>  | <16xf32> |

Please let me know if you see any thing need to be fixed. thanks.

platforms/boards
jsetoain, line 1: This shouldn't be in the commit.

include/aie/Dialect/AIEVec/IR/AIEVecTypeConstraints.td
```diff
@@ -101,6 +101,10 @@ class ShapesCompatibleWithContraction<string lhs, string rhs, string acc> :
 class VectorType<string name> : StrFunc<"cast<VectorType>($" # name #
                                         ".getType())">;
 
+class VectorElementType<string name> :
```
jsetoain, line 104: These already exists in upstream mlir. Check `llvm-project/mlir/include/mlir/IR/OpBase.td`. There's a lot of stuff there you'll find helpful.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -313,16 +313,33 @@ def AIEVec_MulOp:
...
   AIEVec_Op<"mul_elem", [
-    Pure
+    Pure, 
+    IsValidAIE2MulElemShapeAndType<"lhs", "rhs", "acc">
   ]>,
-  Arguments<(ins AnyVector:$lhs, AnyVector:$rhs)>,
-  Results<(outs AnyVector:$result)> {
+  Arguments<(ins AnyVector:$lhs,
+                 AnyVector:$rhs)>,
+  Results<(outs AnyVector:$acc)> {
```
jsetoain, line 321: We don't support `AnyVector`, you should use an operand type constraint for these, and then your op constraint will be simpler.

jsetoain (changes requested):

PR #1490, opened 2024-05-15
[aievec] to-llvm flow for aievec.ext op with I128.I512 extract intrinsic

jamestcl-amd:
This PR add the support for aievec.ext op going through the to-llvm flow with the `extract.I128.I512` intrinsic op.

Changes:
- Update the `extract.I128.I512` intrinsic ops in XLLVM dialect.
- Add aievec-to-llvm conversion pattern/tests for the aievec.ext op for `extract.I128.I512` intrinsic.
- Add test for translating to external llvm.

### DIFF
--- a/test/Conversion/AIEVecToLLVM/test-ext.mlir
+++ b/test/Conversion/AIEVecToLLVM/test-ext.mlir
@@ -69,6 +69,37 @@ func.func @v32i8_ext_v128i8(%arg0 : vector<128xi8>) -> (vector<32xi8>, vector<32
 
 // -----
 
+func.func @v16i8_ext_v64i8(%arg0 : vector<64xi8>) -> (vector<16xi8>, vector<16xi8>) {
+  %0 = aievec.ext %arg0 {index = 0 : i8} : vector<64xi8>, vector<16xi8>
+  %1 = aievec.ext %arg0 {index = 3 : i8} : vector<64xi8>, vector<16xi8>
+  return %0, %1 : vector<16xi8>, vector<16xi8>
+}
+
+// CHECK-LABEL: @v16i8_ext_v64i8
+// CHECK-SAME: %[[ARG0:.*]]: vector<64xi8>
+// CHECK: %[[UNDEF0:.*]] = "xllvm.intr.aie2.v16int32"() : () -> vector<16xi32>
+// CHECK-NEXT: %[[CST0:.*]] = llvm.mlir.constant(0 : i32) : i32
+// CHECK-NEXT: %[[BITCAST0:.*]] = llvm.bitcast %[[ARG0]] : vector<64xi8> to vector<16xi32>
+// CHECK-NEXT: %[[VSHIFT0:.*]] = "xllvm.intr.aie2.vshift.I512.I512"(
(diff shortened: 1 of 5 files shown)


test/Conversion/AIEVecToLLVM/test-ext.mlir
```diff
@@ -333,3 +454,34 @@ func.func @v8f32_ext_v32f32(%arg0 : vector<32xf32>) -> (vector<8xf32>, vector<8x
...
+// CHECK-NEXT: %[[BITCAST0:.*]] = llvm.bitcast %[[ARG0]] : vector<16xf32> to vector<16xi32>
+// CHECK-NEXT: %[[VSHIFT0:.*]] = "xllvm.intr.aie2.vshift.I512.I512"(
+// CHECK-SAME: %[[BITCAST0]], %[[UNDEF0]], %[[CST0]], %[[CST0]]) :
+// CHECK-SAME: (vector<16xi32>, vector<16xi32>, i32, i32) -> vector<16xi32>
+// CHECK-NEXT: %[[EXT0:.*]] = "xllvm.intr.aie2.extract.I128.I512"(
+// CHECK-SAME: %[[VSHIFT0]]) : 
+// CHECK-SAME: (vector<16xi32>) -> vector<4xi32>
+// CHECK-NEXT: %[[RES0:.*]] = llvm.bitcast %[[EXT0]] : vector<4xi32> to vector<4xf32>
+// CHECK-NEXT: %[[UNDEF1:.*]] = "xllvm.intr.aie2.v16int32"() : () -> vector<16xi32>
+// CHECK-NEXT: %[[CST48:.*]] = llvm.mlir.constant(48 : i32) : i32
```
jsetoain, line 479: Is this correct? If I understand correctly, the process to extract a subvector from a given position is to shift the whole vector by the number of *bits* up to the position, and then extracting the lowest part. For `16-bit` element vectors and an extraction from position 3, that's `3 x 16 = 48 bit`, but for `32-bit` element vectors, shouldn't it be `3 x 32 = 96 bit`? Am I misunderstanding something?

jamestcl-amd: The `shift` amount for the `shift` op/intrinsic is in the number of bytes. The `I128.I512` extract intrinsic is special. It always extracts the lowest 128-bit vector from the source 512-bit vector. For `index=0` vector extraction, we actually don't need any vector shift beforehand, so I have updated the conversion pattern for this scenario. As for `index=3` vector extraction, the shift amount is `48 bytes = 384 bits = 3*128 bits`. After the right shift of 48 bytes, the `I128.I512` intrinsic extracts the lowest 128-bit vector, which is correct. I have updated the `aievec.ext/shift` op descriptions, to explicitly clarify the shift amount in bytes and the extraction index to be either 0--1 or 0--3. Let me know if there is anything else I can do :).

jsetoain: Aaaah! Got it, I was trying to make the maths work in my head and failing, thanks for the clarification 🙂

jsetoain (approved):
Looks good to me!

PR #1492, opened 2024-05-16
Tracing improvements

andrej:
image

This updates the tracing utilities in `python/utils/trace.py` as well as the tracing parse scripts in `progamming_examples/utils`.

- When tracing stream switch "port active"-type events, it was not possible before to select which ports we want to generate events. Those events require additional configuration through another register, and we previously had no way to pass in those additional arguments. I adjusted things slightly, so if you wanted to trace Master Port 3 as stream switch event 2 and slave port 2 as stream switch event 1, you could do this like so now: 
   ```python
   from aie.utils.trace import CoreEvent, PortEvent, configure_simple_tracing_aie2
   # ...

   configure_simple_tracing_aie2(
       compute_tile, shim_tile, # ... etc
       events=[
                   PortEvent(CoreEvent.PORT_RUNNING_2, port_number=3, master=True),  
                       # PortEvent is a subclass for the PORT_XXX_0-7 events that configures which port generates events
                   PortEvent(CoreEvent.PORT_RUNNING_1, port_number=2, master=False),
                   CoreEvent.INSTR_LOAD,   
                       # Events that require no further configuration can just use the enum directly
                   22 # <- for backwards compatibility, passing in just the integer instead of enum should still work
               ]
   )
   ```
- Automatically generated enum of all available trace events: `python/utils/trace_events_enum.py` contains an enumeration of all available tracing events. This is automatically generated from [this header file in aie-rt](https://github.com/Xilinx/aie-rt/blob/main-aie/driver/src/events/xaie_events_aie.h) using the new `utils/generate_events_enum.py` script.
- I have replaced the enumeration of (a selection of) events in the parsing scripts and replaced it with a use of the above enum. This means all events should now be named correctly in the output trace JSON. This did change the spelling slightly, e.g. instead of "InstrLoad" it would now be called "INSTR_LOAD". If we're attached to CamelCase, I can adjust the script to automatically rename the enum values.

Please let me know what you think.


fifield:
> * Automatically generated enum of all available trace events: `python/utils/trace_events_enum.py` contains an enumeration of all available tracing events. This is automatically generated from [this header file in aie-rt](https://github.com/Xilinx/aie-rt/blob/main-aie/driver/src/events/xaie_events_aie.h) using the new `utils/generate_events_enum.py` script.

Can you make it work on https://github.com/Xilinx/aie-rt/blob/main-aie/driver/src/events/xaie_events_aieml.h? I think those are the aie2 events.

andrej:
Good catch, fixed it. I got lucky that the event numbers were mostly the same between AIE1 and 2 but was clearly using the wrong header file

andrej:
Looks like there are some tests that need updating. I will look into this next week.

andrej:
I have decided to make it mandatory to specify the port you want to generate events for if an event like PORT_RUNNING_0 is specified. This required updating a test. 

Please let me know if anyone thinks this is a bad idea.

andrej:
@fifield Would you mind having another look at this too, please?

andrej:
I have not yet, but that would be interesting. I'll merge this as it is for now

PR #1494, opened 2024-05-16
Don't compile chess wrappers for every test.

stephenneuendorffer:
Compiling the wrapper is quite slow.  Instead, compile it once and reuse it everywhere.


makslevental:
I already did this https://github.com/Xilinx/mlir-aie/blob/main/python/compiler/util.py#L455 I just didn't announce it because I didn't want to rock the boat 🤷‍♂️

makslevental:
> Yeah I saw, but it shouldn't be a fixed string like that. Can the python binding path read the file instead?

sure

PR #1495, opened 2024-05-17
Reduce concurrency in integration tests to reduce memory usage

fifield:
Replace command lines like,
```
xchesscc_wrapper ... file1.cc file2.cc ..
```
with,
```
xchesscc_wrapper ... -c file1.cc -o file1.cc.o ..
xchesscc_wrapper ... -c file2.cc -o file2.cc.o ..
xchesscc_wrapper ... file1.cc.o file2.cc.o ..
```
because each source file on the xchesscc_wrapper command line gets compiled in parallel, and each of those jobs takes a lot of memory. Instead compile the source files sequentially so that the parallelism can be controlled by lit at a higher level.


fifield:
With this I can run the integration tests with `-j16` on an 8 core 96GB machine, not run out of memory and crash, and be cpu bound instead

stephenneuendorffer (commented):
Would it be simpler just to remove the +P 4 from xchesscc_wrapper?  I'm hesitant to start coding this kind of thing into the individual tests.

fifield:
> Would it be simpler just to remove the +P 4 from xchesscc_wrapper? I'm hesitant to start coding this kind of thing into the individual tests.

Maybe, or expose it to xchesscc_wrapper users (if it isn't already)?  I figured it was a useful default for humans typing `make` but I also didn't know where it was hard coded.

stephenneuendorffer (commented):
Another option would be to add  ALLOW_RETRIES:  (perhaps even in the toplevel python so it doesn't have to be specified everywhere.)

PR #1498, opened 2024-05-17
Update mm.cc to be compatible with iree's current tiling strategy 

newling:
This updated kernel is based on the ukernel in https://github.com/Xilinx/mlir-air/blob/main/test/xrt/09_gemm_extern_vec_4x4/mm.cc

This is the only ukernel which works with the current tiling strategy used when lowering from IREE. 

The striding of the A tensor is different. This replacement ukernel has 4x8 tiles that are _vertically_ contiguous in A, contiguous in memory. The ukernel it replaces has 4x8 tiles which are _horizontally_ contiguous in B, contiguous in memory.

### DIFF
--- a/aie_kernels/mm.cc
+++ b/aie_kernels/mm.cc
@@ -24,6 +24,27 @@
 
 #include "zero.cc"
 
+// Perform matmul where the expected layouts for A and B are as follows.
+//
+// Suppose A is a 64x64 tensor and B is a 64x64 tensor, and r=4, s=8, t=4.
+//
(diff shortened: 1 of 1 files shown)


newling:
@erwei-xilinx please check that my comments about tiling are correct, based on what we discussed yesterday

newling:
I suspect my description is not correct, as the diff suggests it is the B tensor which has a different striding with this change, not the A tensor.

newling:
Abandoning in favour of https://github.com/Xilinx/mlir-aie/pull/1499/files

PR #1501, opened 2024-05-20
Simplify CDO code generation

stephenneuendorffer:
Bootgen requires CDO files to be non-empty.  In some corner cases this isn't guaraunteed, resulting in some tricky workarounds which were incomplete.  A simpler solution is to always emit a few NOOP padding bytes.

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -188,8 +188,8 @@ def emit_partition(mlir_module_str, kernel_id="0x901", start_columns=None):
             module.operation,
             lambda o: isinstance(o.operation.opview, aiedialect.TileOp),
         )
-        min_col = min([t.col.value for t in tiles]) if tiles else 0
-        max_col = max([t.col.value for t in tiles]) if tiles else 0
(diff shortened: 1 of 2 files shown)


makslevental (approved):
Cool I've definitely been bitten by this before.

python/compiler/aiecc/main.py
```diff
@@ -188,9 +188,8 @@ def emit_partition(mlir_module_str, kernel_id="0x901", start_columns=None):
             module.operation,
             lambda o: isinstance(o.operation.opview, aiedialect.TileOp),
         )
-        min_col = min([t.col.value for t in tiles])
-        max_col = max([t.col.value for t in tiles])
-
+        min_col = min([t.col.value for t in tiles]) if tiles else 0
+        max_col = max([t.col.value for t in tiles]) if tiles else 0
```
makslevental, line 192:
```suggestion
        min_col = min([t.col.value for t in tiles], default=0) if tiles else 0
        max_col = max([t.col.value for t in tiles], default=0) if tiles else 0
```

PR #1516, opened 2024-05-28
[aievec] Add new shuffle ops

jsetoain:
This replaces the old `aievec.shuffle` op with a new one with a better syntax that supports all cases supported by the intrinsics, and has strong type guarantees.

For legacy purposes, we leave the old shuffle instruction renamed as `aievec.legacyshuffle`.

### DIFF
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// (c) Copyright 2023 AMD Inc.
+// (c) Copyright 2023-2024 Advanced Micro Devices, Inc. or its affiliates
 //
 //===----------------------------------------------------------------------===//
 // Defines AIE vector operations.
@@ -14,6 +14,7 @@
 #define AIEVEC_OPS
 
 include "aie/Dialect/AIE/IR/AIEAttrs.td"
+include "aie/Dialect/AIEVec/IR/AIEVecAttributes.td"
 include "aie/Dialect/AIEVec/IR/AIEVecTypes.td"
 include "aie/Dialect/AIEVec/IR/AIEVecTypeConstraints.td"
 
@@ -587,8 +588,8 @@ def AIEVec_ShiftOp:
   }];
 }
 
-def AIEVec_ShuffleOp:
-  AIEVec_Op<"shuffle", [
+def AIEVec_LegacyShuffleOp:
+  AIEVec_Op<"legacyshuffle", [
     Pure
   ]>,
   Arguments<(ins AnyVector:$source,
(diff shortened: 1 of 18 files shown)


lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
```diff
@@ -1041,6 +1041,37 @@ static LogicalResult printOperation(CppEmitter &emitter,
...
+    return failure();
+  os << emitter.getOrCreateName(lhs);
+  os << ", ";
+  if (rhs) {
+    if (!emitter.hasValueInScope(rhs))
+      return failure();
+    os << emitter.getOrCreateName(rhs);
+    os << ", ";
+  }
+  os << "eShuffleMode::shuffle_T" << &stringifyEnum(mode).data()[1];
```
jamestcl-amd, line 1066: Not sure if we can do `substr(1)` here, which is easier to understand...

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
```diff
@@ -1620,8 +1627,105 @@ ParseResult ShiftOp::parse(OpAsmParser &parser, OperationState &result) {
...
+                       << "' does not admit a second operand";
+
+  // Verify vector element type
+  auto elemBitWidth =
+      cast<VectorType>(getLhs().getType()).getElementTypeBitWidth();
+  if (modeBitWidth != elemBitWidth)
+    return emitError() << "shuffle mode '" << stringifyEnum(mode)
+                       << "' requires vectors of " << modeBitWidth
+                       << "-bit elements";
+
```
jamestcl-amd, line 1723: Should we also check if the operands and results are in 512-bit vectors?

test/aievec/conv2d_i8_after_polygeist.mlir
```diff
@@ -38,7 +38,7 @@ module attributes {dlti.dl_spec = #dlti.dl_spec<#dlti.dl_entry<"dlti.endianness"
 //      CHECK:    %[[C16:.*]] = arith.constant 16 : index
 //      CHECK:    %[[C0:.*]] = arith.constant 0 : index
 //      CHECK:    %[[T0:.*]] = aievec.upd %[[A1]][%[[C0:.*]]] {index = 0 : i8, offset = 0 : i32} : memref<?xi8>, vector<64xi8>
-//      CHECK:    %[[T1:.*]] = aievec.shuffle %[[T0:.*]] {mode = 0 : i32} : vector<64xi8>, vector<64xi8>
+//      CHECK:    %[[T1:.*]] = aievec.legacyshuffle %[[T0:.*]] {mode = 0 : i32} : vector<64xi8>, vector<64xi8>
```
jamestcl-amd, line 41: Given the mode=0 (`shuffle_T8_64x2_lo`), I'm a bit confused why this can work without a second operand...

test/aievec/conv2d_uij_i8_noinit_aie-ml.mlir
```diff
@@ -80,7 +80,7 @@ func.func @conv2d (%A: memref<18x288xi8>, %B: memref<48xi8>, %C: memref<16x256xi
 //      CHECK:    %[[C8:.*]] = arith.constant 8 : i32
 //      CHECK:    %[[C0:.*]] = arith.constant 0 : index
 //      CHECK:    %[[T0:.*]] = aievec.upd %[[A1]][%[[C0]]] {index = 0 : i8, offset = 0 : i32} : memref<48xi8>, vector<64xi8>
-//      CHECK:    %[[T1:.*]] = aievec.shuffle %[[T0]] {mode = 0 : i32} : vector<64xi8>, vector<64xi8>
+//      CHECK:    %[[T1:.*]] = aievec.legacyshuffle %[[T0]] {mode = 0 : i32} : vector<64xi8>, vector<64xi8>
```
jamestcl-amd, line 83: And this one too...

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -872,4 +873,153 @@ def AIEVec_MatMulOp:
...
+  Arguments<(ins VectorOfBitWidthAndElementTypes<
+                      512, [I8, I16, I32, I64, I128, I256,
+                            I512, BF16, F32]>:$lhs,
+                 Optional<VectorOfBitWidthAndElementTypes<
+                      512, [I8, I16, I32, I64, I128, I256,
+                            I512, BF16, F32]>>:$rhs,
+                 AIEVec_ShuffleModeAttr:$mode)>,
+  Results<(outs AnyVector:$result)> {
+  let summary = "AIE2 shuffle";
+  let description = [{
```
jamestcl-amd, line 889: Thanks for the nice and clear explanation!

lib/Dialect/AIEVec/IR/AIEVecOps.cpp
jsetoain: That's being checked by the type constraints in the op definition, this exclusively verifies whether the shuffle mode is "compatible" with the element type. It's up for debate whether we want to impose this restriction or not, since the hardware is agnostic to the mode (it just shuffles bytes around), but I'd rather be conservative for the time being.

test/aievec/conv2d_i8_after_polygeist.mlir
jsetoain: "Shuffle" always takes two arguments. When translating to C++, the single operand "shuffle" C++ intrinsic calls a two operand version with "undef" in the rhs. For the new `shuffle`, we will have to inject the `undef` when lowering to `LLVM IR`. With the new constraints, I'm just trying to keep things semantically consistent.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -872,4 +873,153 @@ def AIEVec_MatMulOp:
...
+    elements of 32 bits each. The resulting vector would contain either the
+    least (`lo`) or most (`hi`) significant 16 elements of the 32 element vector
+    that would result from selecting, out of the concatenated vectors `lhs:rhs`,
+    8 blocks of 4 elements, each block taking one of every 8 elements starting
+    from the block index.
+
+    That is, for two `vector<16xi32>` operands containing:
+    ```
+    lhs = [0,   1,  2,  3, ..., 15]
+    rhs = [17, 18, 19, 20, ..., 31]
```
david-vc, line 912: rhs should start with 16, i.e., rhs = [16, 17, 18, 19, ..., 31]

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -872,4 +873,153 @@ def AIEVec_MatMulOp:
...
+    ```
+    t16_1x2_flip = [1, 0, 3, 2, ..., 29, 28, 31, 30]
+    ```
+
+    The list of supported shuffle modes, required operands, and associated
+    vector types are the following:
+
+         Shuffle Mode       | Operands           | Types Supported
+        :------------------:|:------------------:|:------------------:
+         t8_8x4             | `lhs`              | `vector<2x32xi8>`
```
david-vc, line 972: The entries in the 'Types Supported' seem inconsistent.
I would expect to see only 1D vectors. So for the entry here, I would expect `vector<64xi8>`.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -872,4 +873,153 @@ def AIEVec_MatMulOp:
...
+         t8_4x8             | ^                  | ^
+         t8_8x8             | ^                  | `vector<64xi8>`
+         t8_16x4            | ^                  |  ^
+         t8_4x16            | ^                  |  ^
+         t8_64x2_lo         | `lhs` & `rhs`      |  ^
+         t8_64x2_hi         | ^                  |  ^
+         t8_2x64_lo         | ^                  |  ^
+         t8_2x64_hi         | ^                  |  ^
+         t16_4x2            | `lhs`              | `vector<4x8xi16>` or `vector<4x8xbf16>`
```
david-vc, line 981: the 3rd column should read `vector<8xi16>` or `vector<8xbf16>`.

david-vc (commented):
Very nice work! It's great to see the symbolic mode in the textual representation.

lib/Targets/AIEVecToCpp/TranslateAIEVecToCpp.cpp
jsetoain: Yup! You caught me being lazy 😂

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
jsetoain: This is something I'd like to discuss. I stumbled upon this "issue" when looking over each of the shuffle modes. For some of them, like this one, it performs the transposition on the two halves of the vector, as if we had two (in this case) 8x4 matrices. I'm inclined to treat it as a flat `64xi8` and this would be yet another set of "special" cases. The ISA is not particularly semantically consistent. It does what it does, except when it doesn't 😉

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
jsetoain: If anything, `vector<32xi16>`/`vector<32xbf16>`. See above.

PR #1517, opened 2024-05-28
Switch to using transaction binary flow with no control packet

nirvedhmeshram:
Requires [this xdna-driver](https://github.com/amd/xdna-driver/commit/58f0afb37e072a60dc382179aef6218c45aedfbf) (or newer)


test/npu-xrt/txn_basic/test.cpp
```diff
@@ -0,0 +1,254 @@
...
+  auto p = patches.begin();
+  for (auto i = instr_v.begin(); i != instr_v.end(); i++) {
+    if (*i == 0xfeedbeef) {
+      uint64_t d = *p++;
+      *i = (d & 0xfffffffc);
+      *(i + 1) = (*(i + 1) & 0xFFFF0000) | ((d >> 32) & 0xffff);
+    }
+    std::cout << "Instruction: " << std::hex << std::setw(8)
+              << std::setfill('0') << *i << std::endl;
+  }
```
fifield, line 191: Currently this is doing address patching manually, but there is also an instruction for this, which we haven't added to mlir-aie yet.

fifield:
Requires [this xdna-driver](https://github.com/amd/xdna-driver/commit/58f0afb37e072a60dc382179aef6218c45aedfbf) (or newer)

test/npu-xrt/txn_basic/test.cpp
fifield: > Currently this is doing address patching manually, but there is also an instruction for this, which we haven't added to mlir-aie yet.

This is now replaced by `aiex.npu.address_patch`

fifield:
The test `npu-xrt/matrix_multiplication_using_dma` is still failing, but I don't understand how it was ever passing. I could be missing something / reading it wrong, but it looks like the `aie.mlir` only returns half the data expected by the `test.cpp`

fifield:
remaining failures:
```
  AIE_TUTORIALS :: basic/matrix_multiplication/whole_array/run_makefile.lit
  AIE_TUTORIALS :: ml/resnet/layers_conv2_x/run_makefile.lit
  AIE :: npu-xrt/matrix_multiplication_using_dma/run-a2x.lit
```

jgmelber:
Which version of the xdna-driver is needed for this? Can this be noted on this PR and the docs updated

jgmelber:
#1544 Updates the docs to point to the required xdna-driver commit to support this

jgmelber (approved):

andrej:
@jgmelber I believe we need to update the version in [`utils/quick_setup.sh`](https://github.com/Xilinx/mlir-aie/blob/1a362e65d325cd9f621f5d52d629744b86c143f5/utils/quick_setup.sh#L65) when you get a chance. With these changes none of the matrix examples work otherwise.

jgmelber:
> @jgmelber I believe we need to update the version in [`utils/quick_setup.sh`](https://github.com/Xilinx/mlir-aie/blob/1a362e65d325cd9f621f5d52d629744b86c143f5/utils/quick_setup.sh#L65) when you get a chance. With these changes none of the matrix examples work otherwise.

Fixed in #1546

PR #1519, opened 2024-05-28
[aievec] to-llvm flow for aievec.min op

jamestcl-amd:
This PR add the support for `aievec.min` op going through the to-llvm flow.

Changes:
- Add `vmin` intrinsic ops to XLLVM dialect.
- Add aievec-to-llvm conversion pattern/tests for the `aievec.min` op.
- Add lit test for translating to external llvm.


include/aie/Dialect/XLLVM/IR/XLLVMAIE2IntrOps.td
```diff
@@ -351,4 +351,49 @@ def VectorMaxLtBf16IntrOp :
...
                    VectorOfLengthAndType<[32], [BF16]>:$rhs)>;
 
+// ----- MIN ELEMENT -----
+
+def VectorMinGe8IntrOp :
+    AIEVec2_IntrOp<"vmin.ge8",
+        [TypeIs<"res",
+            LLVM_StructOf<[
+                VectorOfLengthAndType<[64], [I8]>,
+                I32]>
```
david-vc, line 361: If we have operands with 64 lanes, I would expect that the second result should be I64 instead of I32.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -1498,6 +1498,99 @@ class BroadcastOpConversion
...
+    Location loc = op.getLoc();
+
+    VectorType resultType = cast<VectorType>(op.getResult().getType());
+    Type resultScaTy = resultType.getElementType();
+    unsigned resultBitWidth = resultScaTy.getIntOrFloatBitWidth();
+    int resultLanes = getVectorLaneSize(resultType);
+    int resultVectorSize = resultBitWidth * resultLanes;
+
+    if (resultVectorSize != 512) {
+      op.emitWarning() << "aievec.min conversion with " << resultVectorSize
```
david-vc, line 1517: We need to add a test to exercise this warning message.

david-vc (commented):
Please follow on my comment about the case with 64 lanes.

include/aie/Dialect/XLLVM/IR/XLLVMAIE2IntrOps.td
jamestcl-amd: Good catch. I will update both the min and max ops PRs.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
jamestcl-amd: Sure, will do.

david-vc (approved):
LGTM

PR #1527, opened 2024-05-30
[aievec] Add vector.transpose to aievec.shuffle lowering

jsetoain:


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2801,6 +2801,95 @@ struct LowerVectorContractionOpToAIEVecMatMulPattern
...
+        aievec::ShuffleMode::T16_4X8,
+        id,
+        id,
+        id,
+        id,
+        id,
+        id,
+        id,
+        aievec::ShuffleMode::T16_2X16};
+    auto shuffleMode = aievec::ShuffleMode::T32_4X4;
```
jamestcl-amd, line 2872: The shape isn't checked to be `4x4` in this case?

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2801,6 +2801,95 @@ struct LowerVectorContractionOpToAIEVecMatMulPattern
...
+    // Only permutation of the 2 innermost dimensions are supported.
+    ArrayRef<int64_t> perm = transpOp.getPermutation();
+    for (int64_t i = 0; i < static_cast<int64_t>(perm.size() - 2); ++i)
+      if (perm[i] != i)
+        return failure();
+    if (perm[perm.size() - 1] != static_cast<int64_t>(perm.size() - 2))
+      return failure();
+
+    auto id = aievec::ShuffleMode::T512_1X2_LO;
+    static const aievec::ShuffleMode T8Modes[17] = {
```
jamestcl-amd, line 2836: Perhaps use a `llvm::densemap` or a `std::unordered_map` here? So that we can get rid of all those default `id`s.

jsetoain: Mapping 3-4 values under 17 to another integral value under 256 is a bit of an overkill, maps are slow and tough on caches, so I'd rather use LUTs for this. Alternatively, I could just do if-then-else (it's what I originally had).

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: It's the only possibility left.

jamestcl-amd: The input is `vector.transpose`, so technically, it can be any weird shape. It will eventually lead to an illegal `aievec.shuffle` and error out later on by the `aievec.shuffle` verifier. That sounds reasonable to me.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -2801,6 +2801,95 @@ struct LowerVectorContractionOpToAIEVecMatMulPattern
...
+    for (int64_t i = 0; i < static_cast<int64_t>(resShape.size() - 2); ++i)
+      if (resShape[i] != 1)
+        return failure();
+
+    // Only permutation of the 2 innermost dimensions are supported.
+    ArrayRef<int64_t> perm = transpOp.getPermutation();
+    for (int64_t i = 0; i < static_cast<int64_t>(perm.size() - 2); ++i)
+      if (perm[i] != i)
+        return failure();
+    if (perm[perm.size() - 1] != static_cast<int64_t>(perm.size() - 2))
```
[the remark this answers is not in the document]
jsetoain: Empty and identity permutations fold away, so they can be safely ignored.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
jsetoain: It doesn't even have to be a weird shape, `vector<2x8xi32>` or `vector<2x32xi8>` would end-up with the wrong shuffle mode. I'll have to add some extra restrictions to the allowed shapes, sadly 😩

jamestcl-amd (approved):
LGTM. As mentioned in the comment, we need a follow-up PR on the support of `vector<2x8xi32>` or `vector<2x32xi8>`.

jsetoain:
> LGTM. As mentioned in the comment, we need a follow-up PR on the support of `vector<2x8xi32>` or `vector<2x32xi8>`.

Just for the record, there is no shuffle mode for `vector<2x8xi32>` <-> `vector<8x2xi32>` nor `vector<2x32xi8>` <-> `vector<32x2xi8>`, those cases will need to be handled as a composition of shuffle operations (or using 1024-bit shuffles). I just started diving into the enthralling world of permutation decomposition, let's see where that gets me 😬

PR #1528, opened 2024-05-31
[aievec] add aievec.min/max e2e to-llvm tests 

jamestcl-amd:
This PR updates the existing element-wise `aievec.min/max` e2e test to support the to-llvm flow.

Changes:
- Update the conversion pattern of `aievec.min/max` to the llvm intrinsic. The third operand `cmp` is implicitly treated as `sign` to the vmax and vmin intrinsic. Will confirm this to the owner. 
- Add element-wise `aievec.min/max` e2e tests for the to-llvm flow. This includes updating the testbench.cc and the test script. These tests, like other to-cpp tests, go through the simulator to verify the numeric correctness.

Todo:
- `aievec.min/max` e2e test for bf16 datatype fails at the peano clang. Need further investigation on it.

### DIFF
--- a/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
+++ b/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
@@ -1656,6 +1656,103 @@ public:
   }
 };
 
+class MinOpConversion : public mlir::ConvertOpToLLVMPattern<aievec::MinOp> {
+public:
+  using ConvertOpToLLVMPattern<aievec::MinOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(aievec::MinOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+
+    VectorType resultType = cast<VectorType>(op.getResult().getType());
+    Type resultScaTy = resultType.getElementType();
+    unsigned resultBitWidth = resultScaTy.getIntOrFloatBitWidth();
+    int resultLanes = getVectorLaneSize(resultType);
+    int resultVectorSize = resultBitWidth * resultLanes;
+
+    // aievec.min op has the AllTypesMatch constraint on lhs/rhs/res
+    if (resultVectorSize != 512) {
+      op.emitWarning() << "aievec.min conversion with " << resultVectorSize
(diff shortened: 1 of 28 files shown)


lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -1584,9 +1584,10 @@ class MaxOpConversion : public mlir::ConvertOpToLLVMPattern<aievec::MaxOp> {
     // create xllvm intrinsic
     Value maxOp = nullptr;
     if (llvm::isa<IntegerType>(resultScaTy)) {
-      // create constant for cmp
+      // create constant for third operand `cmp`
+      // Note: `cmp` is implicitly treated as `sign` to the vmax intrinsic
```
david-vc, line 1588: Thanks for the comment.
This is an oversight in the definition of aievec.min and aievec.max. The existing testcases assume signed comparison.
These should have an attribute specifying whether it uses a signed comparison or an unsigned comparison.
In the arith dialect, there are 4 min ops: minsi, minui, mininumF, minNumF. We deliberately wanted a single op to handle both the integer and floating point cases, but we failed to distinguish between minsi and minui.
Regarding minimumF vs minNumF, I did not find in the arch doc whether -0 < +0.

david-vc (approved):
I'm OK to address the missing signed attribute for aievec.min and aievec.max in a later PR.

jsetoain:
Silly question, can't Peano deal with the `arith.min/max` instructions if we just lower them to LLVM IR? It feels like it's the kind of thing that should just work.

jamestcl-amd:
> Silly question, can't Peano deal with the `arith.min/max` instructions if we just lower them to LLVM IR? It feels like it's the kind of thing that should just work.

It works when lowering to generic scalar LLVM, but not with generic vector LLVM (e.g. `@llvm.smax.v32i16`). For the generic vector LLVM, it fails at the legalizer.

jsetoain:
> > Silly question, can't Peano deal with the `arith.min/max` instructions if we just lower them to LLVM IR? It feels like it's the kind of thing that should just work.
> 
> It works when lowering to generic scalar LLVM, but not with generic vector LLVM (e.g. `@llvm.smax.v32i16`). For the generic vector LLVM, it fails at the legalizer.

This should be notified to Peano folks, and they should fix it. Once this works in Peano, we can get rid of this translation 😃

PR #1529, opened 2024-06-03
Attribute to choose shared memory module in ObjFIFO

AndraBisca:
This PR adds a new `via_shared_mem` attribute to the `ObjectFifoCreateOp` that takes as input either 0 or 1 (to represent the Consumer and Producer tiles respectively) which will direct the lowering on which shared memory module to create the objects in, if possible.

This feature can only be used with Object FIFOs that have access to shared memory modules, those that require to use the AXI stream ignore this attribute.

The `object_fifo` python wrapper has been updated with a setter function `set_via_shared_mem(ObjectFifoPort port)` so that users can now set the attribute like below:
```
of1 = object_fifo("of1", T_, C_, 2, T.memref(256, T.i32()))
of1.set_via_shared_mem(ObjectFifoPort.Consume)
```


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1610,7 +1610,10 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [HasParent<"DeviceOp">, Symbol]
         TypeAttrOf<AIE_ObjectFifoType>:$elemType,
         BDDimLayoutArrayAttr:$dimensionsToStream,
         BDDimLayoutArrayArrayAttr:$dimensionsFromStreamPerConsumer,
-        DefaultValuedAttr<BoolAttr, "false">:$via_DMA
+        DefaultValuedAttr<BoolAttr, "false">:$via_DMA,
+        // via_shared_mem==0 means use producer tile's memory module
```
stephenneuendorffer, line 1614: As I understand this, the goal is to control the buffer placement.  However a buffer could be shared but not in the producer or consumer (i.e. in a column, a producer in row 2 communicates to a consumer in row 4 through a buffer in row 3.)  So is this really sufficient?  At some level, even in a non-shared memory situation we have the same question, as to where buffers are allocated.  Perhaps we should really be looking into how to solve some of these buffer placement questions more holistically?

AndraBisca: I fully agree with you Steve. There are a lot of scenarios in which we could use the shared memory module of another tile rather than the one in the producer / consumer. At the moment the lowering will always allocate objects in the producer's shared memory module and doesn't do any analysis to figure out whether that is the best choice. I definitely want to move towards a more general solution in the near future.

The addition in this PR is a momentary solution to at least be able to choose between two available shared memory modules in a column, as we've encountered that situation is Gagan's designs and this addition would enable us to progress faster. However, we don't need to merge this PR into main now. I can keep working on it to cover more cases.

test/python/objFifo.py
```diff
@@ -40,10 +42,13 @@ def objFifo_example():
...
     dev_block = Block.create_at_start(dev.body_region)
     with InsertionPoint(dev_block):
-        S = tile(0, 2)
+        S = tile(0, 0)
         T_ = tile(1, 2)
+        C_ = tile(1, 3)
 
         of0 = object_fifo("of0", S, T_, 2, T.memref(256, T.i32()))
+        of1 = object_fifo("of1", T_, C_, 2, T.memref(256, T.i32()))
+        of1.set_via_shared_mem(ObjectFifoPort.Consume)
```
jgmelber, line 51: I'm not sure if it is clear what this line is really doing? Is it indicating the shared memory is on the consumer side? I'd suggest something like ```of1.set_shared_mem_location()```

test/objectFifo-stateful-transform/via_shared_mem_test.mlir
```diff
@@ -0,0 +1,36 @@
...
+// CHECK:   }
+// CHECK: }
+
+module @viaSharedMem {
+ aie.device(npu1) {
+    %tile12 = aie.tile(1, 2)
+    %tile13 = aie.tile(1, 3)
+
+    aie.objectfifo @of0 (%tile12, {%tile13}, 1 : i32) : !aie.objectfifo<memref<16xi32>>
+    aie.objectfifo @of1 (%tile12, {%tile13}, 1 : i32) {via_shared_mem = 1 : i32} : !aie.objectfifo<memref<16xi32>>
```
jgmelber, line 34: Another idea: ```shared_mem_tile = %tile13```

test/python/objFifo.py
jgmelber: Along the note of the idea in the MLIR test case: ```of1.set_shared_mem_tile(C_)```

test/objectFifo-stateful-transform/via_shared_mem_test.mlir
AndraBisca: I don't think it is possible in MLIR to pass the Value result of an Operation as an attribute to another Operation. Maybe something different we can consider is using %tile13 as input to a new operation to set the shared memory module tile, similar to how it looks in Python.

AndraBisca:
This PR can be merged as is. However, it will have a follow-up PR in the very near future to address some of the good suggestions about how to make it clearer what the feature does, and how to expand the shared memory module choice to other tiles other than the ones in the object FIFO producer / consumers.

PR #1530, opened 2024-06-03
Attribute to specify repeat count in ObjFIFO

AndraBisca:
This PR adds a new `memtile_repeat` attribute to the `ObjectFifoCreateOp`.

This feature can only be used with Object FIFOs that have a Mem tile as a producer.

The `object_fifo` python wrapper has been updated with a setter function `set_memtile_repeat(int num)` so that users can now set the attribute like below:
```
of1 = object_fifo("of1", M, T_, 2, T.memref(256, T.i32()))
of1.set_memtile_repeat(4)
```

This PR is part of a series of PRs that aim to add repeat support to IRON at a higher abstraction level. This PR introduces that feature at the Mem tile level, and future PRs will expand to other tiles.


stephenneuendorffer:
How does this change the semantics of the ObjectFifo?  Or is it just an implementation hint?  If it's changing things semantically, then I think we need to be more careful and it should be part of the op, rather than a later annotation.

jtuyls:
> This PR adds a new memtile_repeat attribute to the ObjectFifoCreateOp.

Why `memtile_repeat`? This should be applicable to cores as well, so it might be better to just call it `repeat_count`?

PR #1537, opened 2024-06-05
Move conditionals to target model

stephenneuendorffer:


fifield:
We also need to update `include/aie-c/TargetModel.h`, `lib/CAPI/TargetModel.cpp` and `python/AIEMLIRModule.cpp` to get python support.

PR #1538, opened 2024-06-06
Changes to NpuDmaMemcpyNdOp and AIEDmaToNpu to support sub-word strides, offsets and sizes

pvasireddy-amd:
This PR removes the manual 32-bit alignment of `strides`, `offsets` and `sizes` for `npu_dma_memcpy_nd`. `AIEDmaToNpu.cpp` now automatically aligns `strides`, `offsets[0]` and `sizes[0]` based on the provided element width. 

The previous condition requiring a fixed element bit width of 32 in `AIEXDialect.cpp`  is replaced with two checks: 1. Minimum data transfer size (number of elements * element width) must be at least 32. 2. Element width remains within 32.

Removed hard-coded 32-bit address granularity  assumption. Added `getAddressGenGranularity()` to determine the appropriate alignment based on the target device.

Manual conversions of `strides`, `offsets` and `sizes` has been removed from the `mlir-aie/programming_examples` directory.

New test cases are added to verify sub-word `strides`, `offsets` and `sizes` in `NpuDmaMemcpyNdOp` and the warning for `objectFifoLink` and updated the existing tests.


include/aie-c/TargetModel.h
```diff
@@ -42,6 +42,9 @@ DEFINE_C_API_STRUCT(AieTargetModel, uint64_t);
 
 MLIR_CAPI_EXPORTED AieTargetModel aieGetTargetModel(uint32_t device);
 
+/// Returns the data bus width for the target model.
+MLIR_CAPI_EXPORTED uint32_t aieGetTargetModelDataBusWidth(AieTargetModel targetModel);
```
jgmelber, line 46: This isn't exactly related to a data bus or stream width, it is the address generation granularity for the DMAs. Perhaps this should be ```aieGetTargetModelAddressGenGranularity```, or something

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -61,6 +61,9 @@ class AIETargetModel {
   /// Return the target architecture.
   virtual AIEArch getTargetArch() const = 0;
 
+  /// Return the data bus width of the device.
+  virtual uint32_t getDataBusWidth() const = 0;
```
jgmelber, line 65: Address generation granularity

makslevental: umm `getMinStrideBitWidth`?

makslevental:
So for which arch is the minimum stride not 32? Some forthcoming one that's not public yet?

include/aie-c/TargetModel.h
pvasireddy-amd: Changed

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -1194,6 +1194,27 @@ struct AIEObjectFifoStatefulTransformPass
...
 
+    //===------------------------------------------------------------------------===//
+    // Check whether the same data type is used in I/O FIFOs of the objectFifoLink
+    //===------------------------------------------------------------------------===//
+    // We compare the data type of the input and the output FIFOs of the 
+    // objectFifoLink and emits a warning if there is a mismatch
+
+    for(auto linkOp : device.getOps<ObjectFifoLinkOp>()){
+      auto fifoIn = linkOp.getInputObjectFifos()[0];
+      auto fifoOut = linkOp.getOutputObjectFifos()[0];
```
AndraBisca, line 1205: There may be multiple fifoIns and fifoOuts  (not at the same time), so for completeness you should check them all.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -1194,6 +1194,27 @@ struct AIEObjectFifoStatefulTransformPass
     std::set<TileOp>
         objectFifoTiles; // track cores to check for loops during unrolling
 
+    //===------------------------------------------------------------------------===//
+    // Check whether the same data type is used in I/O FIFOs of the objectFifoLink
+    //===------------------------------------------------------------------------===//
+    // We compare the data type of the input and the output FIFOs of the 
+    // objectFifoLink and emits a warning if there is a mismatch
```
AndraBisca, line 1201: Minor nitpick: `emit` without the 's'

AndraBisca (approved):
Looks good apart from the two small comments I made, thank you for updating all the tests!

jgmelber:
RTM?

pvasireddy-amd:
Now it is ready to merge. Updated tests since no-prolog branch is now merged into main

andrej:
Is it possible that **offset** is in bytes?

[Multiplying my offsets by two](https://github.com/Xilinx/mlir-aie/pull/1551/commits/f678f39041cb254df2e0818779988f0dd244fe59) is the only way I seem to be able to get my matrix multiplication code to verify with different inputs.

This probably slid by because the verification threshold may be too high currently (15%). But if I give it different inputs, with tighter verification tolerance, it only passes if I multiply offsets by two (size of bf16). I will reduce the verification tolerance in my PR so errors like this don't slide by testing in the future.

I did see [this change](https://github.com/pvasireddy-amd/mlir-aie/blob/6f757afd15ada130f43c1f78070e46a0aa0aeda7/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L246), but somehow multiplying my offsets by two is the only way I can fix this. Maybe it should be the last element of the offsets array that should be modified?

PR #1540, opened 2024-06-06
Generalize low level npu ops

fifield:
Some cleanups enabled by https://github.com/Xilinx/mlir-aie/pull/1517

Rename `npu.shimtile_push_queue` to `npu.push_queue`, remove metadata operand. Add direction, channel, column and row operands. The only place this op is used is in the `npu.dma_memcpy_nd` lowering, which already has the metadata information for col, row, direction, and channel. This change generalizes the op to make it useful elsewhere, and so it can target tile and mem tile dmas. I struggled a little to find a syntax for the op, because I couldn't get the direction enum, e.g. `S2MM`, to parse as an attribute, which would have been the natural extension to the old syntax. So I'm happy to take syntax suggestions. This change also makes the op closer to a version of the op in the python bindings, which also takes direction, channel, column as operands instead of metadata.

```mlir
// before
memref.global "public" @toMem : memref<32xi32>
aiex.npu.shimtile_push_queue {metadata = @toMem, issue_token = true, repeat_count = 0 : i32, bd_id = 3 : i32 }
aie.shim_dma_allocation @toMem (S2MM, 1, 0)

// after
// (col, row, dir:channel)
aiex.npu.push_queue (2, 0, MM2S:0) {issue_token = false, repeat_count = 3 : i32, bd_id = 2 : i32 }
```
Rename `npu.writebd_shimtile` to `npu.writebd`, remove column_num operand, add row operand. This change generalizes the op so that it could be used to target tile and mem tile dmas.

Make `npu.write32` column and row operands optional. Before https://github.com/Xilinx/mlir-aie/pull/1517 this operation mapped to a firmware operation that interprets the address operand as a tile relative offset, then under the hood uses (row, column, offset) to compute the write32 address. For backwards compatibility, if row and column are present, this behavior is preserved by `aie-translate --aie-npu-instgen`. Otherwise, the address operand is used without modification as the write32 address, which matches the underlying txn op.


lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -364,18 +336,23 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
     if (!isMM2S)
       issue_token = BoolAttr::get(ctx, true);
 
-    rewriter.create<NpuWriteBdExShimTileOp>(
-        op->getLoc(), column, column_num, ddr_id, bd_id, buffer_length,
-        buffer_offset, enable_packet, out_of_order_id, packet_id, packet_type,
-        d0_size, d0_stride, d1_size, d1_stride, d2_stride, iteration_current,
-        iteration_size, iteration_stride, next_bd, use_next_bd, valid_bd,
+    rewriter.create<NpuWriteBdOp>(
```
jgmelber, line 339: This would need to be extended further for MemTile with another DMA dimension

jgmelber (approved):
LGTM, it's good to generalize now that we aren't trying to represent the firmware ops with the move to TXN. The write BD op still needs some work wrt MemTile

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
fifield: > This would need to be extended further for MemTile with another DMA dimension

Thanks for caching this. I left actual support and tests for memtile as future work.

PR #1545, opened 2024-06-10
[ObjectFifoStatefulTransform] Reuse mlir::loopUnrollByFactor instead of custom function

jtuyls:
I encountered an error with a `reinterpret_cast` operation inside `aie.core`, operating on an `aie.objectfifo.subview.access` operation. The cloned reinterpret cast would be invalid after the `unrollForLoops` method was called. I debugged this a bit and it seemed operands were set incorrectly in some cases, however, instead of fixing the existing method I replaced it with the upstream `mlir::loopUnrollByFactor` method as it would be better to reuse that one instead I think. With this change the `reinterpret_cast` error got resolved as well.


AndraBisca (approved):
I think I meant to use this at some point during the original design of the object FIFO but couldn't figure out how to do it then. I'm happy to replace the custom unrolling if it works with all of our designs! Thank you!

PR #1548, opened 2024-06-11
[aievec] add BF16 aievec.min/max e2e to-llvm tests

jamestcl-amd:
This PR updates the existing BF16 element-wise aievec.min/max e2e test to support the to-llvm flow.

Changes:
- Add BF16 element-wise `aievec.min/max` and `vector.reduction<minimumf/maximumf>` e2e tests for the to-llvm flow. This includes updating the testbench.cc and the test script. These tests, like other to-cpp tests, go through the simulator to verify the numeric correctness.


test/unit_tests/aievec_tests/bf16_min_reduce/testbench.cc
```diff
@@ -6,7 +6,17 @@
...
 #include <float.h>
 
+#ifdef TO_CPP
 void dut(bfloat16 *restrict in0, bfloat16 *restrict out0);
+#elif TO_LLVM
+extern "C" {
+void dut(bfloat16 *in0_allocated, bfloat16 *in0_aligned, int64_t in0_offset,
+         int64_t in0_sizes_0, int64_t in0_strides_0, bfloat16 *out0_allocated,
+         bfloat16 *out0_aligned, int64_t out0_offset, int64_t out0_sizes_0,
+         int64_t out0_strides_0);
```
jsetoain, line 16: Fyi, there's a way to lower FuncOp to LLVM that interprets `memrefs` as bare pointers, it's the option `use-bare-ptr-memref-call-conv`. It might simplify these examples.

jamestcl-amd: I'm not aware of that. Let me update the example. The CPP should be a lot easier...

jamestcl-amd: Will change all the `testbench.cc` to use bare-pointer functions with the `-convert-func-to-llvm='use-bare-ptr-memref-call-conv=1'` in another PR. Let me know if this PR is good to go. Thanks :)

jsetoain: Don't need the `=1`, just the presence of the flag 🙂

PR #1550, opened 2024-06-11
[aie.objectfifo.link] Enable multiple consumers with dimensionsToStream

jtuyls:
Removes an unnecessary check on `ObjectFifoLinkOp` emitting an error on multiple consumer objectFifos with dimensionsToStream set. The codegen for this is actually supported as demonstrated by the added and updated tests.


makslevental (approved):
LGTM

PR #1134, opened 2024-03-15
Stochastic matmul verify

andrej:

Issue #1547, opened 2024-06-11
ObjectFifo - Assertion Error in Nested Loop with Only One Iteration


andrej:
I'm just making this issue to document an error (and workaround, see below) I'm seeing with the ObjectFifo loop unrolling:

**Summary**

- Three nested loops
- Middle loop only has a single iteration
- ObjectFifo accesses in both the two inner loops
- The above conditions result in an assertion error, causing the compiler to crash

**Error**

```
/usr/include/c++/11/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = xilinx::AIE::BufferOp*; _Alloc = std::allocator<xilinx::AIE::BufferOp*>; std::vector<_Tp, _Alloc>::reference = xilinx::AIE::BufferOp*&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: 
Assertion '__n < this->size()' failed.
Aborted (core dumped)
```
**Compilation Command**

```bash
aiecc.py --aie-generate-cdo --no-compile-host --xclbin-name=bug.xclbin \
                         --aie-generate-npu --npu-insts-name=bug.txt bug.mlir

```

**Code**

The "unique" thing about this code is that we have a loop with only a single iteration. If we make it multiple iterations, the error does not happen. The error also does not happen when we only have two, not three, nested loops.

```mlir
module {
  aie.device(npu1_4col) {

    %c0 = arith.constant 0 : index
    %c1 = arith.constant 1 : index
    %c4 = arith.constant 4 : index
    %c4294967295 = arith.constant 4294967295 : index

    %tile_0_1 = aie.tile(0, 1)
    %tile_0_2 = aie.tile(0, 2)

    aie.objectfifo @fifoA(%tile_0_2, {%tile_0_1}, 2 : i32) : !aie.objectfifo<memref<64x64xbf16>>
    aie.objectfifo @fifoB(%tile_0_1, {%tile_0_2}, 2 : i32) : !aie.objectfifo<memref<64x64xbf16>>

    %core_0_2 = aie.core(%tile_0_2) {

      scf.for %arg0 = %c0 to %c4294967295 step %c1 {
        scf.for %arg1 = %c0 to %c1 step %c1 {
          %0 = aie.objectfifo.acquire @fifoA(Produce, 1) : !aie.objectfifosubview<memref<64x64xbf16>>
[...]

PR #1551, opened 2024-06-11
Matrix Multiplication Improvements

andrej:
- work around object fifo bug #1547 
- in host verification code, accumulate in float instead of bf16 -- this allows verification to pass even for bigger problem sizes, where before the accumulation errors got too big
- if problem size is beyond threshold, stochastically verify instead of exhaustively  (exhaustively verifying takes way too long, which is probably why the failed verification for larger problem sizes went undetected before) -- if this is merged, we can close #1134 which contains a similar change
- add assertions and comments
- improve verification output
- add workaround comment for excessive program size (CDO generation error `[AIE ERROR] _XAie_LoadProgMemSection():230: Overflow of program memory` e.g. for size M=768 K=768 N=256 if `fifo_depth>1`)

cc @jgmelber


andrej:
- make use of changes in #1538 to simplify code (strides) for whole_array design
- allow changing inner tile size (small m, k, n) for whole_array design
- replace complicated verficiation code with the naive implementation; the idea was that the other implementation would be faster, but it was not (significantly); swapped the naive version back in to have easier maintainability and confidence in the code

andrej:
@jgmelber Kindly kick off the tests for this when you get a chance. I fixed a number of issues with the matmuls that slid by because the verification tolerance was a bit high. I reduced the tolerance, so let's see if the tests reveal more issues.

fifield (approved):

jgmelber:
@andrej matrix_vector still needs to be fixed, the rest look okay

andrej:
@jgmelber Hopefully this will pass now

andrej:
@jgmelber Could you kick off the tests on this please?

andrej:
Looks like tests pass but I forgot to format. Another try please @jgmelber ?

PR #1555, opened 2024-06-13
[aievec][nfc] update e2e tests to use bare ptr for to-llvm flow

jamestcl-amd:
As suggested in https://github.com/Xilinx/mlir-aie/pull/1548. We can use `-convert-func-to-llvm='use-bare-ptr-memref-call-conv'` to minimize the changes to `testbench.cc` when handling the to-llvm tests. This PR updates i8/i16/i32 max/min/mul elem tests.


jsetoain (approved):
lgtm!

PR #1557, opened 2024-06-14
General fixes to IRON guide

mariodruiz:
After going through the guide I found a few things that needed fixing.

In particular the table in section 5 for `Basic` examples, the link were broken due to the `$e^x$` equation that I replaced with HTML

In other cases the text didn't match the images exactly.  

In section 4b, there is a piece of text that's not very clear. I leave here to see if anyone can propose a fix. 

The packet IDs, on the other hand, **\<UNCLEAR MEANING AFTER THIS\>** an be variable but must be globally unique to distinguish routes from one another.


jgmelber (approved):
Thanks for the great fixes!

denolf (approved):

denolf:
Thanks Mario.

PR #1508, opened 2024-05-23
POC multiple kernels in one xclbin

fifield:
This adds `--xclbin-input=input.xclbin` flag to aiecc.py. If this flag is set then `--input input.xclbin` is passed to xclbinutil during xclbin creation, and the new kernel (pdi) is appended to the existing xclbin, creating a new xclbin (specified by `--xclbin-name=output.xclbin` as before).

PR #1561, opened 2024-06-14
Modify aie2xclbin tool to allow merging kernels in existing XCLBIN

nirvedhmeshram:
This adds the same feature added on the python side by https://github.com/Xilinx/mlir-aie/pull/1508
Tested with iree-amd-aie which as far as I know is the only client of this tool.
I am also assuming that iree-amd-aie will correctly provide unique kernel ids here as the tool doesn't give any warnings if you merge two kernels of same kernel id but then gives incorrect results at runtime.
Here is a WIP PR of how iree-amd-aie will use this. Will clean that PR once we have a wheel from this one.
https://github.com/nod-ai/iree-amd-aie/pull/418


newling (approved):
xclbinutil is part of XRT AFAIK. Before this PR, could we generate vmfbs without xrt? The instructions here: https://github.com/nod-ai/iree-amd-aie?tab=readme-ov-file#runtime-driver-setup explain how to build iree-amd-aie with runtime support, sounds like XRT is only needed for if you want to runtime support there. But those instructions don't seem complete anyway, as there is no mention of building mlir-aie. 

It'd be nice one day to have a CI to check the "no runtime" workflow too. Anyway, my overarching thought/question is that it's quite nice for developers to work on codegen without XRT. Hopefully this PR doesn't change that. 

My only other nitty comment is that maybe it would be better to pass the path to xrt directly when looking for xclbinutil, instead of using  `auto xclbinutil = sys::findProgramByName("xclbinutil")`, like is done for peano (i.e. there is a std::string PeanoDir) (not requesting a change though)

fifield:
> xclbinutil is part of XRT AFAIK. Before this PR, could we generate vmfbs without xrt? The instructions here: https://github.com/nod-ai/iree-amd-aie?tab=readme-ov-file#runtime-driver-setup explain how to build iree-amd-aie with runtime support, sounds like XRT is only needed for if you want to runtime support there. But those instructions don't seem complete anyway, as there is no mention of building mlir-aie.
> 
> It'd be nice one day to have a CI to check the "no runtime" workflow too. Anyway, my overarching thought/question is that it's quite nice for developers to work on codegen without XRT. Hopefully this PR doesn't change that.
> 

This PR does not change the xclbinutil requirement. It is already required to generate an xclbin.

nirvedhmeshram:
About changing the way it finds xclbinutil, the only way I see that working is adding an extra flag to iree-amd-aie for the XRT path, which I feel will make the iree-compile command further cluttered. Anyway since this point is not directly related to this PR I will land this so I have wheels on Monday to work on the iree-amd-aie side PR. Thanks for the comments.

PR #1568, opened 2024-06-17
Fully unroll for-loops with tripCount < unrollFactor

### LINKED ISSUE #1556: MLIR::loopUnrollByFactor fails for small loops
After #1545 designs with loops that have less iterations than the unroll factor fail silently. 

It was caught by one of @eddierichter-amd designs.

The example [here](https://github.com/Xilinx/mlir-aie/blob/unroll-bug/test/objectFifo-stateful-transform/loop_test_small.mlir) shows this error and expected output. The loop has 4 iterations and the unroll factor is 6.

@jtuyls

AndraBisca:
Fixes #1556


test/objectFifo-stateful-transform/loop_test_small.mlir
```diff
@@ -0,0 +1,97 @@
+//===- loop_test.aie.mlir --------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2021 Xilinx Inc.
+//
+// Date: February 9th 2022
```
fifield, line 9: Test name, copyright and date need updates.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -806,7 +806,17 @@ struct AIEObjectFifoStatefulTransformPass
...
+          // if loop iterations < unrollFactor, unroll the loop fully
+          if (forLoop.getSingleLowerBound() || forLoop.getSingleUpperBound() ||
+              forLoop.getSingleStep()) {
+            int64_t tripCount =
+                constantTripCount(*(forLoop.getSingleLowerBound()),
+                                  *(forLoop.getSingleUpperBound()),
+                                  *(forLoop.getSingleStep()))
+                    .value_or(0);
+            if (tripCount < unrollFactor)
+              unrollFactor = tripCount;
```
fifield, line 818: Should the `if` on line 810 have `&&` conditions? If any of them are false, but one of them is true, then I think `constantTripCount` will fail and `unrollFactor` gets set to zero.

AndraBisca: That makes sense.

Issue #1578, opened 2024-06-25
npu_dma_memcpy `offset` is wrong


andrej:
Assume a microkernel that just writes `11` repeatedly into its output buffer, and that output buffer is transferred via `fifo_c` to the shim. Assume the data type of the output buffer is `i8` and its length is 8 bytes.

For a data transfer like the following (pay attention to the offset):

```
aiex.npu.dma_memcpy_nd(0, 0, %arg0[0, 0, 0, 1][1, 1, 1, 8][0, 0, 0]) {id = 0 : i64, metadata = @fifo_c} : memref<8xi8>
//                                          ^
```

With offset=1, we would expect output like the following (assuming the output buffer is zero-initialized):

```
   0   11   11   11   11   11   11   11
```

Instead, we get:

```
  11   11   11   11   11   11   11   11
```

The results are identical (i.e. the offset seems to have no effect) for offset=1, offset=2, ... all the way up to offset=15. Once we set offset=16, we get:

```
   0    0    0    0   11   11   11   11 
```

At this point, we would expect
```
   0    0    0    0    0    0    0    0
```
as this offset is larger than the buffer, so the buffer should not be touched by the transfer.

Similar issues, although with different numbers, arise with `i16`. Once we use `i32`, everything works as expected. I suspect this is a rounding error somewhere, where the offset is erroneously multiplied by data type size then divided by four. I think offsets may need to be set in bytes.

I have attached a minimal example that should be easy to reproduce. Play with `offset_c` on line 21 in `aie2.py`.

**[minimal.zip](https://github.com/user-attachments/files/15975557/minimal.zip)**

cc @pvasireddy-amd

andrej:
From my initial tests it seems like the offset is indeed expressed in bytes, but the value has to be 4-byte aligned.

That is, an output like
```
   0   11   11   11   11   11   11   11
```
is not possible since that offset is not 4-byte aligned, but if you want to get
```
   0    0    0    0   11   11   11   11 
```
you can, by setting offset=4.

I don't know if there's documentation somewhere to back this up. (@jgmelber ?)

I'll make a PR with the fix once I added a test case.

stephenneuendorffer:
I thought we had a validator for this case?

andrej:
[This validator](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L67)? It currently does not seem to check offsets.

[...]

PR #1580, opened 2024-06-25
Offset and validation fixes for npu_memcpy_nd

andrej:
#1538 introduced a bug (#1578) to the offset calculation in npu_dma_memcpy. This was not caught because there were not sufficient tests for this feature. This PR

 - fixes the erroneous offset bug
 - refactors the stride/wrap/offset calculations into their own functions; this way they can be reused for validation (we want to validate the same values that we will later lower)
 - adds more validation code for npu_dma_memcpy, to make sure input strides/wraps/offsets are valid (instead of just rounding to the nearest valid value, which is probably almost never what the user expects)
 -  adds more test cases for npu_dma_memcpy


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -110,6 +173,34 @@ LogicalResult AIEX::NpuDmaMemcpyNdOp::verify() {
...
+  }
+
+  bool error = false;
+  std::stringstream msg;
+  for (int i = 0; i < 3; i++) {
+    if (raw_strides[i] * elemWidth % addressGranularity != 0) {
+      error = true;
+      msg << "Stride " << i << " is " << raw_strides[i] << " elements * "
+          << (elemWidth / 8) << " bytes = " << (raw_strides[i] * elemWidth / 8)
+          << " bytes, which is not divisible by 4. ";
```
AndraBisca, line 188: Is this 4 because of the address granularity? Should it be a different number if the address granularity is more than 4 bytes?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -110,6 +173,34 @@ LogicalResult AIEX::NpuDmaMemcpyNdOp::verify() {
...
+  if (error) {
+    return emitOpError(msg.str());
+  }
+
+  if (raw_sizes[0] * elemWidth % addressGranularity != 0) {
+    std::stringstream msg;
+    msg << "Transfer sizes must be multiples of " << (addressGranularity / 8)
+        << " bytes. " << raw_sizes[0] << " elements at " << (elemWidth / 8)
+        << " bytes each equal " << (raw_sizes[0] * elemWidth / 8)
+        << " bytes, which is not divisible by 4. ";
```
AndraBisca, line 200: Same as above for '4'

AndraBisca (approved):
Thank you for the fix Andre! Aside from my two small comments this looks good to me!

lib/Dialect/AIEX/IR/AIEXDialect.cpp
andrej: You're right, I should use the addressGranularity variable for the error message here. All hardware currently uses 4, but that doesn't mean it shouldn't be changeable.

fifield (changes requested):
Thanks for cleaning up the error checking and adding more negative tests. I found a few license nits to complain about.

andrej:
Thanks for having a look @fifield and @AndraBisca. I think I've addressed all feedback

jgmelber (approved):

fifield (approved):

PR #1583, opened 2024-06-26
[aie2xclbin] Towards stdout free compilation

newling:
Removes XRT logging leaking into iree-compile. The one outstanding leakage is from bootgen, I probably need to make a PR in that repo to remove the polluting printf there.


lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -788,10 +793,16 @@ LogicalResult AIETranslateToCDODirect(ModuleOp m, llvm::StringRef workDirPath,
 
   AIEControl ctl(aieSim, xaieDebug, targetModel);
   initializeCDOGenerator(endianness, cdoDebug);
-  if (emitUnified)
-    return generateCDOUnified(ctl, workDirPath, targetOp, aieSim, enableCores);
-  return generateCDOBinariesSeparately(ctl, workDirPath, targetOp, aieSim,
-                                       enableCores);
+
```
newling, line 796: No functional change here

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -711,6 +711,8 @@ void initializeCDOGenerator(byte_ordering endianness, bool cdoDebug) {
 
 LogicalResult generateCDOBinary(const StringRef outputPath,
                                 const std::function<LogicalResult()> &cb) {
+
+  // TODO(newling): Get bootgen team to remove print statement in this function.
```
makslevental, line 715: More likely we'll just carry the patch.

makslevental (approved):
LGTM

PR #1584, opened 2024-06-27
Make aiex.npu.dma_memcpy_nd d0 stride explicit

fifield:
The stride of the inner dimension is assumed to be one for `npu.dma_memcpy_nd` (e.g. [here](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L267)). This PR makes it an explicit operand.

PR #1586, opened 2024-06-28
Make aiex.npu.dma_memcpy_nd d0 stride explicit

fifield:
The stride of the inner dimension is assumed to be one for npu.dma_memcpy_nd (e.g. [here](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L267)). This PR makes it an explicit operand.

replaces #1584

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -180,7 +180,8 @@ LogicalResult AIEX::NpuDmaMemcpyNdOp::verify() {
 
   bool error = false;
   std::stringstream msg;
-  for (int i = 0; i < 4; i++) {
+  // FIXME: skip inner dim
+  for (int i = 1; i < 4; i++) {
     if (raw_strides[i] * elemWidth % addressGranularity != 0) {
       error = true;
       msg << "Stride " << i << " is " << raw_strides[i] << " elements * "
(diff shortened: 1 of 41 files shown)


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -157,41 +157,70 @@ LogicalResult AIEX::NpuDmaMemcpyNdOp::verify() {
...
   }
 
-  bool error = false;
-  std::stringstream msg;
-  for (int i = 0; i < 3; i++) {
+  for (int i = 0; i < 4; i++) {
+    // strides[0] == 1 is ok iff the tranfer size is a multiple of
+    // addressGranularity, which is checked below
+    if (i == 0 && raw_strides[i] == 1)
+      continue;
```
andrej, line 214: Looks good. Technically something like
```
sizes=[addressGranularity/2, addressGranularity/2]
strides=[addressGranularity/2, 1]
```
(and similar examples for higher dimensions, with addressGranularity in bytes)
is equivalent to
```
sizes=[addressGranularity]
strides=[1]
```
but would currently error.

I think it's a non-issue as this is clearly a contrived example with no real uses. (Except maybe automatically generated code?)

So this looks good to me with this check.

andrej (approved):
Looks good to me. Have you tested using a d0 `stride != 1` at all?

PR #1588, opened 2024-07-01
Remove aie2xclbin tool

newling:
This has been moved to iree-amd-aie (its only user afaik). 

Also removed some ununsed variables which were making a Werror build fail.


stephenneuendorffer (approved):

fifield:
There are also some mentions of `aie2xclbin` in the tests and in `CODEOWNERS`.

PR #1590, opened 2024-07-02
Add d0 Stride Test (Matrix Transpose)

andrej:
Adds a test for #1586 that makes use of the d0 stride != 1.

@fifield


fifield (approved):
thanks!

PR #1592, opened 2024-07-02
Make quick_setup.sh point to latest wheel

andrej:
As discussed with @jgmelber. This reverts to pulling the latest nightly wheel instead of whatever version was previously fixed in quick_setup.sh.

We decided that using the latest wheel of main should be sufficiently stable, and it is preferrable to have everyone on the latest version. People trying to run bleeding-edge programming examples from the repo with an outdated compiler pulled from quick_setup.sh has lead to a lot of headaches, and keeping the version in quick_setup.sh updated is laborious. 

There are arguments to keep a version hard-coded in quick_setup.sh, s.t. the repository commit and compiler commit are coupled together, but it seems generally using latest seems to be a better trade-off.

Another stage of this could be to only release wheels after testing them and passing all tests (but this might be redundant with the tests we run on main already anyways).

### DIFF
--- a/utils/quick_setup.sh
+++ b/utils/quick_setup.sh
@@ -61,11 +61,9 @@ if test -f "$VPP"; then
   AIETOOLS="`dirname $VPP`/../aietools"
   mkdir -p my_install
   pushd my_install
-  # pip download mlir_aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels/
-  wget -q --show-progress https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024061822+29d5cec-py3-none-manylinux_2_35_x86_64.whl
+  pip download mlir_aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels/
   unzip -q mlir_aie-*_x86_64.whl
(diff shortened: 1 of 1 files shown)


jackl-xilinx (approved):

utils/quick_setup.sh
```diff
@@ -61,11 +61,9 @@ if test -f "$VPP"; then
   AIETOOLS="`dirname $VPP`/../aietools"
   mkdir -p my_install
   pushd my_install
-  # pip download mlir_aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels/
-  wget -q --show-progress https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024061822+29d5cec-py3-none-manylinux_2_35_x86_64.whl
```
stephenneuendorffer, line 65: could we get to the point where we just pip install the packages into the virtual environment, rather than this dance?

stephenneuendorffer (approved):

utils/quick_setup.sh
andrej: I just looked into this and if I do
```
  pip install mlir_aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels/
```
it actually picks up the right mlir-aie version and installs it. However, it does not seem to add `aiecc.py` where it does currently and so it's not found on the PATH.

Given that, is it cool if I leave this as-is for now?

PR #1598, opened 2024-07-03
Parametrize Number of Used AIE Columns in Whole Array Matrix Multiplication Design

andrej:
Adapts the whole array design to make it possible to run matrix multiplication on 1, 2 or 4 columns with configurable parameter `n_aie_cols`. The number of used rows remains fixed at 4.

For 4 columns (default) this should be the same as the previous whole_array design.

For 2 and 1 columns, the mem tiles hold double (or quadruple) the tile size of A and then distribute those to all rows in the array. Thus, it may be necessary to reduce tile size if you run into memory limits with designs with fewer columns.

Also cleans up the code a bit. Adds tests for 1 and 2 column designs as well by running the same design with different parameters.

### DIFF
--- a/programming_examples/basic/matrix_multiplication/whole_array/Makefile
+++ b/programming_examples/basic/matrix_multiplication/whole_array/Makefile
@@ -17,10 +17,11 @@ N?=768
 m?=16
 k?=64
 n?=48
+n_aie_cols?=2
 
 kernels=mm_${m}x${k}x${n}
-aieargs+=-m $m -k $k -n $n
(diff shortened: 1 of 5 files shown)


andrej:
@jgmelber 

Since I cleaned up a lot of the code it may be hard to see what actually changed to make it work on multiple columns.

So if you are curious about that specifically, commit ac4e83274d9d181a0c7193621fe6c42c40bb4785 and the next one show most of the necessary changes for making it work on multiple columns specifically. (Although I did also rename some variables in that first one ...)

jgmelber (approved):

PR #1603, opened 2024-07-09
add issue_token attribute to Python bindings of dma_memcpy_nd

andrej:


andrej:
@jgmelber Can you please verify that the documentation additions I have made are factually correct?

jgmelber (commented):
Can you add a test for this

andrej:
@jgmelber Test added. For this PR, I chose to do synchronous configure-submit-wait-for BD, due to the issue we discussed (if I submit multiple BDs to the same task queue back to back, there is no synchronization for submitting to the task queue, so they get overwritten -- would need chaining of dma_memcpy_nds but that is probably not within the scope of this PR).

jgmelber (approved):

Issue #1604, opened 2024-07-10
Documentation is not deploying


andrej:
The [online documentation](https://xilinx.github.io/mlir-aie/) has not been updated in quite a while.

The responsible [GitHub workflow](https://github.com/Xilinx/mlir-aie/commits/gh-pages/) is failing with this error:
```
Rendering Liquid: AIEXDialect.md github-pages 231 | Error: Liquid syntax error (line 553): Unknown tag 'tile_0_0' 
```

The fundamental issue is that we have example code in the docs that contains things like 
```
  aie.objectfifo @out0(%tile_0_1, {%tile_0_0}, 4 : i32) : !aie.objectfifo<memref<32x32xi32>>
```

The offending part here is `{%tile_0_0}`. This happens to be an expression in the Liquid templating language. GitHub pages/Jekyll uses liquid to turn the markdown files into HTML pages.

Something as simple as putting a space between the `{` and the `%` may help, but not sure. I'm sure there's also some sort of escape sequence in the Liquid template language that would allow us to write `{%` verbatim.

PR #1605, opened 2024-07-10
Fix Documentation Sequence Interfering with Liquid Templating

andrej:
Attempt at getting the online documentation to build and deploy again.

This is a band-aid fix. Issue #1604 should still be properly addressed by properly _automatically_ escaping "{%" sequences in our documentation even if this ends up fixing it for now. 

There is almost certainly no use for us to use Liquid templating in our documentation strings, but stuff like "{%tile_0_0}" is almost inveitably bound to show up again, and the next person to add it might not know that they need to escape it.


AndraBisca (approved):
Thank you for catching and fixing this. I remember now that the issue has come up in the past so we definitely need a more maintainable way of handling this issue.

andrej:
Don't know why that matrix multiplication design test is failing. This PR leaves it unchanged and the tests have been passing up to this point. How do I restart the test?

PR #1606, opened 2024-07-10
Dynamic ObjectFIFO w/o Loop Unrolling -- Example Lowering / Future Test

andrej:
As discussed with @abisca .

Both `aie2.mlir` with the old static lowering (loop unrolling) and the proposed new lowered code `aie2-objfifo-lowered.mlir` produce the same output and pass the `test.cpp`.

From the included readme:

> 
> Contains an example of what a ObjectFIFO lowering may look like that does not unroll loops, but instead chooses the buffers dynamically.
> 
> `aie2.mlir` shows what the high-level ObjectFIFO **input** looks like.
> 
> `aie2-objfifo-lowered.mlir` shows what the **output** should be after lowering with `aie-opt --aie-objectFifo-stateful-transform aie2.mlir`.
> 
> Once implemented, this can be used as a test case.
> 
> **TODO**: Dynamic lock value calculation, see [here](https://github.com/Xilinx/mlir-aie/commit/01f3642ecfe3b39708d31b42bba7adf179935deb) for an example.


AndraBisca:
This looks great, thank you for taking the time to make a test design with this IR. I will try to apply this lowering on a few other examples with different loop patterns that I have been looking at and see how those compare.

andrej:
@AndraBisca feel free to close this if you don't need it any more / incorporated into your work already

AndraBisca:
I'm just about ready to change this PR to target my branch, just need to push it and switch the merge target. I'll update it early next week!

PR #1613, opened 2024-07-12
Add support for constant-padding multi-dimensional BDs in AIE2

erwei-xilinx:
`aie.dma_bd(%buf : memref<256xi32>, 0, 4, [<size = 2, stride = 128>], [<const_pad_before = 1, const_pad_after = 1>], pad_value = 0)`


lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -324,6 +324,33 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
+      bdOp.getPadDimensions();
+  if (padDims) {
+    XAie_DmaPadTensor dmaPadTensor = {};
+    dmaPadTensor.NumDim = padDims->size();
+    dmaPadTensor.PadDesc = static_cast<XAie_PadDesc *>(
+        calloc(dmaPadTensor.NumDim, sizeof(XAie_PadDesc)));
+    if (!dmaPadTensor.PadDesc)
+      return bdOp.emitError("couldn't allocate array of XAie_PadDesc");
+    // libxaie requires stride in multiples of 32b
+    double elementWidthIn32bWords =
```
stephenneuendorffer, line 338: should be integer arithmetic.

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -324,6 +324,33 @@ LogicalResult configureBdInBlock(XAie_DevInst &devInst, XAie_DmaDesc &dmaTileBd,
...
+        static_cast<double>(bdOp.getBufferElementTypeWidthInBytes()) / 4.0;
+    for (size_t i = 0; i < padDims->size(); i++) {
+      // Pass down dimensions in reverse order.
+      int j = padDims->size() - i - 1;
+      uint8_t before;
+      uint8_t after;
+      before = static_cast<uint8_t>(padDims.value()[i].getZeroPadBefore() *
```
stephenneuendorffer, line 345: Can the padding actually be less than one data beat?  Does it work to pad by one byte?

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -758,6 +758,7 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
         2. the transfer length;
         3. the sizes and strides for n-d tensor addressing (described below);
         4. the "bd_id" with which to associate the buffer descriptor (most often left empty).
+        5. the number of zeros to pad before and after every dimension of an n-d tensor (described below);
 
     `offset`, `len`, `size`s and `stride`s are all denominated in element width; e.g., transferring the whole of
```
stephenneuendorffer, line 763: Clarify that this is in units of the element width, like everything else?

include/aie/Dialect/AIE/IR/AIEAttrs.td
```diff
@@ -159,4 +159,26 @@ def BDDimLayoutArrayArrayAttr : ArrayOfAttr<
...
+    "uint16_t" : $zero_pad_after
+  );
+
+  let assemblyFormat = "`<` struct(params) `>`";
+}
+
+def BDPadLayoutArrayAttr : ArrayOfAttr<
+    /*dialect*/AIE_Dialect,
+    /*attrName*/"BDPadLayoutArray",
+    /*attrMnemonic*/"bd_pad_layout_arr",
```
stephenneuendorffer, line 180: spell out 'array'

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -847,6 +848,13 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [
 
     All strides are expressed in multiples of the element width (just like `len` and `offset`)
     **with the caveat that the inner-most dimension's stride must be 1**.
+
+    ## DMA zero padding on AIE-ML Devices
+
+    AIE-ML devices can apply zero padding at the buffer descriptor level, described with pairs of zero count 
+    before and after every dimension, up to three dimensions (four dimensions on memtiles). The zero-padding 
+    counts can be supplied to the `dma_bd` through an optional argument, an array of "tuple-like" attributes 
+    `bd_pad_layout<zero_pad_before, zero_pad_after>`.
```
stephenneuendorffer, line 857: presumably this needs a verifier...
1) what happens if the device doesn't support it?
2) what's the range of valid values?
3) what happens if too many pads are specified?

lib/Targets/AIETargetCDODirect.cpp
erwei-xilinx: Good point. Updated.

lib/Targets/AIETargetCDODirect.cpp
erwei-xilinx: No. In the underlying hw it is counting as either (1) if innermost dimension or if zero padding before, # 32-bit words, or else (2) number of n-1 wraps.

This kind of things really need to be enforced with proper verifiers.

include/aie/Dialect/AIE/IR/AIEOps.td
erwei-xilinx: Description updated.

include/aie/Dialect/AIE/IR/AIEOps.td
erwei-xilinx: Verifier added to check for cases above.

lib/Targets/AIETargetCDODirect.cpp
erwei-xilinx: Verifier updated.

PR #1614, opened 2024-07-12
refactor data structure in Pathfinder

Yu-Zhewen:
The following classes have been refactored:

`SwitchboxNode`: it now creates and holds a connectivity matrix for each switchbox, whose rows and cols are representing the ports into and out of the switchbox. The elements inside this matrix are ternary values, where `-1`: invalid in arch, `0`: empty and available, `1`: occupued.

`ChannelEdge`: still represents a directed connection between two `SwitchboxNode`s, but now the direction (`North/South/West/East`) are inferred from corrdinates. Also `overCapacity` and `usedCapacity` (used for `dijkstraShortestPaths`), are now all moved to the `grid` level instead, with the potential of parallel computation.

`grid`: still stores the mapping between corrdinates and `SwitchboxNode`s . It is also used more often now in `Pathfinder` so that `llvm/ADT/DirectedGraph` is no longer required.

The main benefit of these refactors is making router more aware of the architecture constraints on valid switchbox connections, where before the router only knows the bundle capacities. For example, we used to run the router first and then have a heuristic-based `attemptFixupMemTileRouting`  to do partial rerouting afterwards. Instead, it is now dealt by the router itself internally.


makslevental:
Clever.

erwei-xilinx (approved):

PR #1617, opened 2024-07-12
Avoid test conflicts between peano and chess

stephenneuendorffer:
Basically there are two kinds of tests:
1) tests which can use either compiler
2) tests which specifically use one or the other

We need to be careful in the second case to not mix the two compilers.


fifield (approved):
Typically when I build with peano it is also the default, so it's easy to miss the need for the aiecc command line flags.

hunhoffe:
Reviewing against current main: the peano/chess test-conflict problem was resolved independently by #1876 ("Fix test failures when running with llvm-aie"), which took the same "chess and peano don't mix" conclusion but landed it as commented-out //  UN: cases rather than new DONTRUN markers. The AIE1 chess_compiler_tests/00_itsalive was later removed altogether in #2623. The other bits here didn't end up being needed — 31_stream_core stayed peano-gated, so the chess stream-intrinsic wrappers weren't required. Since the core issue is covered, closing this as superseded. Happy to revisit any particular pieces of this PR as needed.

PR #1623, opened 2024-07-17
Fixing up PLIO compiler support and creating an example

eddierichter-amd:
Adding the necessary passes to have a `plio` attribute in an aie2.py and have that lower to connecting the PLIO on the VCK5000 platform. 

This will be coupled with a PR in the [platforms repo](https://github.com/Xilinx/ROCm-air-platforms) with a platform that has PLIO connected on the VCK5000.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -1538,7 +1538,9 @@ def AIE_ShimDMAAllocationOp : AIE_Op<"shim_dma_allocation", [HasParent<"DeviceOp
     ins FlatSymbolRefAttr:$sym_name,
         DMAChannelDir:$channel_dir,
         AIEI64Attr:$channel_index,
-        AIEI64Attr:$col
+        AIEI64Attr:$col,
+        // If this is set we are using the PLIO in this ShimTile
+        DefaultValuedAttr<BoolAttr, "false">:$plio
   );
 
   let results = (outs);
@@ -1634,7 +1636,8 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [HasParent<"DeviceOp">, Symbol]
         TypeAttrOf<AIE_ObjectFifoType>:$elemType,
         BDDimLayoutArrayAttr:$dimensionsToStream,
         BDDimLayoutArrayArrayAttr:$dimensionsFromStreamPerConsumer,
-        DefaultValuedAttr<BoolAttr, "false">:$via_DMA
+        DefaultValuedAttr<BoolAttr, "false">:$via_DMA,
+        DefaultValuedAttr<BoolAttr, "false">:$plio
   );
 
   let assemblyFormat = [{
--- a/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
@@ -126,9 +126,9 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
           } else if (srcBundle ==
                      WireBundle::PLIO) { // PLIO at start of flows with mux
             ShimMuxOp shimMuxOp = analyzer.getShimMux(rewriter, srcSB.col);
-            addConnection(
-                rewriter, cast<Interconnect>(shimMuxOp.getOperation()),
-                flowOp, srcBundle, srcChannel, WireBundle::North, shimCh);
+            addConnection(rewriter,
+                          cast<Interconnect>(shimMuxOp.getOperation()), flowOp,
+                          srcBundle, srcChannel, WireBundle::North, shimCh);
           }
         }
         for (const auto &[bundle, channel] : setting.dsts) {
--- a/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
@@ -159,7 +159,7 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
                 addConnection(
                     rewriter, cast<Interconnect>(shimMuxOp.getOperation()),
                     flowOp, WireBundle::North, shimCh, bundle, channel);
-              } else if (bundle == WireBundle::PLIO) { 
+              } else if (bundle == WireBundle::PLIO) {
                 ShimMuxOp shimMuxOp = analyzer.getShimMux(rewriter, curr.col);
                 addConnection(
                     rewriter, cast<Interconnect>(shimMuxOp.getOperation()),
--- a/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
@@ -105,7 +105,7 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
         int shimCh = srcChannel;
         // TODO: must reserve N3, N7, S2, S3 for DMA connections
         if (curr == srcSB && analyzer.getTile(rewriter, srcSB.col, srcSB.row)
-                                 .isShimNOCorPLTile()) {
+                                 .isShimNOCTile()) {
           // shim DMAs at start of flows
           if (srcBundle == WireBundle::DMA) {
             shimCh = srcChannel == 0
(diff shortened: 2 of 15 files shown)


jgmelber (approved):

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1634,7 +1636,8 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [HasParent<"DeviceOp">, Symbol]
         TypeAttrOf<AIE_ObjectFifoType>:$elemType,
         BDDimLayoutArrayAttr:$dimensionsToStream,
         BDDimLayoutArrayArrayAttr:$dimensionsFromStreamPerConsumer,
-        DefaultValuedAttr<BoolAttr, "false">:$via_DMA
+        DefaultValuedAttr<BoolAttr, "false">:$via_DMA,
+        DefaultValuedAttr<BoolAttr, "false">:$plio
```
stephenneuendorffer, line 1640: This seems a bit odd to me..  Can't the objectFifo figure out that it's connected to PLIO based on the shimDMAAllocation op?  Could I have an objectFifo connecting 2 PLIOs?

lib/Targets/AIETargetXAIEV2.cpp
```diff
@@ -725,24 +725,44 @@ mlir::LogicalResult AIETranslateToXAIEV2(ModuleOp module, raw_ostream &output) {
...
+          connectOp.emitWarning(
+              "Currently only PLIO channel 0 and 1 are supported.");
+        }
+
+        if (connectOp.getDestBundle() == WireBundle::North)
+          // mux
+          output << "__mlir_aie_try(XAie_PlToAieIntfEnable(" << deviceInstRef
+                 << ", " << tileLocStr("x", "y") << ", "
+                 << connectOp.destIndex() << ", PLIF_WIDTH_64));\n";
+      }
```
stephenneuendorffer, line 765: What about the case of being connected TO a PLIO?

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -125,13 +125,10 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
...
           } else if (srcBundle ==
                      WireBundle::PLIO) { // PLIO at start of flows with mux
-            if (srcChannel == 2 || srcChannel == 3 || srcChannel == 6 ||
-                srcChannel == 7) { // Only some PLIO requrie mux
-              ShimMuxOp shimMuxOp = analyzer.getShimMux(rewriter, srcSB.col);
-              addConnection(
-                  rewriter, cast<Interconnect>(shimMuxOp.getOperation()),
-                  flowOp, srcBundle, srcChannel, WireBundle::North, shimCh);
-            }
+            ShimMuxOp shimMuxOp = analyzer.getShimMux(rewriter, srcSB.col);
```
stephenneuendorffer, line 128: It's unclear to me what the point of this change is, but I don't understand the original code necessarily.  Is this related to the fact that you only 'support' channel 0 and 1 in the code below?

stephenneuendorffer (changes requested):

include/aie/Dialect/AIE/IR/AIEOps.td
eddierichter-amd: I am using this as a way for the user to specify whether the corresponding shim tile(s) of an object FIFO should be connected to GMIO or PLIO. Here is an example: https://github.com/Xilinx/mlir-aie/blob/plio-rebase/programming_examples/basic/passthrough_dmas_plio/aie2-input-plio.py#L49. 

We do also need to encode this in the shimDMAAllocationOP but that is later in the compilation, and I believe usually not exposed via the IRON python bindings. 

I believe an object FIFO could connect 2 PLIOs but have never tested it. It can connect two GMIOs, correct?

lib/Targets/AIETargetXAIEV2.cpp
eddierichter-amd: Currently we are only supporting the 0th and 1st channel as mentioned in https://github.com/Xilinx/mlir-aie/blob/e05bff488e5c3fc940334b0da7a603d0106638f6/lib/Targets/AIETargetXAIEV2.cpp#L757. For AIE -> PL using those channels it is a straightshot to the PL and we actually don't need to configure anything. From PL -> AIE we need a little configuration which is what this line is doing. This is all tested in hardware.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
eddierichter-amd: I don't know how correct the previous code was, but in the pass here: https://github.com/Xilinx/mlir-aie/blob/d08090b51dc480ea1eacb9d16d8770a463093047/lib/Targets/AIETargetXAIEV2.cpp#L709-L767 we use the information in the shim mux to know whether we should enable PLIO or GMIO.

lib/Targets/AIETargetXAIEV2.cpp
stephenneuendorffer: OK, would like to see the proper configuration happen here, since there's no fundamental reason why the other cases can't be implemented as well.  At the very least, this should be an error rather than a warning.

stephenneuendorffer (commented):
Thanks for pulling this together!

lib/Targets/AIETargetXAIEV2.cpp
eddierichter-amd: Good point, it was originally an error but that made one of the plio unit tests fail. The reason that I didn't put them in there is that it is difficult to test as I am not sure how at platform build time to force a PLIO connection to use a specific channel. I will look into it and if that doesn't work, I will at least make it an error and change the unit test that was failing.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -1141,18 +1146,32 @@ struct AIEObjectFifoStatefulTransformPass
...
+        if (producer.getPlio()) {
+          producerWireType = producer.getProducerTileOp().isShimTile()
+                                 ? WireBundle::PLIO
+                                 : WireBundle::DMA;
+          consumerWireType = !(producer.getProducerTileOp().isShimTile())
+                                 ? WireBundle::PLIO
+                                 : WireBundle::DMA;
+        } else {
+          producerWireType = WireBundle::DMA;
+          consumerWireType = WireBundle::DMA;
```
AndraBisca, line 1160: I think this might not be needed as these are the default values assigned at variable creation.

include/aie/Dialect/AIE/IR/AIEOps.td
AndraBisca: The `shimDMAAllocationOp` is currently being generated by the AIEObjectFifoStatefulTransform and as the object FIFO also abstracts the `AIEFlowOps` between tiles I think it can be confusing to know where to add new attributes to guide the configuration. A cleaner solution might be to "allow" `shimDMAAllocationOps` to exist before the object FIFO lowering pass and have the lowering pass check for them and either: adopt whatever configuration they specify for the shim tile (including the channel index), or completing any missing information if it was not provided (following the same generation it currently uses). Another possibility would be to rely on the current `AIEFlowOp` in the same manner as above for a more general solution that would allow users to specify the ports they want for any tile not just the shim tile.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
eddierichter-amd: This if statement is in a loop that the variable creation isn't in, so they might not have the same values as the default.

include/aie/Dialect/AIE/IR/AIEOps.td
eddierichter-amd: Either works for me the only important thing to me is that this is user facing. That being said, that change in allowing `shimDMAAllocationOps` to be before the object FIFO lowering pass sounds like it could require some refactoring. If it is non-trivial, should that be a separate PR? If so, I can create an issue so we can track it.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
AndraBisca: Ah right, that's my bad!

include/aie/Dialect/AIE/IR/AIEOps.td
AndraBisca: If we go this route, it will indeed require a bit of refactoring and I do think it will be better to have it in a separate PR that this PR would then build on. Let's discuss further and see if any of those solution would be preferred.

fifield: > A cleaner solution might be to "allow" shimDMAAllocationOps to exist before the object FIFO lowering pass

+1 because it makes things more composable

lib/Targets/AIETargetXAIEV2.cpp
eddierichter-amd: [0f50197](https://github.com/Xilinx/mlir-aie/pull/1623/commits/0f501974e7a6e94b4a43e1cfd77405ad81a06fb3) Removed the channel limitation in this commit.

AndraBisca (approved):
Approved with the intent to refactor some of the code once the PR (to be linked..) generalizing the use of `shimDMAAllocationOps` as discussed is merged.

Thank you for the changes Eddie!

stephenneuendorffer (approved):

PR #1625, opened 2024-07-17
Fall back routine when bank-aware allocation fails

pvasireddy-amd:
Adds flags for the user to select the allocation scheme. When no allocation scheme is given, bank-aware allocation is selected as default and when it fails, sequential allocation is performed.


include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -27,8 +27,8 @@ def AIEAssignBufferAddresses : Pass<"aie-assign-buffer-addresses", "DeviceOp"> {
   let constructor = "xilinx::AIE::createAIEAssignBufferAddressesPass()";
 
   let options = [
-    Option<"clBasicAlloc", "basic-alloc", "bool", /*default=*/"false",
-            "Flag to enable the basic sequential allocation scheme (not bank-aware).">
+    Option<"clAllocScheme", "alloc-scheme", "std::string", /*default=*/"",
+           "Choose allocation scheme; possibilities: basic-sequential, bank-aware. By default, basic-sequential is selected for the Tile's memory and bank-aware is selected for the rest.">,
   ];
```
AndraBisca, line 32: I think this should be something along the lines of: By default, the compiler will try to use the bank-aware scheme first and, if it fails, will fallback to the basic-sequential scheme.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -195,33 +187,84 @@ bool checkAndAddBufferWithMemBank(BufferOp buffer, int numBanks,
   return false;
 }
 
+// Prints the memory map across banks
+void printMemMap(TileOp tile, SmallVector<BufferOp, 4> allocatedBuffers,
+                 SmallVector<BufferOp, 4> preAllocatedBuffers, int numBanks,
+                 std::vector<BankLimits> &bankLimits, int stacksize) {
+  InFlightDiagnostic error =
+      tile.emitOpError("All requested buffers doesn't fit in the available "
+                       "memory: Bank aware\n");
```
AndraBisca, line 196: Here: `don't `instead of `doesn't`

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -380,16 +451,23 @@ struct AIEAssignBufferAddressesPass
...
         if (auto res = basicAllocation(tile); res.failed())
           return signalPassFailure();
       }
-    } else {
+    } else if (clAllocScheme == "bank-aware") {
       for (auto tile : device.getOps<TileOp>()) {
         if (auto res = simpleBankAwareAllocation(tile); res.failed())
           return signalPassFailure();
       }
+    } else {
```
fifield, line 464: might want to warn or error here if clAllocScheme was given by user but unrecognized

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -27,8 +27,8 @@ def AIEAssignBufferAddresses : Pass<"aie-assign-buffer-addresses", "DeviceOp"> {
   let constructor = "xilinx::AIE::createAIEAssignBufferAddressesPass()";
 
   let options = [
-    Option<"clBasicAlloc", "basic-alloc", "bool", /*default=*/"false",
-            "Flag to enable the basic sequential allocation scheme (not bank-aware).">
+    Option<"clAllocScheme", "alloc-scheme", "std::string", /*default=*/"",
+           "Choose allocation scheme; possibilities: basic-sequential, bank-aware. By default, bank-aware is selected and if it fails, will fall back to basic-sequential scheme.">,
```
fifield, line 31: Could this be made to fit into ~80-100 columns?

fifield (approved):

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -195,14 +187,55 @@ bool checkAndAddBufferWithMemBank(BufferOp buffer, int numBanks,
...
 // attributes and update the nextAddrInBanks vector.
-// If it does not find one with enough space, it will allocate the
-// buffer in the last checked bank (this will be picked up during
-// overflow error checking). Finally, the function returns the index
-// of the next bank to search (which should be given to subsequent
+// If it does not find one with enough space, it will throw an error.
+// Finally, the function returns a pass or a fail.
+// The index of the next bank to search (which should be given to subsequent
 // calls of this function to ensure a round-robin allocation scheme
 // over the available banks).
```
AndraBisca, line 240: The function now only returns a pass or fail, and no longer the index, correct?

AndraBisca (approved):
Looks good to me aside from the one comment!

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
pvasireddy-amd: Thank you Andra, updated the comment accordingly.

PR #1626, opened 2024-07-18
[aievec] Add `vector.fma` lowering

jsetoain:


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -628,6 +630,73 @@ struct ConvertMulAddToAIEVecFMAElemOpPattern
   unsigned shiftParam;
 };
 
+// Convert `vector.fma` to `aievec.mac_elem`. Only `vector<16xf32>` and
+// `vector<16xbf16>` operand types are supported. In the case of vectors with
+// `f32` elemental type, this pattern will try to match `bf16` to `f32`
+// widening ops in the `lhs` and `rhs` operands, or fail otherwise.
+// TODO: When sign extensions are not found, a conversion from `f32` to `bf32`
```
newling, line 637:
```suggestion
// TODO: When sign extensions are not found, a conversion from `f32` to `bf16`
```

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -628,6 +630,73 @@ struct ConvertMulAddToAIEVecFMAElemOpPattern
...
+    Value acc = adaptor.getAcc();
+    if (resElemTy.isBF16())
+      acc = rewriter.create<aievec::UPSOp>(
+          fmaOp.getLoc(), VectorType::get({16}, rewriter.getF32Type()), acc,
+          shiftParam);
+    else {
+      lhs = getSourceOfWideningOp(lhs).value_or(nullptr);
+      rhs = getSourceOfWideningOp(rhs).value_or(nullptr);
+      if (!lhs || !rhs) {
+        LDBG("vector.fma operands are f32, and they don't come from arith.extf "
```
newling, line 671: could use `rewriter.notifyMatchFailure(fmaOp, "reason")` instead maybe?

test/Conversion/VectorToAIEVec/test-vector-fma.mlir
```diff
@@ -0,0 +1,37 @@
...
+  return %0 : vector<16xbf16>
+}
+
+// -----
+
+// CHECK-LABEL: test_fma_f32
+// CHECK-SAME: %[[V0:[a-zA-Z0-9]+]]: vector<16xbf16>,
+// CHECK-SAME: %[[V1:.*]]: vector<16xbf16>,
+// CHECK-SAME: %[[V2:.*]]: vector<16xf32>)
+func.func @test_fma_f32(%v0: vector<16xbf16>,
```
newling, line 27: The 16 lanes referred to in the pattern, does that mean only fma with vectors of 16 elements is supported?

newling (commented):
Thank you for working on this!

test/Conversion/VectorToAIEVec/test-vector-fma.mlir
newling: For the matmuls, I use the table https://github.com/Xilinx/mlir-aie/blob/ea17ecba6006c6a33ce38844d0237be8f38bce47/include/aie/Dialect/AIEVec/IR/AIEVecOps.td#L870 to help tiling in IREE. Is there something similar for fma? Or is it as simple as  "fma must be 256 bits"?

jsetoain: Fma must be 256bit (2x16 bfloat16 elements into a single vector of 16 f32 elements). It should be possible to combine a sequence of two linked 256bit fma into a single 2x256bit fma, but I'd have to experiment a bit to make sure I'm not breaking anything.

newling (approved):

PR #1627, opened 2024-07-18
[aievec] Add `aievec.mac_elem` to LLVM conversion

jsetoain:

PR #1630, opened 2024-07-18
[aievec] Clean-up #1626 and #1627

jsetoain:
These two PRs were rushed to unblock other efforts, this patch cleans up a couple of non-critical things left.


jamestcl-amd (approved):
LGTM. Thanks for the clean up!

PR #1640, opened 2024-07-24
Rename `func.func @sequence` to its own op

andrej:
For the following benefits:
 - Enables use of SSA values defined outside of the sequence function scope, so we can e.g. reference `%tile_0_0` defined in the device
 - Previous point will enable better refactorings of some of the NPU ops to use SSA values instead of integers in the future (helps catch mistakes early)
 - Limit to just one sequence function per device
 - Make sure NPU ops are properly nested inside a sequence

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -20,6 +20,7 @@ include "mlir/IR/EnumAttr.td"
 include "mlir/IR/SymbolInterfaces.td"
 include "mlir/Interfaces/CallInterfaces.td"
 include "mlir/Interfaces/SideEffectInterfaces.td"
+include "mlir/IR/CommonAttrConstraints.td"
 
 def AIEX_Dialect : Dialect {
   let name = "aiex";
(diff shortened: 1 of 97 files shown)


include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -463,9 +464,29 @@ def AIE_SelectOp: AIEX_Op<"select", []>, Results<(outs Index)> {
...
+    Typically, these instructions include configuring the data transfers between host and AIE array on the shims.
+    The input arguments are arguments passed in from the host at kernel invocation time. This may include buffers on the host.
+  }];
+  let arguments = (ins
+    Variadic<AnyType>:$args
+  );
+  let regions = (region
+    AnyRegion:$body
+  );
+  let hasCustomAssemblyFormat = 1;
```
fifield, line 482: I'm not sure you need a custom parser and printer for this op.
Does something like this work?
```
'(' ($args^ `:` type($args))? ')' attr-dict `:` regions
```
I probably don't have the syntax quite right, but something like that should work.

andrej: Three problems I ran into with that are:
 1. In the function declaration, `%ssa` actually declares a new SSA value. But if you do it the way you suggest, it needs to reference an already-defined value. The whole idea though is that it's supposed to be a new one that you can then use in the function body.
 2.  I wouldn't know how to let the region parser know that it can use those values as entry arguments.
 3. I think `aiex.sequence (%a, %b, %c) : (memref<>, memref<>, memref<>)` is ugly compared to `aiex.sequence (%a: memref<>, %b: memref<>, %c: memref<>)`

If you know how to address 1 and 2 I'm happy to live with 3 though.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -592,7 +613,7 @@ def AIE_NpuDmaMemcpyNdOp: AIEX_Op<"npu.dma_memcpy_nd", [
   let hasVerifier = 1;
 }
 
-def AIE_NpuDmaWaitOp: AIEX_Op<"npu.dma_wait", []> {
+def AIE_NpuDmaWaitOp: AIEX_Op<"npu.dma_wait", [HasParent<"RuntimeSequenceOp">]> {
```
fifield, line 616: I'd prefer we not add this restriction to all the npu ops. They should be able to exist outside of a sequence. For example to support incremental lowering into and out of aie/aiex dialect. In fact I think there's a case to be made for the npu ops to live in their own standalone dialect. i.e. `npu.*` not `aiex.npu.*`

andrej: Not sure I fully follow. What does `npu.write32` mean outside of a runtime sequence, for example?

fifield: > Not sure I fully follow. What does `npu.write32` mean outside of a runtime sequence, for example?

It means the same, just in a different container.

andrej: The container `aiex.runtime_sequence` in my mind means "this runs on the LX6/7". And I understand `write32(addr, value)` to be specifically an LX6/7 instruction that only it can execute. (Other parts of the AIE may write to registers, but they'll have different address spaces.)

What other containers would this instruction make sense in? Probably not `aie.core` and things like that, since they can't write to arbitrary other AIEs registers?

fifield: > What other containers would this instruction make sense in? Probably not `aie.core` and things like that, since they can't write to arbitrary other AIEs registers?

`my_dialect.foo`, `func.func`, ...

jgmelber:
Be sure to update the programming_guide and programming_examples README's for this change

fifield (approved):

PR #1642, opened 2024-07-26
Update tests for #1640

fifield:

PR #1644, opened 2024-07-26
update PLIO passthrough DMA example with new sequence function syntax

andrej:
Somehow #1640 was merged in despite failing this test. I don't know why our CI allows this?

This should fix it.


andrej:
Jeff already fixed this #1642

PR #1646, opened 2024-07-26
Reorganize wheels

andrej:
Towards `pip install mlir-aie` just working.

- Move `mlir_aie/bin` to `<packagename>.data/scripts` inside wheel.
  - This means `pip install mlir-aie` will put `aiecc.py`, `aie-opt` etc. into a directory that is already on the user's `PATH`.
- Add `mlir` and `mlir-python-utils` as dependencies in the `mlir-aie` wheel.
  - This means users won't have to install these dependencies manually.


andrej:
I have new found respect for Maks for dealing with the mess that is Python setuptools/distutils. Not very fun.

This might end up being kind of hard to make work. I have not seen any provisions in the `wheel` spec that would allow for installing shared libraries, for example.

PR #1650, opened 2024-07-29
Update aiex.runtime_sequence to add optional sym_name, remove unused args

fifield:
changes to `aiex.runtime_sequence`
* Support an optional name in the printer/parser.
* Remove the unused variadic args from the tablegen description so that the default builder is usable.


andrej:
Looks good to me

andrej (approved):

PR #1653, opened 2024-07-31
Fix Col/Row for BlockWrite Op

jgmelber:
Offset address based on col/row provided for block write NPU TXN stream OP.


fifield (approved):
Thanks. I have this sitting in my local branch, which I forgot to push before merge, along with additional test updates.

PR #1654, opened 2024-07-31
npu op lowering test updates

fifield:
This is a missing commit from #1647, rebased onto #1653


test/Targets/NPU/npu_blockwrite_instgen.mlir
```diff
@@ -27,7 +27,7 @@ module {
       %c1_i32 = arith.constant 1 : i32
       // CHECK: 00000001
       // CHECK: 00000000
-      // CHECK: 0611D0C0
+      // CHECK: 0601D0C0
```
jgmelber, line 30: Check was right, the test was wrong

jgmelber: 03661ca854e8dfbfc6ac16196bf91b04f3ed3415

fifield: I suppose 1 is a better check than 0, but makes no difference to me, I'll just rebase again.

PR #1655, opened 2024-08-01
[aievec] Add reordering of extsi -> broadcast ops

jsetoain:
Since, in MLIR, mixed-type arithmetic operations such as:

```mlir
%result = arith.<op> %operandA, %operandB : <narrow type> to <wide type>
```

Are canonically represented as:

```mlir
%wideA = arith.ext<ty> %operandA : <narrow type> to <wide type>
%wideB = arith.ext<ty> %operandB : <narrow type> to <wide type>
%result = arith.<op> %wideA, %wideB : <wide type>
```

This change should improve the chances of matching mixed-type arithmetic ops by getting the sign extension operations closer to the arithmetic, enabling the `ext/ext -> arith` pattern match when a broadcast op ends up between one of the `ext` ops and the `arith` op.

The rewrite pattern is provided as a template so it can be reused whenever more of these "swappable" `ext -> op -> arith` cases show up.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
@@ -326,6 +326,47 @@ struct HoistCastOpToDataSourcePattern : public RewritePattern {
   }
 };
 
+// This pattern swaps a UnaryOpA followed by UnaryOpB. This pattern can be used
+// to improve pattern matching for mixed-type arithmetic ops, by getting sign
+// extension ops closer to the single-type arithmetic operations.
+template <class UnaryOpA, class UnaryOpB>
+struct SwapUnaryOpsPattern : public OpRewritePattern<UnaryOpB> {
+  using OpRewritePattern<UnaryOpB>::OpRewritePattern;
+  // This function takes the chain of operations A->B, and returns the new type
+  // between B and A after the swap.
+  using InferTypeB2AFnTy = std::function<Type(UnaryOpA aOp, UnaryOpB bOp)>;
+  InferTypeB2AFnTy inferTypeB2A = nullptr;
+
+  SwapUnaryOpsPattern(MLIRContext *context, InferTypeB2AFnTy inferType)
+      : OpRewritePattern<UnaryOpB>(context), inferTypeB2A(inferType) {}
+
+  LogicalResult matchAndRewrite(UnaryOpB bOp,
+                                PatternRewriter &rewriter) const override {
+    static_assert(
+        UnaryOpA::template hasTrait<OpTrait::OneOperand>(),
+        "SwapUnaryOps can only be instantiated for single-operand ops");
+    static_assert(
+        UnaryOpB::template hasTrait<OpTrait::OneOperand>(),
+        "SwapUnaryOps can only be instantiated for single-operand ops");
+    UnaryOpA aOp = bOp.getOperand().template getDefiningOp<UnaryOpA>();
+    if (!aOp)
+      return rewriter.notifyMatchFailure(bOp, UnaryOpB::getOperationName() +
+                                                  " not preceeded by " +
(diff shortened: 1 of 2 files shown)


lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
```diff
@@ -326,6 +326,48 @@ struct HoistCastOpToDataSourcePattern : public RewritePattern {
...
+
+  LogicalResult matchAndRewrite(Operation *op,
+                                PatternRewriter &rewriter) const override {
+    static_assert(
+        UnaryOpA::template hasTrait<OpTrait::OneOperand>(),
+        "SwapUnaryOps can only be instantiated for single-operand ops");
+    static_assert(
+        UnaryOpB::template hasTrait<OpTrait::OneOperand>(),
+        "SwapUnaryOps can only be instantiated for single-operand ops");
+    UnaryOpB bOp = cast<UnaryOpB>(op);
```
newling, line 351: Do we know op is UnaryOpB, should this be a dynamic_cast with a check?

newling (commented):
Thanks! Just the one question

lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
jsetoain: Can it be anything else? I believe that, as it is, this pattern should match only for UnaryOpB.

```c++
  SwapUnaryOpsPattern(MLIRContext *context, InferTypeB2AFnTy inferType)
      : RewritePattern(UnaryOpB::getOperationName(), /*benefit=*/1, context), inferTypeB2A(inferType) {}
```

But maybe I'm not understanding something?

lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
```diff
@@ -788,6 +857,8 @@ void xilinx::aievec::buildCanonicalizeVectorForAIEVec(
   // Add `Vector` code canonicalization passes
   // TODO: Add passes to unroll vector with unsupported types
   // TODO: Add passes to split vectors that won't fit in registers
+  if (decodeTargetBackend(options.targetBackend) == TargetBackend::LLVMIR)
```
jamestcl-amd, line 860: What will break when the backend is C++?

jsetoain: It will likely be immediately undone by the other pattern we have, the one that hoists `ext` ops up to the load/block argument. It's disabled for the LLVMIR backend because it was breaking pattern matching in certain circumstances.

jamestcl-amd: I see. Thanks for the explanation.

lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
jamestcl-amd: I guess `: public mlir::OpRewritePattern<UnaryOpB>` will work as well. So that you can avoid that cast?

jsetoain: Yeah... I guess I was being lazy. Let me fix that 🙂

jamestcl-amd (approved):
LGTM :)

lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp
newling: Ok hadn't noticed the `RewritePattern(UnaryOpB::getOperationName()` but that looks like it would have worked too (I was looking for something like `: public mlir::OpRewritePattern<UnaryOpB>`) Thanks!

PR #1656, opened 2024-08-01
Add parametrizable BD chain syntax and new syntax for BD configuration in sequence

andrej:
Adds the following ops:

 - Intended for future general use, but for now only a lowering for the runtime sequence is available:
   - `aie.bd_chain`  -- Define an abstract sequence of `dma_bd` invocations. Think of this as a 'function definition'.
 - Intended for use in the runtime sequence:
   - `aiex.dma_start_bd_chain` -- Materialize an abstract BD chain with concrete arguments. Think of this as an inlined 'function call'. 
   - `aiex.dma_configure_task` -- New way of configuring a task (task = buffer descriptor configuration + channel + direction) in the sequence function that reuses the same familiar existing syntax used on the mem tiles and core memory modules.
   - `aiex.dma_start_task` -- Push a configured task (see above) to a task queue
   - `aiex.dma_await_task` -- Await a task completion token from a previously submitted task.
   - `aiex.dma_free_task` -- Inform the compiler that buffer descriptor IDs that are part of a configured task can be reused after this call. `dma_await_task` implies `dma_free_task`, but it is still useful to have a freestanding `dma_fre_task` to let the programmer express when it is safe to reuse BDs even without an explicit sync (e.g. due to implicit other synchronization).

Adds the following passes:

 - `--aie-materialize-bd-chains`: Concretize invocations of previously defined abstract BD chains. Essentially inlines BD chains.
 - `--aie-assign-runtime-sequence-bd-ids`: Assign buffer descriptor IDs for BDs configured inside the runtime sequence. Reuses components of the existing  `--aie-assign-buffer-descriptor-ids` pass but adapted to the runtime sequence which is more complicated since it allows reuse of BD IDs.
 - `--aie-dma-tasks-to-npu`: Takes BD configurations expressed in the pre-existing syntax, wrapped in some of the new ops above, and turns them into NPU runtime sequence instructions. Expressing BD configuration using the pre-existing syntax is more powerful than e.g. `nd_dma_memcpy_nd` because it allows to express chains of BDs.

Discussed in #1610.

**Example Use of a BD Chain**

```mlir
  aie.device(npu1_4col) {
    %tile_0_0 = aie.tile(0, 0)
    %tile_0_2 = aie.tile(0, 2)

    aie.bd_chain @simple_chain(%arg0: memref<8xi16>, %arg1: memref<12xi16>, %arg2: memref<8xi16>) {
            aie.dma_bd(%arg0 : memref<8xi16>, 0, 8, [<size=1, stride=0>, <size=2, stride=2>, <size=2, stride=4>, <size=2, stride=1>])
            aie.next_bd ^bd1
        ^bd1:
            aie.dma_bd(%arg1 : memref<12xi16>, 0, 12)
            aie.next_bd ^bd2
        ^bd2:
            aie.dma_bd(%arg2 : memref<8xi16>, 0, 8)
            aie.end
    }

    aiex.runtime_sequence(%arg0: memref<8xi16>, %arg1: memref<12xi16>, %arg2: memref<8xi16>) {
      %t1 = aiex.dma_start_bd_chain @simple_chain(%arg0, %arg1, %arg2) : (memref<8xi16>, memref<12xi16>, memref<8xi16>)  
                                    on (%tile_0_0, MM2S, 0) 
      aiex.dma_await_task(%t1)
    }
  }
```

**Example Use of Syntax Reuse for BD Configuration in the Runtime Sequence**

(The more abstract BD chains lower to this in the `--aie-materialize-bd-chains` pass.)

```mlir
 aie.device(npu1_4col) {
    %tile_0_0 = aie.tile(0, 0)
    %tile_0_2 = aie.tile(0, 2)

    aiex.runtime_sequence(%arg0: memref<8xi16>) {

      // Configure a new task: BDs + channel + direction
      %t1 = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
        aie.dma_bd(%arg0 : memref<8xi16>, 0, 8)
        aie.next_bd ^bb1
      ^bb1:
        aie.dma_bd(%arg0 : memref<8xi16>, 0, 8)
        aie.end
      }
     
      // Push the configured task to queue
      aiex.dma_start_task(%t1)

      // Awaiting a task frees previously used BD IDs and issues a sync instruction
      aiex.dma_await_task(%t1)
    }
  }
```

There remains a good number of things to do, but the PR is already getting large and this is a good point to merge.

Further to do's:
 - Integrate the new operations with Python bindings; expressing BD chains in Python is going to be the most challenging part of this
 - Make it possible to refer to shim DMA allocations (e.g. generated from objectfifos) in the `aiex.dma_start_task` and `aiex.start_bd_chain` operations.


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1102,7 +1102,8 @@ def AIE_MemTileDMAOp: AIE_Op<"memtile_dma", [
 }
 
 def AIE_NextBDOp: AIE_Op<"next_bd", [
-    Terminator, ParentOneOf<["MemOp", "MemTileDMAOp", "mlir::func::FuncOp", "ShimDMAOp"]>
+    Terminator, 
+    //ParentOneOf<["MemOp", "MemTileDMAOp", "mlir::func::FuncOp", "ShimDMAOp"]>
```
fifield, line 1106: remove

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
```diff
@@ -0,0 +1,118 @@
+//===- AIEMaterializeBDChains.cpp ---------------------------------*- C++
+//-*-===//
```
fifield, line 2: reformat

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
```diff
@@ -0,0 +1,118 @@
...
+    // Wrap bd chains in DMAConfigureTaskOp regions before inlining
+    r = device.walk([&](DMAStartBdChainOp start_op) {
+      return inlineUsage(device, start_op);
+    });
+    if (r.wasInterrupted()) {
+      return signalPassFailure();
+    }
+
+    // Verify inlined basic blocks do form a chain reachable from the start;
+    // Remove empty blocks
```
fifield, line 90: Can "remove empty blocks" be a canonicalization pattern?

python/compiler/aiecc/main.py
```diff
@@ -1016,6 +1024,9 @@ async def run_flow(self):
                     progress_bar.task,
                     [
                         "aie-opt",
+                        "--aie-materialize-bd-chains",
+                        "--aie-assign-runtime-sequence-bd-ids",
+                        "--aie-dma-tasks-to-npu",
                         "--aie-dma-to-npu",
```
fifield, line 1030: use DMA_TO_NPU here?
```python
["aie-opt", f"--pass-pipeline={DMA_TO_NPU ()}" ...
```

fifield (changes requested):
Looks pretty good, most of my comments are style and formatting nits.

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -147,4 +147,37 @@ def AIEDmaToNpu : Pass<"aie-dma-to-npu", "AIE::DeviceOp"> {
...
+
+def AIEAssignRuntimeSequenceBDIDs : Pass<"aie-assign-runtime-sequence-bd-ids", "AIE::DeviceOp"> {
+  let summary = "Assign IDs to Buffer Descriptors Configured in the Runtime Sequence";
+
+  let constructor = "xilinx::AIEX::createAIEAssignRuntimeSequenceBDIDsPass()";
+  let dependentDialects = [
+    "xilinx::AIE::AIEDialect",
+    "xilinx::AIEX::AIEXDialect",
+  ];
+}
```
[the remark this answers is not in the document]
andrej: Currently it's not combined because `aie-assign-bd-ids` is an AIE pass, whereas this is an AIEX pass operating on AIEX instructions.

What is different is that this "allocator" allows re-use of BD IDs, making it slightly more complicated. The other one sequentially assigns IDs until it runs out. Other than that they do the same thing, so could be merged if AIEX stuff makes it into AIE.

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
```diff
@@ -0,0 +1,118 @@
...
+      chain_def.erase();
+    }
+
+    // Add a start BDs instruction
+    builder.create<DMAStartTaskOp>(start_op.getLoc(), configure_op.getResult());
+
+    // After fully inlining, remove the original instruction
+    start_op.erase();
+
+    return WalkResult::advance();
```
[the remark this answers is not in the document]
andrej: From my understanding, rewrite patterns are not allowed to inspect sibling ops because `aie-opt` may parallelize them. This needs to inspect sibling ops, namely the `bd_chain` ops and the arguments to `await_task` and `free_task`.

I think I tried as a rewrite pattern first and didn't succeed so chose the "easier way". Let me know if I misunderstood something.

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
andrej: Do canonicalization patterns run after each pass? It would be important that empty blocks are removed before subsequent lowering passes, because it may trip `--aie-dma-tasks-to-npu` up if it finds a basic block containing no `dma_bd` ops.

python/compiler/aiecc/main.py
andrej: yeah, looks like we were previously defining it but not using it. weird

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -30,8 +29,14 @@ def AIEX_Dialect : Dialect {
...
     The expectation is that new ideas can be developed here before migration
     to the more mature AIE dialect.
+  }];
 
+  let extraClassDeclaration = [{
+    static uint64_t getBufferDescriptorAddressRegisterAddress(const AIE::AIETargetModel &tm, unsigned bd_id, unsigned col);
+    static std::pair<llvm::SmallVector<int64_t, 4>, llvm::SmallVector<int64_t, 4>> getHardwareStridesWraps(const AIE::AIETargetModel &targetModel, mlir::MemRefType referencedBufType, llvm::SmallVector<int64_t, 4> inputSizes, llvm::SmallVector<int64_t, 4> inputStrides);
+    static mlir::LogicalResult verifyStridesWraps(mlir::Operation *forOp, mlir::MemRefType referencedBufType, int tileCol, int tileRow, llvm::SmallVector<int64_t, 4> inputSizes, llvm::SmallVector<int64_t, 4> inputStrides, llvm::SmallVector<int64_t, 4> hardwareSizes, llvm::SmallVector<int64_t, 4> hardwareStrides);
   }];
+
```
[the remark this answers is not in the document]
andrej: Moved these to be freestanding functions inside namespace xilinx::AIEX instead of static members of AIEXDialect. Please let me know if this is not what you had in mind.

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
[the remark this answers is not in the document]
fifield: That might be a restriction on Passes, but seems like it's safe here with rewrite patterns.

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
fifield: It turns out we don't have any examples of op canonicalization in mlir-aie (yet).

If you put let `hasCanonicalizer = 1;` in the tablegen description for an op, then you get to define `op::getCanonicalizationPatterns` in your Dialect.cpp, and in our case your `runOnOperation` can be rewritten like this:
```c++
    RewritePatternSet patterns(ctx);
    patterns.insert<MaterializeBDChainsPattern>(ctx);
    DMAConfigureTaskOp::getCanonicalizationPatterns(patterns, ctx);
    (void)applyPatternsAndFoldGreedily(device, std::move(patterns));
```
with the benefit that whatever patterns you put in `op::getCanonicalizationPatterns` will also be used by `-canonicalize`

fifield (approved):

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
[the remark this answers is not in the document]
andrej: turned into a rewrite pattern

jgmelber (approved):

PR #1660, opened 2024-08-05
Bump llvm

fifield:
~~This depends on format changes to `dma_bd` op in #1661~~

Now without syntax changes.

PR #1661, opened 2024-08-05
Change dma bd syntax

fifield:
These are the `dma_bd` op syntax changes from #1137. This or something similar is needed to bump mlir, due to upstream printer changes where default valued optional attributes no longer get printed in the same way. e.g. the `offset` operand is not printed when zero. This PR moves everything other than `dims` and `pad_dims` to the attribute dictionary:
```mlir
// old
aie.dma_bd(%buf01_0 : memref<16xi32>, 64, 128, [<size = 2, stride = 1>, ... ])

// new
aie.dma_bd(%buf01_0 : memref<16xi32>, dims = [<size = 2, stride = 1>, ... ]) {offset = 64 : i32, len = 128 : i32 }
```

### DIFF
--- a/docs/AIEDesignPatterns.md
+++ b/docs/AIEDesignPatterns.md
@@ -11,19 +11,19 @@ We can use the AIE Cores as below to perform some operations
 
 Define a tile and a buffer
 ```
-%tile13 = AIE.tile(1, 3)
-%buf13_0 = AIE.buffer(%tile13) { sym_name = "a" } : memref<256xi32>
+%tile13 = aie.tile(1, 3)
+%buf13_0 = aie.buffer(%tile13) { sym_name = "a" } : memref<256xi32>
 ```
 
 Perform some operations on the buffer in the core
 ```
-%core13 = AIE.core(%tile13) {
+%core13 = aie.core(%tile13) {
 	%val1 = constant 7 : i32
 	%idx1 = constant 3 : index
 	%2 = addi %val1, %val1 : i32
 	memref.store %2, %buf13_0[%idx1] : memref<256xi32>
 
-	AIE.end
+	aie.end
 }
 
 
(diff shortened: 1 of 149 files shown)


stephenneuendorffer (commented):
If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified.  Then we don't require a syntax change.

I'm leery of these ops turning into a string of attributes, although it's arguable that they already are.

fifield:
> If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified. Then we don't require a syntax change.
> 
> I'm leery of these ops turning into a string of attributes, although it's arguable that they already are.

Well the simplest thing was to recycle the changes in the old (and now this) PR and fixup the tests over coffee.  But I did want to reopen the discussion on what to do.

fifield:
> If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified. Then we don't require a syntax change.

Another option is to put a keyword on everything in the `assemblyFormat` string:
```
(`,` `offset` `=` $offset^)?  (`,` `len` `=` $len^)? ...
```

stephenneuendorffer:
> > If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified. Then we don't require a syntax change.
> 
> Another option is to put a keyword on everything in the `assemblyFormat` string:
> 
> ```
> (`,` `offset` `=` $offset^)?  (`,` `len` `=` $len^)? ...
> ```

That at least elides the types compared to using the attribute list.

andrej:
> > If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified. Then we don't require a syntax change.
> 
> Another option is to put a keyword on everything in the `assemblyFormat` string:
> 
> ```
> (`,` `offset` `=` $offset^)?  (`,` `len` `=` $len^)? ...
> ```

This option would have my vote over the attr dict. Both options seem fine though.

andrej (approved):

jgmelber:
> > > If I understand the LLVM change, the simplest thing is to implement a custom printer which prints the offset if the length is specified. Then we don't require a syntax change.
> > 
> > 
> > Another option is to put a keyword on everything in the `assemblyFormat` string:
> > ```
> > (`,` `offset` `=` $offset^)?  (`,` `len` `=` $len^)? ...
> > ```
> 
> This option would have my vote over the attr dict. Both options seem fine though.

+1

fifield:
It seems like folks would be most happy with, let's call it option 3, which is `keyword = value`, but this still requires a small fix of the printer. In the process of learning that I cut-n-pasted the parser/printer, modified them to keep the existing syntax, and called it a day (https://github.com/Xilinx/mlir-aie/pull/1660), call that option 2. So I'm closing this one (option 1), which is the least favored.

PR #1664, opened 2024-08-05
npu maskwrite operation

fifield:
This adds `aiex.npu.maskwrite32` to enable use of the npu firmware operation of the same name.

The e2e test works, but it still needs some mlir unit tests.


jgmelber:
Have you tested using this for the DMA push to queue lowering?

fifield:
> Have you tested using this for the DMA push to queue lowering?

No. I have only tried what is in the test in this PR.

PR #1668, opened 2024-08-07
[aievec] Fix xfer inboud attributes

jsetoain:
After the latest LLVM version bump, the pass that set transfer op `inbound` attributes to true was not working properly.


jamestcl-amd (approved):
Thanks a lot for the quick fix!

PR #1675, opened 2024-08-09
[aievec] Lower broadcasts to insert/splat

jsetoain:


jsetoain:
Together with the new lowering, I've included a bit of clean-up here and there. No functional changes, at most simplifications of boolean expressions.

david-vc (approved):
LGTM.

test/dialect/AIEVec/precanonicalization-aieml-llvmir.mlir
```diff
@@ -1,23 +1,43 @@
...
 // CHECK-LABEL: @extsi_to_broadcast_swap(
 // CHECK-SAME: %[[VIN:.*]]: vector<8xi8>
 func.func @extsi_to_broadcast_swap(%v: vector<8xi8>) -> vector<4x8xi32> {
-    // CHECK: %[[BCAST:.*]] = vector.broadcast %[[VIN]] : vector<8xi8> to vector<4x8xi8>
-    // CHECK: %[[EXT:.*]] = arith.extsi %[[BCAST]] : vector<4x8xi8> to vector<4x8xi32>
+    // CHECK: %[[ZV:.*]] = arith.constant dense<0> : vector<4x8xi8>
+    // CHECK: %[[I0:.*]] = vector.insert %[[VIN]], %[[ZV]] [0] : vector<8xi8> into vector<4x8xi8>
+    // CHECK: %[[I1:.*]] = vector.insert %[[VIN]], %[[I0]] [1] : vector<8xi8> into vector<4x8xi8>
+    // CHECK: %[[I2:.*]] = vector.insert %[[VIN]], %[[I1]] [2] : vector<8xi8> into vector<4x8xi8>
+    // CHECK: %[[BC:.*]] = vector.insert %[[VIN]], %[[I2]] [3] : vector<8xi8> into vector<4x8xi8>
```
jamestcl-amd, line 22: One quick question, I don't see the pattern in this PR that can canonicalize a `vector.broadcast` to four `vector.insert`. Is it  a pattern in our existing codebase or from upstream MLIR? Maybe I'm missing something...

jsetoain: I'm using [the upstream lowering of `vector.broadcast`](https://github.com/Xilinx/mlir-aie/blob/ca6c1eaf315055d6624e0fcece1118046d64af4a/lib/Dialect/AIEVec/Transforms/VectorToVectorConversions.cpp#L757), it already does what we need it to do. It also lowers scalar broadcast to splat, hence my changes around that. And I introduced another change that converts "reshape" broadcasts (the ones that simply add leading unit dimensions) because, otherwise, those get converted to inserts that then get canonicalized back to broadcast.

PR #1676, opened 2024-08-09
Switch to txn 1.0 format

fifield:
Update transaction op encoding to 1.0


jgmelber (approved):
This looks good to me. It requires #1687 correct?

fifield:
> This looks good to me. It requires #1687 correct?

It requires a newish driver like the one documented in that PR but I don't know the exact version required.

PR #1678, opened 2024-08-12
Interleave sync instructions in matmul designs

andrej:
to go from this
old trace

to this
new trace

(note the shorter lock stalls after each four/five tile computations)


programming_examples/basic/matrix_multiplication/single_core/opsync_trace_new.json
fifield, line 1: Is it necessary to check in the 2 trace json files?

andrej: oops no, definitely not intended. thanks for catching.

fifield (approved):

PR #1681, opened 2024-08-12
Cascade Matmul Sweep using IRON

Yu-Zhewen:
Using the whole NPU array with broadcast in rows and cascade in columns. The objective is to explore whether this approach yields any performance benefits over the existing `whole_array` configuration. The performance sweep results can be found through this [link](https://gist.github.com/Yu-Zhewen/da3fed9feb278b973f35fb78c2d3a484). No gain is observed so far, but a larger `K` leads to closer performance as original `whole_array`.


programming_examples/basic/matrix_multiplication/cascade/test.cpp
```diff
@@ -0,0 +1,11 @@
+//===- test.cpp -------------------------------------------000---*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2023, Advanced Micro Devices, Inc.
```
erwei-xilinx, line 7: 2024

programming_examples/basic/matrix_multiplication/cascade/aie2.py
```diff
@@ -0,0 +1,388 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2023 AMD Inc.
```
erwei-xilinx, line 6: 2024

PR #1689, opened 2024-08-13
`aiex.configure_task`: Allow constant buffer references (rather than runtime sequence args)

andrej:
With this PR, you can reference buffers on the mem tile, as long as the `dma_configure_task` configures the task on a tile that has access to that buffer. As of now, that's not possible. Only address_patch instructions are emitted for the shim, so only runtime sequence arguments can be used.

This PR is part of a chain of changes that are independent but build on each other. GitHub shows the diff to `main`, so later PRs also contain the changes of previous PRs.

To see minimal diffs (compared to previous PR in chain), click the arrow links between the PRs below:

**#1698  [-->](https://github.com/andrej/mlir-aie/compare/bd-chain-tests..configure-task-allow-other-bufs)  #1689 (this PR)**  [-->](https://github.com/andrej/mlir-aie/compare/configure-task-allow-other-bufs..shim-dma-alloc)  #1693  [-->](https://github.com/andrej/mlir-aie/compare/shim-dma-alloc..py)  #1696  [-->](https://github.com/andrej/mlir-aie/compare/py..bd-py)  #1699

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEInterfaces.td
+++ b/include/aie/Dialect/AIE/IR/AIEInterfaces.td
@@ -134,6 +134,15 @@ def AIETarget : OpInterface<"AIETarget"> {
 }
 
 // Don't delete - see AIEDialect::myVerifyOffsetSizeAndStrideOp
-def MyOffsetSizeAndStrideOpInterface: OpInterfaceTrait<"::xilinx::AIE::MyOffsetSizeAndStrideOpInterface"> {}
+def MyOffsetSizeAndStrideOpInterface : OpInterfaceTrait<"::xilinx::AIE::MyOffsetSizeAndStrideOpInterface"> {}
+
+// Ops may specify that the accessibility checks of their child ops are not verified.
+// For example, by default, ops will verify that all contained aie.buffers are
+// accessible from within the tile from which they are used. However, in abstract
+// definitions not tied to any tile, it may still be useful to be able to refer
+// to buffers and skip those checks.
+def SkipAccessibilityCheckTrait : NativeOpTrait<"SkipAccessibilityCheckTrait"> {
+  string cppNamespace = "::xilinx::AIE";
+}
 
 #endif // AIE_INTERFACES
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -172,7 +172,9 @@ struct UsesAreAccessible {
       if (user->getParentWithTrait<SkipAccessibilityCheckTrait>()) {
         continue;
       }
-      if (auto element = getParentTileElement(user)) {
+      TileElement element;
+      if ((element = llvm::dyn_cast<TileElement>(user)) ||
+          (element = getParentTileElement(user))) {
         auto tileID = element.getTileID();
         if (!targetModel.isLegalMemAffinity(tileID.col, tileID.row, thisID.col,
                                             thisID.row))
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -172,21 +172,23 @@ struct UsesAreAccessible {
       if (user->getParentWithTrait<SkipAccessibilityCheckTrait>()) {
         continue;
       }
-      TileElement element;
-      if ((element = llvm::dyn_cast<TileElement>(user)) ||
-          (element = getParentTileElement(user))) {
-        auto tileID = element.getTileID();
-        if (!targetModel.isLegalMemAffinity(tileID.col, tileID.row, thisID.col,
-                                            thisID.row))
-          return (op->emitOpError("in Column ")
-                  << thisID.col << " and Row " << thisID.row
-                  << " is accessed from an unreachable tile in Column "
-                  << tileID.col << " and Row " << tileID.row)
-                     .attachNote(user->getLoc())
-                 << "user";
-      } else {
+      TileElement element = llvm::dyn_cast<TileElement>(user);
+      if (!element) {
+        element = getParentTileElement(user);
+      }
+      if (!element) {
         // This should probably be caught elsewhere as well.
         return op->emitOpError("is accessed outside of a tile")
+                   .attachNote(user->getLoc())
+               << "user";
+      }
+      auto tileID = element.getTileID();
+      if (!targetModel.isLegalMemAffinity(tileID.col, tileID.row, thisID.col,
+                                          thisID.row)) {
+        return (op->emitOpError("in Column ")
+                << thisID.col << " and Row " << thisID.row
+                << " is accessed from an unreachable tile in Column "
+                << tileID.col << " and Row " << tileID.row)
                    .attachNote(user->getLoc())
                << "user";
       }
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -155,7 +155,7 @@ static TileElement getParentTileElement(Operation *op) {
   return llvm::dyn_cast<TileElement>(parent);
 }
 
-struct UsesAreAccessable {
+struct UsesAreAccessible {
   static LogicalResult verifyTrait(Operation *op) {
     auto thisElement = cast<TileElement>(op);
     auto thisID = thisElement.getTileID();
@@ -1291,7 +1291,7 @@ int64_t BufferOp::getAllocationSize() {
 TileOp BufferOp::getTileOp() { return cast<TileOp>(getTile().getDefiningOp()); }
 
 LogicalResult BufferOp::verify() {
-  if (UsesAreAccessable::verifyTrait(*this).failed())
+  if (UsesAreAccessible::verifyTrait(*this).failed())
     return failure();
   return success();
 }
@@ -2012,7 +2012,7 @@ int LockOp::colIndex() { return getTileOp().colIndex(); }
 int LockOp::rowIndex() { return getTileOp().rowIndex(); }
 
 LogicalResult LockOp::verify() {
-  if (auto result = UsesAreAccessable::verifyTrait(*this); result.failed())
+  if (auto result = UsesAreAccessible::verifyTrait(*this); result.failed())
     return result;
 
   if (getLockID().has_value()) {
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -164,10 +164,15 @@ struct UsesAreAccessible {
     for (auto *user : users) {
       // AIE.useLock may be used in a device to set the lock's default value
       // Allow in a toplevel module for backward compatibility
-      if (llvm::isa_and_nonnull<DeviceOp, ModuleOp>(user->getParentOp()))
-        return success();
+      if (llvm::isa_and_nonnull<DeviceOp, ModuleOp>(user->getParentOp())) {
+        continue;
+      }
+      // If any parent prescribes that accessibility checks be skipped,
+      // skip the check for that user.
+      if (user->getParentWithTrait<SkipAccessibilityCheckTrait>()) {
+        continue;
+      }
       if (auto element = getParentTileElement(user)) {
-
         auto tileID = element.getTileID();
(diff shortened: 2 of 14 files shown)


test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-5.mlir
```diff
@@ -0,0 +1,35 @@
...
+    %tile_0_0 = aie.tile(0, 0)
+    %tile_0_1 = aie.tile(0, 1)
+    %tile_0_2 = aie.tile(0, 2)
+    // CHECK: %{{.*}} = aie.buffer(%tile_0_1) {address = [[ADDR1:[0-9]+]] {{.*}}}
+    %buf0 = aie.buffer(%tile_0_1) : memref<32xi8> 
+    // CHECK: %{{.*}} = aie.buffer(%tile_0_1) {address = [[ADDR2:[0-9]+]] {{.*}}}
+    %buf1 = aie.buffer(%tile_0_1) : memref<32xi8> 
+
+    aiex.runtime_sequence(%arg0: memref<32xi8>) {
+      %t1 = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
```
jgmelber, line 23: Should this be `%tile_0_1` ?

test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-4.mlir
```diff
@@ -0,0 +1,30 @@
...
+  aie.device(npu1_4col) {
+    %tile_0_0 = aie.tile(0, 0)
+    %tile_0_1 = aie.tile(0, 1)
+    %tile_0_2 = aie.tile(0, 2)
+    %buf = aie.buffer(%tile_0_1) { address = 0xBEEF : i32 } : memref<32xi8> 
+
+    aiex.runtime_sequence(%arg0: memref<32xi8>) {
+      // CHECK: aiex.npu.writebd {bd_id = 0 : i32, buffer_length = 4 : i32, buffer_offset = 4 : i32, column = 0 : i32, d0_size = 1 : i32, d0_stride = 0 : i32, d1_size = 2 : i32, d1_stride = 1 : i32, d2_stride = 0 : i32, enable_packet = 0 : i32, iteration_current = 0 : i32, iteration_size = 0 : i32, iteration_stride = 0 : i32, lock_acq_enable = 0 : i32, lock_acq_id = 0 : i32, lock_acq_val = 0 : i32, lock_rel_id = 0 : i32, lock_rel_val = 0 : i32, next_bd = 0 : i32, out_of_order_id = 0 : i32, packet_id = 0 : i32, packet_type = 0 : i32, row = 0 : i32, use_next_bd = 0 : i32, valid_bd = 1 : i32}
+      // CHECK: aiex.npu.write32 {address = 118788 : ui32, value = 48879 : ui32}
+      %t1 = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
```
jgmelber, line 22: Should this be `%tile_0_1` ?

jgmelber (commented):
I am still unclear about what this PR allows regarding accessing buffers allocated to a tile for DMA operations on tiles within the `runtime_sequence` region.

test/bd-chains-and-dma-tasks/materialize-bd-chains/good-1.mlir
```diff
@@ -28,7 +28,7 @@ module {
     aiex.runtime_sequence(%arg0: memref<8xi16>, %arg1: memref<12xi16>, %arg2: memref<8xi16>) {
       %t1 = aiex.dma_start_bd_chain @simple_chain(%arg0, %arg1, %arg2) : (memref<8xi16>, memref<12xi16>, memref<8xi16>)  
                                     on (%tile_0_0, MM2S, 0) 
-      // CHECK: %[[task1:[0-9]+]] = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
+      // CHECK: %[[task1:.+]] = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
```
jgmelber, line 31: These changes unrelated to the functionality of this PR make it harder to review

andrej:
Thanks for the review.

> I am still unclear about what this PR allows regarding accessing buffers allocated to a tile for DMA operations on tiles within the runtime_sequence region.

This:

```
aiex.runtime_sequence(...) {
    %0 = aiex.dma_configure_task( %tile , ...) {
        aie.dma_bd( %buffer , ...)
    }
}
```

will be **allowed** iff `%buffer` is accessible in `%tile`. That is, `%buffer` has been either allocated directly on `%tile` or on one of its neighbor tiles that share memory with it.*

If allowed, the lowering will write the allocated buffer address into the appropriate BD configuration register.

That being said, whether this works is not yet tested. The intent for this PR is to enable @pvasireddy-amd to use this syntax. Since she has experimented with runtime configuration from the shim, she will know best what special other register writes/configuration may be necessary to make this work.

*To contrast, only the following is _currently_ implemented:
```
aiex.runtime_sequence( %arg ) {
    %0 = aiex.dma_configure_task( %shim_tile, ...) {   // _only_ shim tiles can be configured
        aie.dma_bd( %arg, ..)                          // _only_ DDR buffers (i.e. runtime_sequence arguments) are allowed
    }
}
```

test/bd-chains-and-dma-tasks/materialize-bd-chains/good-1.mlir
andrej: Sorry about that, factored out into PR #1698 .
The diff of _this_ PR without those test changes is [here](https://github.com/andrej/mlir-aie/compare/bd-chain-tests..configure-task-allow-other-bufs)

test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-5.mlir
andrej: Yes, this is wrong. Thanks for paying close attention. This should also generate an error I believe, unless we can access memory from mem tiles from the shim directly?

andrej: [This](https://github.com/Xilinx/mlir-aie/blob/d23a3ab31b6dac9fc93be05a697f0ffafb0dcc00/lib/Dialect/AIE/IR/AIETargetModel.cpp#L353) allowed this to pass without error, specifically since `isMemNorth` evaluates to true for tile (0, 0) (I believe).

Since this is encoded in the target model ... can the shim tile in fact access mem tile memory directly?

test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-4.mlir
andrej: Yes it should. Let's discuss this in the other thread

test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-5.mlir
jgmelber: The shim tile cannot access memtile memory

jgmelber (approved):

jgmelber (approved):

test/bd-chains-and-dma-tasks/dma-tasks-to-npu/good-5.mlir
jgmelber: Please make a fix with a test if you can

include/aie/Dialect/AIE/IR/AIEInterfaces.td
```diff
@@ -134,6 +134,15 @@ def AIETarget : OpInterface<"AIETarget"> {
...
+def MyOffsetSizeAndStrideOpInterface : OpInterfaceTrait<"::xilinx::AIE::MyOffsetSizeAndStrideOpInterface"> {}
+
+// Ops may specify that the accessibility checks of their child ops are not verified.
+// For example, by default, ops will verify that all contained aie.buffers are
+// accessible from within the tile from which they are used. However, in abstract
+// definitions not tied to any tile, it may still be useful to be able to refer
+// to buffers and skip those checks.
+def SkipAccessibilityCheckTrait : NativeOpTrait<"SkipAccessibilityCheckTrait"> {
+  string cppNamespace = "::xilinx::AIE";
+}
```
fifield, line 146: If the op is associated with a buffer, then isn't it already associated with at tile? Can you give an example of where this is not true? It seems like there should be a simpler way to to verification.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -164,10 +164,17 @@ struct UsesAreAccessable {
...
+        continue;
+      }
+      // If any parent prescribes that accessibility checks be skipped,
+      // skip the check for that user.
+      if (user->getParentWithTrait<SkipAccessibilityCheckTrait>()) {
+        continue;
+      }
+      TileElement element;
+      if ((element = llvm::dyn_cast<TileElement>(user)) ||
+          (element = getParentTileElement(user))) {
```
fifield, line 177: we could probably clean up the nesting here instead of making it more obfuscated,
```c++
for (...) {
  ...
  auto element = llvm::dyn_cast<TileElement>(user);
  if (!element)
    element = getParentTileElement(user);

  if (!element)
    return op->emitOpError("is accessed outside of a tile")

  auto tileID = element.getTileID();
  if (!targetModel.isLegalMemAffinity(tileID.col, tileID.row, thisID.col, thisID.row))
    return op->emitOpError(" is accessed from an unreachable tile in Column")
}
return success();

File: include/aie/Dialect/AIE/IR/AIEInterfaces.td (hunk 4 above)
Comment (andrej) [reply include/aie/Dialect/AIE/IR/AIEInterfaces.td:146]: > If a op refers a buffer, then isn't it already associated with at tile?
This is true for all ops except the newly introduced `aie.bd_chain`.

> Can you give an example of where this is not true? 
`aie.bd_chain` chains are not associated with any tile _until_ they are concretized (started) on a particular tile, channel, direction.

Before this change, it would be impossible to define an abstract BD chain that referenced a buffer on a tile. After this, you can describe the abstract chain.

The compiler will then throw an error if you try to start that BD chain on a tile that does not have access to the buffer defined within the chain, or allow it if you start it on a 

Here's an example:

```
%some_tile = aie.tile(1, 2)
%other_tile = aie.tile(3, 2)
%buffer_on_some_tile = aie.buffer(%some_tile)

// Abstract BD chain _definition_ should be allowed:
aie.bd_chain @abstract(...) {
  // ...
  aie.dma_bd(%buffer_on_some_tile)
  // ...
}

// Once we concretize it ...
aiex.runtime_sequence () {
  // ... this should be allowed:
  aiex.start_bd_chain @abstract(...) on (%some_tile, MM2S, 0)  
   // ... but this will throw an error, because a buffer referenced within the chain is not accessible on %other_tile:
   aiex.start_bd_chain @abstract(...) on (%other_tile, MM2S, 0)
}
```

lib/Dialect/AIE/IR/AIEDialect.cpp
andrej: Obfuscating my code is how I make myself indispensable. Just kidding I cleaned it up, thanks for the suggestion

include/aie/Dialect/AIE/IR/AIEInterfaces.td
fifield: Got it. The purpose of `SkipAccessibilityCheckTrait` is to exempt `bd_chain` from the requirement that buffers and locks are contained in a `TileElement`. `AIE::DeviceOp` and `ModuleOp` are also exempt if they are the immediate parent but we don't (can't) add the trait to those. All because `bd_chain` can't be a `TileElement` since it can exist outside a `TileOp`. `TileElement`'s only use is to provide tile id and asm printing. If `getTileID` could return `std::optional<AIE::TileID>` then could `bd_chain` be a (logical) `TileElement`?

andrej: Yes, that would be an alternative way to go about this. Then we would need to update the verifier to instead of checking for that trait, checks if `getTileID` returned an empty optional.

Any reason you would prefer this?

It would feel weird to me to make `bd_chain` a `TileElement`, since it can be instantiated on other tiles as well.

fifield: > Any reason you would prefer this?

In general I would prefer to enable logical cores, buffers, locks, etc, but obviously we won't do it in this PR.

PR #1691, opened 2024-08-14
enable writebd_tokens test

fifield:
@pvasireddy-amd This version of `writebd_tokens` works for me.  I had to explicitly set up packet flows for task complete tokens in order to sync with the memtile. By default the compiler enables sync with the shim tile, not with anything else. I also needed a small fix in the cdo code generator.

### DIFF
--- a/lib/Targets/AIETargetCDODirect.cpp
+++ b/lib/Targets/AIETargetCDODirect.cpp
@@ -639,7 +639,8 @@ struct AIEControl {
 
         bool isdma = connectOp.getDestBundle() == WireBundle::DMA;
         // assume a connection going south from row zero gets wired to shimdma
-        // by a shimmux. TODO: fix the assumption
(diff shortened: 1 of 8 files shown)


lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -638,10 +638,6 @@ struct AIEControl {
         }
 
         bool isdma = connectOp.getDestBundle() == WireBundle::DMA;
-        // assume a connection going south from row zero gets wired to shimdma
-        // by a shimmux. TODO: fix the assumption
-        if (!isdma && (switchboxOp.rowIndex() == 0))
-          isdma = connectOp.getDestBundle() == WireBundle::South;
```
fifield, line 644: assumption fixed.

fifield: > assumption fixed.

Actually that broke other stuff. I pushed a different fix.

PR #1693, opened 2024-08-14
`aiex.configure_task`: Allow referencing `shim_dma_allocation`

andrej:
This PR is part of a chain of changes that are independent but build on each other. GitHub shows the diff to `main`, so later PRs also contain the changes of previous PRs.

To see minimal diffs (compared to previous PR in chain), click the arrow links between the PRs below:

**#1698  [-->](https://github.com/andrej/mlir-aie/compare/bd-chain-tests..configure-task-allow-other-bufs)  #1689  [-->](https://github.com/andrej/mlir-aie/compare/configure-task-allow-other-bufs..shim-dma-alloc)  #1693 (this PR)**  [-->](https://github.com/andrej/mlir-aie/compare/shim-dma-alloc..py)  #1696  [-->](https://github.com/andrej/mlir-aie/compare/py..bd-py)  #1699

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -956,7 +956,6 @@ def AIE_DMAAwaitTaskOp : AIEX_Op<"dma_await_task", [HasParent<"RuntimeSequenceOp
   }];
 }
 
-
 def AIE_DMAStartBdChainOp: AIEX_Op<"dma_start_bd_chain", [HasParent<"RuntimeSequenceOp">, TileElement]>,
     Results<(outs Index:$result)>
   {
@@ -994,6 +993,24 @@ def AIE_DMAStartBdChainOp: AIEX_Op<"dma_start_bd_chain", [HasParent<"RuntimeSequ
     AIE::TileOp DMAStartBdChainOp::getTileOp() { return cast<AIE::TileOp>(getTile().getDefiningOp()); }
   }];
 
+}
+
+def AIE_DMAStartBdChainForOp: AIEX_Op<"dma_start_bd_chain_for", [HasParent<"RuntimeSequenceOp">]>,
+    Results<(outs Index:$result)>
+  {
+  let summary = "As dma_start_bd_chain, but specify tile, direction and channel by reference to a Shim DMA allocation op";
+
+  let arguments = (
+    ins FlatSymbolRefAttr:$symbol,
+        Variadic<AnyType>:$args,
+        FlatSymbolRefAttr:$alloc,
+        DefaultValuedOptionalAttr<BoolAttr, "false">:$issue_token,
+        DefaultValuedOptionalAttr<I32Attr, "0">:$repeat_count
+  );
+
+  let assemblyFormat = [{ 
+     $symbol `(` $args `)` `:` `(` type($args) `)` ` ` `for` ` ` $alloc attr-dict 
+  }];
 }
 
 #endif // AIEX_OPS
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -869,6 +869,18 @@ def AIE_DMAConfigureTaskOp : AIEX_Op<"dma_configure_task", [HasParent<"RuntimeSe
   let hasCanonicalizeMethod = 1;
 }
 
+def AIE_DMAConfigureTaskForOp : AIEX_Op<"dma_configure_task_for", [HasParent<"RuntimeSequenceOp">]>, Results<(outs Index:$result)> {
+  let arguments = (
+    ins FlatSymbolRefAttr:$alloc,
+        DefaultValuedOptionalAttr<BoolAttr, "false">:$issue_token,
+        DefaultValuedOptionalAttr<I32Attr, "0">:$repeat_count
+  );
+
+  let regions = (region AnyRegion:$body);
+
+  let assemblyFormat = [{ $alloc regions attr-dict }];
+}
+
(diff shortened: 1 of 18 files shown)


include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -868,6 +869,18 @@ def AIE_DMAConfigureTaskOp : AIEX_Op<"dma_configure_task", [HasParent<"RuntimeSe
   let hasCanonicalizeMethod = 1;
 }
 
+def AIE_DMAConfigureTaskForOp : AIEX_Op<"dma_configure_task_for", [HasParent<"RuntimeSequenceOp">]>, Results<(outs Index:$result)> {
```
jgmelber, line 872: Can you add a summary documenting this OP?

andrej: Added

jgmelber (approved):

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1104,6 +1111,30 @@ bool isLegalTileConnection(TileOp tile, const AIETargetModel &targetModel,
       tile.colIndex(), tile.rowIndex(), srcBundle, srcChan, dstBundle, dstChan);
 }
 
+TileOp TileOp::getOrCreate(mlir::OpBuilder builder, DeviceOp device, int col,
+                           int row) {
+  TileOp tile = nullptr;
+  // Find matching predefined tile at device top level, ...
+  auto tile_ops = device.getOps<AIE::TileOp>();
+  for (auto it = tile_ops.begin(); it != tile_ops.end(); ++it) {
+    TileOp t = *it;
```
fifield, line 1120:
```C++
  for (auto t : device.getOps<AIE::TileOp>()) {
    if (t.getRow() == row && t.getCol() == col) {
      tile = t;
      break;
    }
  }
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1104,6 +1111,30 @@ bool isLegalTileConnection(TileOp tile, const AIETargetModel &targetModel,
...
+    if (t.getRow() == row && t.getCol() == col) {
+      tile = t;
+      break;
+    }
+  }
+  // ... or if undefined, create a new tile op
+  if (!tile) {
+    auto s = builder.saveInsertionPoint();
```
fifield, line 1128: use insertion guard instead of saving/restoring insertion pt
```c++
OpBuilder::InsertionGuard guard(builder);

File: lib/Dialect/AIE/IR/AIEDialect.cpp [hunk 4]
```diff
@@ -2240,6 +2271,22 @@ LogicalResult BDChainOp::verify() {
...
+
+ShimDMAAllocationOp ShimDMAAllocationOp::getForSymbol(DeviceOp device,
+                                                      llvm::StringRef symbol) {
+  auto alloc_ops = device.getOps<ShimDMAAllocationOp>();
+  for (auto it = alloc_ops.begin(); it != alloc_ops.end(); ++it) {
+    AIE::ShimDMAAllocationOp a = *it;
+    if (a.getSymName() == symbol) {
+      return a;
+    }
+  }
```
fifield, line 2286:
```c++
  for (auto allocOp : device.getOps<ShimDMAAllocationOp>())
    if (allocOp.getSymName() == symbol)
      return allocOp;
```
I also suggest taking a look at `ShimDMAllocationGetter` in AIEDmaToNpu.cpp to determine if these can be combined. That version does a symbol lookup to find the op instead of iterating over all ops. In any case it was identified as a hot spot.

test/bd-chains-and-dma-tasks/materialize-bd-chains/good-2.mlir
fifield, line 1: there's a missed `CHECK` update on like 43. The [diff ](https://github.com/andrej/mlir-aie/compare/configure-task-allow-other-bufs..shim-dma-alloc#diff-0d20bc2f3f3d5a3976582a835d6b359b29d591df6391d42537f9b651fc283621R43) between this and the previous stacked commit shows this PR undoing the fix.

andrej: I am amazed you found this. However, after rebasing everything the wrong `CHECK` seems to have vanished (I'm not clear as to why). Can you confirm?

lib/Dialect/AIE/IR/AIEDialect.cpp
andrej: Thanks for pointing to that. I tried using the `ShimDMAAllocationGetter`, however it seems to only work if the symbol is defined somewhere. The `shim_dma_allocation` op is not the op that defines the symbol, it is merely a user. I don't know what's actually supposed to define the symbol, but I suppose I could see from the objectfifo lowering.

Could this be ok as a future PR to clean that up or would you like to see it in this one?

andrej:
@fifield @jgmelber I realized that the new shim-dma-allocation-aware ops did not play well with the BD ID allocation algorithm. The easiest way to fix this was to add a separate pass that factors out the code that turns shim-dma-allocation-referencing ops into their regular counterparts. I also added a test for what broke before this change.

Please double-check that those changes in the latest commit are ok with you.

lib/Dialect/AIE/IR/AIEDialect.cpp
fifield: > Could this be ok as a future PR to clean that up or would you like to see it in this one?

Yeah future PR. I just wanted flag that there was a similar helper class elsewhere.

fifield (approved):

PR #1694, opened 2024-08-15
group packet flows with overlapped source or destination

Yu-Zhewen:
Before doing the routing, the Pathfinder will now assign `group_id` to packet-switched flows. Flows with any overlapping source/destination ports will be assigned with the same id. Only flows with the same group_id can share channels, which is to prevent deadlocks during arbitration. No impact on circuit-switched flows.


erwei-xilinx (approved):

erwei-xilinx:
Cc: @fifield , @jgmelber , @jamesroxbypb. 

We found out that if two packet flows have different sources and different destinations, it is unsafe to allow them to share switchbox channels along the way. This is because the masters and slaves to the two flows could have potentially conflicting cyclostatic scheduling schemes which, since they now share the same route with no buffering capability, could lead to deadlock.

In this PR, @Yu-Zhewen tries to make the packet flow routing more stable by focing the router to choose distinct routes for the above scenario. With that change, one of the congestion tests which was previously passing is now failing due to less route sharing in the new scheme.

PR #1700, opened 2024-08-15
Minor cleanup of 'keep_pkt_header' attribute

fifield:
Minor cleanup of the `keep_pkt_header` attribute. The default behavior should be the same.

* Make the `keep_pkt_header` attr a formal part of the tablegen definition for `aie.masterset` ops
* Use the auto-generated getKeepPktHeader API instead of looking up attribute 
* Make the attribute a true override*. Previously it was interpreted as true whenever present, no matter what it was set to, and not always respected. Now it will override any default.

\* in the AIETargetCODirect backend only, it doesn't seem to be mentioned in the airbin or xaie targets.


erwei-xilinx (approved):
Thanks for fixing that! Yeah before this `keep_pkt_header` will behave as true even if it was set to false.

PR #1705, opened 2024-08-19
Packet flow-based control packet routing overlay

erwei-xilinx:
This PR adds two passes, `AIEAssignTileCtrlIDsPass` and `AIEGenerateColumnControlOverlayPass` to generate a column-wise control packet routing overlay for (1) TCT routing from tiles to shim south 0 port, and (2) control packet for tile reconfiguration.

- `AIEAssignTileCtrlIDsPass` assigns controller ids to tiles in the design, which is a unique identifier per tile used as control packet header for any control packet either entering or leaving the tile. Prior assignment of controller ids shall override this assignment.
- `AIEGenerateColumnControlOverlayPass` generates packet-flow routings for control packets entering or leaving each tile. Each packet flow shall take the tile's controller id as packet header, if the tile has controller id already assigned; else it generates new unique packet headers for control packet flows.


lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,255 @@
...
+                                     SmallVector<int> availableShimChans,
+                                     SmallVector<AIE::TileOp> ctrlTiles,
+                                     WireBundle ctrlWireBundle,
+                                     int coreOrMemChanId,
+                                     int unusedPacketIdFrom, bool isShimMM2S) {
+    // Create packet flows
+    SmallVector<int> thresholdsToNextShimChannel;
+    for (int i = 1; i < (int)availableShimChans.size() + 1; i++)
+      thresholdsToNextShimChannel.push_back(ctrlTiles.size() /
+                                            availableShimChans.size() * i);
```
fifield, line 190: Can you add comments explaining what `thresholdsToNextShimChannel` is doing?

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,255 @@
...
+  // op
+  SmallVector<int> getAvailableShimChans(DeviceOp device, int numShimChans,
+                                         TileOp shimTile,
+                                         WireBundle shimWireBundle,
+                                         bool isShimMM2S) {
+    SmallVector<int> availableShimChans;
+    for (int i = 0; i < numShimChans; i++) {
+      bool isShimChanReservedByAIEFlowOp = false;
+      device.walk([&](AIE::FlowOp fOp) {
+        if (isShimMM2S && fOp.getSource() == shimTile &&
```
fifield, line 227: Could also iterate over the users of `shimTile`? To avoid walking over all ops in the device `numShimChans` times.

test/npu-xrt/memtile_dmas/writebd_tokens/aie.mlir
```diff
@@ -14,21 +14,12 @@ module {
...
-      aie.packet_dest<%tile_0_0, "South" : 0>
-    }
-
     aiex.runtime_sequence(%arg0: memref<4096xi32>, %arg1: memref<4096xi32>, %arg2: memref<4096xi32>) {
 
       // BD0, DMA_S2MM_0_Task_Queue
       aiex.npu.writebd {bd_id = 0 : i32, buffer_length = 4096 : i32, buffer_offset = 0 : i32, column = 0 : i32, d0_size = 0 : i32, d0_stride = 0 : i32, d1_size = 0 : i32, d1_stride = 0 : i32, d2_stride = 0 : i32, enable_packet = 0 : i32, iteration_current = 0 : i32, iteration_size = 0 : i32, iteration_stride = 0 : i32, lock_acq_enable = 0 : i32, lock_acq_id = 0 : i32, lock_acq_val = 0 : i32, lock_rel_id = 0 : i32, lock_rel_val = 0 : i32, next_bd = 0 : i32, out_of_order_id = 0 : i32, packet_id = 0 : i32, packet_type = 0 : i32, row = 0 : i32, use_next_bd = 0 : i32, valid_bd = 1 : i32}
       aiex.npu.address_patch {addr = 0x1d004 : ui32, arg_idx = 2 : i32, arg_plus = 0 : i32}
-      aiex.npu.maskwrite32 {address = 0x1d200 : ui32, column = 0 : i32, row = 0 : i32, mask = 0x00000F00 : ui32, value = 0x200 : ui32}
+      aiex.npu.maskwrite32 {address = 0x1d200 : ui32, column = 0 : i32, row = 0 : i32, mask = 0x00000F00 : ui32, value = 0x100 : ui32}
```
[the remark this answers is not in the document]
erwei-xilinx: Yes the pass `AIEAssignTileCtrlIDsPass` assigns `controller_id` attributes to tiles in the design. If user written code, then the user could look at the IR after `AIEAssignTileCtrlIDsPass`, and write the `aiex.npu.maskwrite32` op by looking up the `controller_id`s from the tile ops. Alternatively, the user could also manually assign `controller_id`s to the tile ops, and then manually writhe the `aiex.npu.maskwrite32` ops based on the controller ids.

If compiler generated, the `value` field of this `aiex.npu.maskwrite32` op is expected to have been generated by looking up the `controller_id` on the tile op.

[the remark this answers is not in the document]
fifield: "Run the compiler passes and interpret the intermediate results" is not a great answer but if the user's choice of `controller_id` is respected then that's ok for now. Please update the test to set `controller_id` explicitly so that it does not depend on the implementation details of the pass.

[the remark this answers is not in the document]
erwei-xilinx: > "Run the compiler passes and interpret the intermediate results" is not a great answer but if the user's choice of `controller_id` is respected then that's ok for now. Please update the test to set `controller_id` explicitly so that it does not depend on the implementation details of the pass.

Board tests updated to show user manually assigning controller ids.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    int designUnusedPacketIdFrom = getUnusedPacketIdFrom(device);
+    for (int col : occupiedCols) {
+      AIE::TileOp shimTile = nullptr;
+      builder.setInsertionPointToStart(device.getBody());
+      if (tiles.count({col, 0}))
+        shimTile = tiles[{col, 0}];
+      else
+        shimTile = builder.create<AIE::TileOp>(builder.getUnknownLoc(), col, 0);
```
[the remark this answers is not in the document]
erwei-xilinx: Really great to have that utility method. Code updated to use it.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
erwei-xilinx: Sure, comment added.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
erwei-xilinx: Great idea. Code optimized as suggested.

test/npu-xrt/add_one_ctrl_packet/test.cpp
```diff
@@ -145,7 +145,7 @@ int main(int argc, const char *argv[]) {
 
   uint32_t beats = 1 - 1;
   uint32_t operation = 0;
-  uint32_t stream_id = 0;
+  uint32_t stream_id = 5;
```
fifield, line 148: I didn't think this had any effect for writes without return packets, i.e. the first two packets in this test. Did you find otherwise?

erwei-xilinx: Yeah with this PR the `stream_id` must match the controller id of tile(0, 2), because the pass `aie-generate-column-control-overlay` also generates the packet-flow routing between tile(0, 2) CTRL and tile(0, 0) SOUTH0 for TCT routing, with packet header being controller id. 

I got deadlock if I do not change the stream_id to match with controller id of tile(0, 2).

erwei-xilinx: Maybe having the `aie-generate-column-control-overlay` generating TCT routings for all tiles to shim SOUTH0 is dangerous... Let me see if I limit this pass to route TCT from SHIM CTRL only...

erwei-xilinx: > I'm referring to the shim -> tile control packets, which are the ones getting stream_id = 5 here. For these the packet header id is set by the shim bd in the block write. For the return packets of the control port read, the controller id gets set to 2 in the code below (line 181), and that is the id which must match the routing configuration.

I see. You are right. After I limit the pass to generate TCT routings for shim tiles only, then stream_id again resumed to having no effect.

PR #1712, opened 2024-08-21
revert routing coefficient

Yu-Zhewen:
fixed test hanging for `mlir-air/airhost/40_air_8x4_2d_square` caused by last router refactoring @erwei-xilinx


erwei-xilinx (approved):

fifield:
> fixed test hanging for `mlir-air/airhost/40_air_8x4_2d_square` caused by last router refactoring @erwei-xilinx

Thanks for the quick fix. It seems like the only change is to a constant in the code, which is pretty obscure. Is there some simple test we can write to prevent the hang from returning during future refactoring?

Yu-Zhewen:
> > fixed test hanging for `mlir-air/airhost/40_air_8x4_2d_square` caused by last router refactoring @erwei-xilinx
> 
> Thanks for the quick fix. It seems like the only change is to a constant in the code, which is pretty obscure. Is there some simple test we can write to prevent the hang from returning during future refactoring?

Those coefficients will have zero impact if the routing problem is simple, so we probably should  write more stress tests to protect "routability." (however, it might be challenging to design these tests from scratch without real examples). Of course, we can definitely add this failed air test into the mlir-aie unit tests.

PR #1724, opened 2024-08-26
fix build error when NDEBUG is defined

fifield:
This fixes an error experienced when following https://xilinx.github.io/mlir-aie/Building.html. In particular `clone-llvm-local.sh` builds LLVM with asserts turned off, something the mlir-aie CI does not completely do, resulting in a build error which is not caught by the mlir-aie checks.

build error:
```
/home/runner/work/mlir-aie/mlir-aie/lib/Targets/AIETargetCDODirect.cpp:141:21: error: ‘llvm::raw_ostream& operator<<(llvm::raw_ostream&, const XAie_Packet&)’ defined but not used [-Werror=unused-function]
  141 | static raw_ostream &operator<<(raw_ostream &os, const XAie_Packet &packet) {
      |                     ^~~~~~~~
```
This happens when some debug statements inside of `LLVM_DEBUG` macros are optimized away, leaving some `operator<<` functions unused. This only happens if `NDEBUG` is defined.

### DIFF
--- a/lib/Targets/AIETargetCDODirect.cpp
+++ b/lib/Targets/AIETargetCDODirect.cpp
@@ -109,6 +109,8 @@ static const std::map<WireBundle, StrmSwPortType>
         {WireBundle::Trace, StrmSwPortType::TRACE},
 };
 
+#ifndef NDEBUG
+
 // https://stackoverflow.com/a/32230306
 template <typename H1>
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):
Works for me, thank you @fifield!

PR #1728, opened 2024-08-27
Configure AIE using control packets

erwei-xilinx:
- Cherry-picked control packet op implementation in https://github.com/Xilinx/mlir-aie/pull/1709
- Added some fixups to ensure that `txn2mlir.py` can now generate control packet ops that work.
- Add a board test showing memtile, core tile and core program memory reconfigured using control packets streamed in from DDR.


python/compiler/txn2mlir.py
```diff
@@ -137,49 +139,107 @@ def device_body():
...
-                            )
-                            npu_blockwrite(addr, d)
-                        elif op[0] == 0x03:
-                            addr = op[1]
-                            value = op[2]
-                            mask = op[3]
-                            npu_maskwrite32(addr, value, mask)
-                        else:
-                            raise Exception(f"Unhandled op: {op:#x}")
+                if mlir_ctrl_pkt:
```
fifield, line 142: This is a nice quick way to get a proof of concept but I think npu op to control packet conversion should be an mlir pass in C++ if it is going to be supported in mlir-aie.  Converting runtime sequence mlir to a binary blob then to control packet mlir is a bit hacky.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -791,6 +791,27 @@ def AIE_NpuAddressPatchOp: AIEX_Op<"npu.address_patch", []> {
   }];
 }
 
+def AIE_NpuControlPacketOp: AIEX_Op<"npu.control_packet", []> {
```
fifield, line 794: `npu` prefix is not part of the original PR. This is not an npu op. Control packets are part of AIE, so should be `aie.control_packet` (for now in `aiex`)

fifield (changes requested):
This should wait for https://github.com/Xilinx/mlir-aie/pull/1709 to land. That one is still a draft because it lacks proper tests.

test/npu-xrt/ctrl_packet_reconfig/ctrl_pkts.txt
fifield, line 1: Ideally this file should be generated as part of the test

python/compiler/txn2mlir.py
erwei-xilinx: Totally agreed. I'm thinking that perhaps ideally there should be an mlir pass that converts write32/blockwrite/maskedwrite32 ops into control packet ops, and another pass that converts control packet ops to dma_memcpy_nd ops represeting shim dma issuing control packet writes to tiles. Let me create an issue to extend this discussion.

erwei-xilinx:
> This should wait for #1709 to land. That one is still a draft because it lacks proper tests.

Agreed. Will rebase after that one lands on main with more tests.

test/npu-xrt/ctrl_packet_reconfig/ctrl_pkts.txt
erwei-xilinx: Agreed. Locally this file is generated by mlir passes. Let me create a separate PR for those passes that I'm working on locally.

python/compiler/txn2mlir.py
fifield: Sounds good. This txn2mlir.py script is mostly intended to be a disassembler and debugging aid. It is not intended to be load bearing.

erwei-xilinx: Issue created: https://github.com/Xilinx/mlir-aie/issues/1729

include/aie/Dialect/AIEX/IR/AIEX.td
erwei-xilinx: Great point. Npu prefix is removed. Makes sense that this feature should eventually be made available for AIE1 too.

fifield (approved):

python/compiler/txn2mlir.py
fifield: related to https://github.com/Xilinx/mlir-aie/pull/1740

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -641,4 +641,24 @@ LogicalResult AIEX::DMAStartBdChainOp::verify() {
...
+// NpuControlPacketOp
+//===----------------------------------------------------------------------===//
+
+uint32_t AIEX::NpuControlPacketOp::getRow() {
+  const auto &targetModel = AIE::getTargetModel(*this);
+  uint32_t addr = getAddress();
+  uint32_t rowInt =
+      (addr & (0xff << targetModel.getRowShift())) >> targetModel.getRowShift();
+  return rowInt;
+}
```
fifield, line 656: This is called `getRow`, and there's a `getColumn` below, but they aren't actually getting row and column, they are doing some adddress calculations. And they are unused in this PR? Can we remove from this PR and add them when there's a use?

erwei-xilinx: Sure methods removed from this PR and will be added to a separate PR that follows.

PR #1731, opened 2024-08-27
Remove unrequired REQUIRES

fifield:
Remove `REQUIRES: ryzen_aie` from a bunch of tests that only run `aie-opt` and do not need npu hardware.


jgmelber (approved):

andrej (approved):
Makes sense. Must have blindly copied that in from somewhere. Thanks

PR #1733, opened 2024-08-28
Update trace wrappers and add trace test to section-4b

jackl-xilinx:


python/utils/trace.py
```diff
@@ -265,6 +268,32 @@ def configure_simple_tracing_aie2(
     for addr, value in all_reg_writes.items():
         npu_write32(column=int(tile.col), row=int(tile.row), address=addr, value=value)
 
+
+def configure_broadcast_core_aie2(tile, num, event):
```
fifield, line 272: This appears to be an unused function?

python/utils/trace.py
```diff
@@ -200,11 +190,24 @@ def configure_simple_tracing_aie2(
...
+    # Reorder events so they match the event order for display
+    ordered_events = []
+    ordered_events.append(events[3])
+    ordered_events.append(events[2])
+    ordered_events.append(events[1])
+    ordered_events.append(events[0])
+    ordered_events.append(events[7])
+    ordered_events.append(events[6])
+    ordered_events.append(events[5])
+    ordered_events.append(events[4])
```
fifield, line 202:
```suggestion
    ordered_events = [events[p] for p in [3, 2, 1, 0, 7, 6, 5, 4]]
```

programming_guide/section-4/section-4b/Makefile
```diff
@@ -59,7 +59,7 @@ trace: ${targetname}.exe build/trace.xclbin
 	${powershell} ./$< -x build/trace.xclbin -i build/insts.txt -k MLIR_AIE -t ${trace_size}
 	../../../programming_examples/utils/parse_trace.py --filename trace.txt --mlir build/aie_trace.mlir --colshift 1 > trace_4b.json
 
-trace_py: build/final.xclbin
+trace_py: build/trace.xclbin
```
fifield, line 62: Looks like I missed this in #1651 :(

jackl-xilinx: On a side note, I'd like to add more trace testing to make sure changes in the trace.py and parse_trace.py don't break things. I changed the order of events so that order listed in the events array now match the order on the output (instead of matching the config register order which is reversed).

python/utils/trace.py
jackl-xilinx: Well, these are function wrappers. So I use them to wrap other more complex trace configs.

PR #1735, opened 2024-08-28
Add 24.04 runner

fifield:
Trying out ubuntu 24.04 runner


fifield:
Seems to work out of the box:
https://github.com/Xilinx/mlir-aie/actions/runs/10604845842/job/29392381759?pr=1735

keryell (approved):
Thanks!
That seems consistent with the fact that I use Ubuntu 24.04 on my laptop with last may.

PR #1739, opened 2024-08-29
(WiP) More restrictive column control overlay generation

erwei-xilinx:
To ensure control overlay consistency when reconfiguring across multiple designs.
- Control packet flows were routed before other flows, to ensure consistent switchbox master/slave port/arbiter/msel allocation across designs.
- Control packet flows were sorted to ensure routing determinism.
- New pass `AIECtrlPacketInferTilesPass` which infers tile ops from control packet op's addresses; tile ops are handles used to generate column control overlays.
- New pass `AIECtrlPacketToDmaPass` which converts control packet ops into memcpy ops (and sync ops) representing data movements, sending control packets from DDR via shim to tiles.
- Reformatted test `ctrl_packet_reconfig` to generate control packets using compiler passes, rather than copy-pasting from a text file.


fifield:
The new passes really should be separate from the PR updating the overlay pass.

erwei-xilinx:
> The new passes really should be separate from the PR updating the overlay pass.

Ok let me break this PR up into multiple PRs. Will change the PR to draft.

erwei-xilinx:
Closes as all commits were merged into main by other PRs.

PR #1740, opened 2024-08-29
various updates to aie-translate txn backend

fifield:
* port txn2mlir.py binary to mlir conversion into AIETargetCDODirect.cpp, add it back into python via pybind
* aie-translate -aie-generate-txn respects -o option instead of dumping to workdir
* aie-translate -aie-generate-txn respects -aie-output-binary to select between mlir (default) and binary output
* aiecc --generate-txn now generates an mlir file containing a runtime sequence of npu txn ops instead of a binary blob. hardcoded to workdir/txn.mlir for now
* update test/npu-xrt/add_one_two_txn

### DIFF
--- a/python/compiler/txn2mlir.py
+++ b/python/compiler/txn2mlir.py
@@ -65,7 +65,7 @@ def parse_txn(data, verbose=False):
                 _, addr, size = struct.unpack("III", data[i + 4 : i + 16])
                 print_log(f"addr: {addr:#x}")
                 print_log(f"size: {size}")
-                operations.append((opc, addr, data[i + 16 : i + size]))
+                operations.append((opc, addr, data[i + 16 : i + size - 16]))
                 i = i + size
             elif opc == 0x03:
                 print_log("opcode: MASKWRITE (0x03)")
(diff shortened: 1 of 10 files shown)


erwei-xilinx (approved):

python/compiler/txn2mlir.py
```diff
@@ -65,7 +65,7 @@ def parse_txn(data, verbose=False):
                 _, addr, size = struct.unpack("III", data[i + 4 : i + 16])
                 print_log(f"addr: {addr:#x}")
                 print_log(f"size: {size}")
-                operations.append((opc, addr, data[i + 16 : i + size]))
+                operations.append((opc, addr, data[i + 16 : i + size - 16]))
```
erwei-xilinx, line 68: What's the reason for dropping the last 16 data? If we do that, the control packets being generated will come with missing data, and fail (I tested locally with this e2e [test](https://github.com/Xilinx/mlir-aie/pull/1739/files#diff-02a23a98f065c1c6af63f615f8bb801d435144da9d14b0f2ea75fb79255ca298)).

erwei-xilinx: I reverted this in this commit: https://github.com/Xilinx/mlir-aie/pull/1739/commits/5310c72164e145c2eeff83a092d4b2f91e901f1c, in order to try to get the PR to work.

fifield: hmm, that change was unintentional. Probably a merge mistake by me since I see you already fixed it once.  You are correct though, the end of the data should be `i + 16 + size - 16 = i + size`

PR #1742, opened 2024-09-03
update upload/download-artifact action

fifield:
Fix deprecated Node.js version warnings. Also, according to README, actions/download-artifact@v3 is scheduled for deprecation on November 30, 2024.


fifield:
Currently failing because of this update to upload action:
> Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once. Otherwise you will encounter an error.

erwei-xilinx:
Landed in https://github.com/Xilinx/mlir-aie/pull/2023

PR #1744, opened 2024-09-04
Enable objectfifo memtile repeat examples

hunhoffe:
The objectfifo memtile repeat programming examples were not functional due to minor bugs. This PR fixes both examples and enables their lit tests, as well as adds a bit of documentation for the distribute repeat example.


jgmelber (approved):
LGTM. As discussed, I think the repeat count should represent the absolute count the task is repeated rather than "additional count", but this should be a separate PR.

PR #1752, opened 2024-09-05
[Ctrl Pkt Reconfig E2E] Enforcing determinism in control network routing and packet arbitration

erwei-xilinx:
To ensure control overlay consistency when reconfiguring across multiple designs.
- Control packet flows were routed before other flows, to ensure consistent switchbox master/slave port/arbiter/msel allocation across designs.
- Control packet flows were sorted to ensure routing determinism.


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -653,7 +653,8 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
 
   let arguments = (
     ins AIEI8Attr:$ID,
-        OptionalAttr<BoolAttr>:$keep_pkt_header
+        OptionalAttr<BoolAttr>:$keep_pkt_header,
+        OptionalAttr<BoolAttr>:$ctrl_pkt_flow
```
fifield, line 657: What does `ctrl_pkt_flow=true` do? Is it documented?  Does it make it "route before other flows"?

test/npu-xrt/ctrl_packet_reconfig/run.lit
```diff
@@ -8,3 +8,4 @@
 // RUN: clang %S/test.cpp -o test.exe -std=c++11 -Wall %xrt_flags -lrt -lstdc++ -lboost_program_options -lboost_filesystem
 // RUN: %run_on_npu ./test.exe -x aie.xclbin -k MLIR_AIE -i insts.txt | FileCheck %s
 // CHECK: PASS!
+// XFAIL: *
```
fifield, line 11: I guess this is because you are stacking PRs? (my suggestion...)

include/aie/Dialect/AIE/IR/AIEOps.td
erwei-xilinx: Yes exactly, this flag labels packet flow ops which get routed before others, so that the goal is they always get routed to use the same master/slave/arbiter/msel. When control packets reconfigure the control packet flow, things tend to fall apart leading to deadlock.

Let me add documentation.

test/npu-xrt/ctrl_packet_reconfig/run.lit
erwei-xilinx: This test was the control packet reconfiguration prototype which loads control packets from a txt file that I generated offline. The goal is to have compiler passes generate the txt, but the passes are in this PR https://github.com/Xilinx/mlir-aie/pull/1753. Once we land both PRs then I will land a separate PR to rewrite this test.

include/aie/Dialect/AIE/IR/AIEOps.td
erwei-xilinx: Doc. added.

erwei-xilinx: Make sense. For now the ctrl packet flows were only special in that they get routed first. I renamed that attr to "priority_route".

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -426,16 +445,87 @@ void AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder) {
...
+    case WireBundle::Ctrl:
+      return 10;
+    }
+    return -1;
+  };
+  auto getUniqueIdPerFlowPerSB =
+      [getWireBundleAsInt](int flowID, WireBundle srcBundle,
+                           SmallVector<PhysPort, 4> dests) {
+        int currMultiplier =
+            16; // TODO: Will fail when WireBundle enum eventually exceeds 16
```
fifield, line 481: Try `AIE::getMaxEnumValForWireBundle()` ?

erwei-xilinx: Ah really nice to have that method. PR updated.

Issue #926, opened 2024-01-11
Fix github.io link and generation of aie dialect (https://xilinx.github.io/mlir-aie/AIEDialect.html)


jackl-xilinx:
While https://xilinx.github.io/mlir-aie/AIEDialect.html is invalid, https://xilinx.github.io/mlir-aie/AIE.html is valid. I think maybe that's because the AIEDialect.td changed to AIE.td or something like that but the places that reference it (like the [sidebar](https://github.com/Xilinx/mlir-aie/blob/e10cb3043807a9858ff75289627d1bb97f2855c0/docs/_layouts/default.html#L30) and [README](https://github.com/Xilinx/mlir-aie/blob/e10cb3043807a9858ff75289627d1bb97f2855c0/docs/README.md)) didn't change along with it. If you look at the new location, you'll see it's mostly empty and I think it has to do with the how AIE.td was changed where included *.td files were removed around [this commit](https://github.com/Xilinx/mlir-aie/commit/2270ac4fd3fbd4dd8259d6c093b32d118e3a4fa2). You can see here that the commit removed the includes for things like AIEInterfaces.td and replaced it with OpBase.td which I don't see the file for. I think the included td files add the doc definitions.

makslevental:
Docs are fixed https://xilinx.github.io/mlir-aie/AIE.html

jackl-xilinx:
Awesome! Thanks @makslevental

Issue #948, opened 2024-01-22
Documentation has been removed


gabrielrodcanal:
Hi,
The documentation for the dialects was removed a while ago. I'd like to ask if this was intentional and if not can it be restored? If you check https://github.com/Xilinx/mlir-aie/tree/main/docs you'll see that all the links to "Generated code documentation" are broken.

abisca:
Related to #926.

makslevental:
Have no clue how those links ever worked but all of those markdown files are on the `gh-pages` branch not the `main` branch: https://github.com/Xilinx/mlir-aie/tree/gh-pages.

makslevental:
Docs are fixed https://xilinx.github.io/mlir-aie/AIE.html

PR #1754, opened 2024-09-05
[NFC] Another attempt to fix naming of AIE dialect documentation

keryell:
Keep the "Dialect" in "AIEDialect" since it is the convention used with all the other dialects of the project.


keryell:
This is yet-another-attempt to fix https://github.com/Xilinx/mlir-aie/issues/926 and https://github.com/Xilinx/mlir-aie/issues/948

jgmelber (approved):

PR #1755, opened 2024-09-06
Disable region simplification in `--aie-substitute-shim-dma-allocations` pass

andrej:
The pattern rewriter in this pass previously also performed region simplification, which caused an assertion error if the MLIR contained blocks with input arguments, such as generated by:

```
          %7 = scf.if %6 -> (memref<64x32xf32>) {
            scf.yield %C_L1L2_0_0_buff_0 : memref<64x32xf32>
          } else {
            scf.yield %C_L1L2_0_0_buff_1 : memref<64x32xf32>
          }
```

This fixes that issue. I don't fully understand why the issue exists with region simplification, but the point of this pass is not to simplify regions anyways, so it is fine to disable.

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIESubstituteShimDMAAllocations.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIESubstituteShimDMAAllocations.cpp
@@ -67,13 +67,12 @@ struct AIESubstituteShimDMAAllocationsPass
 
     // Convert DMAConfigureTaskForOps that reference shim DMA allocations
     // to regular DMAConfigureTaskOps
-    ConversionTarget target(getContext());
-    target.addLegalDialect<AIEXDialect>();
-    target.addIllegalOp<DMAConfigureTaskForOp>();
+    GreedyRewriteConfig rewriter_config = GreedyRewriteConfig();
+    rewriter_config.enableRegionSimplification =
+        GreedySimplifyRegionLevel::Disabled;
     RewritePatternSet patterns(&getContext());
     patterns.insert<DMAConfigureTaskForOpPattern>(&getContext());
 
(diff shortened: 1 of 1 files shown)


fifield:
Seems like red-flag that the IR is left in a bad state somewhere. Do you have a failing (fixed) test you can add to the PR?

andrej:
Oops, forgot about this one. Yes I believe this is still needed. I can work on a test.

andrej:
@fifield I finally got around to creating a fairly minimal test. I can confirm without this PR, the added test still fails on current MLIR-AIE. Sorry it took so long.

Merging this should allow #1772 to pass.

fifield:
> @fifield I finally got around to creating a fairly minimal test. I can confirm without this PR, the added test still fails on current MLIR-AIE. Sorry it took so long.
> 
> Merging this should allow #1772 to pass.

Thanks, out of curiosity I'd still like to take a look at why it was crashing without the modification.

fifield:
> > @fifield I finally got around to creating a fairly minimal test. I can confirm without this PR, the added test still fails on current MLIR-AIE. Sorry it took so long.
> > Merging this should allow #1772 to pass.
> 
> Thanks, out of curiosity I'd still like to take a look at why it was crashing without the modification.

The error has little to do with the pass changed in this CR. It has to do with a missing terminator at the end of the `aie.device`. Still investigating, but I'll note that the `aie.device` verifier implementation is currently `return success();`

andrej:
Please let me know if/when you find the root cause and if you'd still be OK with merging this PR even if it doesn't directly address the root cause.

andrej:
Should we still merge this after #1936? While the error doesn't exist any more after #1936, I think it's kind of weird that this pass by default performs region simplification when what it purports to do is something else. Or is that normal in MLIR/LLVM world?

PR #1756, opened 2024-09-06
[Ctrl Pkt Reconfig E2E] New `-aie-generate-ctrlpkt` comilation mode in `aiecc.py`

erwei-xilinx:
- Rewrite control packet reconfiguration end-to-end test to have control packet raw data compiler generated.
- Directly translating to `aiex.control_packet` ops in `aiecc.py`, without going through `txn2mlir.py` prototype pass.
- Rename `aieTranslateToControlPackets` to `aieTranslateControlPacketsToUI32Vec` to avoid confusion between converting *to* control packet to converting *from* control packets.


include/aie/Targets/AIETargets.h
```diff
@@ -61,8 +61,8 @@ AIETranslateToCDODirect(mlir::ModuleOp m, llvm::StringRef workDirPath,
 mlir::LogicalResult
 AIETranslateToTxn(mlir::ModuleOp m, llvm::raw_ostream &output,
                   llvm::StringRef workDirPath, bool outputBinary = false,
-                  bool aieSim = false, bool xaieDebug = false,
-                  bool enableCores = true);
+                  bool outputCtrlpkt = false, bool aieSim = false,
```
fifield, line 64: A new function e.g. `AIETranslateToControlPackets` would be better than overloading the existing function with a third output format that is not "`ToTxn`"

lib/Targets/AIETargetCDODirect.cpp
```diff
@@ -1189,17 +1189,135 @@ xilinx::AIE::AIETranslateBinaryToTxn(mlir::MLIRContext *ctx,
...
   if (failed(result))
     return result;
 
+  if (outputCtrlpkt) {
+    auto new_module = AIETranslateBinaryToCtrlpkt(m.getContext(), bin);
+    if (!new_module)
+      return failure();
+    new_module->print(output);
+    return success();
+  }
```
fifield, line 1319: Should also support `outputBinary=true`

lib/Targets/AIETargets.cpp
```diff
@@ -159,6 +159,12 @@ void registerAIETranslations() {
           "Select binary (true) or text (false) output for supported "
           "translations. e.g. aie-npu-instgen, aie-ctrlpkt-to-bin"));
 
+  // static llvm::cl::opt<bool> outputCtrlpkt(
+  //     "aie-output-ctrlpkts", llvm::cl::init(false),
+  //     llvm::cl::desc(
+  //         "Select control packet output for supported translations. "
+  //         "e.g. aie-npu-instgen, aie-ctrlpkt-to-bin"));
```
fifield, line 166: remove

include/aie/Targets/AIETargets.h
```diff
@@ -37,10 +37,10 @@ mlir::LogicalResult AIETranslateGraphXPE(mlir::ModuleOp module,
...
                                       llvm::raw_ostream &output);
 mlir::LogicalResult AIETranslateToNPU(mlir::ModuleOp, std::vector<uint32_t> &);
-mlir::LogicalResult AIETranslateToControlPackets(mlir::ModuleOp module,
-                                                 llvm::raw_ostream &output);
-mlir::LogicalResult AIETranslateToControlPackets(mlir::ModuleOp,
-                                                 std::vector<uint32_t> &);
+mlir::LogicalResult AIETranslateControlPacketsToText(mlir::ModuleOp module,
+                                                     llvm::raw_ostream &output);
+mlir::LogicalResult AIETranslateControlPacketsToText(mlir::ModuleOp,
+                                                     std::vector<uint32_t> &);
```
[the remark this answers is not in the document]
erwei-xilinx: Make sense. PR updated to rename to `...ToUI32Vec`. Also updated the PR description to highlight this name change.

include/aie/Targets/AIETargets.h
erwei-xilinx: Agreed. A new function is created to avoid overloading the txn workflow. It still calls the same `translateToTxn` though, as internally we still translate the device op to transaction binary, before it gets further translated into control packet ops.

lib/Targets/AIETargets.cpp
erwei-xilinx: Done.

fifield (approved):

PR #1765, opened 2024-09-11
Add aie.device to transaction ops mlir pass

fifield:
This moves the aie.device to transaction ops translation out of aie-translate and into a MLIR pass, since it is a mlir to mlir conversion. This will be followed by PRs to similarly move the aie.device to control packet conversion to a pass and to remove redundancies.

* Add pass `convert-aie-to-transaction` which runs on `aie.device` ops. The pass inserts into an `aie.device` a new `aiex.runtime_sequence` containing the transaction op sequence corresponding to the configuration (transaction binary) for that `aie.device`.
* Add a new folder `aie/Conversion/AIEToConfiguration` to house aie dialect to configuration dialect(s) conversion.
* Relax the restriction that `aie.device` can only contain one `aiex.runtime_sequence`. Instead it is allowed to contain multiple, e.g. one transaction sequence for configuration and one for control.
* Add `-aie-sequence-name=<sym_name>` options to `aie-translate` to say which runtime sequence to translate, if there are multiple.

### DIFF
--- a/include/aie/Conversion/AIEToConfiguration/AIEToConfiguration.h
+++ b/include/aie/Conversion/AIEToConfiguration/AIEToConfiguration.h
@@ -1,4 +1,4 @@
-//===- AIEToTransaction.h ---------------------------------------*- C++ -*-===//
+//===- AIEToConfiguration.h -------------------------------------*- C++ -*-===//
 //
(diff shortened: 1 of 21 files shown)


jsetoain (approved):
👍

PR #1767, opened 2024-09-12
Offset calculation should be purely based on offsets and strides

erwei-xilinx:
… not memref shape.

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -306,19 +306,19 @@ int64_t AIEX::NpuDmaMemcpyNdOp::getOffsetInBytes() {
       llvm::map_to_vector(llvm::reverse(getMixedOffsets()), [](OpFoldResult s) {
         return getConstantIntValue(s).value();
       });
-  size_t stride = 1;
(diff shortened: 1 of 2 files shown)


fifield:
Do you have a test to show the incorrect/fixed behavior?

andrej (approved):
I agree that this makes more sense. I think a lot of examples (all of them, since you did not have to update any?) just use the lowest-dimension offset so did not have to worry about this inconsistency. However maybe @jgmelber has thoughts as well?

erwei-xilinx:
> Do you have a test to show the incorrect/fixed behavior?

Yeah I have an example driven from AIR. Let me make a test to show the fixed behavior.

erwei-xilinx:
> I agree that this makes more sense. I think a lot of examples (all of them, since you did not have to update any?) just use the lowest-dimension offset so did not have to worry about this inconsistency. However maybe @jgmelber has thoughts as well?

Yeah I think you are right. Existing examples in both AIR and AIE never really triggered the highest wrap-and-stride offset dimensions, until when I recently was debugging a batch matmul from AIR.

erwei-xilinx:
> Do you have a test to show the incorrect/fixed behavior?

Test added.

PR #1768, opened 2024-09-12
Move control packet translation from aie-translate to aie-opt

fifield:
The same as #1765 and #1766, but for control packet translation. Depends on the latter. This completes the move of code for generating transaction or packet based configuration out of the cdo (xclbin) backend and into mlir passes.


test/npu-xrt/ctrl_packet_reconfig/run.lit
```diff
@@ -3,12 +3,17 @@
...
+// RUN: %python aiecc.py --no-aiesim --aie-generate-cdo --no-compile-host --xclbin-name=aie1.xclbin aie1_overlay.mlir
+//
+// RUN: aie-opt -aie-generate-column-control-overlay="route-shim-to-tile-ctrl=true" %S/aie2.mlir -o aie2_overlay.mlir
+// RUN: %python aiecc.py --no-aiesim --aie-generate-ctrlpkt --aie-generate-npu --no-compile-host --npu-insts-name=aie2_run_seq.txt aie2_overlay.mlir
+//
+// RUN: aie-translate -aie-ctrlpkt-to-bin -aie-sequence-name=configure aie2_overlay.mlir.prj/ctrlpkt.mlir -o ctrlpkt.txt
+//
+// RUN: aie-opt -aie-ctrl-packet-to-dma -aie-dma-to-npu aie2_overlay.mlir.prj/ctrlpkt.mlir -o ctrlpkt_dma_seq.mlir
+// RUN: aie-translate -aie-npu-instgen -aie-sequence-name=configure ctrlpkt_dma_seq.mlir -o ctrlpkt_dma_seq.txt
+//
```
fifield, line 16: This refactors the e2e test in a couple ways:
1) I found it more intuitive and easier to debug for `run.lit` to insert the control overlays explicitly using `aie-opt` vs. leaving it up to the block box of `aiecc`.
2) The `aie-ctrl-packet-infer-tiles` step is no longer needed, because conversion to control packets no longer discards the information being inferred.

cc: @erwei-xilinx

erwei-xilinx: Thanks for the refactor! (1) Yeah totally. The original impl. was in aiecc only so that it makes the ctrlpkt workflow similar to the txn workflow in our code base, so that it's easier to maintain in the short term. Separating it into aie-opt makes better sense. (2) Really nice that we can now remove that experimental aie-ctrl-packet-infer-tiles pass.

Issue #1751, opened 2024-09-05
std::runtime_error -- what():  qds_device::wait() unexpected command state


hecmay:
I was running cascade matrix multiple example from programming examples: https://github.com/Xilinx/mlir-aie/tree/main/programming_examples/basic/matrix_multiplication/cascade

I noticed that the code only works when `M` or `m` is greater than 64. When I set them to smaller values, say `M = 32, n = 32`, the host code will throw the following error:

```bash
Running Kernel (iteration 0).
terminate called after throwing an instance of 'std::runtime_error'
  what():  qds_device::wait() unexpected command state
Aborted (core dumped)
make: *** [/home/ubuntu/mlir-aie-test/programming_examples/basic/matrix_multiplication/cascade/../makefile-common:114: run] Error 134
```

No error from XCLBIN compilation process, and it seems to be something wrong with runtime? Any idea how this can be fixed?

hecmay:
Running into exactly the same error in matrix vector sample code if the `n_cores` is > 1: https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/basic/matrix_multiplication/matrix_vector/aie2.py#L21

hecmay:
I am not sure if this error is caused by my environment setup. I am using [Phoenix Point Mini PC: Minisforum UM790 Pro : AMD Ryzen™ 9 7940HS](https://store.minisforum.com/products/minisforum-um790-pro?variant=43865372786933). I followed every single step in this README: https://github.com/Xilinx/mlir-aie/blob/main/docs/buildHostLin.md

```bash
Linux kernel: 6.10
Vitis 2023.2
AMDXDNA: 2.18.0_20240825, 537a509a3ab1b698c9c9f6ebcd88035b2fe8359b
```

Can anyone reproduce the issue? Any help would be highly appreciated. Thanks! @stephenneuendorffer @fifield @hunhoffe @Yu-Zhewen @makslevental

PisonJay:
Me too. Getting same error with Ryzen AI 9 365.

PisonJay:
I guess that current infrastructure only supports XDNA1 (AIE2) architecture. Strix Point, XDNA2 (with code name AIEP) is not supported yet. So does [Peano], listing XNDA2 as "coming soon". The only usable runtime is ONNX runtime from [Ryzen AI SDK](https://ryzenai.docs.amd.com/en/latest/inst.html), only available on Windows at current time.

hecmay:
[...]

PR #1770, opened 2024-09-12
Fix offset bug in matrix multiply cascade

hunhoffe:
I had introduced a bug in both matrix multiply whole array and matrix multiply cascade. I fixed it for matrix multiply whole array but did not realize the bug affected cascade as well until recently.

### DIFF
--- a/programming_examples/basic/matrix_multiplication/cascade/aie2.py
+++ b/programming_examples/basic/matrix_multiplication/cascade/aie2.py
@@ -230,6 +230,7 @@ def my_matmul(M, K, N, m, k, n, n_aie_cols, dtype_in_str, dtype_out_str):
                         (t, 1),
                     ],
                 )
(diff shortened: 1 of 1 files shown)


hunhoffe:
I believe the change I committed fixes the bug, but am running the whole matrix multiply `sweep.sh` on `cascade` locally to be more sure before I set this PR as ready for review. This may be related to https://github.com/Xilinx/mlir-aie/issues/1751 but I did not verify (although I do know the bug can present with that type of error). This bug would not have affected `matrix_vector`, as `matrix_vector` does not contain any `object_fifo_link`s with offsets, so at most it partially addresses the issue.

hunhoffe:
I tried config:
```
M?=32
K?=512
N?=512
m?=32
k?=64
n?=32
```
and it passed. This may address (one of) the issues in the PR (namely, `M = 32, n = 32`).

PR #1771, opened 2024-09-13
Add end-to-end test of new BD chain syntax

andrej:
#1656 introduced a new syntax for configuring BDs from the runtime sequence by reusing the same syntax as we use for static configuration elsewhere in the array. It introduced a number of compiler pass tests, ensuring the correct lowering is produced.

This PR serves two purposes:
 - Give an illustrative example of how to use this syntax in the case of a matrix multiplication by re-implementing the shim data movement of single_core/aie2.py using the alternative syntax in single_core/aie2_alt.py.
 - Provide an end-to-end test for the CI to ensure this syntax continues to be supported and works properly.


jgmelber (commented):
Add documentation to the README and a few more comments but otherwise looks good and thanks for adding this!

andrej:
Added two paragraphs in README.md. The `aie2_alt.py` already had some comments outlining differences to `aie2.py`. Let me know if there's anything else you think should be documented that currently isn't.

jgmelber (approved):

PR #1772, opened 2024-09-13
add example with run-time-parametrized matmul

andrej:
Reconfigures different matrix sizes by swapping out insts.txts and using a run-time parameter for K. At the BD level, since objectFifo currently does not support dynamic loop bounds.


jgmelber:
#2160 helps here too

jgmelber:
@andrej keep open or close?

andrej:
@jgmelber I'll close. If we want to revisit this, it'll be easier since the ObjectFIFO now supports dynamic loop bounds with the `aiecc.py --dynamic-objFifos` flag.

PR #1775, opened 2024-09-18
refactor chess interface

stephenneuendorffer:
To properly support other architectures, we need to call xchesscc with --aiearch as a parameter.  This makes it more awkward to used the chess-clang hack.  It was always a hack anyway, so remove it and instead require users to pass "+Wclang,xir" when feeding LLVMIR to chess.

Lastly, clean up a small issue where we really should be using chess-llvm-link to link together downgraded LLVMIR for chess, rather than mixing and matching our LLVM tools with chess tools.


python/compiler/aiecc/main.py
```diff
@@ -445,7 +445,10 @@ async def chesshack(self, task, llvmir, aie_target):
         await self.do_call(
             task,
             [
-                "chess-llvm-link",
+                # The path below is cheating a bit since it refers directly to the AIE1
+                # version of llvm-link, rather than calling the architecture-specific
+                # tool version.
+                opts.aietools_path + "/tps/lnx64/target/bin/LNa64bin/chess-llvm-link",
```
fifield, line 451: So it does not matter which version is used?

stephenneuendorffer: The different synopsys versions are currently pretty close to one another, and 'HEAD' is diverging relatively quickly...  This is at some level kicking the can down the road, but it's better than what we have.

python/compiler/aiecc/main.py
```diff
@@ -513,9 +514,9 @@ async def process_core(
                     file_core_llvmir_chesslinked = await self.chesshack(task, file_core_llvmir, aie_target)
                     if self.opts.link and self.opts.xbridge:
                         link_with_obj = await extract_input_files(file_core_bcf)
-                        await self.do_call(task, ["xchesscc_wrapper", aie_target.lower(), "+w", self.prepend_tmp("work"), "-d", "-f", "+P", "4", file_core_llvmir_chesslinked, link_with_obj, "+l", file_core_bcf, "-o", file_core_elf])
+                        await self.do_call(task, ["xchesscc_wrapper", aie_target.lower(), "+w", self.prepend_tmp("work"), "-d", "+Wclang,-xir", "-f", file_core_llvmir_chesslinked, link_with_obj, "+l", file_core_bcf, "-o", file_core_elf])
                     elif self.opts.link:
-                        await self.do_call(task, ["xchesscc_wrapper", aie_target.lower(), "+w", self.prepend_tmp("work"), "-c", "-d", "-f", "+P", "4", file_core_llvmir_chesslinked, "-o", file_core_obj])
+                        await self.do_call(task, ["xchesscc_wrapper", aie_target.lower(), "+w", self.prepend_tmp("work"), "-c", "-d", "+Wclang,-xir", "-f", file_core_llvmir_chesslinked, "-o", file_core_obj])
```
keryell, line 519: While you are here you could add a comment about what these options do without having to dive into Synopsys documentation since it is pretty ad-hoc.
Actually this is repeated 3 times, so perhaps a function?

tools/chess-clang/xchesscc_wrapper
```diff
@@ -23,29 +20,19 @@ fi
...
+LIBDIR=${AIETOOLS}/data/${AIEARCH}/lib
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-export RDI_DATADIR=$AIETOOLS/data
-export UNWRAPPED_XCHESSCC=$AIETOOLS/bin/unwrapped/lnx64.o/xchesscc
-export LD_LIBRARY_PATH=$AIETOOLS/lib/lnx64.o:$AIETOOLS/lnx64/tools/dot/lib:$LD_LIBRARY_PATH
-# Carefully crafted path so that we can inject other scripts into the chess path, namely chess-clang
-export PATH=$DIR:$AIETOOLS/bin/unwrapped/lnx64.o:$AIETOOLS/tps/lnx64/target/bin/LNa64bin
-$UNWRAPPED_XCHESSCC -p me -C Release_LLVM -D__AIENGINE__ $EXTRA_DEFS -Y clang=$DIR/chess-clang -P $LIBDIR -d -f $@
+xchesscc --aiearch ${AIEARCH} -p me -C Release_LLVM -D__AIENGINE__ $EXTRA_DEFS -P $LIBDIR -d -f $@
```
keryell, line 38: Adding some comments for simple mortals wanting to understand?

tools/chess-clang/CMakeLists.txt
```diff
@@ -10,24 +10,4 @@ if(NOT DEFINED CMAKE_CROSSCOMPILE)
...
-                    COMMAND ${CMAKE_COMMAND} -E copy ${CMAKE_CURRENT_SOURCE_DIR}/${file}
-                    ${PROJECT_BINARY_DIR}/bin/${file}
-                    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/${file})
-  endforeach()
-
-  # install too!
-  foreach(file ${FILES})
-    install(PROGRAMS ${file} DESTINATION ${SCRIPT_INSTALL_PATH})
-  endforeach()
 endif()
```
keryell, line 13: Missing EOL.
Curious, this project has an `.editorconfig` stating
```
insert_final_newline = true
```

tools/chess-clang/xchesscc_wrapper
```diff
@@ -23,29 +20,19 @@ fi
...
-  then
-    echo "xchesscc_wrapper needs a valid Vitis installation"
-    exit 1
-fi
-
-TARGETVAR=TARGET_${TARGET}_LIBDIR
-LIBDIR=${!TARGETVAR}
+LIBDIR=${AIETOOLS}/data/${AIEARCH}/lib
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
keryell, line 37: Useless line now?

keryell (commented):
Thanks for the generalization.

tools/chess-clang/xchesscc_wrapper
jgmelber: Unrelated to @keryell's comment. The `--aiearch` string is different than the string needed above for the `LIBDIR` target directory. I am taking a look at this now.

PR #1776, opened 2024-09-18
Have AIECtrlPacketToDmaPass create dynamic `MemRefType` for ctrl packet stream

erwei-xilinx:
…instead of hard coded.


fifield:
> …instead of hard coded.

Could it also be `<?xi32>`?  Does the size even matter since the length of transfer is encoded in the memcpy_nd?

erwei-xilinx:
> > …instead of hard coded.
> 
> Could it also be `<?xi32>`? Does the size even matter since the length of transfer is encoded in the memcpy_nd?

Good point. Yeah it should work. PR updated. I also modified the verifier for `AIEX::NpuDmaMemcpyOp` slightly just to enable dynamic shape on its memref, but I don't think the memcpy op fundamentally requires the memref to be static.

PR #1777, opened 2024-09-18
Fixup pathfinder pktflow sorting by sorting on src coords and port only

erwei-xilinx:
Previous sorting method tries to sort on src and dest coords only, but this does not guarantee determinism in control packet routings because:
- port (wirebundle, channel) is not taken into account,
- control packet flows could have fused with data movement packet flows, leading to inconsistent control packet flow routings if all destinations (which could include both CTRL and DMA, NSWE) were taken into account when sorted,
- "Flow" struct in PathFinder assumes uniqueness in source; sorting the source coords, wirebundle and channels is sufficient to enforce determinism in the sorted order.

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
@@ -354,25 +354,13 @@ void Pathfinder::sortFlows(const int maxCol, const int maxRow) {
       normalFlows.push_back(f);
   }
   std::sort(priorityFlows.begin(), priorityFlows.end(),
-            [maxCol, maxRow](const auto &lhs, const auto &rhs) {
+            [maxCol](const auto &lhs, const auto &rhs) {
               int lhsUniqueID = lhs.src.coords.col;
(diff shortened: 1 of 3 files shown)


lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
```diff
@@ -693,3 +686,32 @@ Pathfinder::findPaths(const int maxIterations) {
...
+    return 7;
+  case WireBundle::NOC:
+    return 8;
+  case WireBundle::Trace:
+    return 9;
+  case WireBundle::Ctrl:
+    return 10;
+  }
+  return -1;
+}
```
fifield, line 717: I happened to be playing with this earlier on a branch:
```c++
int AIE::getWireBundleAsInt(WireBundle bundle) {
  return static_cast<typename std::underlying_type<WireBundle>::type>(bundle);
}
```

erwei-xilinx: That's really neat! Added into this PR. Thanks for the suggestion, @fifield.

PR #1781, opened 2024-09-19
Faster (static) method to get unique id for creating global op

erwei-xilinx:
Optimize the existing method of getting unique id for creating `memref.global` ops, as it was running in multiple minutes if the number of NpuWriteBdOps in input IR goes over a thousand.

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
@@ -16,9 +16,6 @@
 #include "mlir/Transforms/DialectConversion.h"
 #include "llvm/ADT/DenseMap.h"
 
-#include <charconv>
-#include <string>
-
 using namespace mlir;
 using namespace xilinx;
 using namespace xilinx::AIEX;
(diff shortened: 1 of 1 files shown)


lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -634,10 +675,10 @@ struct WriteBdToBlockWritePattern : OpConversionPattern<NpuWriteBdOp> {
       std::string name = "blockwrite_data_";
       rewriter.setInsertionPoint(
           op->getParentOfType<AIEX::RuntimeSequenceOp>());
-      int id = 0;
-      while (dev.lookupSymbol(name + std::to_string(id)))
-        id++;
-      name += std::to_string(id);
```
fifield, line 640: Maybe the issue was that it starts at zero each time when it doesn't have to?  If `id` was a static member variable of `WriteBdToBlockWritePattern`, initialized once to zero, it would probably have the same effect as this PR.

erwei-xilinx: Good point. PR updated. Thanks for the suggestion, @fifield!

PR #1786, opened 2024-09-23
Use numpy.ndarray and numpy dtypes instead of memref types in Python API

hunhoffe:
### Summary ###
This PR optionally allows a Python user to use numpy `ndarray` and numpy `dtype` types instead of `memref` and MLIR types in their designs.

Specifically, changes in this PR include:
* Porting all `programming_examples` to use numpy types (except those examples using bfloat16 as numpy doesn't support that type directly - those examples use the existing `memref`/MLIR types without modification)
* Port portions of all python tests to use the numpy tests; only portions were ported because we still support both methods
* A new example, `programming_examples/basic/passthrough_pykernel`, as an example that shows how to create/call a function defined in Python

### Example ###
For example, to define a memref type in the past (such as for use in initializing an object FIFO) as well as external function argument types, you would write something like:
```python
        # define types
        memRef_ty = T.memref(lineWidthInBytes, T.ui8())

        passThroughLine = external_func(
            "passThroughLine", inputs=[memRef_ty, memRef_ty, T.i32()]
        )
```

Now, you can instead write a type definition like so:
```python
        line_ty = np.ndarray[(lineWidthInBytes,), np.dtype[np.int8]]

        passThroughLine = external_func(
            "passThroughLine", inputs=[line_ty, line_ty, np.int32]
        )
```

### Requirements ###
* Python 3.10 or higher for type annotations with bracket notation

Ideally, we would be able to update to numpy 2.1 or higher for shape annotations for `numpy.ndarray` type. However, >=numpy2 doesn't seem to work well with pyxrt, so instead I do some manual checks to make sure the shape is a valid value.

Internally, some of the code would look better if we bumped all the way to python 3.12+ so we could use type annotations with generics with bracket notation, but so far that is not necessary.


keryell (approved):
Nice!
If at some point you want to do the same at the C++ level, the C++23 equivalent of `memref` is `std::mdspan` http://eel.is/c++draft/views#multidim

fifield:
Leaving some notes here regarding numpy 2 if someone else wants to pick it up the investigation. The CI failures with numpy 2 in this PR could be because older pybind11 doesn't have good (any?) numpy 2 support. With that in mind I tried recompiling _everything_ (xrt, llvm/mlir, mlir-aie) with numpy 2.1 and pybind11 3.13 on ubuntu 22.04 and it still didn't work. The data returned by pyxrt buffer object reads seems to be wrong (but the C++ version of the same host code works). I didn't dig too much deeper

PR #1788, opened 2024-09-23
Bump cmake modules

denolf:
use latest cmakeModulesXilinx that now also contain support to only find AIETools.


fifield (approved):

eddierichter-amd (approved):
Works for me on VCK5000.

PR #1791, opened 2024-09-24
Replace `npu_sync` with `dma_wait` in programming examples

hunhoffe:
This PR changes the programming examples and programming guide to use `dma_wait` instead of `npu_sync`. The goal of this change is primarily readability/usability. In about half the examples, I only wait on the output dma. In the other half, I issue tokens from the input dmas and wait on both input and output dmas.

To further simplify the interface of specifying object fifos to `npu_dma_memcpy_nd` and `dma_wait`, I changed `dma_wait` to be a variable argument function so that you can wait on more than one thing (in order given) with just one function call.


jgmelber (approved):
Thanks for doing this refactor!!

PR #1792, opened 2024-09-24
Cleanup xchesscc and find it rather than Vitis

jgmelber:
Find xchesscc in aiecc 
Clean up `xchesscc_wrapper` library path for arch string 
Update docs for find `xchesscc`/`aiecompiler` only

Building on #1775 and #1788

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -1189,9 +1189,7 @@ def run(mlir_module, args=None):
         xchesscc_path = os.path.dirname(xchesscc_bin_path)
         os.environ["AIETOOLS"] = xchesscc_path
         print("Found xchesscc at " + xchesscc_path)
-        os.environ["PATH"] = os.pathsep.join(
-            [os.environ["PATH"], xchesscc_bin_path]
-        )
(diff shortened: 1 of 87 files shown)


python/compiler/aiecc/main.py
```diff
@@ -1174,34 +1178,19 @@ def run(mlir_module, args=None):
...
+        if xchesscc_path:
+            xchesscc_bin_path = os.path.dirname(os.path.realpath(xchesscc_path))
+            xchesscc_path = os.path.dirname(xchesscc_bin_path)
+            os.environ["AIETOOLS"] = xchesscc_path
+            print("Found xchesscc at " + xchesscc_path)
+            os.environ["PATH"] = os.pathsep.join([os.environ["PATH"], xchesscc_bin_path])
+            opts.aietools_path = xchesscc_path
     else:
-        print("Vitis not found...")
+        print("xchesscc not found...")
```
fifield, line 1193:
```
if AIETOOLS not in env:
  set AIETOOLS
else:
  xchesscc not found
```
so it's an error if AIETOOLS is already set?

fifield (approved):

PR #1800, opened 2024-09-26
Call kernels directly in Python, index into MemRefs, other small python changes

hunhoffe:
This PR contains several small steps to clean up Python code in examples a bit to make it seem less MLIR-like and more Python-like.

### 1. Calling External Functions
Previously, we declared an external function and called it like so:
```python
my_func = external_function(...
...
call(my_func, ...)
```

This PR overloads the `__call__` method of `external_function` so you can call an `external_function` directly:
```python
my_func = external_function(...
...
my_func( ...)
```

Also, there were two different implementations of the `call` function, so I deduplicated this function. I also changed the type conversion of integers to use the expected type of the function for that integer value; this means that the function can take in a python `int` and it will be converted to the expected `int16` or `int32` or etc. value instead of defaulting to `int32`.

### 2. Indexing

There was already code in the `python/extras/ext/dialects` to access `MemRef`s similar to python arrays using slice (`arr[n:m]`) and index (`arr[n]`) notation. However, this wasn't used for a few reasons:
* The neat accessing methods don't apply to `ObjectFifoSubviewAccessOp` objects
* The neat accessing methods don't apply to `Buffer` and `ExternalBuffer` objects

This PR changes the convenience methods of `object_fifo.access()` to return one or more MemRefs (hiding the `ObjectFifoSubviewAccessOp` from the Python user). I make `Buffer` and `ExternalBuffer` extensions of `MemRef`s so they too can inherit the `MemRef` methods. There are also methods `buffer` and `external_buffer` that already returned the `MemRef` - I'm wondering if I should revert my changes or get rid of these methods, since it seems weird to have both `buffer`/`Buffer` and `external_buffer`/`ExternalBuffer`.

### 3. Remove calls to `constant` functions where possible

There is a lot of work already done to auto-convert Python values into the correct MLIR type; in places where these functions were still used, I removed them.


programming_examples/basic/vector_vector_add/aie2.py
```diff
@@ -62,10 +61,7 @@ def core_body():
                     elem_in2 = of_in2.acquire(ObjectFifoPort.Consume, 1)
                     elem_out = of_out.acquire(ObjectFifoPort.Produce, 1)
                     for i in range_(n):
-                        v0 = memref.load(elem_in1, [i])
-                        v1 = memref.load(elem_in2, [i])
-                        v2 = arith.addi(v0, v1)
-                        memref.store(v2, elem_out, [i])
+                        elem_out[i] = elem_in1[i] + elem_in2[i]
```
AndraBisca, line 64: This is great!

AndraBisca (approved):
These changes look great to me!

Regarding your question about buffer/Buffer and external_buffer/ExternalBuffer I think we only need these classes because of naming consistency (name starting with capital letter vs all low caps as the rest of the names).

jgmelber (approved):
Great changes. This looks good to me

PR #1801, opened 2024-09-27
Initialize buffer with values

jgmelber:
Enables buffer initialization through CDO lowering 
Adds CDO target tests for initialization 
Adds e2e test of vector vector add where one vector comes from a compile time initialized vector in a memtile

### DIFF
--- a/lib/Targets/AIERT.cpp
+++ b/lib/Targets/AIERT.cpp
@@ -18,6 +18,7 @@ extern "C" {
 #include "xaiengine/xaie_elfloader.h"
 #include "xaiengine/xaie_interrupt.h"
 #include "xaiengine/xaie_locks.h"
+#include "xaiengine/xaie_mem.h"
 #include "xaiengine/xaie_plif.h"
 #include "xaiengine/xaie_ss.h"
 #include "xaiengine/xaie_txn.h"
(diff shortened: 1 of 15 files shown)


test/Targets/AIETargetCDODirect/initbuffer_int8.mlir
fifield, line 1: What do you think about changing the tests into mlir transform tests:
```
$ aie-opt --convert-aie-to-transaction initbuffer_int8.mlir 
module {
  aie.device(npu1_1col) {
    memref.global "private" constant @blockwrite_data : memref<2xi32> = dense<[-50200320, 117835012]>
    aiex.runtime_sequence @configure() {
      %0 = memref.get_global @blockwrite_data : memref<2xi32>
      aiex.npu.blockwrite(%0) {address = 1048576 : ui32} : memref<2xi32>
    }
    %tile_0_1 = aie.tile(0, 1)
    %mem_buff_0 = aie.buffer(%tile_0_1) {address = 0 : i32, mem_bank = 0 : i32, sym_name = "mem_buff_0"} : memref<8xi8> = dense<[0, 1, 2, -3, 4, 5, 6, 7]>
  }
}
```
This has the side benefit of checking the actual addresses and data, instead of checking debug messages.

fifield: > ```
> $ aie-opt --convert-aie-to-transaction initbuffer_int8.mlir 

This requires https://github.com/Xilinx/mlir-aie/pull/1802

lib/Targets/AIERT.cpp
```diff
@@ -373,6 +374,55 @@ LogicalResult AIERTControl::initLocks(DeviceOp &targetOp) {
...
+                  byteSize,
+              bytes.begin());
+          byteVec.insert(byteVec.end(), bytes.begin(), bytes.end());
+        }
+      } else
+        llvm::outs() << "buffer op type not supported for initialization "
+                     << bufferOp << "\n";
+      TRY_XAIE_API_FATAL_ERROR(XAie_DataMemBlockWrite, &devInst, tileLoc,
+                               bufferOp.getAddress().value(), byteVec.data(),
+                               byteVec.size());
```
fifield, line 418: what will happen when this `else` falls through to the `XAie_DataMemBlockWrite` call? Should the `else` cause return of `failure()` ?

test/Targets/AIETargetCDODirect/initbuffer_int8.mlir
jgmelber: It could be done. The existing couple tests for this backend use debug messages. I don't feel strongly one way or another

fifield (approved):

PR #1805, opened 2024-09-27
Cleanup aie2 simulation lit commands and remove meaningless run_on_board

fifield:
Don't pretend to compile aie2 simulation tests for a board. `%run_on_board` is already deprecated in the `lit.cfg.py` anyway.


jgmelber (approved):

stephenneuendorffer (commented):
Don't you still need to compile the host code in order to run the aie simulator?

fifield:
> Don't you still need to compile the host code in order to run the aie simulator?

yes, but I learned last week that we already compile the host code separately during `--aiesim`, just in a slightly different way in order to link in the simulation libraries. I suppose this is because simulation is a flag and not a proper target,

PR #1814, opened 2024-10-02
Using objectFifo link to access the shared memory between compute tiles

pvasireddy-amd:
Accessing the memory of the adjacent tile using objectFifo link.


lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -217,12 +217,50 @@ struct AIEObjectFifoStatefulTransformPass
...
 
-    // Only test for this objfifo belonging to a LinkOp if we are in the shared
-    // memory case; otherwise, we will return `true` in any case.
+    // Check if the operation uses shared memory for linking.
+    // If shared memory is available and the link operation is valid,
+    // mark the operation for splitting and determine if it is used in a link
+    // operation. If the link operation is a distribute or join operation, DMAs
+    // are required even if shared memory is available. If the link has
+    // different memref types, then shared memory cannot be used, and DMAs are
+    // needed.
```
AndraBisca, line 226: Maybe something more like this would be clearer: 

Check if the objectfifo operation can use shared memory for linking.  If the link operation is a distribute or a join operation, or if the link has different memref types, DMAs are required even if shared memory is available and the objectfifo should be split. Otherwise check if the via_shared_memory attribute of the objectfifo operation is set and apply it.

AndraBisca (changes requested):

test/objectFifo-stateful-transform/link_via_shared_mem.mlir
```diff
@@ -0,0 +1,74 @@
...
+        %tile12 = aie.tile(1, 2)
+        %tile22 = aie.tile(2, 2)
+
+        aie.objectfifo @of1 (%tile20, {%tile12}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+        aie.objectfifo @of2 (%tile12, {%tile22}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+
+        aie.objectfifo.link [@of1] -> [@of2] ([] [])
+
+        %ext_buff_in = aie.external_buffer {sym_name = "ext_buff_in"} : memref<16xi32>
+        aie.objectfifo.register_external_buffers @of1 (%tile20, {%ext_buff_in}) : (memref<16xi32>)
```
AndraBisca, line 72: Instead of this I think I'd be more insightful to have a core running on tile22 with objectfifo accesses to of2 (you can use a dummy function just to see what they lower to) to see which buffers are used.

test/objectFifo-stateful-transform/link_via_shared_mem2.mlir
```diff
@@ -0,0 +1,74 @@
...
+        %tile12 = aie.tile(1, 2)
+        %tile22 = aie.tile(2, 2)
+
+        aie.objectfifo @of1 (%tile20, {%tile12}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+        aie.objectfifo @of2 (%tile12, {%tile22}, 2 : i32) {via_shared_mem = 0 : i32} : !aie.objectfifo<memref<16xi32>>
+
+        aie.objectfifo.link [@of1] -> [@of2] ([] [])
+
+        %ext_buff_in = aie.external_buffer {sym_name = "ext_buff_in"} : memref<16xi32>
+        aie.objectfifo.register_external_buffers @of1 (%tile20, {%ext_buff_in}) : (memref<16xi32>)
```
AndraBisca, line 72: Same as above.

test/objectFifo-stateful-transform/link_via_shared_mem3.mlir
```diff
@@ -0,0 +1,71 @@
...
+        %tile12 = aie.tile(1, 2)
+        %tile22 = aie.tile(2, 2)
+
+        aie.objectfifo @of1 (%tile20, {%tile12}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+        aie.objectfifo @of2 (%tile12, {%tile22}, 2 : i32) {via_shared_mem = 1 : i32} : !aie.objectfifo<memref<16xi32>>
+
+        aie.objectfifo.link [@of1] -> [@of2] ([] [])
+
+        %ext_buff_in = aie.external_buffer {sym_name = "ext_buff_in"} : memref<16xi32>
+        aie.objectfifo.register_external_buffers @of1 (%tile20, {%ext_buff_in}) : (memref<16xi32>)
```
AndraBisca, line 69: In this case tile12 does not have access to tile22 memory right? So tile12's memory module is used for shared memory instead. It would be interesting to show what happens if it did, so maybe in another test you can use tile13 instead of tile22.

AndraBisca (approved):
Aside from these comments, and the extra test which I think would be really insightful, this looks good!

PR #1817, opened 2024-10-03
Numpy bfloat16 workaround

hunhoffe:
This PR conditionally installs `bfloat16` where possible and ignores errors otherwise. This is to allow some `bfloat16` integration in the mlir-aie python code. A workaround is necessary because the actual `bfloat16` library does not seem to compile well in vscode environments, such as used in the CI. While another long-term solution is likely necessary, this workaround allows one to create designs that look as though `bfloat16` is working properly.

The conditional installation looks like this in setup scripts and CI actions:
```bash
 pip install -r python/requirements_bfloat16.txt || echo "Failed to install bfloat16; this is ok!"
```

The conditional import in the python looks like this:
```python
class _pseudo_bfloat16(np.float16):
    def __init__(*args, **kwargs):
        raise TypeError(
            "This is a placeholder class for bfloat16. Install bfloat16 python package in order to instantiate bfloat16 values"
        )

try:
    from bfloat16 import bfloat16 as real_bfloat16

    bfloat16 = real_bfloat16
except:
    # Numpy doesn't support bfloat16 at this time. Using hacky extension of np.float16 as a placeholder, for now.
    print(
        "Warning! bfloat16 python package not available, so using placeholder bfloat16 type",
        file=sys.stderr,
    )
    bfloat16 = _pseudo_bfloat16
```

And finally, programming examples can import a version of `bfloat16` (either real or placeholder) like so:
```python
from aie.extras.util import bfloat16
```

This PR also diasables some of the AIE distros github action, to disable MacOS builds/tests, as those aren't currently supported. They didn't work before this PR, however, the action only runs when you change the workflow definition, so it was hidden from most PRs/CI runs.

### DIFF
--- a/programming_examples/basic/matrix_multiplication/single_core/aie2.py
+++ b/programming_examples/basic/matrix_multiplication/single_core/aie2.py
@@ -4,9 +4,9 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # (c) Copyright 2023 AMD Inc.
-
-import sys
 import argparse
+import numpy as np
+import sys
 
 from aie.extras.context import mlir_mod_ctx
 from aie.dialects.aie import *
@@ -14,6 +14,15 @@ from aie.dialects.aiex import *
 import aie.utils.trace as trace_utils
 from aie.utils.trace import PortEvent
 from aie.extras.dialects.ext.scf import _for as range_
+from aie.extras.util import bfloat16
+
+dtype_map = {
+    "bf16": bfloat16,
+    "i8": np.int8,
+    "i16": np.int16,
+    "f32": np.float32,
+    "i32": np.int32,
+}
 
 
 def main():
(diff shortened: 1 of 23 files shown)


hunhoffe:
@andrej tagging because this changes the matmul examples!

programming_examples/basic/matrix_multiplication/single_core/aie2.py
```diff
@@ -73,24 +82,8 @@ def my_matmul(M, K, N, m, k, n, dtype_in_str, dtype_out_str):
...
-    elif dtype_out_str == "i8":
-        dtype_out = T.i8
-    elif dtype_out_str == "i16":
-        dtype_out = T.i16
-    elif dtype_out_str == "f32":
-        dtype_out = T.f32
-    elif dtype_out_str == "i32":
-        dtype_out = T.i32
+    dtype_in = dtype_map[dtype_in_str]
+    dtype_out = dtype_map[dtype_out_str]
```
andrej, line 86: much better! thanks for cleaning up my mess!

andrej (approved):
Thanks for tagging me. I only looked at the matmul changes. They look good, but please keep in mind that the CI currently only tests a limited number of data types. You can see which ones the CI tests in the `run_makefile*.lit` files. (If the `.lit` does not specify, the test will use the default data type in the `makefile-common`.)

Since the matmul tests are already some of the slowest in the CI, and the changes seem very symmetrical for all data types, it probably doesn't make sense to add a bunch more CI tests. But I would advise to test everything with all data types at least locally! Especially the bf16 one.

hunhoffe:
@andrej Thanks for taking a look! I did test locally (only for default shape) through changing the input datatype in `makefile-common` to bf16

fifield (approved):

PR #1828, opened 2024-10-09
Use mlir-python-extras as a dependency

hunhoffe:
This PR removes code in `mlir-aie/python/extras` that is currently present in [`mlir-python-extras`](https://github.com/makslevental/mlir-python-extras) and uses `mlir-python-extras` as a dependency. To differentiate more clearly between what is in `mlir-python-extras` and what is added, I also separated the import paths; the `extras` from the external library are found at `aie.extras`, as before. The additional code that is useful for this project but potentially not upstream-able/applicable for `mlir-python-extras` has been moved to `aie.helpers`. I'm open to naming suggestions if there are better ideas than this!

The installation of `mlir-python-extras` is not as clean as I would like, because to build the python api wheels, dependencies are installed via the `requirements.txt` file; however, the requirements are parsed in a way that is incompatible with requirements installed via github repos (which is how `mlir-python-extras` is installed). I think the truly correct way to fix this is to change how we handle dependencies in the wheels; I'm not sure if this is in the scope of this PR or not. I'm happy to get advice.


hunhoffe:
Waiting on https://github.com/makslevental/mlir-python-extras/pull/101

jgmelber (approved):
LGTM

hunhoffe:
I was going to wait to merge until https://github.com/Xilinx/mlir-aie/pull/1835 was in, so this PR would be tested with that, but it's taking a while so I did the test locally with quick_setup and it worked for me, so I'll go ahead and merge this now.

PR #1831, opened 2024-10-11
Example of doing type verification with numpy types in design

hunhoffe:
This PR is an example of how to verify qualities of data types using normal numpy checks. I also made a mistake when running matmul in setting the dtype in `makefile-common`, and this type of check catches the error sooner with a more readable message.


programming_examples/basic/matrix_multiplication/whole_array/aie2.py
```diff
@@ -76,6 +76,13 @@ def my_matmul(M, K, N, m, k, n, n_aie_cols, dtype_in_str, dtype_out_str, b_col_m
     dtype_in = dtype_map[dtype_in_str]
     dtype_out = dtype_map[dtype_out_str]
 
+    assert np.issubdtype(dtype_in, np.integer) == np.issubdtype(
+        dtype_out, np.integer
+    ), f"Input dtype ({dtype_in}) and output dtype ({dtype_out}) must either both be integral or both be float"
+    assert (
+        np.dtype(dtype_out).itemsize >= np.dtype(dtype_in).itemsize
+    ), f"Output dtype ({dtype_out}) must be equal or larger to input dtype ({dtype_in})"
+
```
hunhoffe, line 85: @andrej Hello! Does this type of check seem reasonable to you? If so, I can add it to single core, etc and try to get this PR merged in.

jgmelber (approved):
Should this check be in all of the `matrix_multiplication` examples?

hunhoffe:
I was originally waiting to hear if @andrej  had an opinion either way, but the code is minimal so I went ahead and added the checks to the other matmul designs which have configurable input/output types

andrej:
Sorry for being slow to reply. I think any reasonable use case would have equal or higher precision for the output data type with regards to the input data types so I think this is a fine check.

hunhoffe:
@andrej no worries on timing, thanks for taking a look!

PR #1835, opened 2024-10-14
Add quick_setup.sh test to github action

hunhoffe:
Test the `quick_setup.sh` environment/build using the programming examples and programming guide on Ryzen AI.


.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -169,3 +169,39 @@ jobs:
...
+        # click on "launch_tmate_terminal_for_debug"
+        if: github.event_name == 'workflow_dispatch'
+                && inputs.launch_tmate_terminal_for_debug
+
+      - name: Run commands
+        run: |
+
+          pip cache purge
+          source /opt/xilinx/xrt/setup.sh
+          python -m pip install virtualenv
```
fifield, line 196: Other workflows use venv. It is already installed, is part of python, and does not pip install anything to the gihub user's site packages.

hunhoffe: Makes sense to me! I've changed quick_setup to use venv instead of virtualenv.

fifield: > Makes sense to me! I've changed quick_setup to use venv instead of virtualenv.

Note it was slightly religious statement I made (venv over virtualenv). There may be cases where virtualenv is less friction because ubuntu doesn't install venv by default and user might not have sudo to apt install it.

hunhoffe: I don't have a strong preference either way (venv vs virtualenv), but I think it makes sense to be consistent. Is there anyone else who may want to weigh in?

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -134,38 +134,17 @@ jobs:
...
-            -DCMAKE_MODULE_PATH=$PWD/../cmake/modulesXilinx \
-            -DMLIR_DIR=$PWD/../mlir/lib/cmake/mlir \
-            $CMAKE_ARGS
-
           ninja install
           ninja check-reference-designs
           ninja check-programming-guide
-
-          popd
+
```
fifield, line 149: nit: whitespace

PR #1837, opened 2024-10-14
Add github workflow for testing with Ryzen AI Software.

fifield:
Add CI testing for building with aietools from [Ryzen AI Software](https://www.amd.com/en/developer/resources/ryzen-ai-software.html). The tests currently run in an Ubuntu 22 docker with xrt and xdna-driver pre-installed. Ryzen AI Software pre-download into the docker and is installed as part of the build process. It is not pre-installed because I read [somewhere ](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#troubleshooting) that docker layers are limited to 10GB.

* aiecc.py: Add detection of aietools installed via Ryzen AI Software.
* `xchesscc_wrapper`: Add `-aietools` flag to override default aietools path (which is derived from xchesscc path). 
* bump cmake in `requirements.txt` to >= 3.29 to be compatible with venv created by Ryzen AI Software installer

### DIFF
--- a/.github/workflows/buildAndTestRyzenAISw.yml
+++ b/.github/workflows/buildAndTestRyzenAISw.yml
@@ -1,0 +1,133 @@
+name: Build and Test with Ryzen AI Sofware
+
+on:
+  push:
+    branches:
+      - main
+      - ryzen-ai-sw-test
+#  pull_request:
+  workflow_dispatch:
+    inputs:
+      AIE_COMMIT:
+        description: 'which aie commit to build'
+        type: string
+        required: false
+        default: ''
+      LIT_FILTER:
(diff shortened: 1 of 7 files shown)


fifield:
https://github.com/Xilinx/mlir-aie/actions/runs/11334837297/job/31521822240

fifield:
> The tests currently run in an Ubuntu 22 docker with xrt and xdna-driver pre-installed

I tested this container on a hawkpoint Linux box and the container can access the npu (i.e. mlir-aie tests pass) with `-v /dev:/dev --device-cgroup-rule 'c 261:* rmw'` on the `docker run` command line.

PR #1839, opened 2024-10-15
Filter tests based on available Vitis cmake components

fifield:
Not all aietools installations have support for every AIE architecture and not all installations contain `aiesimulator`. This PR adds lit test filtering based on which vitis cmake components are found (`AIE`, `AIE2`, `AIE2p`, etc), and adds filtering based on whether aiesimulator is installed.

For lit testsing each component found corresponds to an `"aietools_<lower case component>"` feature being available. For example if AIE2 support is found by cmake, then `aietools_aie2` will be an available feature. Available components are also stored in the `config.vitis_components` list. This PR generally filters test by entire test directories by querying `config.vitis_components` and setting `config.unsupported` where appropriate. 

Example lit script usage:
```
// REQUIRES: aiesimulator, aietools_aie2, chess
```

This PR also removes `config.has_libxaie` which seems to have overlapping but bit rotted functionality. 
This PR also fixes use of `config.unsupported` to match upstream usage by only setting it to `True` instead of to a directory name.

### DIFF
--- a/test/unit_tests/aievec_tests/lit.local.cfg
+++ b/test/unit_tests/aievec_tests/lit.local.cfg
@@ -1,7 +1,9 @@
+#
 # This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2023 AMD Inc.
+# Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
 
-config.unsupported = True
+if 'AIE' not in config.vitis_components:
+    config.unsupported = True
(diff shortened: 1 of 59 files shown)


stephenneuendorffer (approved):

test/unit_tests/aievec_tests/lit.local.cfg
```diff
@@ -0,0 +1,9 @@
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# Copyright (C) 2024, Advanced Micro Devices, Inc. All rights reserved.
+
+if 'AIE' not in config.vitis_components:
```
fifield, line 8: This is wrong, there are both aie1 and aie2 tests here.

jgmelber (approved):

test/unit_tests/aievec_tests/lit.local.cfg
fifield: > This is wrong, there are both aie1 and aie2 tests here.

...and they use `xca_udm_dbg` which I think we can assume is present if `aiesimulator` is found.

PR #1840, opened 2024-10-15
Match dynamic object fifo python examples style to other examples in programming examples.

hunhoffe:


AndraBisca (approved):
Looks great, thank you Erika!

PR #1844, opened 2024-10-16
[WIP] Use Peano for Programming Guide and Examples

jgmelber:


jackl-xilinx (approved):
Looks good. I was worried if peano might break some of the larger reference designs but if it's running through lit and compiling and running in peano, then that works for me. Nice work!

fifield (approved):

PR #1852, opened 2024-10-18
Try adding Python 3.12 to build wheels

jgmelber:


fifield:
FWIW this adds yet another job to the RyzenAI runners which are already a bottleneck.

Why can't the wheels be built on the generic runners?

PR #1854, opened 2024-10-18
Enable buildAndTestRyzenAISw workflow for PRs

fifield:
It has been working for pushes to main, enable for PRs too


fifield:
I'm going to leave this as not required because it takes >20min to run

PR #1859, opened 2024-10-21
Quick setup for Python 3.10 or 3.12

jgmelber:


jgmelber:
@mariodruiz

hunhoffe (approved):
Tested locally with both 3.10 and 3.12

mariodruiz:
I tested in NPU cloud, I checked `programming_guide/section-3` and `programming_examples/ml/bottleneck` both seem to work.

Is there any other piece I should test?

jgmelber:
Thanks @mariodruiz and @hunhoffe !

utils/quick_setup.sh
```diff
@@ -44,7 +44,7 @@ fi
 # if an install is already present, remove it to start from a clean slate
 rm -rf ironenv
 rm -rf my_install
-python3.10 -m venv ironenv
+python3 -m venv ironenv
```
jgmelber, line 47: Maybe this should still enforce python3.10 or python3.12, just in case users have those in their environment satisfying the check but they aren't the default

jgmelber: Agreed. That would be clear and safe

PR #1869, opened 2024-10-22
Strix npu support

fifield:
Support for strix devices.


aie_runtime_lib/AIE2P/aiesim/genwrapper_for_ps.cpp
fifield, line 1: needs license

jgmelber (approved):

PR #1881, opened 2024-10-24
Build wheels in docker container

fifield:
This is so that the actions can run on general purpose runners instead of on the Ryzen AI runners. It is slower, but this workflow does not need to be fast and the Ryzen AI runners are a limited resource. This PR also updates the build to use Vitis AIE Essentials instead of Vitis 2023.2 so that AIE2P is better supported.

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -1,4 +1,4 @@
-name: Build wheels on/for Ryzen AI
+name: Build wheels for Ryzen AI
 
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


fifield:
I triggered the workflow manually and tested the wheel on ubuntu 24 hawkpoint using the quick_setup instructions. It seemed to work.

https://github.com/Xilinx/mlir-aie/releases/download/latest-wheels/mlir_aie-0.0.1.2024102416+835e2ec-py3-none-manylinux_2_35_x86_64.whl

jgmelber (approved):

jgmelber:
Agreed. These are all great reasons to switch from running this on Ryzen AI runners

PR #1882, opened 2024-10-24
Linear transfer without transformation but with repeat

hunhoffe:
This PR relaxes some constraints made by checks in `AIEXDialect.cpp` and `AIEDmaToNpu.cpp` to allow one to continue to special-case the linear transfer without transform but to allow `repeat_count`/`iteration_stride`. 

To be specific, cases such as `size=[1, 1, 1, N], strides=[0, 0, 0, 1]` were special-cased so that when `N` exceeded the maximum allowable size, the transformation was still considered legal. This special case can still be done in this case: `size=[repeat_count, 1, 1, N], strides=[iteration_stride, 0, 0, 1]` but was previously disallowed by checks. 

I've relaxed the checks to allow this case and also added an example of this case in `test/npu-xrt`.

I also tried to fix-up some documentation to reflect the `repeat_count`/`iteration_stride` information.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -857,6 +857,10 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
           // access/store element at/to index (i * 16 /*stride_2*/ + j * 1 /*stride_1*/ + k * 2 /*stride_0*/)
     ```
 
+    Note that an additional dimension of sizes/strides is accepted (5th dimension for memtiles, 4th otherwise);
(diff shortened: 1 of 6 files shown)


hunhoffe:
I am wondering if it should be equally legal (with/without `repeat_count`/`iteration_stride`) to have:
```
size=[1, 1, N, 1], strides=[0, 0, 1, 1]
```
where `N` is too large.

Because this is logically equivalent to:
```
size=[1, 1, 1, N], strides=[0, 0, 0, 1]
```

Maybe I will try this in a subsequent PR.

jgmelber (approved):

PR #1900, opened 2024-11-01
Fully open-source AIE flow

jgmelber:
Use open source aie_api 
No longer rely on Vitis for any components to compile for AIE

Add aie_api as a submodule
Install aie_api with build 
Use aie_api from install for Peano programming examples and programming guide

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 [submodule "third_party/bootgen"]
 	path = third_party/bootgen
 	url = https://github.com/Xilinx/bootgen.git
+[submodule "third_party/aie_api"]
+	path = third_party/aie_api
+	url = https://github.com/jgmelber/aie_api
--- a/programming_examples/makefile-common
+++ b/programming_examples/makefile-common
@@ -3,7 +3,7 @@ AIETOOLS_DIR ?= $(shell realpath $(dir $(shell which xchesscc))/../)
 AIE_INCLUDE_DIR ?= ${AIETOOLS_DIR}/data/versal_prod/lib
 AIE2_INCLUDE_DIR ?= ${AIETOOLS_DIR}/data/aie_ml/lib
 
-AIEOPT_DIR ?= $(shell realpath $(dir $(shell which aie-opt))/../)
+AIEOPT_DIR ?= $(shell realpath $(dir $(shell which aie-opt))/..)
 
 WARNING_FLAGS = -Wno-parentheses -Wno-attributes -Wno-macro-redefined
 
--- a/programming_examples/makefile-common
+++ b/programming_examples/makefile-common
@@ -11,8 +11,8 @@ CHESSCC1_FLAGS = -f -p me -P ${AIE_INCLUDE_DIR} -I ${AIETOOLS_DIR}/include
 CHESSCC2_FLAGS = -f -p me -P ${AIE2_INCLUDE_DIR} -I ${AIETOOLS_DIR}/include -D__AIENGINE__=2 -D__AIEARCH__=20
 CHESS_FLAGS = -P ${AIE_INCLUDE_DIR}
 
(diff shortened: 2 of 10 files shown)


fifield (approved):

programming_examples/makefile-common
```diff
@@ -3,6 +3,8 @@ AIETOOLS_DIR ?= $(shell realpath $(dir $(shell which xchesscc))/../)
 AIE_INCLUDE_DIR ?= ${AIETOOLS_DIR}/data/versal_prod/lib
 AIE2_INCLUDE_DIR ?= ${AIETOOLS_DIR}/data/aie_ml/lib
 
+AIEOPT_DIR ?= $(shell realpath $(dir $(shell which aie-opt))/../)
```
keryell, line 6:
```suggestion
AIEOPT_DIR ?= $(shell realpath $(dir $(shell which aie-opt))/..)
```

runtime_lib/aie_api/aieapi.cmake
```diff
@@ -0,0 +1,13 @@
+
+function(add_aie_api_headers SRCPATH BUILDPATH INSTALLPATH)
+  message("Installing aie_api includes from ${SRCPATH} in ${BUILDPATH}")
+
+  # copy header files into build area
+  install(DIRECTORY ${SRCPATH}/ DESTINATION ${BUILDPATH}/aie_api)
```
keryell, line 6: I am curious about all these trailing `/` in the project.

keryell (commented):
Could you improve the PR description (which will end up as the eventual commit message for the eternity) about why we need this new API?

jgmelber:
> Could you improve the PR description (which will end up as the eventual commit message for the eternity) about why we need this new API?

Changed the description and added a couple lines. We use this API in the single core code. It was previously distributed with Vitis. Now that this is open sourced and we can use Peano. We no longer have to depend on Vitis.

programming_examples/makefile-common
jgmelber: Copy the entire contents of the directory in cmake

runtime_lib/aie_api/aieapi.cmake
jgmelber: Copy the entire contents of the directory in cmake

.gitmodules
```diff
@@ -11,3 +11,6 @@
 [submodule "third_party/bootgen"]
 	path = third_party/bootgen
 	url = https://github.com/Xilinx/bootgen.git
+[submodule "third_party/aie_api"]
+	path = third_party/aie_api
+	url = https://github.com/jgmelber/aie_api
```
stephenneuendorffer, line 16: This is the best URL?

stephenneuendorffer (approved):

.gitmodules
jgmelber: For now, I have a fork that comments out an include to adf.h. A follow on will remove any defines to __AIENGINE__, that will gate the include. We use the deprecated convention that flag defines something about the version of the AIE version and redundantly define the correct __AIE_ARCH__.

runtime_lib/aie_api/aieapi.cmake
keryell: Interesting. You mean that without a `/` it would only create an empty directory?

jgmelber:
```
A trailing slash will leave the last component empty and install the contents of the input directory to the destination. 
```

https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html

keryell: Interesting. With CMake I learn all my life long!
Actually this is the equivalent of using `/.` in as with normal Unix copy-related commands.

PR #1901, opened 2024-11-02
Make pre-allocated buffers work again

fifield:
I ran into problems on a branch where I was trying to manually set the address of buffers and the bank aware allocator silently failed and unexpectedly overwrote the manually set addresses. This PR fixes the problem by,

* Fail, don't warn, when bank aware buffer allocation conflicts with buffers with 'address' attribute set.
* Make basic buffer allocation scheme aware of 'address' attributes
* Small aiecc update to print errors to user when python bindings are used to run pass pipeline. For example during buffer allocation.
* I also snuck in a comment and indentation fixup for most of the other allocation tests

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
@@ -26,7 +26,7 @@ using namespace xilinx::AIE;
 //===----------------------------------------------------------------------===//
 LogicalResult checkAndPrintOverflow(TileOp tile, int address,
                                     int maxDataMemorySize, int stacksize,
-                                    SmallVector<BufferOp, 4> buffers) {
+                                    SmallVector<BufferOp> &buffers) {
   if (address > maxDataMemorySize) {
     InFlightDiagnostic error =
         tile.emitOpError("allocated buffers exceeded available memory\n");
@@ -65,8 +65,8 @@ LogicalResult basicAllocation(TileOp tile) {
   else
     maxDataMemorySize = targetModel.getLocalMemorySize();
 
-  SmallVector<BufferOp, 4> buffers;
-  std::list<BufferOp> allocated_buffers;
+  SmallVector<BufferOp> buffers;
+  SmallVector<BufferOp> allocated_buffers;
   // Collect all the buffers for this tile. If the buffer has an address, add
   // it to allocated_buffers. Otherwise, add it to buffers.
   device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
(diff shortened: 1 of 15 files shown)


fifield:
TODO: some tests

python/compiler/aiecc/main.py
```diff
@@ -334,7 +334,12 @@ def run_passes(pass_pipeline, mlir_module_str, outputfile=None, verbose=False):
         print("Running:", pass_pipeline)
     with Context() as ctx, Location.unknown():
         module = Module.parse(mlir_module_str)
-        PassManager.parse(pass_pipeline).run(module.operation)
+        pm = PassManager.parse(pass_pipeline)
+        try:
+            pm.run(module.operation)
+        except Exception as e:
+            print("Error running pass pipeline: ", pass_pipeline, e)
+            raise(e)
```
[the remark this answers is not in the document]
keryell: Interesting suggestion... I guess the tool should suggest
```suggestion
            raise e
```

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -66,16 +66,28 @@ LogicalResult basicAllocation(TileOp &tile) {
...
   });
-  // Sort by allocation size.
+
+  // Sort buffers by allocation size.
   std::sort(buffers.begin(), buffers.end(), [](BufferOp a, BufferOp b) {
     return a.getAllocationSize() > b.getAllocationSize();
   });
 
+  // Sort allocated_buffers by address
+  allocated_buffers.sort([](BufferOp a, BufferOp b) {
```
keryell, line 87: Interesting, I did not remember `std::list` had a `sort()` member!
Any reason for not using a `std::vector`, besides being a Halloween programming exercise? :ghost: :jack_o_lantern: :smile:

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -150,25 +171,29 @@ void setAndUpdateAddressInBank(BufferOp buffer, int64_t start_addr,
...
+                             std::vector<int64_t> &nextAddrInBanks,
+                             std::vector<BankLimits> &bankLimits) {
+  auto addrAttr = buffer->getAttrOfType<IntegerAttr>("address");
+  if (!addrAttr)
+    return false;
+
+  int addr = addrAttr.getInt();
+  for (int i = 0; i < numBanks; i++) {
+    // if the address is not within the bank, continue
+    if (addr < bankLimits[i].startAddr && addr >= bankLimits[i].endAddr)
```
fifield, line 185:
```suggestion
    if (addr < bankLimits[i].startAddr || addr >= bankLimits[i].endAddr)
```

Issue #1884, opened 2024-10-25
Dynamic Object-FIFO lowering not functional


andrej:
Hi @AndraBisca @hunhoffe, I hope you guys are well.

I'm looking at the [new Object-FIFO tests](https://github.com/Xilinx/mlir-aie/tree/main/test/npu-xrt/dynamic_object_fifo). I believe none of these tests use the dynamic lowering, as the `--dynamic-objFifos` option [is missing from the aiecc.py call](https://github.com/Xilinx/mlir-aie/blob/088876dba42205b7dfcb94e5d1fc280f9d543c5b/test/npu-xrt/dynamic_object_fifo/ping_pong/aie2.py#L12). This is confirmed by looking at the lowered MLIR, which does not contain any of the `scf.index_switches` that the dynamic lowering is supposed to emit.

If I do try to enable the dynamic lowering by adding the command line option, compilation fails at the buffer address assignment stage (`--aie-assign-buffer-addresses`) with this error:

```
aie-opt: /project/llvm-project/mlir/lib/IR/Types.cpp:134: unsigned int mlir::Type::getIntOrFloatBitWidth() const: Assertion `isIntOrFloat() && "only integers and floats have a bitwidth"' failed.
```
<details>
<summary>Backtrace</summary>
<pre>
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: aie-opt "--pass-pipeline=builtin.module(lower-affine,aie-canonicalize-device,aie.device(aie-assign-lock-ids,aie-register-objectFifos,aie-objectFifo-stateful-transform{ dynamic-objFifos=1 },aie-assign-bd-ids,aie-lower-cascade-flows,aie-lower-broadcast-packet,aie-lower-multicast,aie-assign-tile-controller-ids,aie-generate-column-control-overlay{ route-shim-to-tile-ctrl=0 },aie-assign-buffer-addresses{ basic-alloc=0 }),convert-scf-to-cf)" ./aie2.mlir
 #0 0x00005642438bbf7b llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) Signals.cpp:0:0
 #1 0x00005642438b947b llvm::sys::RunSignalHandlers() Signals.cpp:0:0
 #2 0x00005642438b95a5 SignalHandler(int) Signals.cpp:0:0
[...]

PR #1907, opened 2024-11-05
Fixing dynamic objectFifo 

pvasireddy-amd:
Issue #1884


pvasireddy-amd:
The sliding_window_conditional example is the only one that fails, but this issue appears unrelated to dynamic objectFifos. The example defines 3 input objectFifos, but only 2 are created when lowered using the --aie-objectFifo-stateful-transform pass. Currently still investigating.

pvasireddy-amd:
Since the problem is unrelated to the dynamic objectFifos, I have pushed a working MLIR version of the `sliding_window_conditional` example into `test/npu-xrt/dynamic_object_fifo/sliding_window_conditional`. A separate PR will follow with the necessary fixes to enable running the Python version of this example.

AndraBisca (approved):
Thank you for the fix!
Regarding the sliding_window_conditional example, it appears that the if-statement does not lower from IRON to MLIR as there is no wrapper similar to the for_ loop statement. As we plan to add this change in a subsequent PR I am happy to merge this fix in already.

jgmelber (approved):

hunhoffe:
For the record, mlir-python-extras should contain some wrappers for if statements e.g., here: https://github.com/makslevental/mlir-python-extras/blob/b19f9b4246a0e092a2f30eaa6dde8584adb64bc5/mlir/extras/dialects/ext/scf.py#L363

I have not attempted to use them in IRON, but wanted to mention them for whoever goes down that path later on!

PR #1908, opened 2024-11-05
Enable one of the npu-xrt/e2e tests

hunhoffe:
I came across the set of `test/npu-xrt/e2e` tests more or less by accident today. I am specifically interested in (eventually) creating an end-to-end notebook, and looking at these tests seemed like a decent place to start exploring. These tests are all currently disabled.

In this PR:
* I ported one of the tests to be supported (compiled with peano), which now runs correctly in the CI as far as I can tell
* I removed all unsupported tests and the compiler util file that seems pretty outdated. I don't think there was an easy path to updating all these things.

As a note for the future, some of these examples and the compiler utils showed some neat/interesting ways to interact with the compiler in Python, which may be a good idea to revisit in the future!


fifield (approved):
Thanks for the cleanup. The removed tests and utils can always be recovered from the repo if anyone wants to revive them.

PR #1911, opened 2024-11-08
Add llvm rtti support in AIETargetModel

fifield:
Add [basic LLVM RTTI support](https://llvm.org/docs/HowToSetUpLLVMStyleRTTI.html) to AIETargetModel. This allows querying the target model object hierarchy in the llvm way with `isa<>`, etc., which is sometimes nicer than checking `AIETargetModel::getTargetArch()` or other properties. As someone who has spent time updating code for strix and beyond I think it is nice to have this option.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIETargetModel.h
+++ b/include/aie/Dialect/AIE/IR/AIETargetModel.h
@@ -53,8 +53,31 @@ using TileID = struct TileID {
 };
 
 class AIETargetModel {
+
 public:
-  AIETargetModel() = default;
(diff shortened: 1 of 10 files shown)


stephenneuendorffer (approved):
Makes sense to me... Can you add corresponding tests to make sure the enum gets maintained properly?

keryell (approved):

fifield:
> Makes sense to me... Can you add corresponding tests to make sure the enum gets maintained properly?

test added.

PR #1913, opened 2024-11-12
[WIP][CIR] Single-source C++ lowering with ClangIR MLIR CIR

keryell:
This is a huge WIP for now and far from review.
It requires a specific branch of ClangIR instead of the usual Clang/LLVM/MLIR and it is not clear how to have this in current CI.


keryell:
This relies on https://github.com/llvm/clangir/pull/1203 and https://github.com/llvm/clangir/pull/1164

PR #1914, opened 2024-11-12
Cleanup extra code in headers

fifield:
Cleanup some extra code in a couple header files. I bumped into this because of some compiler warnings regarding redefinition of `DEBUG_TYPE`, which shouldn't be in header files anyway.

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.h
@@ -11,27 +11,8 @@
 #ifndef AIE_ASSIGN_BUFFER_DESCRIPTOR_IDS_H
 #define AIE_ASSIGN_BUFFER_DESCRIPTOR_IDS_H
 
-#include <optional>
+#include "aie/Dialect/AIE/IR/AIETargetModel.h"
 
-#include "aie/Dialect/AIE/IR/AIEDialect.h"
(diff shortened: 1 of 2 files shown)


jgmelber (approved):

include/aie/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.h
```diff
@@ -11,27 +11,8 @@
...
-#include "aie/Dialect/AIE/IR/AIEDialect.h"
-#include "aie/Dialect/AIE/Transforms/AIEPasses.h"
-
-#include "mlir/Pass/Pass.h"
-
-#define DEBUG_TYPE "aie-assign-bd-ids"
-
-using namespace mlir;
-using namespace xilinx;
 using namespace xilinx::AIE;
```
keryell, line 16: Thanks for all the seasonal cleaning!
Even the last one is too much for my taste, but it would be probably too extreme for now. :smiley:

keryell (approved):

include/aie/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.h
fifield: > Thanks for all the seasonal cleaning! Even the last one is too much for my taste, but it would be probably too extreme for now. 😃

yeah it's a slippery slope. Before you know it you're "cleaning" every file in the repo.

PR #1915, opened 2024-11-12
Remove CallableOpInterface from MemOp and MemTileOp

fifield:
This PR removes `CallableOpInterface` from `MemOp` and `MemTileOp`.

`MemOp` and `MemTileOp` implement the `CallableOpInterface` but the way it's implemented is problematic, because the implementation returns pointers to the stack, for example:
```
llvm::ArrayRef<mlir::Type> getArgumentTypes() { return getOperand().getType(); }
```
This could be resolved by stashing a `FunctionType` as an attribute, as the various flavors of `FuncOp` do, and safely returning the types. But it also appears that the comment explaining why the ops have `CallableOpInterface`:
```
// MemOps are not actually Callable, but we want to inline code into them, so we have to
// implement CallableOpInterface
```
is not accurate, because I can remove the `CallableOpInterface` from these two ops completely and nothing seems to break.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -1046,8 +1046,6 @@ def AIE_DMAOp: AIE_Op<"dma", [
   }];
 }
 
-// MemOps are not actually Callable, but we want to inline code into them, so we have to
-// implement CallableOpInterface
 def AIE_MemOp: AIE_Op<"mem", [
     TileElement, FlowEndPoint,
     IsCoreTile, HasValidBDs, HasValidDMAChannels,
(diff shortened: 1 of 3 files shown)


stephenneuendorffer (approved):
No real objections.  I think the goal of this is clear, but it's not something we've ever used aggressively.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1046,10 +1046,8 @@ def AIE_DMAOp: AIE_Op<"dma", [
   }];
 }
 
-// MemOps are not actually Callable, but we want to inline code into them, so we have to
-// implement CallableOpInterface
```
keryell, line 1050: Is this some kind of design documentation or precondition that should be kept somewhere, perhaps around the operation description?

PR #1920, opened 2024-11-14
Add tracing support for Conv2D example

asyms:
This PR adds support for tracing in the Conv2D example in `programming_examples/ml/conv2d`.

It updates: 

- `aie2.py` to set up a packet-based tracing flow 
- `test.py` to correctly separate the data and tracing buffers
- `Makefile` to add new make rules for tracing

All changes were done accordingly to how the tracing is setup for the programming_examples/basic/vector_scalar_mul.

One caveat: Compilation for Conv2D doesn't function (for me) without adding the AIETOOLS_DIR/include to the programming_examples/makefile-common. Not sure if this is a common issue or if something is wrong on my end regarding my peano install.


programming_examples/ml/conv2d/aie2.py
jackl-xilinx, line 1: `N_in_bytes = 32*32*8*8` is the output buffer size in bytes which serves as the offset for where trace is written to. In this example, you can rewrite this to be `N_in_bytes = tensorSize` since that is what we want. In the vector_scalar_mul example, it is x2 because the output datatype for that example is int16 whereas this one is int8. So we need to scale the output since N_in_bytes is in terms of bytes.

jackl-xilinx (approved):
Great addition Arne! You can make the changes I suggested for clarify and then we can merge this later today.

programming_examples/ml/conv2d/aie2.py
asyms: Okay that makes sense! I have updated it.

asyms:
> Can you please enable testing of the trace functionality in the lit test?

@jgmelber Is this directed at me or Jack because if me I have no idea how to go about this.

jackl-xilinx:
@asyms  I can push the change if you give me permission to push to your branch. For reference, it needs code similar to https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/basic/vector_scalar_mul/run_makefile.lit#L14

asyms:
@jackl-xilinx Thanks! I have invited you, should have push access once you accept.

PR #1870, opened 2024-10-22
TilerHelper and 2D Tiling Visualizations

hunhoffe:
## Description

This PR introduces some classes to help with data transformations/tiling and data visualization. The main classes introduced are:
* `TensorTile`: A class representing information for specifying a 'chunk' of data, suitable for use in describing a `npu_dma_memcpy_nd` operation; also includes code for visualizing the `TensorTile`.
* `TensorTileSequence`: A wrapper class for a list of `TensorTile`s; also includes code for visualization (including animations).
* `TensorTiler2D`: A class that helps with data tiling. Provides `simple_tiler()`, `group_tiler()`, and `step_tiler()` methods to aid in generating iterative patterns of tiling over the tensor.

I also include:
* A (non-exhaustive, but also not completely trivial) set of tests in `test/python/tensortiler`
* A minor edit of a few programming examples (`dma_transpose`, `matrix_scalar_add`, `row_wise_bias_add` `matrxi_multiplication/whole_array`) to show how to use the `TensorTiler2D`
* Two new programming examples (in `programming_examples/basic/tiling_exploration`) and an introductory notebook that use the device to create an access order map that can be compared to the map provided by the `TensorTiler2D`. These examples are meant to verify the `TensorTiler2D` can produce valid access patterns but also illustrate how to use the tensor tiling code.
* A change to `npu_dma_memcpy_nd` to support taking a `TensorTile` directly instead of sizes/strides/offset

Some incidental things that got included in this PR:
* First programming example with no inputs, only an output, which required some minor changes to support

## Note for future contributions
* Visualization of transformation chaining (e.g., toStream AND fromStream w/ ObjectFIFOLink)
* Helper class for transformation chaining?
* Support for overlapping data patterns (e.g., image transforms)
* Integration of visualization into runtime sequence and/or memtile data movement.

## Notes

Credit to Andre for his really excellent tool here: https://andreroesti.com/data-layout-viz/data_layout.html
This is my attempt at making something similar that can be integrated into a development tool directly, with the additional step of generating sizes/strides for you.

PR #1924, opened 2024-11-15
Mat mul whole array implementation using tiler helper tools

hunhoffe:
The changes to mat mul whole array in https://github.com/Xilinx/mlir-aie/pull/1870 did not change implementation of the design at all, they just added some hooks to help with visualization. 

The implementation changes in this PR actually use the `TensorTiler2D` to generate offsets/sizes/strides in a new copy of the mat mul whole array design.

This is useful for comparing how the code looks in each approach, but also the B tiles produced by the tiler are functionally equivalent to the B tiles produced by the original design, but the sizes/strides are different. I plan to use this branch to benchmark if the difference results in any performance changes between the two implementations.

Update: benchmark with sweep does not show performance difference, so I am going to try to merge.


jgmelber (approved):
I am looking forward to a section in the programming guide about the tiler functionality

PR #1925, opened 2024-11-15
Fix typo in lit file for mat mul whole array test

hunhoffe:
I don't think this typo affected the tests in any way, but it still seems best to be correct.


andrej (approved):
Ooops .. good catch

PR #1929, opened 2024-11-17
Add `hasProperty` method to targetModel

erwei-xilinx:
The `hasProperty` method checks whether a target model has a given device property (implemented as `enum`).
List of initial properties added:
- UsesSemaphoreLocks
- IsNPU
- IsVirtualized
- UsesMultiDimensionalBDs


include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -236,6 +236,9 @@ class AIETargetModel {
   // There are several special cases for handling the NPU at the moment.
   virtual bool isNPU() const { return false; }
 
+  // Return true if this device is using semaphore locks.
+  virtual bool isUsingSemaphoreLocks() const { return false; }
```
fifield, line 240: There aren't any `hasProperty()` functions in `AIETargetModel` but this seems like it could be one. "is using" is a bit awkward to me.

fifield (commented):
This is probably better than the current convention which is to test for `!aie1`. Should we update such cases in this PR or gradually?

jgmelber:
Isn't it enough to check if the code is targeting AIE1. Otherwise it has semaphore locks?

erwei-xilinx:
> Isn't it enough to check if the code is targeting AIE1. Otherwise it has semaphore locks?

Functionality wise, yes we can definitely check for aie1. It's just that when other projects such as air use mlir-aie's target models, it isn't so intuitive for outside developers as there isn't a logical connection between "the device is not aie1" and "the device is using semaphore locks".

include/aie/Dialect/AIE/IR/AIETargetModel.h
erwei-xilinx: Good point. Method is reimplemented with `hasProperty(UsesSemaphoreLocks)`. Thanks for the pointer.

erwei-xilinx:
> This is probably better than the current convention which is to test for `!aie1`. Should we update such cases in this PR or gradually?

Sure, I updated some cases with the `hasProperty` method in this PR.

lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -647,6 +652,21 @@ bool AIE2TargetModel::isLegalTileConnection(int col, int row,
   return false;
 }
 
+bool AIE2TargetModel::hasProperty(ModelProperty prop) const {
+  uint64_t properties = 0;
+  properties |= AIETargetModel::UsesSemaphoreLocks;
+  properties |= AIETargetModel::UsesMultiDimensionalBDs;
+  properties |= AIETargetModel::IsNPU;
+  return (properties & prop) == prop;
+}
```
fifield, line 661: AIE2 does not imply `IsNPU`. It should get inherited from `BaseNPUTargetModel` or `BaseNPU3TargetModel`.

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -232,9 +246,8 @@ class AIETargetModel {
   // Run consistency checks on the target model.
   void validate() const;
 
-  // Return true if this is an NPU-based device
-  // There are several special cases for handling the NPU at the moment.
-  virtual bool isNPU() const { return false; }
+  // Return true if this device has a given property.
+  virtual bool hasProperty(ModelProperty Prop) const = 0;
```
fifield, line 250:
```suggestion
  bool hasProperty(ModelProperty Prop) const { return false; }
```

But it might also be worth looking at how rtti works in AIETargetModel. Using a similar scheme, each subclass would add it's bitvector of features via the call to the base class constructor. The base class would OR the bitvector with it's own, so the subclass only needs to add new features not present in the base class.
```c++
#include <iostream>
#include <cstdlib>

class ClassA {
private:
  const uint32_t features;
public:
  ClassA(uint32_t f) : features(f) {}
  uint32_t getFeatures() const { return features; };
};

class ClassB : public ClassA {
  static constexpr uint32_t class_b_features = 0x1;
public:
  ClassB(uint32_t f) : ClassA(f | class_b_features) {}
  ClassB() : ClassA(class_b_features) {}
};

class ClassC : public ClassB {
  static constexpr uint32_t class_c_features = 0x2;
public:
  ClassC() : ClassB(class_c_features) {}
};

class ClassD : public ClassB {
  static constexpr uint32_t class_d_features = 0x4;
public:
  ClassD() : ClassB(class_d_features ) {}
};

int main(int argc, char *argv[]) {
  std::cout << "B " << ClassB().getFeatures() << "\n";
  std::cout << "C " << ClassC().getFeatures() << "\n";
  std::cout << "D " << ClassD().getFeatures() << "\n";
}
```
```bash
$ ./a.out
B 1
C 3
D 5
```

lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -289,6 +289,11 @@ bool AIE1TargetModel::isLegalTileConnection(int col, int row,
   return false;
 }
 
+bool AIE1TargetModel::hasProperty(ModelProperty prop) const {
+  uint64_t properties = 0;
+  return (properties & prop) == prop;
+}
+
```
fifield, line 296: This doesn't add anything and can be covered by the base class `AIETargetModel` ?

lib/Dialect/AIE/IR/AIETargetModel.cpp
erwei-xilinx: Thanks for finding the bug. Fixed.

include/aie/Dialect/AIE/IR/AIETargetModel.h
erwei-xilinx: Good point. PR updated. Thanks!

lib/Dialect/AIE/IR/AIETargetModel.cpp
erwei-xilinx: Agreed. Now removed in PR.

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -232,9 +246,11 @@ class AIETargetModel {
   // Run consistency checks on the target model.
   void validate() const;
 
-  // Return true if this is an NPU-based device
-  // There are several special cases for handling the NPU at the moment.
-  virtual bool isNPU() const { return false; }
+  // Return true if this device has a given property.
+  uint32_t ModelProperties = 0;
```
fifield, line 250: This should be private

erwei-xilinx: Good point. Let me fix this in a separate PR.

erwei-xilinx: PR created: https://github.com/Xilinx/mlir-aie/pull/1944

PR #1936, opened 2024-11-20
Change aie.device op from NoTerminator to SingleBlockImplicitTerminator

fifield:
There is a block merging pattern used by the greedy pattern rewriter which [queries a values's parent block's terminator](https://github.com/llvm/llvm-project/blob/08af115d97e39223b5cc8bdbc56b1dfb758bf6d3/mlir/lib/Transforms/Utils/RegionUtils.cpp#L631). If the value is an aie.buffer in an aie.device, the getTerminator() call will assert because aie.device has no terminator. This is the root cause of the bug in https://github.com/Xilinx/mlir-aie/pull/1755. Another fix would be to modify the upstream code to first check that the block is expected to have a terminator


python/dialects/aie.py
```diff
@@ -290,7 +290,7 @@ def __init__(
         )
 
     def get_name(self):
-        return self.result.get_name()
+        return self.sym_name.value if self.sym_name else ""
```
andrej, line 293: Good change. Before this, I had to have  `some_buf.get_name()[1:]` somewhere in my code, I think because it contained the `@` in front of the name. Good to be aware of this though in case it breaks other code anywhere (hopefully nobody did what I did).

test/python/tile_array.py
```diff
@@ -347,15 +347,15 @@ def npu():
...
-        # CHECK: (1, 0) <Channel: buffer=MemRef(%buffer_3_2, memref<10x10xi32>) producer_lock=Scalar(%buffer_3_2_producer_lock = aie.lock(%tile_3_2) {sym_name = "buffer_3_2_producer_lock"}) consumer_lock=Scalar(%buffer_3_2_consumer_lock = aie.lock(%tile_3_2) {sym_name = "buffer_3_2_consumer_lock"})>
-        # CHECK: (1, 1) <Channel: buffer=MemRef(%buffer_3_3, memref<10x10xi32>) producer_lock=Scalar(%buffer_3_3_producer_lock = aie.lock(%tile_3_3) {sym_name = "buffer_3_3_producer_lock"}) consumer_lock=Scalar(%buffer_3_3_consumer_lock = aie.lock(%tile_3_3) {sym_name = "buffer_3_3_consumer_lock"})>
+        # CHECK: (0, 0) <Channel: buffer=MemRef(%33, memref<10x10xi32>) producer_lock=Scalar(%34 = "aie.lock"(%14) <{sym_name = "_producer_lock"}> : (index) -> index) consumer_lock=Scalar(%35 = "aie.lock"(%14) <{sym_name = "_consumer_lock"}> : (index) -> index)>
+        # CHECK: (0, 1) <Channel: buffer=MemRef(%36, memref<10x10xi32>) producer_lock=Scalar(%37 = "aie.lock"(%15) <{sym_name = "_producer_lock"}> : (index) -> index) consumer_lock=Scalar(%38 = "aie.lock"(%15) <{sym_name = "_consumer_lock"}> : (index) -> index)>
```
andrej, line 356: Just out of curiosity, what is happening here? Why is it fine for all these locks to have the same symbol name?

andrej (approved):
Very nice, thank you for investigating the root cause and fixing this. I double-checked with #1772 and this changeset allows that to compile.

test/python/tile_array.py
fifield: Ah, good catch, I'll investigate. There's a difference in how the ir is getting printed, likely because it isn't considered "valid" yet without a terminator when the print()s happen. But there seems to be some difference in how names/symbols are resolved too which I don't fully understand.

jgmelber (approved):

keryell:
> There is a block merging pattern used by the greedy pattern rewriter which [queries a values's parent block's terminator](https://github.com/llvm/llvm-project/blob/08af115d97e39223b5cc8bdbc56b1dfb758bf6d3/mlir/lib/Transforms/Utils/RegionUtils.cpp#L631). [...] Another fix would be to modify the upstream code to first check that the block is expected to have a terminator

Yes, go for it for upstream!
We need more people caring about the commons.

PR #1945, opened 2024-11-25
Fix topological sort use

keryell:
Probably this code was never exercised since the API was wrongly used (topologicalSort does not mutate its parameter) and it was probably not the right function at the first place.


keryell:
This caused me quite some debug time since I took this code as a word of truth in https://github.com/Xilinx/mlir-aie/pull/1913/files#diff-9a9c67284c8006f81d3dcfb1e27c01b2de924a2d5d41f27c9fd640af38fe5a0bR88 :smiley_cat:

AndraBisca (approved):
Thank you for the fix Ronan! :)

jgmelber (commented):
Is there a test case you can provide for this?

keryell:
> Is there a test case you can provide for this?

No, since I do not use this code here, this is just an educated guess with my fix.
It looks like @makslevental wrote this last year
abbcb9a05 lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp (Maksim Levental)      2023-11-29
https://github.com/Xilinx/mlir-aie/commit/abbcb9a05a37e33217db8a0771594b81e3c4b95f#diff-9b64fc073f494cbde2d58fb41a69634d2cbb29b6b5430216c16649c000b7250a
so perhaps he could comment?

PR #1947, opened 2024-11-29
Add programming example with BD level syntax and init values

AndraBisca:


jgmelber (commented):
Can you add more comments in the Python design files of the `@mem` blocks in particular? 

@eddierichter-amd does this look good for vck5000?

eddierichter-amd:
> Can you add more comments in the Python design files of the `@mem` blocks in particular?
> 
> @eddierichter-amd does this look good for vck5000?

Thanks for the notice! I am not sure and will need to check this out. My intuition is that it should be good, only thing perhaps is making sure we can capture the AIE1 lock functionality, but I imagine that was already taken care of outside of this PR.

test/npu-xrt/vec_vec_add_tile_init/aie.mlir
jgmelber, line 1: Why remove this test?

AndraBisca: I removed it initially to avoid having the same design in two places, however the design I propose for the programming examples is a little bit different so I can put it back.

PR #1951, opened 2024-12-03
Removing dynamic object fifo examples

pvasireddy-amd:
Examples are in test/npu-xrt/dynamic_object_fifo folder


AndraBisca (approved):
Looks good! These are exact duplicates of the ones in test/npu-xrt/ so I think it's good to move them!

PR #1961, opened 2024-12-09
Fix for symbol duplicate error in static L1 initialization

pvasireddy-amd:
With static L1 buffer initialization, neighboring cores end up creating identical buffer symbols, which results in a 'Symbol Duplication' error. Temporary fix.


AndraBisca:
I think it would make sense to create a dedicated test for this issue + fix instead of adding extra code to existing tests. Just for clarity.

AndraBisca:
Sounds good! :)

jgmelber (approved):

AndraBisca (approved):

Issue #1964, opened 2024-12-10
Some questions about the number of Compute tiles?


ngdymx:
Hi team,

I read the tutorial and got that the device is present in HawkPoint (e.g., 8040HS) SOCs. has 5 Columns and 6 Rows, as shown below:

![image](https://github.com/user-attachments/assets/d1b00f12-c0b8-4f6b-ae66-1c599000ad65)

However, I cannot control the leftmost column CTS. I want to confirm whether we cannot access it. Is there any way to access and control that column? Thanks a lot!

Additionally, the following is noted about device partitions:

![image](https://github.com/user-attachments/assets/ff5951cd-05f2-47ff-8982-17bcc74183d2)

jgmelber:
Hi! If you use the device type `npu1` you should be able to  access the whole 5 column device. Remember that you will need to start your shimDMAs at column 1 in that case as there is not a shimTile in the leftmost column.

ngdymx:
Hi @jgmelber,

Thanks.

I try it, but get the error shown in the following figure.

<img width="950" alt="image" src="https://github.com/user-attachments/assets/addd8d63-4a4c-4f60-a971-2d1b3fcd0744">

Could you help me with it? Is there any example I can follow? Thanks a lot!

jgmelber:
@ngdymx can you share the entire design file, or at least the `@sequence`?

ngdymx:
Hi @jgmelber,

Sorry for the late reply. I write a passthrough kernel to test it. It works well when I set the dev is ```AIEDevice.npu1_1col```.  When I change the dev to ```AIEDevice.npu1```, I will get the error I reported yesterday.  I need to rewrite the sequence function but don't know how to do it. Could you help me with it? Thanks!

Case 1:
```python
dev = AIEDevice.npu1_1col
/*************************/
ShimTile = tile(0, 0)
ComputeTile = tile(0, 2)

# To/from AIE-array data movement
@runtime_sequence(tensor_ty, tensor_ty)
def sequence(A, C):
  npu_dma_memcpy_nd(metadata=of_in, bd_id=1, mem=A, sizes=[1, 1, 1, N], issue_token=True)
  npu_dma_memcpy_nd(metadata=of_out, bd_id=0, mem=C, sizes=[1, 1, 1, N], issue_token=True)
  dma_wait(of_in, of_out)
```
The following is the  result: 
[...]

PR #1971, opened 2024-12-12
Remove (X, Y) coordinates from NpuDmaMemcpyNdOp

pvasireddy-amd:
Issue #1964 
Requires #1997

This PR fixes an issue where the X and Y input attributes of the `NpuDmaMemcpyNdOp` given through the python bindings were always (0, 0), which may not be the coordinates of the Shim tile that the operation will be mapped to (and were not intended to represent that, by design). Instead, the lowering uses the metadata input of the `NpuDmaMemcpyNdOp` to find the corresponding `ShimDMAAllocationOp`. As this operation has the column coordinate (and the row coordinate is implicit for Shim tiles) this PR removes the X and Y coordinates from the `NpuDmaMemcpyNdOp` altogether and fully relies on the metadata to find the information.


pvasireddy-amd:
Debug status:
1. One error is related to changing device name from npu1_1col to npu1 to perform a passthrough from ShimTile (0,0) to ComputeTile (0,2). This results in `'aiex.npu.dma_memcpy_nd' op Unsupported tile type at (0, 0). Must be ShimNOC, Mem or Core.` error. This is because of using `getX()` and `getY()` instead of column and row information which is obtained from the metadata of `dma_memcpy_nd` in this [line](https://github.com/Xilinx/mlir-aie/blob/aff6da32b03c2e236806e90e77eff4f553be054b/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L395). Tile (0,0) is considered as ShimPLTile when entire NPU is used. This check needs to be moved.
2. The second case is changing the ShimTile and ComputeTile to any other column other than zeroth column and using appropriate device name for the same passthrough example. This results in runtime error. One temporary fix is to declare at least one tile (can be ShimTile or ComputeTile with empty body) in the zeroth column. However, this doesn't address the issue completely.

lib/Dialect/AIEX/Transforms/AIECtrlPacketToDma.cpp
```diff
@@ -125,7 +125,7 @@ struct AIECtrlPacketToDmaPass : AIECtrlPacketToDmaBase<AIECtrlPacketToDmaPass> {
 
               StringRef metadata = builder.getStringAttr(shimDmaAllocName);
               builder.create<NpuDmaMemcpyNdOp>(
-                  builder.getUnknownLoc(), 0, 0, newBlockArg,
+                  builder.getUnknownLoc(), newBlockArg, SmallVector<Value>{},
```
erwei-xilinx, line 128: There seems to be one extra redundant `SmallVector<Value>{}` argument passed to the builder.

AndraBisca: Nice catch! Thank you Erwei! :)

test/dialect/AIEX/bad_npu_nd.mlir
```diff
@@ -9,26 +9,29 @@
 //
 //===----------------------------------------------------------------------===//
 
-// RUN: aie-opt --split-input-file --verify-diagnostics %s
+// RUN: aie-opt --split-input-file --aie-dma-to-npu --verify-diagnostics %s
```
fifield, line 12: Is the pass added here just to run verification?  Why the move of the verifier from the dialect to the pass?

pvasireddy-amd: Previously, the `verifyStridesWraps` function was part of `NpuDmaMemcpy::verify()`, but it relied on `getX()` and `getY()` to retrieve row and column values. As a result, the function has been moved into the pass, where it now obtains row and column values from `ShimDmaAllocOp`. This change means that some tests in `bad_npu_nd.mlir` related to sizes and strides, which were previously verifiable directly from the dialect, can no longer be verified this way and now require the pass.

fifield: > Previously, the `verifyStridesWraps` [function](https://github.com/Xilinx/mlir-aie/blob/a60888210c76266cc932d5a0cc922fceef0322f0/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L395) was part of `NpuDmaMemcpy::verify()`, but it relied on `getX()` and `getY()` to retrieve row and column values. As a result, the function has been moved into the [pass](https://github.com/Xilinx/mlir-aie/blob/fd2b3ce22f9039e47317b237b7196f922786935c/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L369), where it now obtains row and column values from `ShimDmaAllocOp`. This change means that some tests in `bad_npu_nd.mlir` related to sizes and strides, which were previously verifiable directly from the dialect, can no longer be verified this way and now require the pass.

But all we've done by moving the verification is hard code the row to zero and get the column from the info op. I don't understand how that is dependent on being inside the pass?  Can't we do the same hardcode + info op lookup in the previous code location?

fifield (approved):

PR #1972, opened 2024-12-13
Cleanup import path for IRON `range_` import

hunhoffe:
Previously, even new-style IRON designs needed to reach into a dialect-specific import path to get a handle to the `range_` function, which is used for looping. I reimported this function with a simpler path, `from aie.iron.controlflow import range_`


jgmelber (approved):

abisca (approved):
I like that this import path more clearly references IRON and is visible even without building the repo.

AndraBisca (approved):

PR #1979, opened 2024-12-16
Minimal Example Showcasing Errors with ObjectFIFOs

iavssw:
Minimal Example Showcasing Errors with ObjectFIFOs @ programming_examples/basic/passthough_kernel. The Passthough Kernel alternative times out when the design is modified to link two objectFIFOs with different sizes with tranformations. This problem only problematic when objectFIFOs with different sizes are linked with tranformations. Without transformation works as expected. Documented the BD design output for reference.


hunhoffe:
Closing this out as part of a backlog cleanup since it's fallen behind main. Please don't hesitate to reopen with a rebase if you'd like to continue — we'd be glad to review.

PR #1983, opened 2024-12-16
Remove explicit AIENGINE and AIEARCH defines from compiler driver

jgmelber:
__AIENGINE__ and __AIEARCH__ do not need to be explicitly defined by the compiler, the compiler flags for the architecture (ie. `--aiearch aie_ml`) handle the architecture selection expression. 

This is a step towards using the general open source [aie_api](https://github.com/Xilinx/aie_api) rather than my fork that [comments out the include for adf headers](https://github.com/jgmelber/aie_api/commit/ced5f060498394cf3520195a43eb83cf3f52ac13).

This PR builds off of #1982: [diff](https://github.com/Xilinx/mlir-aie/compare/kernel-noarch...compiler-noarch)


stephenneuendorffer (approved):
Thanks for doing this!  Can you add a note that older Vitis versions will no longer work? I assume we're only testing with 2024.2 at this point?

fifield:
> Thanks for doing this! Can you add a note that older Vitis versions will no longer work? I assume we're only testing with 2024.2 at this point?

We currently testing with vitis 2023.2 and with the vitis essentials from ryzen ai sw 1.3 (out of date ea version).

jgmelber:
Superseded by #3262, which completes the de-fork of `aie_api` (moves the submodule to stock `Xilinx/aie_api` and suppresses the Vitis-only `<adf.h>` include via `-D__AIE_API_AIE_ADF_HPP__`).

Note on the defines cleanup this PR attempted: verified on hardware that against current upstream `aie_api`, `--aiearch` sets `__AIE_ARCH__` (the selector aie_api actually reads) but does **not** set `__AIENGINE__`. `__AIENGINE__` is still required — upstream `aie.hpp` gates the entire real implementation on it (`#if !__AIENGINE__` pulls in only forward declarations), so dropping it breaks kernels with `no member named 'load_v'`. #3262 therefore keeps `-D__AIENGINE__` and drops only the genuinely-unused `__AIEARCH__`/`__AIE_ARCH__` EXTRA_DEFS from `xchesscc_wrapper`.

PR #1988, opened 2024-12-17
Update pyproject.toml

fifield:
Add new llvm/mlir requirement `nanobind` to whl build dependencies. I'm not sure if this is the correct place to put it.

### DIFF
--- a/utils/mlir_wheels/pyproject.toml
+++ b/utils/mlir_wheels/pyproject.toml
@@ -9,6 +9,7 @@ requires = [
     "numpy",
     "dataclasses",
     "mlir-native-tools",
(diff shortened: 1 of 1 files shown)


jgmelber:
Is this also in the requirements.txt? I suppose if it is in the wheels that covers the dependency, so maybe that isn't necessary...

fifield:
> Is this also in the requirements.txt? I suppose if it is in the wheels that covers the dependency, so maybe that isn't necessary...

I think `nanobind` will be required for mlir-aie's cmake even if the bindings aren't ported to use nanobind, because it uses mlir's `mlir_configure_python_dev_packages`

PR #1991, opened 2024-12-18
Update mlir whl setup.py

fifield:
fix for aarch64 build


fifield:
https://github.com/Xilinx/mlir-aie/actions/runs/12394052492

PR #2000, opened 2024-12-20
Get columns from device rather than tiles

jgmelber:


jgmelber:
This is a more complete fix for the bug fixed in #1997

PR #2005, opened 2025-01-07
Use target model and dynamically create wrapper classes in iron device.py

fifield:
Two updates to iron device.py:
* use target model to query number of rows and columns instead of hard coding
* create device wrapper classes automatically by iterating AIEDevice enum instead of typing boilerplate

### DIFF
--- a/python/iron/device/device.py
+++ b/python/iron/device/device.py
@@ -65,7 +65,7 @@ class Device(Resolvable):
         self._tiles: list[list[Device.__DeviceTile]] = []
 
         # Create all "physical" tiles belonging to the device at initialization to
-        # ensure only one "physical" tile object is every created corresponding to the same
+        # ensure only one "physical" tile object is ever created corresponding to the same
         # coordinates.
         tm = get_target_model(device)
         for c in range(tm.columns()):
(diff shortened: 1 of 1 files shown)


python/iron/device/device.py
```diff
@@ -128,116 +129,59 @@ class NPUBase(Device):
     * The 2nd+ tiles in each column are compute tiles
     """
 
-    def __init__(self, cols: int, rows: int) -> None:
+    def __init__(self, device: AIEDevice) -> None:
         """Initialize a device based on numbers of rows and columns.
```
AndraBisca, line 133: We could remove "based on number of rows and columns"

python/iron/device/device.py
```diff
@@ -53,32 +55,31 @@ def op(self, op: TileOp):
...
+            device (AIEDevice): aie device
         """
-        self._cols = cols
-        self._rows = rows
+        self._device = device
         self._tiles: list[list[Device.__DeviceTile]] = []
 
         # Create all "physical" tiles belonging to the device at initialization to
         # ensure only one "physical" tile object is every created corresponding to the same
         # coordinates.
```
AndraBisca, line 69: Not directly related to changes introduced by this PR but it caught my eye: "ever" instead of "every"?

AndraBisca (approved):
Aside from two small comments on the documentation, LGTM!

jgmelber (approved):

PR #2009, opened 2025-01-08
Change WireBundle::Ctrl to WireBundle::Control. Decouple WireBundle and StrmSwPortType.

fifield:
This changes replaces the abbreviated `WireBundle::Ctrl` with `WireBundle::Control` and removes an implicit coupling between the `WireBundle` enum names and the aie-rt `StrmSwPortType` enum names.

Previously the xaie target (e.g. `aie-translate --aie-generate-xaie`) used the TableGen'd `wireBundleStringify` functions to emit `StrmSwPortType` enums directly as C++ source code. This PR replaces uses of `wireBundleStringify` in `AIETargetXAIEV2` with a new utility function. This corrects a potential bug related to the `PL` port and allows the renaming of the control port.


fifield:
replaced by #2010

stephenneuendorffer (commented):
While I'm normally very much in favor of avoiding abbreviations, I'm not 100% on board with this. The WireBundle names come from the architecture and at some point it's not good to have that inconsistency.  But I guess there has to be some friction somewhere, the only question is where is the least confusing place for it to be?

PR #2010, opened 2025-01-08
Change WireBundle::Ctrl to WireBundle::TileControl. Decouple WireBundle and StrmSwPortType

fifield:
This changes replaces the abbreviated `WireBundle::Ctrl` with `WireBundle::TileControl` and removes an implicit coupling between the `WireBundle` enum names and the aie-rt `StrmSwPortType` enum names.

Previously the xaie target (e.g. `aie-translate --aie-generate-xaie`) used the TableGen'd `wireBundleStringify` functions to emit `StrmSwPortType` enums directly as C++ source code. This PR replaces uses of `wireBundleStringify` in `AIETargetXAIEV2` with a new utility function. This corrects a potential bug related to the `PL` port and allows the renaming of the control port.


stephenneuendorffer:
While I'm normally very much in favor of avoiding abbreviations, I'm not 100% on board with this. The WireBundle names come from the architecture and at some point it's not good to have that inconsistency. But I guess there has to be some friction somewhere, the only question is where is the least confusing place for it to be?

fifield:
> While I'm normally very much in favor of avoiding abbreviations, I'm not 100% on board with this. The WireBundle names come from the architecture and at some point it's not good to have that inconsistency. But I guess there has to be some friction somewhere, the only question is where is the least confusing place for it to be?

As far as I know the `Ctrl` name comes from aie-rt `StrmSwPortType` enum. I see the port referred to as "tile_control" in the arch spec. Maybe that would be better?

stephenneuendorffer:
I think it came from the AIE1 arch spec.  TileControl indeed might be better?  Given that it is itself inconsistent, I remove my objection :)

fifield:
> I think it came from the AIE1 arch spec. TileControl indeed might be better? Given that it is itself inconsistent, I remove my objection :)

Updated to TileControl to be more....consistent.

PR #2025, opened 2025-01-24
[asm] Specialize tile SSA names according to the tile kind

keryell:
Specialize the SSA value names of the tiles according to their real functions to make text assembly clearer, like:
    %tile_1_3 = aie.tile(1, 3)
    %shim_noc_tile_0_0 = aie.tile(0, 0)
    %mem_tile_0_1 = aie.tile(0, 1)


keryell:
This is the followup of Erwei's presentation today showing thousands of lines of MLIR-AIE generated from MLIR-AIR. :smile:

erwei-xilinx (approved):

stephenneuendorffer (approved):

PR #2028, opened 2025-01-28
[CI] test rollback of setup-cpp action

fifield:
Failures like https://github.com/Xilinx/mlir-aie/actions/runs/13007517333/job/36302736888?pr=2026 are caused by a problem with the setup-cpp action. `v1` of that action appears to track `main` and there have been recent breaking changes.  This PR fixes the action to a specific commit to avoid using setup-cpp `main`.


fifield:
https://github.com/aminya/setup-cpp/issues/332

PR #2038, opened 2025-02-06
Update downgrade_ir_for_chess after llvm update

fifield:
https://github.com/llvm/llvm-project/pull/123181 replaces `nocapture` with `capture(none)`, which causes a problem for chess, giving errors like:
```
input.llchesshack.ll:74:52: error: expected ')' at end of argument list
declare <8 x float> @llvm.masked.load.v8f32.p0(ptr captures(none), i32 immarg, <8 x i1>, <8 x float>) #1
                                                   ^
```
For example https://github.com/Xilinx/mlir-aie/actions/runs/13077904662/job/36494451051

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -384,6 +384,7 @@ def downgrade_ir_for_chess(llvmir_chesslinked):
             "memory(argmem: write, inaccessiblemem: write)",
             "inaccessiblemem_or_argmemonly writeonly",
         )
+        .replace("captures(none)", "nocapture")
     )
     return llvmir_chesslinked
 


erwei-xilinx (approved):

keryell (approved):

stephenneuendorffer (approved):

keryell:
Does this depend on the Vitis version by the way? Do we assume that MLIR AIE will work only for a specific version of Vitis? :thinking:

fifield:
> Does this depend on the Vitis version by the way? Do we assume that MLIR AIE will work only for a specific version of Vitis? 🤔

Yes this is a good point. For new enough versions of chess `downgrade_ir_for_chess` does not need to be called. I don't know which, if any, versions of Vitis contain such versions of chess.

PR #2048, opened 2025-02-14
Quick install script cleanups

jgmelber:
Remove llvm/mlir installation from the script as it is not needed. 
Use pip install for Peano. 
Update the `env_setup.sh` for changes

TODO:
- [x] Update READMEs for these changes
- [x] Fix GHA workflows


jgmelber:
https://github.com/Xilinx/mlir-aie/blob/90f92bdfa7b867956286b638216f60d66e90ed12/.github/workflows/buildAndTestRyzenAI.yml#L161-L166

We need the build directory to run the lit tests, I will have to patch the workflow since I stop installing LLVM from the milr wheel in the `quick_setup.sh` script.

jackl-xilinx (approved):

PR #2052, opened 2025-02-19
Update build from wheels script to download the mlir wheel

jgmelber:


fifield:
Is this one tested anywhere?

jgmelber:
> Is this one tested anywhere?

I don't think it is. It exists for documentation and convenience...but it should be tested

jgmelber:
> > Is this one tested anywhere?
> 
> I don't think it is. It exists for documentation and convenience...but it should be tested

I stand corrected. Thanks @fifield

erwei-xilinx (approved):

PR #2059, opened 2025-02-21
Refactor host test code and Makefile to help prevent buffer mismatch hangs

jackl-xilinx:
This PR tries to refactor the host test code and Makefile to put the buffer size used in the testbench (test.cpp) and design_alt.py in one place (Makefile -> CMakeList.txt) so that we don't see program hangs from buffer mismatches. This uses an testbench wrapper xrt_test_wrapper.h (we'll need one for 1 input, 1 output as well) to simplify for users what they need to edit to change the testbench to their own design. This is reduced to two things.
* data type in test.cpp
* buffer sizes (in bytes) in Makefile

They would also modify test.cpp to change the init and functional verification functions but this should prevent mismatch in buffer sizes as the buffer size in Makefile is passed to the desgin_alt.py as well as the test.cpp.

### DIFF
--- a/programming_examples/basic/vector_scalar_mul/CMakeLists.txt
+++ b/programming_examples/basic/vector_scalar_mul/CMakeLists.txt
@@ -30,7 +30,9 @@ else()
     set(XRT_LIB_DIR C:/Technical/xrtNPUfromDLL CACHE STRING "Path to xrt_coreutil.lib")
 endif()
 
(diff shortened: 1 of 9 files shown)


jackl-xilinx:
I guess they can still get the data types wrong but that should break the data movement since size in bytes is used as the main buffer size measurement.

jgmelber (approved):

PR #2066, opened 2025-02-24
Repeat Bds

AndraBisca:
This PR changes the lowering of ObjectFifo repeat_count:
- previously the lock values of locks used by the BDs were adjusted to allow BDs to execute multiple times,
- with this PR, the locks are adjusted as before however the BD chain is unrolled to achieve the repeat, such that each BD only executes once; this in particular handles a race condition in the ping-pong case,
- additionally, there was a race condition in the join and distribute patterns; this PR solves that by generating separate pairs of locks for each input channel for the join and each output channel for the distribute,
- finally, this PR removes the generation of the repeat_count attribute in the `DMAStartOp`.


AndraBisca:
I commented out trace generation in programming_examples/ml/relu/ as it would fail sometimes in the CI, and I wasn't able to reproduce the problem locally. I made #2131 to track this.

PR #2075, opened 2025-02-28
Remove aie-only-generate-npu aiecc option, cleanup aie-generate-cdo use

fifield:
* Remove aiecc `aie-only-generate-npu` option. Other options like `--no-compile` can be used instead.
* Rename aiecc `--aie-generate-npu` to `--aie-generate-npu-insts`
* Update tests and examples to use aiecc `--aie-generate-xclbin` instead of `--aie-generate-cdo` where an xclbin is what's desired
* New aiecc options `--aie-generate-pdi` and `--pdi-name` for when pdi output is wanted, not cdo or xclbin. e.g. with rocr runtime

### DIFF
--- a/programming_examples/basic/dma_transpose/Makefile
+++ b/programming_examples/basic/dma_transpose/Makefile
@@ -43,9 +43,9 @@ build/aie.mlir: ${srcdir}/${aie_py_src}
 
 build/final.xclbin: build/aie.mlir
 	mkdir -p ${@D}
-	cd ${@D} && aiecc.py --aie-generate-cdo --no-compile-host --xclbin-name=${@F} \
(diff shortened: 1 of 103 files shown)


jgmelber:
This is fine to remove, the `only-generate-npu` flag was a start towards infrastructure that could be used for JIT but isn't really exercised by anyone AFAIK.

fifield:
> This is fine to remove, the `only-generate-npu` flag was a start towards infrastructure that could be used for JIT but isn't really exercised by anyone AFAIK.

the idea was to make it uniform: `--aie-generate-<thing>` and `--<thing>-name` and let aiecc figure out what steps to run. The purpose of `only-generate-npu` was to create an early exit to avoid running steps.

PR #2102, opened 2025-03-05
Set default ENABLE_RTTI=true for nightly builds

erwei-xilinx:
Minor fixup: Short commit hash for user-specified aie commit.

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -35,6 +35,7 @@ env:
   DEBIAN_FRONTEND: noninteractive
   VITIS: /opt/ryzen_ai-1.3.0/vitis_aie_essentials
   XILINXD_LICENSE_FILE: /opt/xilinx/Xilinx.lic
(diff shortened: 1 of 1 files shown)


fifield:
For my understanding, what is the motivation and implication of having rtti on? Usually llvm/mlir is fairly opinionated about having it off, since it has its own.

erwei-xilinx:
> For my understanding, what is the motivation and implication of having rtti on? Usually llvm/mlir is fairly opinionated about having it off, since it has its own.

My understanding is that our air or aie projects never used any rtti-specific features, so it doesn't make any differences for the two projects themselves; it makes a difference when they get built together with some other mlir projects, therefore the rtti/no-rtti wheel variants. Would be interested in hearing opinions on whether this is correct or not.

stephenneuendorffer:
I believe RTTI is required by the python bindings.

erwei-xilinx:
> I believe RTTI is required by the python bindings.

I see, good to know. Thanks!

PR #2103, opened 2025-03-05
Find NPU2 and set to default if Strix in Makefiles

jgmelber:


jgmelber:
This likely means the "non Strix" `.lit` files will also be "Strix" tests on Strix devices. Either we can remove the Strix specific versions and rely on the CI for both architectures or hardcode `npu` for the "non Strix" `.lit` tests. @fifield

jgmelber:
> This likely means the "non Strix" `.lit` files will also be "Strix" tests on Strix devices. Either we can remove the Strix specific versions and rely on the CI for both architectures or hardcode `npu` for the "non Strix" `.lit` tests. @fifield

This will only be true for the `examples` workflow

jgmelber:
> > This likely means the "non Strix" `.lit` files will also be "Strix" tests on Strix devices. Either we can remove the Strix specific versions and rely on the CI for both architectures or hardcode `npu` for the "non Strix" `.lit` tests. @fifield
> 
> This will only be true for the `examples` workflow

Actually, no, we use the wheels but not the `quick_setup.sh` script that sets the environment variable based on the current device architecture reported by `xrt-smi`.

This is fine.

PR #2106, opened 2025-03-06
Add worker stack size to python bindings

### LINKED ISSUE #2045: Add python bindings for stack size in core definition
Currently in mlir, we can pass in a parameter to the core object to specify stack size as follows:

```
AIE.core(%tile) {
      AIE.end
} { stackSize = 2048 : i32, elf_file = "core_33.elf" }
```

But this option doesn't exist for the python bindings. This should be added.

L-roro:
Closes #2045

Note that the order of the positional arguments does not correspond to the order used in mlir. I did not change this to avoid breaking all previous tests using positional arguments when declaring a core.

### DIFF
--- a/python/dialects/aie.py
+++ b/python/dialects/aie.py
@@ -245,10 +245,11 @@ Device = DeviceOp
 
 class Core(CoreOp):
     # Until https://github.com/llvm/llvm-project/pull/73620 gets figured out.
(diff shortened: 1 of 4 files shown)


jgmelber (approved):

AndraBisca (approved):
Looks good!

PR #2119, opened 2025-03-07
Add per tile allocation scheme

### LINKED ISSUE #1985: Allow choosing allocation scheme per tile
Right now we are able to choose an allocation scheme (between basic-sequential and bank-aware) using the --alloc-scheme flag in the aiecc utility. It would be interesting to have a flag attached to the CoreOp, similar to the stack_size flag.

L-roro:
Closes #1985

Instead of adding this possibility per core, I added it per tile, since it would be legitimate to do this on a tile that does not have a coreOp. 

To avoid misuse, I added a check to TileOp::isVerify to make sure that no allocation scheme is defined for a shim tile.

I updated the python bindings to be able to take advantage of this new feature. It may be passed as an optional parameter to Tiles or even workers in the IRON API. Additionally I removed the default flag for the allocator for the aiecc compiler to make it more transparent to the user (now, if no allocator is chosen, the behavior remains the same but a warning will be generated by the buffer placement pass if a tile has no associated placement algorithm). Specifying an allocation scheme for a worker overrides the flag given to aiecc.

Unfortunately, these modifications allow for a very unfortunate possibility:
my_worker = Worker(None, placement=Tile(1, 2, allocation_scheme="bank-aware"), allocation_scheme="basic-sequential")

In this case, the worker allocation scheme overrides the tile allocation scheme as indicated in the doc. Maybe throwing an error would be wiser.


fifield:
An important use case is to allocate buffers in specific locations and have the allocator not overwrite those addresses. Currently this is accomplished by setting the address attribute on the buffer and ensuring that the allocator does not overwrite the address by passing `-alloc-scheme=basic-sequential` to aiecc. The bank aware allocator does not respect buffer addresses set by the user so the default allocator can't be used. With the proposed change, will I need to set the address attribute on the buffers as well as the allocation_scheme attribute on the tile to which the buffer belongs?

lromero434:
With the proposed change, you will still be able to pass the `-alloc-scheme=basic-sequential` to aiecc and set the address attribute on the buffer to avoid the bank aware allocation and there is no need to specify it on the tile. On the previous commit I added an additional test as an example of this.

In case it was not clear enough: There is a hierarchy to the allocations schemes and all can be used simultaneously. Going from highest to lowest priority:
IRON:
-Worker allocation scheme
-Tile allocation scheme

MLIR:
-tile
-aiecc flag

This hierarchy allows to indicate a default allocation scheme for all tiles through the aiecc flag, that may be overridden later for a specific tile in the mlir.

The proposed changes add on top of previous behavior and do not modify it with one exception. Previously, if no aiecc flag was given, the default allocation was bank-aware. Now, if no flag is given and no allocation scheme is given for a tile, the allocator will issue a warning and try first bank-aware followed by basic sequential in case of an error.

fifield:
>Now, if no flag is given and no allocation scheme is given for a tile, the allocator will issue a warning and try first bank-aware followed by basic sequential in case of an error.

Do you mean it is required to provide an aiecc flag (or tile attribute?) otherwise the user experiences a warning message?

L-roro:
Yes, that is the behavior on this change. I believe that was the intended behavior in the AIEAssignBuffers.cpp code before I touched it, but it never happened because there always was a default flag that was being passed to it. Is this warning undesirable?

test/python/tile_allocation_scheme.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (C) 2023, Advanced Micro Devices, Inc.
```
fifield, line 1:
```suggestion
# Copyright (C) 2025, Advanced Micro Devices, Inc.
```

fifield (approved):

AndraBisca (approved):
LGTM!

PR #2130, opened 2025-03-14
Expose burst length

L-roro:
Expose burst size parameter to npu_dma_memcpy_nd and shim_dma_single_bd_task in python and remove hardcoded configuration of burst length.

The modified operations and passes to accomplish this are illustrated below (marked in red):
exposing_burst_length

As shown in the diagram, DMABDOp must be modified for this behavior. It belongs to the AIE dialect and not AIEX. To avoid undesired behavior, DMABDOps need checks to avoid specifying the burst length on tiles unrelated to ShimNOCs. This has been added. Unfortunately, this check has to be added twice because DMABDOps' verification waits until they are enclosed in AIE operations to be executed and does not recognize AIEX operations. Because of this, all DMABDOps generated from the AIEX dialect (which includes all chains) are currently never verified. I imagine this is a problem that will be addressed later down the line, but I believe this might be undesirable behavior even during development ([link](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/IR/AIEDialect.cpp#L1830)).

NpuControlPackets do not have this new parameter exposed.

There is an unused python function writebd_shimtile (and _npu_writebd_shimtile) in aiex.py that has a seemingly outdated (previous to my modifications) reimplementation of the behavior that has been modified in this proposal. I am unsure about whether it should be removed or updated.

This proposal also adds documentation to the modified python functions, updates the programming guide to reflect the changes and modifies shim_dma_single_bd_task to mimic the behavior of npu_dma_memcpy_nd when no sizes are given.


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -890,6 +891,7 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
         DefaultValuedOptionalAttr<AIEI32Attr, "0">:$pad_value,
         OptionalAttr<AIEI32Attr>:$bd_id,
         OptionalAttr<PacketInfoAttr>:$packet,
+        DefaultValuedOptionalAttr<AIEI32Attr, "0">:$burst_length,
```
jgmelber, line 894: The default should be the maximum burst length

L-roro: A value of 0 is interpreted as the maximum burst length for the chosen architecture (either 256 or 512 bytes for the currently supported architectures). This is explained in the documentation and the available values can be changed in this newly defined [method](https://github.com/Xilinx/mlir-aie/pull/2130/files#diff-4dbaf32e6067f9c4e74657de29003aa4e8a27f8676148111624530b231fdf17dR273). Is this choice misleading?

jgmelber (approved):

jackl-xilinx (approved):

PR #2132, opened 2025-03-14
Rename programming examples from _alt to _placed

L-roro:


AndraBisca (approved):
There seems to be a reference to _alt left in:
./section-2/section-2d/DMATasks.md:All programming examples have an `*_alt.py` version that is written using DMA task operations.

Otherwise looks good to me!

_Edit: just noticed that only the basic/ examples have been renamed; will review again when the others are updated_

AndraBisca (approved):

PR #2145, opened 2025-03-20
Update aiert based targets to use burst length from target model

fifield:
This PR updates the `AIRRTControl` class to use the shim burst length api introduced in https://github.com/Xilinx/mlir-aie/pull/2130. The helper functions to compute burst length are also updated to live in AIEDialect.cpp and to return length in bytes (for interfacing with aie-rt) as well the register encodings.

Mixed in is a minor cleanup of `namespece xilinx::AIE` blocks in AIEDialect.cpp which revealed some declaration errors.

### DIFF
--- a/lib/Dialect/AIE/IR/AIETargetModel.cpp
+++ b/lib/Dialect/AIE/IR/AIETargetModel.cpp
@@ -291,9 +291,8 @@ bool AIE1TargetModel::isLegalTileConnection(int col, int row,
 }
 
 std::vector<std::pair<uint32_t, uint32_t>>
-AIE1TargetModel::getBurstEncodingsAndLengths() const {
-  return {std::pair(0x00000000, 64), std::pair(0x40000000, 128),
-          std::pair(0x80000000, 256)};
+AIE1TargetModel::getShimBurstEncodingsAndLengths() const {
+  return {std::pair(0, 64), std::pair(1, 128), std::pair(2, 256)};
 }
 
 ///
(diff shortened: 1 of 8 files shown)


lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -291,9 +291,8 @@ bool AIE1TargetModel::isLegalTileConnection(int col, int row,
 }
 
 std::vector<std::pair<uint32_t, uint32_t>>
-AIE1TargetModel::getBurstEncodingsAndLengths() const {
-  return {std::pair(0x00000000, 64), std::pair(0x40000000, 128),
-          std::pair(0x80000000, 256)};
+AIE1TargetModel::getShimBurstEncodingsAndLengths() const {
+  return {std::pair(0, 64), std::pair(1, 128), std::pair(2, 256)};
```
fifield, line 295: The encoding would not have been correct for AIE1 although no code path for AIE1 uses it.

lib/Dialect/AIE/IR/AIETargetModel.cpp
```diff
@@ -291,9 +291,8 @@ bool AIE1TargetModel::isLegalTileConnection(int col, int row,
 }
 
 std::vector<std::pair<uint32_t, uint32_t>>
-AIE1TargetModel::getBurstEncodingsAndLengths() const {
-  return {std::pair(0x00000000, 64), std::pair(0x40000000, 128),
-          std::pair(0x80000000, 256)};
```
fifield, line 296: This encoding is not correct for AIE1

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -588,7 +559,8 @@ struct WriteBdToBlockWritePattern : OpConversionPattern<NpuWriteBdOp> {
       words[3] |= op.getD0Stride() & 0xfffff;
 
       // DMA_BDX_4
-      words[4] = burstLengthEncoding; // burst length;
+      words[4] = (getShimBurstLengthEncoding(tm, op.getBurstLength()) & 0x3)
+                 << 30;
```
fifield, line 563: I've moved the left shift out of the "encoding" returned by the target model into this code because this is where all the other mask and shift information is encoded in code.

stephenneuendorffer (commented):
Got tests?

fifield:
> Got tests?

I've updated a cdo target test to check that setting the burst length has the desired result. The same test already checked that the default behavior wasn't changed for cdo (or xclbin) target.

PR #2152, opened 2025-03-22
Emit transaction binary spec 0.1

jgmelber:
Swaps the transaction binary generation from specification 1.0 to 0.1 to align with other tools

### DIFF
--- a/lib/Targets/AIETargetNPU.cpp
+++ b/lib/Targets/AIETargetNPU.cpp
@@ -69,7 +69,7 @@ void appendSync(std::vector<uint32_t> &instructions, NpuSyncOp op) {
 
 void appendWrite32(std::vector<uint32_t> &instructions, NpuWrite32Op op) {
 
-  auto words = reserveAndGetTail(instructions, 3);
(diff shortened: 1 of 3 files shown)


lib/Targets/AIETargetNPU.cpp
```diff
@@ -196,9 +203,11 @@ xilinx::AIE::AIETranslateNpuToBinary(ModuleOp module,
...
 
   // setup txn header
-  uint8_t major = 1;
-  uint8_t minor = 0;
-  uint8_t devGen = 3;
+  uint8_t major = 0;
+  uint8_t minor = 1;
+  uint8_t devGen = 3;                      // NPU (PHX HWK)
+  if (llvm::isa<AIE::NPU2TargetModel>(tm)) // FIXME
+    devGen = 4;                            // NPU2 (STX KRK)
```
fifield, line 210: What needs to be fixed here?

jgmelber: Fixed in #2156

PR #2128, opened 2025-03-12
Use TXN bin rather than text format 

jgmelber:
- [x] programming examples
- [x] programming guide
- [x] npu-xrt tests

Control packet tests still use `.txt` format rather than `.bin`

PR #2168, opened 2025-03-31
Change aie-translate runtime sequence default from text to binary

fifield:
Follow on to https://github.com/Xilinx/mlir-aie/pull/2128. The next step could be to remove the text support entirely.


jgmelber:
> Follow on to #2128. The next step could be to remove the text support entirely.

I am a fan of completely deprecating text, I think most users have bumped

jgmelber (approved):

fifield:
> > Follow on to #2128. The next step could be to remove the text support entirely.
> 
> I am a fan of completely deprecating text, I think most users have bumped

Some tests depend on text format, but we could find another way to check the output.

PR #2174, opened 2025-04-01
Bump XDNA-driver and NPU find logic

jgmelber:
Done

### DIFF
--- a/utils/env_setup.sh
+++ b/utils/env_setup.sh
@@ -37,6 +37,15 @@ if [[ $PEANO_INSTALL_DIR == "" ]]; then
   export PEANO_INSTALL_DIR="$(pip show llvm-aie | grep ^Location: | awk '{print $2}')/llvm-aie"
 fi
 
+NPU=`xrt-smi examine | grep -E "NPU Phoenix|NPU Strix|NPU Strix Halo|NPU Kracken"`
+# Check if the current environment is NPU2
+# npu4 => Strix, npu6 => Kracken
+if echo "$NPU" | grep -qiE "NPU Strix|NPU Strix Halo|NPU Kracken"; then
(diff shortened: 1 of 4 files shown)


jgmelber:
@fifield this will require updates to the drivers on the runners

utils/env_setup.sh
```diff
@@ -37,6 +37,15 @@ if [[ $PEANO_INSTALL_DIR == "" ]]; then
   export PEANO_INSTALL_DIR="$(pip show llvm-aie | grep ^Location: | awk '{print $2}')/llvm-aie"
 fi
 
+NPU=`/opt/xilinx/xrt/bin/xrt-smi examine | grep -E "NPU Phoenix|NPU Strix|NPU Strix Halo|NPU Kracken"`
```
fifield, line 40: Should we be sourcing the setup script here instead?  To put xrt-smi in PATH...

jgmelber, line 40:
```suggestion
XRTSMI=`which xrt-smi`
if ! test -f "$XRTSMI"; then 
  source /opt/xilinx/xrt/setup.sh
fi
NPU=`/opt/xilinx/xrt/bin/xrt-smi examine | grep -E "NPU Phoenix|NPU Strix|NPU Strix Halo|NPU Kracken"`
```

jgmelber: Suggestion below

fifield (approved):
I'd be ok with merging as is, because the runners are already updated and this PR is needed to make them usable by other PRs.

jgmelber:
> I'd be ok with merging as is, because the runners are already updated and this PR is needed to make them usable by other PRs.

I'll move the todos to a follow up PR

PR #2190, opened 2025-04-09
Add support for trace to iron python.

jackl-xilinx:


jackl-xilinx:
@jgmelber No, there isn't as I only tested it locally. That's a good idea. I'll create a few test cases so they're captured. I also would need to add some documentation if this PR is approved for trace support in the near term.

jackl-xilinx:
@jgmelber I added an example that adds additional trace config param in the ml/eltwise_add design [here](https://github.com/Xilinx/mlir-aie/pull/2190/files#diff-f6a5d82445012ea35aa047fca101b9bfadaa87c3c38f1720bd2e1bf0de716ec5).

AndraBisca (approved):

PR #2199, opened 2025-04-11
Slightly reorganize npu TargetModels to prevent missing overrides

fifield:
While porting some code to strix I encountered some odd compilation errors using aiecc, and it was insisting on compiling some things for aie2, which led me to the discovery of this:
```sh
$ echo 'aie.device(npu2_4col){aie.end}' | aie-translate -aie-generate-target-arch
AIE2
```
which is incorrect. While this one was correct:
```sh
$ echo 'aie.device(npu2){aie.end}' | aie-translate -aie-generate-target-arch
AIE2p
```
The error is from missing overrides on the npu2 virtual device AIETargetModel classes. The way the AIETargetModel hierarchy is set up, the NPU base classes are generic AIE2 devices, and the npu1 and npu2 device classes inherit from these same base classes. This reduces code because npu1 and npu2 are similar. But this also means any npu2 leaf class must separately implement any npu2/aie2p overrides. This does not reduce code and can lead to missing overrides causing the error(s) I observed.

This PR attempts to fix the situation by adding a npu2 specific `BaseNPU2TargetModel` base class for implementing aie2p overrides in one place. And the npu1 classes are renamed from "NPU" to "NPU1".

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIETargetModel.h
+++ b/include/aie/Dialect/AIE/IR/AIETargetModel.h
@@ -574,9 +574,9 @@ public:
   }
 };
 
-class BaseNPUTargetModel : public AIE2TargetModel {
+class BaseNPU1TargetModel : public AIE2TargetModel {
 public:
-  BaseNPUTargetModel(TargetModelKind k) : AIE2TargetModel(k) {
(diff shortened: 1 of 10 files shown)


jgmelber (approved):

AndraBisca:
Thank you for the fix and for the explanation!

PR #2201, opened 2025-04-13
Add IRON single-source Python enablement modules

mawad-amd:
This PR introduces core modules to enable single-source Python programming with IRON, similar in style to Triton. Specifically, it includes:

1. A JIT compilation module
3. A PyTorch-like Tensor class for NPUs
5. An update to the vector-vector add example to use the new flow

### JIT Compilation:

To support the unplaced variant, I added to the JIT decorator an argument to specify whether the kernel is using explicit or implicit placement. For a placed kernel, the JIT compiled kernel looks like:

```python
@iron.jit(is_placed=True)
def my_kernel():
   # code ..

# launched via
my_kernel(args..)
```

For an unplaced version:

```python
@iron.jit(is_placed=False)
def my_kernel():
   # Code ...
   return Program(dev, rt).resolve_program(SequentialPlacer())

# launched via
my_kernel(args..)
```
Some feedback on the unplaced variant would be appreciated. Since the `Program` abstraction requires the device and runtime sequence, I can't move them to the JIT module. Also the Program abstraction builds the MLIR inside the `resolve_program` call. I am not very excited about the kernel returning the context but this is the best I came up with. I _think_ it would look cleaner if the launch looked like:

```python
@iron.jit(is_placed=False)
def my_kernel():
   # Code ...
   # No return...

my_kernel[SequentialPlacer()](args..)
```
But this would require some invasive changes to the `Program` abstraction.

### Tensors for NPU:
PyTorch-like tensor with few initial APIs that follow PyTorch API and semantics. For example:
```python
import aie.iron as iron
random_values = iron.rand((64,), dtype=np.float32, device="npu")
zero_like_random = iron.zeros_like(random_values)
cpu_np_array = zero_like_random.numpy()
```

### Other notes:
I also completely removed the files for VCK5000.


programming_examples/basic/vector_vector_add/run_placed.lit
jgmelber, line 1: Split the npu and npu2 tests to separate lit files so it's easy to tell if one target fails at runtime.

python/iron/jit.py
```diff
@@ -0,0 +1,150 @@
...
+
+import os
+import tempfile
+import hashlib
+import numpy as np
+
+try:
+    import pyxrt as xrt
+except ImportError:
+    print("pyxrt not available; skipping module.")
```
mawad-amd, line 17: @jgmelber : It looks like some runners do not have access to `pyxrt`. I need some guidance on how to gracefully handle that. Looks like I added the first `pyxrt` dependency inside the python directory. 

It also looks like the internal runners do not have `pyxrt` installed. Is that correct or am I doing something wrong here?
https://github.com/Xilinx/mlir-aie/actions/runs/14424671764/job/40452761094#step:4:8382

jgmelber: @fifield as far as I know this shouldn't be an issue. The runners have pyxrt?

jgmelber:
@mawad-amd can you also port the unplaced variant?

jgmelber:
> 1. Check `.lit` file working as intended

I have intentionally broken the tests before to verify this...it can be hard to tell at times if they "just work". :)

mawad-amd:
> > 1. Check `.lit` file working as intended
> 
> I have intentionally broken the tests before to verify this...it can be hard to tell at times if they "just work". :)

Looks like it's running and failing with an error `Exception: Kernel returned ert_cmd_state.ERT_CMD_STATE_ABORT`. I will debug this. https://github.com/Xilinx/mlir-aie/actions/runs/14452090618/job/40527022834?pr=2201#step:4:6013

programming_examples/basic/vector_vector_add/run_placed.lit
```diff
@@ -0,0 +1,7 @@
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// REQUIRES: ryzen_ai, peano 
+//
+// RUN: dev=$(if [ "$NPU2" = "1" ]; then echo npu2; else echo npu; fi); \
+// RUN: %run_on_npu python3 %S/vector_vector_add_placed.py --device=$dev --column=0
```
jgmelber, line 7:
```suggestion
// RUN: %run_on_npu python3 %S/vector_vector_add_placed.py --device=npu --column=0
```

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -7,42 +7,72 @@
...
+    help="Number of elements (default: 32)",
+)
+args = parser.parse_args()
+
+device_map = {
+    "npu": AIEDevice.npu1_1col,
+    "npu2": AIEDevice.npu2_1col,
+    "xcvc1902": AIEDevice.xcvc1902,
+}
+dev = device_map[args.device]
```
jgmelber, line 51: It would be cool if pyxrt would return useful information about the device to set this within Python rather than command line

fifield: > It would be cool if pyxrt would return useful information about the device to set this within Python rather than command line

I was adding similar functionality to the equivalent mlir-air xrt module last week: https://github.com/Xilinx/mlir-air/pull/943/files#diff-1805a0355b6edc61a7102bb4ea5208a5a97374fb4393ce17bdd03a757e267ef3R82-R118, using the logic from lit.cfg.py, to support Strix.

python/iron/jit.py
mawad-amd: I think these actions run on runners without XRT?
https://github.com/Xilinx/mlir-aie/actions/runs/14426235707/workflow#L28
I added a small workaround in [d082a62](https://github.com/Xilinx/mlir-aie/pull/2201/commits/d082a627b9fab3747ba9498171512643ff795b8b) for now..

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
mawad-amd: We could start with something like that. In offline chat Joe mentioned it would be nice to add these to XRT Python bindings.

jgmelber: I am talking to @stsoe about enabling this

jgmelber: Exposing via XRT and PyXRT specifically

mawad-amd:
All green. Thanks Joe for the fix in https://github.com/Xilinx/mlir-aie/pull/2201/commits/cf39c93785dceec2f50b23d6606c0605405055cc. I will finish my todos and we should have PR ready tomorrow.

jgmelber:
> All green. Thanks Joe for the fix in https://github.com/Xilinx/mlir-aie/pull/2201/commits/cf39c93785dceec2f50b23d6606c0605405055cc. I will finish my todos and we should have PR ready tomorrow.

Great! Yeah the NPU2 environment variable isn't set by the GitHub action workflow if I remember correctly.

python/iron/jit.py
fifield: Right, the stock github runners will not have xrt. Sorry I did not look to see which action was hitting problems.

jgmelber:
> I also removed the files for VCK5000.

No problem with this

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -7,32 +7,64 @@
...
+# The JIT-compiled kernel relies on inputs from the command line.
+# Because of that, we need to parse the arguments before the JIT is invoked.
+parser = argparse.ArgumentParser()
+parser.add_argument(
+    "-v", "--verbose", action="store_true", help="Enable verbose output"
+)
+parser.add_argument(
+    "-d",
+    "--device",
+    choices=["npu", "npu2", "xcvc1902"],
```
jgmelber, line 28: OK to deprecate xcvc1902 (vck5000) for this example.
```suggestion
    choices=["npu", "npu2"],
```

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -7,32 +7,64 @@
...
-    else:
-        raise ValueError("[ERROR] Device name {} is unknown".format(sys.argv[1]))
+device_map = {
+    "npu": NPU1Col1(),
+    "npu2": NPU2Col1(),
+    # "xcvc1902": XCVC1902(),
+    # Commented out because of the error:
+    # TypeError: Can't instantiate abstract class XCVC1902 without an
+    # implementation for abstract methods 'get_compute_tiles',
+    # 'get_mem_tiles', 'get_shim_tiles'
```
jgmelber, line 54: OK to deprecate xcvc1902 (vck5000) for this example.
```suggestion
```

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -68,5 +100,34 @@ def core_body(of_in1, of_in2, of_out):
     return Program(dev, rt).resolve_program(SequentialPlacer())
 
 
-module = my_vector_add()
-print(module)
+def main():
```
jgmelber, line 103: Maybe some comments to each line here since it is simplified

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -70,21 +101,45 @@ def core_body():
...
             dma_await_task(out_task)
             dma_free_task(in1_task, in2_task)
 
 
-with mlir_mod_ctx() as ctx:
-    my_vector_add()
-    res = ctx.module.operation.verify()
-    if res == True:
-        print(ctx.module)
+def main():
```
jgmelber, line 115: Comments in here too for each "section"

python/iron/jit.py
```diff
@@ -0,0 +1,177 @@
...
+import hashlib
+import numpy as np
+import pyxrt as xrt
+
+from aie.extras.context import mlir_mod_ctx
+from ..utils.compile import compile_mlir_to_binary
+from ..utils.xrt import read_insts_binary
+
+
+IRON_CACHE_DIR = os.path.expanduser("~/.iron/cache")
```
jgmelber, line 20: This is cool! Maybe a comment explaining the cache functionality

python/iron/jit.py
```diff
@@ -0,0 +1,177 @@
...
+
+    pass
+
+
+def jit(*, is_placed=True, use_cache=True):
+    """
+    Decorator to compile an IRON kernel into a binary to run on the NPU.
+
+    Parameters:
+    - is_placed (bool): Whether the kernel is using explicit or implicit placement Defaults to True.
```
jgmelber, line 124: "implicit" also known as "deferred" placement

python/iron/jit.py
```diff
@@ -0,0 +1,177 @@
...
+        module_hash = hash_module(mlir_module)
+        kernel_dir = os.path.join(IRON_CACHE_DIR, f"{module_hash}")
+        mlir_path = os.path.join(kernel_dir, "aie.mlir")
+
+        # Ensure cache directory exists
+        os.makedirs(kernel_dir, exist_ok=True)
+
+        # Write MLIR to file if not already cached
+        inst_filename = "insts.bin"
+        xclbin_filename = "final.xclbin"
```
jgmelber, line 150: It would be nice to allow users to change these hardcoded defaults. Additionally the "kernel" name that is accessible via aiecc.py

python/utils/compile.py
```diff
@@ -0,0 +1,51 @@
...
+    ]
+
+    try:
+        subprocess.run(
+            cmd,
+            cwd=mlir_dir,
+            check=True,
+            stdout=sys.stdout if debug else subprocess.DEVNULL,
+            stderr=sys.stderr if debug else subprocess.DEVNULL,
+        )
```
jgmelber, line 49: It's possible to call aiecc.py directly without needing to use a subprocess

jgmelber: I will follow up with a link

jgmelber:
> I _think_ it would look cleaner if the launch looked like:
> 
> ```python
> @iron.jit(is_placed=False)
> def my_kernel():
>    # Code ...
>    # No return...
> 
> my_kernel[SequentialPlacer()](args..)
> ```
> 
> But this would require some invasive changes to the `Program` abstraction.

We could take this up in a "Discussion" or through a follow up PR. I do like this proposed syntax as well.

python/utils/compile.py
jgmelber: [Link](https://github.com/Xilinx/mlir-aie/blob/4147c9babe655e4f744bb352e0494fb677ac429b/test/python/aiecc_simple.py#L12)

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
mawad-amd: Added https://github.com/Xilinx/mlir-aie/pull/2201/commits/450790240fe29e9e5ca1479aded004098e018e31

python/iron/jit.py
mawad-amd: Added https://github.com/Xilinx/mlir-aie/pull/2201/commits/450790240fe29e9e5ca1479aded004098e018e31 Let me know if you have more comments.

python/iron/jit.py
mawad-amd: Do you want me to change the argument name? I am open to ideas.

programming_examples/basic/vector_vector_add/vector_vector_add.py
mawad-amd: Added 4507902

python/utils/compile.py
mawad-amd: I tried this and there is one issue. I need to store the xclbin in some output location and hence I set the `cwd`. I can't see an option in the module to achieve that behavior. I also can't pass something like `--xclbin-name=/home/muhaawad/.iron/cache/8b8ec456a41df8f8/final.xclbin` it seems.

python/iron/jit.py
jgmelber: I am thinking about it. The name makes a lot of sense in the placed example and then in your proposed syntax. The current unplaced example returns a "placed" module since the placer is run within `resolve` in the `return` line.

python/utils/compile.py
jgmelber: I don't think I fully understand the limitations. Is there something that can to be added to aiecc.py to help?

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
jgmelber: This can be recorded in an issue and left to a future PR

python/utils/compile.py
mawad-amd: Possibly. I will need to check the code. Can we leave this to different PR? The issue simply is that I can't tell aiecc.py to store the output files inside some directory.

jgmelber: Yes. Track this in an issue referencing this PR. Include the syntax proposal and runtime detection of device type.

jgmelber: @fifield has worked extensively with calling the compiler drivers from Python host code

python/iron/jit.py
mawad-amd: Any proposals for how to do that? I am not currently exposing where the kernel cache is too. The path to storing the (xclbin, inst, mlir) files is simply:
```
~/.iron/cache/<MLIR-module-hash>/
```
So, not only we will need to set the filenames itself, but also some control over the directory too.

Is the reason for that so people can just copy the xclbin and use them via other mechanisms?

jgmelber: People may want a more descriptive rather than generic name for the binaries. Agreed they may want to change the cache location too. Maybe expose as arguments with defaults set the way you have them now?

mawad-amd:
All pending suggestions and improvements are tracked by https://github.com/Xilinx/mlir-aie/issues/2206. If anyone has other thoughts, please leave a comment on the issue. Thanks!

python/utils/compile.py
fifield: > I tried this and there is one issue. I need to store the xclbin in some output location and hence I set the `cwd`. I can't see an option in the module to achieve that behavior. I also can't pass something like `--xclbin-name=/home/muhaawad/.iron/cache/8b8ec456a41df8f8/final.xclbin` it seems.

What was the issue you faced with "--xclbin-name"?  The aiecc run function is a thin wrapper around the command line tool, so you should be able to pass any command line. another example: https://github.com/Xilinx/mlir-air/blob/main/python/air/compiler/aircc/main.py#L578

mawad-amd: I take this back. I just tried it again and it seems to be working. Let me create the PR to avoid using a new subprocess.

PR #2213, opened 2025-04-18
Passing arguments through iron.jit decorator

### LINKED ISSUE #2211: Passing arguments to the JIT'd function
The following would allow passing arguments to the function and optional positional decorator arguments:
```python
import functools

def jit(function=None, is_placed=True, use_cache=True):

    def decorator(fn):

        @functools.wraps(fn)
        def wrapped_function(*args, **kwargs):
            fn(*args, **kwargs) # this would be the compilation / launch phase

        return wrapped_function

    if function:
        return decorator(function)

    return decorator

@jit
def test(v):
    print(v)

@jit(is_placed=False)
def test2():
    print(111)
```

So for the vector add, it'll be something like:
```python
@iron.jit(is_placed=False)
def vector_vector_add(input0, input1, output):
    # infer elements, do error checking etc. using input0, input1, output

def main():
    vector_vector_add(device, input0, input1, output) # we already do that but we need the device too
```

The drawbacks are
1. one additional wrapping / indirection and
2. the dependency on `functools`.

Thoughts @jgmelber, @mawad-amd?

ypapadop-amd:
This PR allows to pass arguments through the `iron.jit` decorator to the called function.

Closes #2211

### DIFF
--- a/programming_examples/basic/vector_vector_add/vector_vector_add.py
+++ b/programming_examples/basic/vector_vector_add/vector_vector_add.py
@@ -9,8 +9,8 @@
 import argparse
 import sys
 import numpy as np
+import aie.iron as iron
 
-from aie import iron
 from aie.iron import ObjectFifo, Program, Runtime, Worker
 from aie.iron.placers import SequentialPlacer
 from aie.iron.device import NPU1Col1, NPU2Col1
--- a/programming_examples/basic/vector_vector_add/vector_vector_add.py
+++ b/programming_examples/basic/vector_vector_add/vector_vector_add.py
@@ -112,8 +112,8 @@ def main():
 
     # Construct two input random tensors and an output zeroed tensor
     # The three tensor are in memory accessible to the NPU
-    input0 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
-    input1 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
+    input0 = iron.randint(0, 100, (args.num_elements,), dtype=np.int32, device=args.device)
+    input1 = iron.randint(0, 100, (args.num_elements,), dtype=np.int32, device=args.device)
     output = iron.zeros_like(input0)
 
     # JIT-compile the kernel then launches the kernel with the given arguments. Future calls
--- a/programming_examples/basic/vector_vector_add/vector_vector_add.py
+++ b/programming_examples/basic/vector_vector_add/vector_vector_add.py
@@ -4,78 +4,29 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
-import numpy as np
-import sys
-import argparse
+# (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
 
+import argparse
+import sys
+import numpy as np
+
+from aie import iron
 from aie.iron import ObjectFifo, Program, Runtime, Worker
 from aie.iron.placers import SequentialPlacer
-from aie.iron.device import NPU1Col1, NPU2Col1, XCVC1902
+from aie.iron.device import NPU1Col1, NPU2Col1
 from aie.iron.controlflow import range_
-
-import aie.iron as iron
-
-# The JIT-compiled kernel relies on inputs from the command line.
-# Because of that, we need to parse the arguments before the JIT is invoked.
-parser = argparse.ArgumentParser()
-parser.add_argument(
-    "-v", "--verbose", action="store_true", help="Enable verbose output"
-)
-parser.add_argument(
-    "-d",
-    "--device",
-    choices=["npu", "npu2"],
-    default="npu",
-    help="Target device",
-)
-parser.add_argument(
-    "-c", "--column", type=int, default=0, help="Column index (default: 0)"
-)
-parser.add_argument(
-    "-n",
-    "--num-elements",
-    type=int,
-    default=32,
-    help="Number of elements (default: 32)",
-)
-args = parser.parse_args()
-
-device_map = {
-    "npu": NPU1Col1(),
-    "npu2": NPU2Col1(),
-}
-device = device_map[args.device]
-
-# Construct two input random tensors and an output zeroed tensor
-# The three tensor are in memory accessible to the NPU
-input0 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
-input1 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
-output = iron.zeros_like(input0)
 
 
 @iron.jit(is_placed=False)
-def vector_vector_add(
-    # input0, input1, output
-):
+def vector_vector_add(device, input0, input1, output):
     if input0.shape != input1.shape:
         raise ValueError(
             f"Input shapes are not the equal ({input0.shape} != {input1.shape})."
         )
-    if input0.dtype != input1.dtype:
-        raise ValueError(
-            f"Input data types are not the same ({input0.dtype} != {input1.dtype})."
-        )
-
     if input0.shape != output.shape:
         raise ValueError(
             f"Input and output shapes are not the equal ({input0.shape} != {output.shape})."
         )
-    if input0.dtype != output.dtype:
-        raise ValueError(
-            f"Input and output data types are not the same ({input0.dtype} != {output.dtype})."
-        )
-
     if len(np.shape(input0)) != 1:
         raise ValueError("Function only supports vectors.")
     num_elements = np.size(input0)
@@ -85,6 +36,15 @@ def vector_vector_add(
             f"Number of elements ({num_elements}) must be a multiple of {n}."
         )
     N_div_n = num_elements // n
+
+    if input0.dtype != input1.dtype:
+        raise ValueError(
+            f"Input data types are not the same ({input0.dtype} != {input1.dtype})."
+        )
+    if input0.dtype != output.dtype:
+        raise ValueError(
+            f"Input and output data types are not the same ({input0.dtype} != {output.dtype})."
+        )
     dtype = input0.dtype
 
     # Define tensor types
@@ -125,10 +85,40 @@ def vector_vector_add(
 
 
 def main():
+    device_map = {
+        "npu": NPU1Col1(),
+        "npu2": NPU2Col1(),
+    }
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", help="Enable verbose output"
+    )
+    parser.add_argument(
+        "-d",
+        "--device",
+        choices=["npu", "npu2"],
+        default="npu",
+        help="Target device",
+    )
+    parser.add_argument(
+        "-n",
+        "--num-elements",
+        type=int,
+        default=32,
+        help="Number of elements (default: 32)",
+    )
+    args = parser.parse_args()
+
+    # Construct two input random tensors and an output zeroed tensor
+    # The three tensor are in memory accessible to the NPU
+    input0 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
+    input1 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
+    output = iron.zeros_like(input0)
 
     # JIT-compile the kernel then launches the kernel with the given arguments. Future calls
     # to the kernel will use the same compiled kernel and loaded code objects
-    vector_vector_add(input0, input1, output)
+    vector_vector_add(device_map[args.device], input0, input1, output)
 
     # Check the correctness of the result
     e = np.equal(input0.numpy() + input1.numpy(), output.numpy())
@@ -148,11 +138,11 @@ def main():
     # Otherwise, exit with a failure code
     if not errors:
         print("\nPASS!\n")
-        exit(0)
+        sys.exit(0)
     else:
         print("\nError count: ", errors)
         print("\nFailed.\n")
-        exit(-1)
+        sys.exit(-1)
 
 
 if __name__ == "__main__":
--- a/programming_examples/basic/vector_vector_add/vector_vector_add.py
+++ b/programming_examples/basic/vector_vector_add/vector_vector_add.py
@@ -41,26 +41,55 @@ parser.add_argument(
 )
 args = parser.parse_args()
 
-if args.num_elements % 16 != 0:
-    raise ValueError("num_elements must be a multiple of 16.")
-
 device_map = {
     "npu": NPU1Col1(),
     "npu2": NPU2Col1(),
 }
-dev = device_map[args.device]
-num_elements = args.num_elements
-data_type = np.int32
+device = device_map[args.device]
+
+# Construct two input random tensors and an output zeroed tensor
+# The three tensor are in memory accessible to the NPU
+input0 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
+input1 = iron.rand((args.num_elements,), dtype=np.int32, device=args.device)
+output = iron.zeros_like(input0)
 
 
 @iron.jit(is_placed=False)
-def vector_vector_add():
+def vector_vector_add(
+    # input0, input1, output
+):
+    if input0.shape != input1.shape:
+        raise ValueError(
+            f"Input shapes are not the equal ({input0.shape} != {input1.shape})."
+        )
+    if input0.dtype != input1.dtype:
+        raise ValueError(
+            f"Input data types are not the same ({input0.dtype} != {input1.dtype})."
+        )
+
+    if input0.shape != output.shape:
+        raise ValueError(
+            f"Input and output shapes are not the equal ({input0.shape} != {output.shape})."
+        )
+    if input0.dtype != output.dtype:
+        raise ValueError(
+            f"Input and output data types are not the same ({input0.dtype} != {output.dtype})."
+        )
+
+    if len(np.shape(input0)) != 1:
+        raise ValueError("Function only supports vectors.")
+    num_elements = np.size(input0)
     n = 16
+    if num_elements % n != 0:
+        raise ValueError(
+            f"Number of elements ({num_elements}) must be a multiple of {n}."
+        )
     N_div_n = num_elements // n
+    dtype = input0.dtype
 
     # Define tensor types
-    tensor_ty = np.ndarray[(num_elements,), np.dtype[data_type]]
-    tile_ty = np.ndarray[(n,), np.dtype[data_type]]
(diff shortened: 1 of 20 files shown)


jgmelber:
I like this!

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -4,63 +4,52 @@
...
-# (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
-import numpy as np
-import sys
+# (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
+
 import argparse
+import sys
+import numpy as np
 
+from aie import iron
```
jgmelber, line 13: Don't we want `import aie.iron as iron` ?

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -4,70 +4,54 @@
...
-import numpy as np
-import sys
-import argparse
+# (c) Copyright 2024-2025 Advanced Micro Devices, Inc. or its affiliates
 
+import argparse
+import sys
+import numpy as np
 
+from aie import iron
```
jgmelber, line 13: Don't we want `import aie.iron as iron` ?

mawad-amd:
I don't think this will work, unfortunately. The operator `()` expects Tensors that match what is inside the runtime sequence because these will be passed to the kernel. I need to add some unittests.

https://github.com/Xilinx/mlir-aie/blob/d291a69b3e91f4d82f4fe74b2adc1c175d50dda6/python/iron/jit.py#L83-L103

mawad-amd:
I like the `iron.jit` though. I tried to implement previously but had issues.

ypapadop-amd:
> I don't think this will work, unfortunately. The operator `()` (i.e., `vector_vector_add(device_map[args.device], input0, input1, output)` expects Tensors that match what is inside the runtime sequence because these will be passed to the kernel. I need to add some unittests.
> 
> https://github.com/Xilinx/mlir-aie/blob/d291a69b3e91f4d82f4fe74b2adc1c175d50dda6/python/iron/jit.py#L83-L103

You mean because of the `device` arg? That can be fixed if we always assume that the first argument is device (or make it positional).

mawad-amd:
Particularly for the device, I think it's best if we do things like:

```python
@iron.jit
def kernel()
   device = iron.get_current_device()

iron.set_device(device)
```

Or via contexts:

```python
@iron.jit
def kernel()
   device = iron.get_current_device()

with iron.device('npu'):
    kernel()

```

But it's likely we will need to pass more arguments to the kernel that we can't set as globals. We could possibly start adding annotations (e.g., `iron.kernel_argument`) or something like that. 

I think it's best if we just meet and discuss to keep the momentum going. I will schedule something.

not-jgm:
> We can talk more on Tuesday.

Sounds good to me

jgmelber:
> We can talk more on Tuesday.

Sounds good to me

ypapadop-amd:
I can see us controlling not just which device, but how many tiles to use etc. Which I think makes it equivalent to what numba does with kernel invocation (https://numba.readthedocs.io/en/stable/cuda/kernels.html).

But I'd like us to think what's appropriate for our devices, not how the GPU world converged around CUDA.

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -92,20 +81,48 @@ def core_body(of_in1, of_in2, of_out):
...
+        help="Number of elements (default: 32)",
+    )
+    args = parser.parse_args()
 
     # Construct two input random tensors and an output zeroed tensor
     # The three tensor are in memory accessible to the NPU
-    input0 = iron.rand((num_elements,), dtype=data_type, device="npu")
-    input1 = iron.rand((num_elements,), dtype=data_type, device="npu")
+    input0 = iron.randint(
+        0, 100, (args.num_elements,), dtype=np.int32, device=args.device
```
mawad-amd, line 116:
```suggestion
        0, 100, (args.num_elements,), dtype=np.int32, device='npu'
```

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -113,16 +97,53 @@ def sequence(A, B, C):
...
+        help="Number of elements (default: 32)",
+    )
+    args = parser.parse_args()
 
     # Construct two input random tensors and an output zeroed tensor
     # The three tensor are in memory accessible to the NPU
-    input0 = iron.rand((num_elements,), dtype=data_type, device="npu")
-    input1 = iron.rand((num_elements,), dtype=data_type, device="npu")
+    input0 = iron.randint(
+        0, 100, (args.num_elements,), dtype=np.int32, device=args.device
```
mawad-amd, line 132:
```suggestion
        0, 100, (args.num_elements,), dtype=np.int32, device='npu'
```

mawad-amd:
I like the syntax of `kernel[configs](kernel_arg)` because then users won't be confused about kernel arguments, their order and how they relate to the runtime sequence. Eventually, we want the user to get a nice error when they mess that up (also the asserts you added, these are great but should be eventually done by compiler). 

I also like your `config` argument, although, I would prefer it to be the last argument. I wonder if we can get rid of `tensor_ty = np.ndarray[(num_elements,), np.dtype[dtype]]` and other `numpy` arrays and reuse the input tensors too (not sure how the objectfifo work though).

programming_examples/basic/vector_vector_add/vector_vector_add.py
ypapadop-amd: I thought IRON supported other devices than NPUs (e.g., PCIe cards).

mawad-amd: The Tensor abstraction are designed and tested only for RyzenAI NPUs.

ypapadop-amd:
Here's how numba does it: https://numba.pydata.org/numba-doc/dev/roc/examples.html

Triton: https://triton-lang.org/main/getting-started/tutorials/03-matrix-multiplication.html#sphx-glr-getting-started-tutorials-03-matrix-multiplication-py (I think they infer the device from the inputs). In triton, functions with `@triton.jit` if you pass them as parameters to other functions, which is very convenient.

ypapadop-amd:
Here's the separation I have in my head:

1. Function arguments should be passed to function parameters, e.g., `def vector_vector_add(input0, input1, output)` should be called with `vector_vector_add(input0, input1, output)`.
2. Attributes that affect how the code is compiled, e.g., `is_placed`, should go as decorator arguments, e.g., `@iron.jit(is_placed=True)`. I think that also includes any type constraints, like numba does (`@roc.jit('(float32[:,:], float32[:,:], float32[:,:])'`) in https://numba.pydata.org/numba-doc/dev/roc/examples.html) but I'm not sure if we need something like this right now.
3. Attributes that are affecting run-time behaviour, e.g., to use a cached version or not, where is the cache, which device we target, etc. should be passed somehow else, e.g., the `config` variable, a singleton that is visible upon import etc.

The distinction between 2 and 3 is that 2 has to do with the code that follows (think of C++ strong typing) whereas 3 is more on where and how that code will be called.

jgmelber:
> I added a `iron.set_default_device()` / `iron.get_default_device()` that eliminates the `config` parameter. However, the column_id is still a problem to express via a config.

`column_id` should be removed. The NPU runtime firmware handles this well. This is an artifact from supporting cvk5000 in this example previously.

programming_examples/basic/vector_vector_add/vector_vector_add.py
ypapadop-amd: I replaced them again with "npu" because the only accepted values in iron.Tensor are "npu" or "cpu". I'll leave it to you to decide what's the correct approach.

jgmelber: This is in relation to XRT buffer objects...currently we only use NPUs with XRT. We can look into extending to PCIe devices if that becomes a target we choose to support later on

jgmelber:
😄 https://github.com/Xilinx/mlir-aie/pull/2213/commits/57376150f68b73d9d07e87f7d4a9bbc7332ab3a1

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -92,17 +81,43 @@ def core_body(of_in1, of_in2, of_out):
...
 
     # Construct two input random tensors and an output zeroed tensor
     # The three tensor are in memory accessible to the NPU
-    input0 = iron.rand((num_elements,), dtype=data_type, device="npu")
-    input1 = iron.rand((num_elements,), dtype=data_type, device="npu")
+    input0 = iron.randint(0, 100, (args.num_elements,), dtype=np.int32, device="npu")
+    input1 = iron.randint(0, 100, (args.num_elements,), dtype=np.int32, device="npu")
     output = iron.zeros_like(input0)
 
+    iron.set_current_device(device_map[args.device])
```
jgmelber, line 119: An eventual PyXRT interface for below will help @mawad-amd: 

```
// Get a device handle
unsigned int device_index = 0;
auto device = xrt::device(device_index);
std::cout << device.get_info<xrt::info::device::name>() << std::endl;
```

mawad-amd:
This is looking awesome, thanks Yiannis for the improvements. And thanks for fixing the tensor random initialization bug too. I suggest adding one comment on why the column id is 0 in the code for the readers (I didn't know about that). Other than that, the PR is looking great.

programming_examples/basic/vector_vector_add/vector_vector_add.py
ypapadop-amd: Do we have a way to get the current device at startup? Maybe we can avoid setting it at the beginning.

jgmelber: What do you mean startup?

mawad-amd: We don't have one yet, but we discussed and we we want to add it.
https://github.com/Xilinx/mlir-aie/pull/2201#discussion_r2042762849

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -113,13 +97,40 @@ def sequence(A, B, C):
 
 
 def main():
+    device_map = {
+        "npu": AIEDevice.npu1_1col,
```
ypapadop-amd, line 101: Why do we have both `AIEDevice` and `Device`? Is there a path to having them be an alias to the same object?

programming_examples/basic/vector_vector_add/vector_vector_add.py
```diff
@@ -92,17 +81,43 @@ def core_body(of_in1, of_in2, of_out):
         rt.drain(of_out.cons(), C, wait=True)
 
     # Place program components (assign them resources on the device) and generate an MLIR module
-    return Program(dev, rt).resolve_program(SequentialPlacer())
+    return Program(iron.get_current_device(), rt).resolve_program(SequentialPlacer())
```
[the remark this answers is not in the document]
jgmelber: Right. As @mawad-amd mentioned, we discussed it and it would take a PR into the XRT repo to add a python binding for the C++ interface exposing the device name as `xrt-smi examine` does. We could also use a sys call from that utility directly. We do this in the setup scripts to set an environment variable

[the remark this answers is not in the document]
jgmelber: https://github.com/Xilinx/mlir-aie/blob/main/utils/env_setup.sh#L52-L63

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
jgmelber: They do wind up generating the same MLIR. I think the placed IRON examples would just need to import them. As far as I know that should be possible

python/iron/config.py
```diff
@@ -0,0 +1,21 @@
+# config.py -*- Python -*-
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
+
+config = {}
```
jgmelber, line 9: Can/should this use `xrt-smi` to set a default based on the system? Then it can be overridden by `set_current_device`

jgmelber:
> the PR is looking great

Agreed!

python/iron/config.py
ypapadop-amd: Seems reasonable, let me try it.

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
ypapadop-amd: I'll leave it as future work. It'll be nice to constraint the `set_default_device` function.

python/iron/config.py
jgmelber: Willing to try myself... what do I pass to `set_current_device()` "npu" or "npu2"?

programming_guide/mini_tutorial/exercise_4/exercise_4b/exercise_4b.py
```diff
@@ -18,22 +19,23 @@
 
 import aie.iron as iron
 
-dev = NPU2()
-
-# Define tensor types
+# Define tensor shape
 data_height = 3
```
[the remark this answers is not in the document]
erwei-xilinx: I _think_ the pointer is the only thing that really matters. My understanding is that the tensor shape is only used for a verifier, checking whether the wrap-and-stride access pattern goes out of bound. The amount of data being copied is coded in `TensorAccessPattern`'s sizes.

python/iron/config.py
jgmelber: #2226

programming_guide/mini_tutorial/exercise_4/exercise_4b/exercise_4b.py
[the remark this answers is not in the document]
ypapadop-amd: Still, we lack the mechanism to pass this information in. Another example would be if you have a function that can generate vector-add or vector-sub using one extra argument. It can't be part of the arguments, since those are supposed to be tensors only. How do we pass that in?

test/python/set_current_device.py
```diff
@@ -0,0 +1,22 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 AMD Inc.
+
+# RUN: %python %s | FileCheck %s
```
mawad-amd, line 7: Not all CI runners have `xrt` and so not all of them get the `iron.get_current_device` function. You could move the [import outside the try-except block](https://github.com/Xilinx/mlir-aie/blob/ypapadop/decorator-arg-passthrough/python/iron/__init__.py#L15) but I do recommend we keep this here since we will need xrt when we add query/enumartion functionality.

```suggestion
# RUN: %run_on_npu %python %s | FileCheck %s
```

test/python/set_current_device.py
```diff
@@ -0,0 +1,22 @@
...
+
+from aie.iron.device import NPU2
+
+import aie.iron as iron
+
+
+# CHECK: NPU2
```
mawad-amd, line 14: I recommend we transition to `pytest` since it makes more sense for us here. I added some unittests for the Tensor class using pytest here: https://github.com/Xilinx/mlir-aie/pull/2227/files

mawad-amd: Then we can have:
```python
# RUN: %run_on_npu %pytest
# RUN: %run_on_2npu %pytest
```
which would make a lot of sense for the `detect_npu_device`.

ypapadop-amd: Done.

ypapadop-amd: I missed the fact it's still open. I'll wait for it to be merged.

jgmelber: Merged

test/python/set_current_device.py
```diff
@@ -0,0 +1,19 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 AMD Inc.
+
+# RUN: %run_on_npu %pytest %s
+# RUN: %run_on_2npu %pytest %s
+
+import pytest
```
mawad-amd, line 10:
```suggestion
```
I don't think you need this import here?

python/iron/config.py
jgmelber: I am unfamiliar. @mawad-amd feel free to add this

mawad-amd: I will add it in a different pull request.

ypapadop-amd:
> This is looking awesome, thanks Yiannis for the improvements. And thanks for fixing the tensor random initialization bug too. I suggest adding one comment on why the column id is 0 in the code for the readers (I didn't know about that). Other than that, the PR is looking great.

What's the comment here? 0 implies runtime figures out placement? @jgmelber

jgmelber:
I thought column_id was removed?

ypapadop-amd:
> I thought column_id was removed?

It did, I'm just asking if an additional comment is needed.

jgmelber:
> It did, I'm just asking if an additional comment is needed.

I don't think so

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
```diff
@@ -4,76 +4,60 @@
...
-        tile_ty = np.ndarray[(n,), np.dtype[data_type]]
+        tensor_ty = np.ndarray[(num_elements,), np.dtype[dtype]]
+        tile_ty = np.ndarray[(n,), np.dtype[dtype]]
 
         # AIE Core Function declarations
 
         # Tile declarations
-        ShimTile = tile(column_id, 0)
-        ComputeTile2 = tile(column_id, 2)
+        ShimTile = tile(0, 0)
```
mawad-amd, line 59: A comment on the 0 here would be nice.

mawad-amd (approved):
Thanks, Yiannis!

programming_examples/basic/vector_vector_add/vector_vector_add_placed.py
jgmelber: I mean the first column of any partition is 0

jgmelber: Additionally only 1 column devices are in the device_map specified in main.

PR #2217, opened 2025-04-22
Update build_drivers.sh for tag 1.4

jgmelber:
Use xdna-driver commit for tag 1.4


fifield:
I don't know if you noticed, but this reverts the phx firmware to the December version which we had fewer issues with...
https://github.com/amd/xdna-driver/pull/474/files

jgmelber:
Yeah I wonder if we should use this

jgmelber:
I tried this locally but XRT didn't build the base package, just npu and that didn't work on my system

PR #2226, opened 2025-04-25
Detect device from xrt-smi in JIT

jgmelber:
Concept for detecting the device based on the current system. Makes the assumption XRT is installed. The device type set is the "whole array" either `NPU1Col4` or `NPU2`, this is not efficient but might be sufficient.


ypapadop-amd (approved):
LGTM. 

Is there a way to separate the device from the columns in general?

jgmelber:
> LGTM.
> 
> Is there a way to separate the device from the columns in general?

Not currently. All of this opens some questions on handling the partitioning. We detect columns at compile time in `aiecc.py` for XCLBIN metadata. We should think about this.

ypapadop-amd:
> > LGTM.
> > Is there a way to separate the device from the columns in general?
> 
> Not currently. All of this opens some questions on handling the partitioning. We detect columns at compile time in `aiecc.py` for XCLBIN metadata. We should think about this.

Maybe we can create a new set of types (or reuse only NPU1 / NPU2) + column information. Then instead of invoking `get_current_device()` we can do something like `get_current_device_and_column()` and maybe pass a flag to return `AIEDevice` or `Device`. At least that'll fix the user facing interface.

python/iron/__init__.py
```diff
@@ -12,6 +12,7 @@
     # imports the .jit module.
     import pyxrt
     from .jit import jit
+    from .config import set_current_device, get_current_device
```
maawad, line 15:
```suggestion
    from .config import set_current_device, get_current_device, detect_npu_device
```

mawad-amd, line 15:
```suggestion
    from .config import set_current_device, get_current_device, detect_npu_device
```

python/iron/config.py
```diff
@@ -0,0 +1,75 @@
...
+            text=True,
+        )
+        output = result.stdout
+
+        # Match strings for NPU2
+        # Set's generic "whole array" devices, this is overkill...
+        if any(
+            keyword.lower() in output.lower()
+            for keyword in [
+                "NPU Strix",
```
mawad-amd, line 31: (I know we will transition to a more robust way in the future) I guess I need to update my system for the nice naming:
```
xrt-smi examine
System Configuration
  OS Name              : Linux
  Release              : 6.10.3-061003-generic
  Machine              : x86_64
  CPU Cores            : 16
  Memory               : 15782 MB
  Distribution         : Ubuntu 24.04 LTS
  GLIBC                : 2.39
  Model                : Venus series
  BIOS Vendor          : American Megatrends International, LLC.
  BIOS Version         : 1.09

XRT
  Version              : 2.18.0
  Branch               : HEAD
  Hash                 : 1c405960ff6733ffe637311a7d7177ad7a185cc2
  Hash Date            : 2024-12-03 04:34:17
  xocl                 : unknown, unknown
  xclmgmt              : unknown, unknown
WARNING: xclmgmt version is unknown. Is xclmgmt driver loaded? Or is MSD/MPD running?
  amdxdna              : 2.18.0_20241203, 17b334dbf7682950a0077e04ed8c7ea386fb3a1d
  NPU Firmware Version : 1.5.1.362

Device(s) Present
|BDF             ||Name          |
|----------------||--------------|
|[0000:c5:00.1]  ||RyzenAI-npu1  |
```

test/python/set_current_device.py
```diff
@@ -0,0 +1,21 @@
...
+# (c) Copyright 2025 AMD Inc.
+
+# RUN: %python %s | FileCheck %s
+
+from aie.iron.device import NPU2
+
+import aie.iron as iron
+
+# CHECK: NPU2
+def main():
```
mawad-amd, line 14: Now I know how to add unit tests :)

mawad-amd (approved):
This looks good. I would like us to support and use often index-based device management similar to PyTorch and Numba (future improvement we can add).

https://pytorch.org/docs/stable/generated/torch.cuda.device.html
https://numba.pydata.org/numba-doc/dev/cuda/device-management.html

test/python/set_current_device.py
```diff
@@ -0,0 +1,21 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 AMD Inc.
+
+# RUN: %python %s | FileCheck %s
```
mawad-amd, line 7:
```suggestion
# RUN: %run_on_npu %python %s | FileCheck %s
```

python/iron/config.py
```diff
@@ -6,8 +6,56 @@
...
+                "NPU Strix Halo",
+                "NPU Kracken",
+                "RyzenAI-npu4",
+                "RyzenAI-npu6",
+            ]
+        ):
+            return NPU2()
+        else:
+            return NPU1Col4()
```
fifield, line 47: It would be nice for future proofing to explicitly match Phoenix devices here and fail if it doesn't match a known string.

PR #2228, opened 2025-04-25
Add bfp16 support

L-roro:
This is a proposal to add bfp16 support to IRON.  It is currently incomplete since I have not been able to make Peano work for the examples I have written, but they are apparently working with chess. I open this draft pr in order to get some feedback from the team.

The logic for the lowering is to simply transform the new block float type into an integer type of the appropriate length from the builtin mlir dialect. This lowering is done before the lowering of the object fifos into DMABDOps to avoid conflicts with functions calls that assume they are dealing with builtin types beyond that point in the pipeline.

There are currently no additional transformations being done and therefore management of the bfps is let at the discretion of the programmer. Currently, bfps are referenced as a unit from IRON (1 bfp represents 8 or 16 floats) to avoid dividing a bfp during data movement. This is in contrast to the AIE API which defines its vectors by referencing the number of floats represented. See bfp_conversion/kernel.cc for example and additional explanations.

Currently, different bfp types are instantiated uniquely in IRON (bfp16ebs8, bfp16ebs16). Nevertheless, they translate into a single MLIR type that encompasses all such types and identifies them through their AIE API name. I believe this is a reasonable compromise between syntax complexity and maintainability, that allows both for general optimizations for all block types or for individual ones if needed (despite not supporting any currently).

### DIFF
--- a/programming_examples/basic/block_datatypes/bfp_conversion/test.cpp
+++ b/programming_examples/basic/block_datatypes/bfp_conversion/test.cpp
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// Copyright (C) 2023, Advanced Micro Devices, Inc.
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
 //
 //===----------------------------------------------------------------------===//
 
--- a/programming_examples/basic/block_datatypes/helper.h
+++ b/programming_examples/basic/block_datatypes/helper.h
@@ -1,3 +1,13 @@
+//===- helper.h -------------------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
 #include <cassert>
 #include <cfloat>
 #include <cmath>
--- a/python/_mlir_libs/_aie.pyi
+++ b/python/_mlir_libs/_aie.pyi
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-import numpy as np
-
 from .ir import Operation, DialectRegistry, Type
 
 __all__ = [
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -796,7 +796,7 @@ uint64_t AIEX::BlockFloatType::getABIAlignment(
     const mlir::DataLayout &dataLayout,
     mlir::DataLayoutEntryListRef params) const {
   // For the purposes of the data movement operations, we want all types to be
-  // packed <=> ABI alignemnt is 1.
+  // packed <=> ABI alignment is 1.
   return 1;
 }
 
--- a/programming_examples/basic/block_datatypes/bfp_conversion/CMakeLists.txt
+++ b/programming_examples/basic/block_datatypes/bfp_conversion/CMakeLists.txt
@@ -5,7 +5,6 @@
 # (c) Copyright 2025 Advanced Micro Devices, Inc.
 
 # parameters
-# -DBOOST_ROOT: Path to Boost install
 # -DXRT_INC_DIR: Full path to src/runtime_src/core/include in XRT cloned repo
 # -DXRT_LIB_DIR: Path to xrt_coreutil.lib
 # -DTARGET_NAME: Target name to be built
--- a/programming_examples/basic/block_datatypes/vector_passthrough/CMakeLists.txt
+++ b/programming_examples/basic/block_datatypes/vector_passthrough/CMakeLists.txt
@@ -2,10 +2,9 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2023 Advanced Micro Devices, Inc.
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
 
 # parameters
-# -DBOOST_ROOT: Path to Boost install
 # -DXRT_INC_DIR: Full path to src/runtime_src/core/include in XRT cloned repo
 # -DXRT_LIB_DIR: Path to xrt_coreutil.lib
 # -DTARGET_NAME: Target name to be built
--- a/programming_examples/basic/block_datatypes/vector_passthrough/README.md
+++ b/programming_examples/basic/block_datatypes/vector_passthrough/README.md
@@ -1,4 +1,4 @@
-<!---//===- README.md --------------------------*- Markdown -*-===//
+<!---//===- README.md --------------------------------------*- Markdown -*-===//
 //
 // This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 // See https://llvm.org/LICENSE.txt for license information.


include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -52,6 +52,21 @@ using TileID = struct TileID {
...
+
+  bool operator==(const BfpType &rhs) const {
+    return std::tie(blockSize, mantissaBits, exponentBits, subtileShiftBits) ==
+           std::tie(rhs.blockSize, rhs.mantissaBits, rhs.exponentBits,
+                    rhs.subtileShiftBits);
+  }
+
+  bool operator!=(const BfpType &rhs) const { return !(*this == rhs); }
+};
+
```
fifield, line 69: Why this structure instead of using `BlockFloatType`?

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -34,8 +36,39 @@ def AIEX_Dialect : Dialect {
   let extraClassDeclaration = [{
   }];
 
+  let useDefaultTypePrinterParser = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// AIEX Types
+//===----------------------------------------------------------------------===//
+def AIEX_BlockFloatingPointType : TypeDef<AIEX_Dialect, "blockFloat", [
```
fifield, line 45:
```suggestion
def AIEX_BlockFloatingPointType : TypeDef<AIEX_Dialect, "BlockFloat", [
```
Types are usually capitalized in MLIR.

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -301,6 +316,10 @@ class AIETargetModel {
   // their corresponding lengths in bytes (second).
   virtual std::vector<std::pair<uint32_t, uint32_t>>
   getShimBurstEncodingsAndLengths() const = 0;
+
+  // Returns a bfpType with the corresponding caracteristics or std::nullopt if
+  // unsupported.
+  virtual std::optional<BfpType> getBfpType(std::string bfpName) const = 0;
```
fifield, line 322: I think type "getters" should be part of the type or dialect interface, e.g.:
```c++
AIEX::BlockFloatType AIEX::BlockFloatType::get(std::string &bfpName) {
 ...
}
```
In the TargetModel I would expect something more like
```
bool TargetModel::isSupported(mlir::Type t) { 
  ...
}
```

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -34,8 +36,39 @@ def AIEX_Dialect : Dialect {
...
+  let parameters = (
+    ins StringRefParameter<"">:$block_type
+  );
+
+  let assemblyFormat = "`<` $block_type `>`";
+
+  // Indicate that our type will not add additional verification to the parameters.
+  // Since this type needs information about the targetModel to be verified, 
+  // checks are done in the bfp lowering pass instead of the verifier.
+  let genVerifyDecl = 0;
```
fifield, line 66: The type should be valid (or not) independent of TargetModel. It's use can be validated later on during some legalization or TargetModel based checks.

lib/Dialect/AIEX/Transforms/AIETransformBfpTypes.cpp
```diff
@@ -0,0 +1,236 @@
...
+        // returning a std::nullopt will allow the converter to keep trying the
+        // remaining conversions (thus reaching the default one in this case)
+        return nullptr;
+      }
+
+      return mlir::IntegerType::get(blockType.getContext(),
+                                    bfpType->blockSize * bfpType->mantissaBits +
+                                        bfpType->subtileShiftBits +
+                                        bfpType->exponentBits);
+    });
```
fifield, line 69:
```suggestion
      bool isSupported = targetModel.isSupportedType(blockType.getBlockType());
      if (!isSupported ) {
        llvm::errs() << "Block type " << blockType.getBlockType()
                     << " is not supported in the specified model\n";
        // Note that returning a nullopt here will stop the conversion while
        // returning a std::nullopt will allow the converter to keep trying the
        // remaining conversions (thus reaching the default one in this case)
        return nullptr;
      }

      return mlir::IntegerType::get(blockType.getContext(),
                                    blockType.getBlockSize() * blockType.getMantissaBits() +
                                    blockType.getSubtileShiftBits() +
                                    blockType.getExponentBits());
    });
```

lib/Dialect/AIEX/Transforms/AIETransformBfpTypes.cpp
```diff
@@ -0,0 +1,236 @@
...
+      if (!conversion) {
+        conversionFailed = true;
+        return op->emitError()
+               << "Failed to convert operand type: " << operand.getType();
+      }
+      operand.setType(conversion);
+    }
+
+    // Operation attributes
+    // Note: For some reason, the attribute list looks like it is immutable and
```
fifield, line 181: Attributes represent constant data, are unique, and are cached, so they are immutable.

programming_examples/basic/block_datatypes/bfp_conversion/run_strix_makefile.lit
```diff
@@ -0,0 +1,9 @@
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// REQUIRES: ryzen_ai, peano
+//
+// RUN: mkdir -p test_stx
+// RUN: cd test_stx
```
fifield, line 7: `test_stx` appears to be reused for mulitple lit tests in this same directory. The reason to `mkdir` here is to avoid file collisions for tests running in the same directory, so the directories made here need to be unique.

programming_examples/basic/block_datatypes/vector_passthrough/run_strix_makefile.lit
```diff
@@ -0,0 +1,9 @@
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// REQUIRES: ryzen_ai, peano
+//
+// RUN: mkdir -p test_stx
+// RUN: cd test_stx
```
fifield, line 7: See previous comment about `mkdir`

python/helpers/util.py
```diff
@@ -17,6 +19,23 @@
...
+    def get():
+        return CustomTypes.blockFloatType.get("bfp16ebs8")
+
+
+class bfp16ebs16(np.generic):
+    """Custom type to be used in IRON that is translated to a generic blockFloatType"""
+
+    @staticmethod
+    def get():
+        return CustomTypes.blockFloatType.get("bfp16ebs16")
```
fifield, line 36: Please add some simple unit test(s) for the python bindings in `test/python`

programming_examples/basic/block_datatypes/bfp_conversion/run_strix_makefile.lit
L-roro: I did not catch that detail, thank you. Should be corrected now.

python/helpers/util.py
L-roro: Done

include/aie/Dialect/AIE/IR/AIETargetModel.h
fifield: The refactored code looks good to me. The string interface avoids the dependency issue and could be cleaned up later if/when the type is moved to AIE dialect.

jamesroxbypb:
I'd really like us to think about the naming here, and making it consistent with how others describe these block types.

I don't really like 1 bfp16ebs8 representing 8 different values and I think if we then started to get into "we want a vector of these", having v8bfp16ebs8 mean 64 different values, I just think we'll end up in a world of hurt.

This was always going to be weird.  There isn't really the notion of a "single" bfp16 value.  It has to come as part of a group, whereas you can have a single int8, so v64int8 straight up works.  I just think v64bfp16ebs8 meaning 64 bfp16 numbers which happen to share exponents in group size of 8, is more natural v8bfp16ebs8 for the same thing.

python/_mlir_libs/_aie.pyi
```diff
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import numpy as np
+
```
fifield: is this needed? I don't see any use of `np`

programming_examples/basic/block_datatypes/helper.h
fifield, line 1: File needs license/copyright.

programming_examples/basic/block_datatypes/bfp_conversion/test.cpp
```diff
@@ -0,0 +1,221 @@
+//===- test.cpp -------------------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2023, Advanced Micro Devices, Inc.
```
fifield:
```suggestion
// Copyright (C) 2025, Advanced Micro Devices, Inc.
```

python/_mlir_libs/_aie.pyi
L-roro: Definitely not, it was a leftover from when I was trying out things, sorry

programming_examples/basic/block_datatypes/helper.h
L-roro: Added

programming_examples/basic/block_datatypes/bfp_conversion/test.cpp
L-roro: I will stop making that mistake at some point in my life, I swear

fifield (approved):

AndraBisca (approved):

PR #2234, opened 2025-04-27
Use `cxxopts` instead of Boost

### LINKED ISSUE #2231: Remove Boost dependency
We don't really need Boost just for command line options

mawad-amd:
This PR removes the Boost dependency and uses [cxxopts](https://github.com/jarro2783/cxxopts) instead for command line arguments. `cxxopts` is a single-file library and has similar API to Boost’s `program_options`.

I copied the entire header here ([link with sha](https://github.com/jarro2783/cxxopts/blob/dbf4c6a66816f6c3872b46cc6af119ad227e04e1/include/cxxopts.hpp)), but we can fetch it at build time if people prefer that.

Closes #2231

### DIFF
--- a/.github/workflows/buildAndTestAieToolsHsaBuildOnly.yml
+++ b/.github/workflows/buildAndTestAieToolsHsaBuildOnly.yml
@@ -83,7 +83,7 @@ jobs:
                 | tee /etc/apt/preferences.d/rocm-pin-600
             apt update
             apt install -y rocm-hip-runtime-dev5.6.0 python3.10-venv \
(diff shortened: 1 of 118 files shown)


jgmelber:
This is great. Will review in more detail.

Can you explain what `cxxopts.hpp` is and where it comes from in the PR description?

jgmelber:
Thanks for updating 2f856dab97b5bf01d919db7328d5c8fca21a8e54.

In another PR, I'd like to remove the buildHost*.md files. Linux is covered...we need a fresh Windows doc once we can commit to support.

jgmelber (approved):

PR #2239, opened 2025-04-28
Fixing Krackan spelling

eddierichter-amd:
xrt-smi reports NPU as "Krackan" which means with the current spelling no device is found.


jgmelber (approved):

jgmelber:
Source: 

https://github.com/amd/xdna-driver/blob/a7a23f661c3003b9a016dacb262ab1496420ecba/src/driver/amdxdna/npu6_regs.c#L8-L12

PR #2251, opened 2025-04-30
Experiment with trace

mawad-amd:
- WIP experimentation with JIT and Trace. 
- `CoreFunction` is a wrapper around the C++ core code (ignore for now)

Notes from the discussion with Jack:
* `enable_trace` needs to internally figure out that the trace buffer is the last argument. At the moment, it is assumed to be the fifth argument and hence we have this dummy argument at kernel launch.

Run with:
```cmd
python programming_examples/basic/vector_scalar_mul/vector_scalar_mul.py
```

### DIFF
--- a/programming_examples/basic/vector_scalar_mul/vector_scalar_mul.py
+++ b/programming_examples/basic/vector_scalar_mul/vector_scalar_mul.py
@@ -99,7 +99,7 @@ def vector_scalar_mul(input, factor, output, dummy_input0, trace):
     # Runtime operations to move data to/from the AIE-array
     rt = Runtime()
     with rt.sequence(tensor_ty, scalar_ty, tensor_ty) as (A, F, C):
-        rt.enable_trace(trace.numel())
+        rt.enable_trace(trace.numel() * np.dtype(trace.dtype).itemsize)
         rt.start(worker)
         rt.fill(of_in.prod(), A)
         rt.fill(of_factor.prod(), F)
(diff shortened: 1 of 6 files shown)


programming_examples/basic/vector_scalar_mul/vector_scalar_mul.py
```diff
@@ -72,51 +99,39 @@ def core_body(of_in, of_factor, of_out, scale_fn):
     # Runtime operations to move data to/from the AIE-array
     rt = Runtime()
     with rt.sequence(tensor_ty, scalar_ty, tensor_ty) as (A, F, C):
-        rt.enable_trace(trace_size)
+        rt.enable_trace(trace.numel() * np.dtype(trace.dtype).itemsize)
```
mawad-amd, line 102: @jackl-xilinx Does `enable_trace` take the number pf bytes?

mawad-amd:
@jackl-xilinx, whenever you have time, could you please try running this? Here is the trace I get too [trace.txt](https://github.com/user-attachments/files/19971254/trace.txt). After that we can see how we can get rid of the dummy variable.

mawad-amd:
Looking at this again, I think we might as well completely hide the tracing. Consider something like:
```python
iron.enable_tracing()

# Magically insert the trace at kernel launch and into the RT sequence
vector_scalar_mul(input, factor, output)

iron.stop_tracing("trace.bin")
```

mawad-amd:
Better approach is implemented in https://github.com/Xilinx/mlir-aie/pull/2541. Closing this one.

PR #2255, opened 2025-04-30
Update windows instructions and Makefile

jackl-xilinx:


jgmelber (commented):
Can you remove the requirement on Vitis from these docs like the front page Linux instructions?

jgmelber:
Will `${getwslpath}` be applied to the other programming examples?

jackl-xilinx:
> Will `${getwslpath}` be applied to the other programming examples?

Yes, I'm applying those  now.

jackl-xilinx:
> Will `${getwslpath}` be applied to the other programming examples?

Yes, I'm applying these now.

programming_examples/basic/matrix_multiplication/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ if (NOT WSL)
     set(XRT_INC_DIR /opt/xilinx/xrt/include CACHE STRING "Path to XRT cloned repo")
     set(XRT_LIB_DIR /opt/xilinx/xrt/lib CACHE STRING "Path to xrt_coreutil.lib")
 else()
+    set(BOOST_INCLUDE_DIRS C:/Technical/thirdParty/boost_1_83_0 CACHE STRING "Path to Boost install")
```
mawad-amd, line 26: Why are we bringing Boost back? Is there issues with cxxopts?

jgmelber: XRT needs the headers 😢

utils/setup_files/device.h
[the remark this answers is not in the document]
jgmelber: XRT somewhere/somehow? @jackl-xilinx has the details

programming_examples/basic/matrix_multiplication/CMakeLists.txt
jackl-xilinx: @mawad-amd XRT seems to have a dependancy on boost and its headers in windows have references it. If we don't have boost, we get the following host linking errors. If you have a way to work around it, let me know. I'd love to not have to unzip the boost source (no need to build it though).
```
   1>Checking Build System
  Building Custom Rule C:/Technical/mlir-aie/programming_examples/basic/vector_scalar_mul/CMakeLists.txt
  test_utils.cpp
  test.cpp
C:\Technical\XRT\src\runtime_src\core\include\xrt\detail\any.h(21,11): error C1083: Cannot open include file: 'boost/any.hpp': No such file or directory [C
:\Technical\mlir-aie\programming_examples\basic\vector_scalar_mul\_build\vector_scalar_mul_in1_size.vcxproj]
  (compiling source file '../../../../runtime_lib/test_lib/test_utils.cpp')

C:\Technical\XRT\src\runtime_src\core\include\xrt\detail\any.h(21,11): error C1083: Cannot open include file: 'boost/any.hpp': No such file or directory [C
:\Technical\mlir-aie\programming_examples\basic\vector_scalar_mul\_build\vector_scalar_mul_in1_size.vcxproj]
  (compiling source file '../test.cpp')
```

mawad-amd: Are we compiling with C++17? Seems like C++17 and later should not need Boost according to:
https://github.com/Xilinx/XRT/blob/508aafdd91997e54591f44aa0a8fdec39955c38e/src/runtime_src/core/include/xrt/detail/any.h#L13-L22

mawad-amd: hmm I see `set(CMAKE_CXX_STANDARD 23)` so this linker error shouldn't be there. Maybe `__cplusplus` is not correctly defined for MSVC.

mawad-amd: Ok, I got it to work. You need to use the flag `/Zc:__cplusplus` for MSVC to defined the `__cplusplus` macro that XRT is using. XRT should use `_MSVC_LANG` when the platform is windows and we wouldn't need this. I think for now, we can add
```cmake
if (MSVC)
    add_compile_options(/Zc:__cplusplus)
endif()
```
And remove the Boost dependency. See an example that is correctly compiling [here](https://godbolt.org/z/355hnEss4) and one that is failing (as expected) [here](https://godbolt.org/z/EbfeqE55e).

mawad-amd: Some links for reference:

- [/Zc:__cplusplus](https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170)
- [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html)
- [_MSVC_LANG](https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170)

utils/setup_files/device.h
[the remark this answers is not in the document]
jackl-xilinx: After you build XRT, it is found under `build/Release/opt/xilinx/xrt/include/detail/version.h`

programming_examples/basic/matrix_multiplication/CMakeLists.txt
jackl-xilinx: That's fantastic @mawad-amd . Thanks for finding this solution!

PR #2256, opened 2025-05-01
Memcpy and Scale-Shift programming examples

jgmelber:
This adds two programming examples: memcpy and scale_shift. 

Memcpy is designed to use every column's shimDMA in-out pairs to fully saturate DDR bandwidth. It is a superset of `passthrough_kernel` and `passthrough_dmas`. As such, it can be used as a microbenchmark or as a template for multi-core unary operations. 

Scale Shift is a multi-core example that time-multiplexes the cores to perform first a scale, followed by a bias addition to shift: A * B + C = D, where each is a vector of `bfloat16` values. A "Runtime Parameter" `rtp` is used to switch between `*` and `+` in each core at runtime. `WorkerRuntimeBarrier`s are used to synchronize between the runtime sequence and each `Worker`.


jgmelber:
#2284 should be merged first

PR #2261, opened 2025-05-02
Adding compile MLIR to PDI function

ypapadop-amd:
This PR adds a function to compile MLIR to a PDI via aiecc.


ypapadop-amd:
I don't know if you also want me to move https://github.com/Xilinx/mlir-aie/blob/710ab57b832a4c4fb5b7b9bb6e62535dc28d34a0/python/utils/compile.py#L13 here.

mawad-amd (approved):
Looks good, thanks. I will move the other util compile function to here (there is other ongoing work there and it would be easy if I just do it).

ypapadop-amd:
@jgmelber do we need to have flags for PDI / xclbin, Peano/chess or distinct functions?

jgmelber:
> @jgmelber do we want flags for PDI / xclbin, Peano/chess or distinct functions?

I think we do want options/flags for compilation, but we can wait and see what @mawad-amd is designing

jgmelber (approved):

PR #2267, opened 2025-05-03
draft: no running event for memory bank trace

joeldushouyu:


joeldushouyu:
Note:  The tracing on each CT's memory banks is enabled when the same CT object presents in the list more than once.
As reflected in the change in "vector_scalar_mult" demo

joeldushouyu:
@jackl-xilinx  I have some questions I am unsure of

According to the [AXI4-Stream Interconnect](https://docs.amd.com/r/en-US/am020-versal-aie-ml/AXI4-Stream-Interconnect) in the documentation, there is both a "Core Trace" and "Memory Trace", as shown in the picture.

Thus, I wonder did my change in  [configure_packet_tracing_flow()](https://github.com/Xilinx/mlir-aie/pull/2267/files#diff-a1cba15602cd8cbf39b9921f4402bc176c773a638487c9358745e6474cb64ed8R1140) correctly creates a packet_flow path from "Memory Trace" back to Shimtile by connecting Trace-port 1 back to Shimtile?

Because currently, the output trace json file contains the event headers for my mem_trace, but there is not mem_trace events being recorded. In other words,  there is no event with **PID:1.**

Or is there something else that could be wrong that causes this issue?


Screenshot from 2025-05-03 18-14-53

jackl-xilinx:
@joeldushouyu I pulled this PR and tested it and was able to get core mem events in my trace (vector_scalar_mul oh a phoenix machine) so from that perspective, it seems to work - see below. 
image

I will say that the trace signals coming out isn't what I expected since I see a continuous stream of dma s2mm start tasks which shouldn't be happening. I don't know if this is an issue with misconfiguration in the wrapper code or odd behavior in the trace units. Also, this trace output with the core mem event only occurs on 1 out of 5 runs, which also isn't right. But this explains why you didn't see any core mem events. If you run it a few more times, you might.

I also browsed through your changes and on first glance, appears correct to me. The only thing I'm not sure about is how we chose to indicate the tracing of a core mem by having a tile repeated in the tiles_to_trace array. This seems very unintuitive though I understand why you chose this option since we don't have core mems as objects we can reference. For now, I think that's fine 
and since you've enabled tracing core mem's directly with the wrappers you added, that essentially adds the feature we wanted to add. We just need to investigate further on the unexpected trace results.

joeldushouyu:
I see, I will probably take some time when I am free to try investigate further.
Also, I agree the way of indicating tracing a core mem probably is not the best way. I will probably need to come up with a better method later, after we can figure out the issues with the trace.

joeldushouyu:
In this commit 3de67d563149304bbbe7a71d398def16d2ba1952
I purposefully changed the order of MemEvent. Thus, Event0 correspond to **DM_PARITY_ERROR_BANK_4**

Here is the result of the trace
image

In other words, the same thing happened here.  The 0's in the bitstream got reinterpred as Event0: in this case DM_PARITY_ERROR_4
```json
pkt type 1 :
2,0 [240, 0, 0, 0, 0, 4, 226, 255, 220, 0, 38, 31, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 254, 0, 0,0 .. repeats 0 for about 9000 values
```

I mean it is impossible to have DM_PARITY_ERROR for the whole time, yet the result of vector_scalar_mul is correct, right?

joeldushouyu:
In this commit: 9d373023195ac17b4b469a6c10992ae92c7d9851

I included some lock event that I know is going to happened, below is the trace I got.
image

Below is the commands interpret by the parse_trace.py file

```
pkt type 1 :
2,0
	 {'type': 'Start', 'timer_value': 340036}
	 {'type': 'Single2', 'event': 6, 'cycles': 2838}
	 {'type': 'Single1', 'event': 4, 'cycles': 555}
	 {'type': 'Single0', 'event': 6, 'cycles': 1}
	 {'type': 'Single1', 'event': 4, 'cycles': 552}
	 {'type': 'Single0', 'event': 6, 'cycles': 0}
	 {'type': 'Single1', 'event': 4, 'cycles': 553}
	 {'type': 'Single0', 'event': 6, 'cycles': 0}
	 {'type': 'Single1', 'event': 4, 'cycles': 553}
	 {'type': 'Single0', 'event': 0, 'cycles': 0}...repeats for 9000 lines
	```

In this case, the trace for memtile looks mostly correct.
The only thing that stands out to me is that there are still a thinly bit of

**Comment by joeldushouyu** [comment]
The reason why there should be no "DM_PARITY_ERROR_BANK_4" is obvious.

But below is my theory for why the trace does not see any events such as **DMA_MM2S_0_START_TASK**, **DMA_S2MM_1_START_TASK** and **DMA_MM2S_1_START_TASK**.

Looking at the MLIR file, all trace configuration is configured under runtime_sequenc, as shown below
```mlir
    aiex.runtime_sequence @sequence(%arg0: memref<4096xi16>, %arg1: memref<1xi32>, %arg2: memref<4096xi16>) {
      aiex.npu.write32 {address = 213200 : ui32, column = 0 : i32, row = 2 : i32, value = 2038038528 : ui32}
      aiex.npu.write32 {address = 213204 : ui32, column = 0 : i32, row = 2 : i32, value = 1 : ui32}
      aiex.npu.write32 {address = 213216 : ui32, column = 0 : i32, row = 2 : i32, value = 1260724762 : ui32}
      aiex.npu.write32 {address = 213220 : ui32, column = 0 : i32, row = 2 : i32, value = 556608591 : ui
```
And given the MLIR instruction under runtime_sequenc is completed to a instruction.bin file together with the bitstream file.

I am guessing that during launch, it will first
1, Load the bitstream file that contains the code/setup for CT and MT
2. Load the instruction.bin file that send/receive data through the Shimtile

Ineffectively, this means that the S2MM/MM2S on CT configured and enabled during loading of bitstream (step 1) before
trace unit (at step 2) was able to set up to capture those events.

Thus, this is the reason that explain why the trace unit never sees events such as **DMA_MM2S_0_START_TASK**, **DMA_S2MM_1_START_TASK** and **DMA_MM2S_1_START_TASK**?  @jackl-xilinx

jackl-xilinx:
Thanks @joeldushouyu for doing such a comprehensive analysis and apologies for the delayed response as this week had been very busy. Your efforts are very helpful and I see that the empty packets are being picked up by the parser rather than being ignored so I'm looking more closely at the parser to see how to fix it. I'll update this and propose a fix in the next day or so.

jackl-xilinx:
@joeldushouyu I think I found the issue. It's misinterpreting data that is all 0's as packets with a single event. I've added code to trim these non-packets from the end of the file. I also noticed the enum list of events for coremem is missing some numbers in the middle which sometimes can get the ordering to be off. I've added some placeholders to prevent this. I ran the design with the updates from https://github.com/Xilinx/mlir-aie/pull/2306 and I get sensible event traces. I will say that the start_task still doesn't show up and I think what you said about them occurring before trace is enabled might be on the right track. I chose a different set of events and the resulting waveform looks sensible.

                    coremem_events=[
                        MemEvent.DMA_S2MM_0_FINISHED_BD,
                        MemEvent.DMA_S2MM_0_STREAM_STARVATION,
                        MemEvent.DMA_S2MM_1_FINISHED_BD,
                        MemEvent.GROUP_ERRORS, 
                        MemEvent.GROUP_LOCK, 
                        MemEvent.DMA_MM2S_0_FINISHED_BD,
                        MemEvent.DMA_MM2S_0_FINISHED_TASK,
                        MemEvent.GROUP_MEMORY_CONFLICT,
                    ],

This gives me:
image

which seems reasonable as to what is occurring in the core memory. When my PR is merged, give it a try and see if it works for your experiments. If it does, this PR should be good to merge as well.

jackl-xilinx:
@joeldushouyu My PR fix has been merged. Please go ahead and try it out with this fix and see if things work on your end. If it does, I'll go ahead and merge this PR.

joeldushouyu:
> The only thing I'm not sure about is how we chose to indicate the tracing of a core mem by having a tile repeated in the tiles_to_trace array. This seems very unintuitive though I understand why you chose this option since we don't have core mems as objects we can reference. For now, I think that's fine and since you've enabled tracing core mem's directly with the wrappers you added, that essentially adds the feature we wanted to add. We just need to investigate further on the unexpected trace results.

@jackl-xilinx  Should this be resolved in this PR also? or leave it as a different PR?

jackl-xilinx:
> @jackl-xilinx Should this be resolved in this PR also? or leave it as a different PR?

@joeldushouyu Let's leave it as a separate PR. Please file an github issue to track that so we can improve it.

joeldushouyu:
> @jackl-xilinx One other quick note, Somehow, I can't build (linkage error) the new commit (merged with main) on my laptop. I am going try to revert to find a working commit

It seems to me the error first appeared at the first merge commit.

jackl-xilinx (approved):

jackl-xilinx:
> > @jackl-xilinx One other quick note, Somehow, I can't build (linkage error) the new commit (merged with main) on my laptop. I am going try to revert to find a working commit
> 
> It seems to me the error first appeared at the first merge commit.

Hmm.. I'm not sure why you're seeing a build conflict. Let's see if we see it on the lit tests.

jackl-xilinx (approved):

jackl-xilinx:
@joeldushouyu Were you able to get around the compilation issue and test out the updated main successfully?

joeldushouyu:
> @joeldushouyu Were you able to get around the compilation issue and test out the updated main successfully?

@jackl-xilinx 
Unfortunately, I still get aiecc.py related error when trying to run the example, as shown below. I haven't spend the time to dig into which recent commits that introduce this issue yet.


```bash

Error running pass pipeline:  builtin.module(aie.device(aie-materialize-bd-chains,aie-substitute-shim-dma-allocations,aie-assign-runtime-sequence-bd-ids,aie-dma-tasks-to-npu,aie-dma-to-npu,aie-lower-set-lock)) Failure while executing pass pipeline:
error: "-":160:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":160:7: see current operation: "aiex.dma_free_task"(%21) : (index) -> ()
 note: "-":155:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.
error: "-":161:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":161:7: see current operation: "aiex.dma_free_task"(%19) : (index) -> ()
 note: "-":145:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.
error: "-":162:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":162:7: see current operation: "aiex.dma_free_task"(%20) : (index) -> ()
 note: "-":150:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.

Traceback (most recent call last):
  File "/home/shouyud/mlir-aie/install/bin/aiecc.py", line 12, in <module>
    main()
  File "/home/shouyud/mlir-aie/install/python/aie/compiler/aiecc/main.py", line 1369, in main
    run(module_str)
  File "/home/shouyud/mlir-aie/install/python/aie/compiler/aiecc/main.py", line 1343, in run
    asyncio.run(runner.run_flow())
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/shouyud/mlir-aie/install/python/aie/compiler/aiecc/main.py", line 1144, in run_flow
    npu_insts_module = run_passes_module(
                       ^^^^^^^^^^^^^^^^^^
  File "/home/shouyud/mlir-aie/install/python/aie/compiler/aiecc/main.py", line 346, in run_passes_module
    raise e
  File "/home/shouyud/mlir-aie/install/python/aie/compiler/aiecc/main.py", line 343, in run_passes_module
    pm.run(mlir_module.operation)
aie._mlir_libs._site_initialize.<locals>.MLIRError: Failure while executing pass pipeline:
error: "-":160:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":160:7: see current operation: "aiex.dma_free_task"(%21) : (index) -> ()
 note: "-":155:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.
error: "-":161:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":161:7: see current operation: "aiex.dma_free_task"(%19) : (index) -> ()
 note: "-":145:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.
error: "-":162:7: 'aiex.dma_free_task' op does not reference a valid configure_task operation.
 note: "-":162:7: see current operation: "aiex.dma_free_task"(%20) : (index) -> ()
 note: "-":150:12: Lower this operation first using the --aie-substitute-shim-dma-allocations pass.
```

joeldushouyu:
@jackl-xilinx  Maybe we should merge this pr?
I am not sure if the compile error I got is due to something wrong with my environment(like driver version), maybe it will work fine for binary release version?

jackl-xilinx:
> @jackl-xilinx Maybe we should merge this pr? I am not sure if the compile error I got is due to something wrong with my environment(like driver version), maybe it will work fine for binary release version?

Yes, I agree. I think your error might be environment related. Thanks again for all your help and work to add this feature! It's much appreciated.

PR #2276, opened 2025-05-06
Support for changing cwd in aiecc PDI compile wrapper

ypapadop-amd:
This is to find object files are in a different directory than the current one.

### DIFF
--- a/python/iron/compile/compile.py
+++ b/python/iron/compile/compile.py
@@ -6,11 +6,16 @@
 #
 # (c) Copyright 2025 Advanced Micro Devices, Inc.
 
+import os
 import aie.compiler.aiecc.main as aiecc
 
 
 def compile_mlir_module_to_pdi(
-    mlir_module: str, insts_path: str, pdi_path: str, options: list[str]
+    mlir_module: str,
+    insts_path: str,
+    pdi_path: str,
+    options: list[str],
+    cwd=None,
 ):
     """
     Compile an MLIR module to instruction and PDI files using the aiecc module.
@@ -20,8 +25,8 @@ def compile_mlir_module_to_pdi(
         insts_path (str): Path to the instruction binary file.
         pdi_path (str): Path to the PDI file.
         options (list[str]): List of additional options.
+        cwd: Overrides the current working directory.
     """
-
     args = [
         "--aie-generate-pdi",
         "--aie-generate-npu-insts",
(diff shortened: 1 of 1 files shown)


mawad-amd (approved):
Just to confirm: you had issues with mlir that looks like:
```mlir
module {
  aie.device(npu1_4col) {
  } {link_with = "/some/director/kernel.o"}
}
```

but this works:
```mlir
module {
  aie.device(npu1_4col) {
  } {link_with = "kernel.o"}
}
```

We would really like to have `aiecc` handle object files at some path.

ypapadop-amd:
> Just to confirm: you had issues with mlir that looks like:
> 
> ```mlir
> module {
>   aie.device(npu1_4col) {
>   } {link_with = "/some/directory/kernel.o"}
> }
> ```
> 
> but this works:
> 
> ```mlir
> module {
>   aie.device(npu1_4col) {
>   } {link_with = "kernel.o"}
> }
> ```
> 
> We would really like to have `aiecc` handle object files at some path.

I had problem with the latter, since the module expected to find the .o in the same directory the module is in. The absolute path works, the relative does not.

fifield:
Should the change be to aiecc instead? e.g. `aiecc --kernel-object-search-path` or some other (better named) option?

ypapadop-amd:
> Should the change be to aiecc instead? e.g. `aiecc --kernel-object-search-path` or some other (better named) option?

Perhaps, but I can't tell where is the search path for object files.

jgmelber (approved):
`cwd` is okay, a descriptive, non acronym, name better

ypapadop-amd:
> `cwd` is okay, a descriptive, non acronym, name better

Only chosen since other Python packages (e.g., subprocess) use it extensively. Would you prefer current_working_directory / current_workdir, something else?

fifield:
> > Should the change be to aiecc instead? e.g. `aiecc --kernel-object-search-path` or some other (better named) option?
> 
> Perhaps, but I can't tell where is the search path for object files.

The linker scripts are generated by `aie-translate` in `AIETargetLdScript.cpp` and `AIETargetBCF.cpp` for peano and chess, respectively.

ypapadop-amd:
> > > Should the change be to aiecc instead? e.g. `aiecc --kernel-object-search-path` or some other (better named) option?
> > 
> > 
> > Perhaps, but I can't tell where is the search path for object files.
> 
> The linker scripts are generated by `aie-translate` in `AIETargetLdScript.cpp` and `AIETargetBCF.cpp` for peano and chess, respectively.

If I understand it correctly, the file to be linked is given by `tile.getCoreOp().getLinkWith()`.

I could pass an additional working dir StringRef like in https://github.com/Xilinx/mlir-aie/blob/78d88e1374fe097f04bb0eff863e9e1243935164/include/aie/Targets/AIERT.h#L215 but the problem is that there'll be no way of differentiating between absolute and relative paths, see https://github.com/Xilinx/mlir-aie/blob/78d88e1374fe097f04bb0eff863e9e1243935164/lib/Targets/AIETargetLdScript.cpp#L174

I don't know if that'd be an expected behavior.

PR #2277, opened 2025-05-06
Add key dependencies to the mlir-aie wheel

erwei-xilinx:


fifield:
Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?

cc: @jgmelber

PR #2278, opened 2025-05-06
Add key dependencies to the mlir-aie wheel (#2277)

erwei-xilinx:
Enables pip installing mlir-aie wheel with installation of some key dependencies, including rich and aiofiles.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,4 +1,4 @@
-PyYAML>=5.3.1, <=6.0.1
+PyYAML>=6.0.2
 aiofiles
 cmake>=3.29, <4.0
 dataclasses>=0.6, <=0.8
--- a/utils/mlir_aie_wheels/setup.py
+++ b/utils/mlir_aie_wheels/setup.py
@@ -247,6 +247,12 @@ MLIR_AIE_SOURCE_DIR = Path(
     )
 ).absolute()
 
+def parse_requirements(filename):
+    with open(filename) as f:
+        lines = f.read().splitlines()
+        # Remove comments and empty lines
+        return [line.strip() for line in lines if line.strip() and not line.startswith('#')]
+
 setup(
     version=version,
     author="",
@@ -271,8 +277,5 @@ setup(
     cmdclass={"build_ext": CMakeBuild},
     zip_safe=False,
     python_requires=">=3.10",
-    install_requires=[
-        "rich",
-        "aiofiles"
-    ],
+    install_requires=parse_requirements(Path(MLIR_AIE_SOURCE_DIR) / "python" / "requirements.txt"),
 )
--- a/utils/mlir_aie_wheels/setup.py
+++ b/utils/mlir_aie_wheels/setup.py
@@ -271,4 +271,8 @@ setup(
     cmdclass={"build_ext": CMakeBuild},
     zip_safe=False,
     python_requires=">=3.10",
+    install_requires=[
+        "rich",
+        "aiofiles"
+    ],
 )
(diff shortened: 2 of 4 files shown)


jgmelber:
Great! Please update the READMEs, `utils/quick_setup.sh`, `requirements.txt` once this works

erwei-xilinx:
> Great! Please update the READMEs, `utils/quick_setup.sh`, any `requirements.txt` once this works

Will do, and will verify with the new wheel being generated from this branch.

erwei-xilinx:
@jgmelber https://github.com/Xilinx/mlir-aie/blob/c2fa2fa4a8443076c9c0f3917797f765481b05af/python/requirements.txt#L1 Is the  PyYAML <= 6.0.1 requirement intended, and can we update to 6.0.2? Currently the wheel is successfully installing mlir-aie wheel with its dependencies, but throws an error because mlir-python-extras requires >=6.0.2.

jgmelber:
> @jgmelber https://github.com/Xilinx/mlir-aie/blob/c2fa2fa4a8443076c9c0f3917797f765481b05af/python/requirements.txt#L1 Is the  PyYAML <= 6.0.1 requirement intended, and can we update to 6.0.2? Currently the wheel is successfully installing mlir-aie wheel with its dependencies, but throws an error because mlir-python-extras requires >=6.0.2.

It's fine to increase if tests pass. The specific versions are intentional but occasionally need to be bumped

fifield:
adding my comment from other PR here:

Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?

erwei-xilinx:
> adding my comment from other PR here:
> 
> Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?

No I couldn't find one such workflow which tests the wheel. We were testing the wheel from triton-air, and that's where we realized it doesn't pip install with its dependencies. But it would definitely be useful to have one CI workflow that checks the wheel.

erwei-xilinx:
> Great! Please update the READMEs, `utils/quick_setup.sh`, any `requirements.txt` once this works

`utils/quick_setup.sh` updated. I couldn't update the README as it is focused on the released 1.0 wheel, which doesn't yet include this change.

erwei-xilinx:
> It's fine to increase if tests pass. The specific versions are intentional but occasionally need to be bumped

I bumped the pyyaml version and confirmed that all tests are passing.

fifield:
> > adding my comment from other PR here:
> > Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?
> 
> The buildAndTestRyzenAI.yml calls quick_setup.sh which installs the wheel, and perform checks. We were testing the wheel from triton-air, and that's where we realized it doesn't pip install with its dependencies.

That action builds, installs and tests mlir-aie from source. The wheel is only used for running quick_setup.sh.

fifield:
> adding my comment from other PR here:
> 
> Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?

https://github.com/Xilinx/mlir-aie/issues/2285

erwei-xilinx:
> > > adding my comment from other PR here:
> > > Do we need a CI workflow to test this? i.e. test the wheel without installing the requirements.txts first?
> > 
> > 
> > The buildAndTestRyzenAI.yml calls quick_setup.sh which installs the wheel, and perform checks. We were testing the wheel from triton-air, and that's where we realized it doesn't pip install with its dependencies.
> 
> That action builds, installs and tests mlir-aie from source. The wheel is only used for running quick_setup.sh.

Ok, good to know. So I think that answers my issue https://github.com/Xilinx/mlir-aie/issues/2286

PR #2281, opened 2025-05-07
Use correct functions to retrieve DMA channels for shim

AndraBisca:


lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -106,15 +106,20 @@ class DMAChannelAnalysis {
...
-      else
-        maxChannelNum = targetModel.getNumDestSwitchboxConnections(
-            tileOp.getCol(), tileOp.getRow(), WireBundle::DMA);
+    if (dir == DMAChannelDir::MM2S) {
+      maxChannelNum =
+          tileOp.isShimTile()
+              ? targetModel.getNumSourceShimMuxConnections(
+                    tileOp.getCol(), tileOp.getRow(), WireBundle::DMA)
+              : targetModel.getNumSourceSwitchboxConnections(
+                    tileOp.getCol(), tileOp.getRow(), WireBundle::DMA);
```
fifield, line 115: Does `TileOp::getNumSourceConnections` / `TileOp::getNumDestConnections` work here?

AndraBisca: Ah I must've missed these! It's exactly what should be used. Let me update the PR. Thank you Jeff!

AndraBisca: Although I do find the comment regarding the reversed direction a bit confusing in these functions. For example, both of them use `getNumDestShimMuxConnections`..

PR #2290, opened 2025-05-08
Remove `pip install -r requirements.txt` from `quick_setup.sh`

erwei-xilinx:
… as the new wheel should install with dependency check.


fifield (approved):

jgmelber (approved):
Please update the READMEs too for this change

PR #2298, opened 2025-05-09
Guide Section 1: replace GlobalBuffer with LocalBuffer

AndraBisca:


programming_guide/section-1/README.md
```diff
@@ -39,18 +39,23 @@ class Worker(ObjectFifoEndpoint):
...
+# Task for the worker to perform
+def core_fn():
+    local = LocalBuffer(
+        data_ty,
+        name="local",
+        initial_value=np.array(range(data_size), dtype=np.int32),
+    )
     for i in range_(data_size):
-        buff_in[i] = buff_in[i] + 1
+        local[i] = local[i] + 1
```
denolf, line 52: Initialization of the local buffer might make more sense
local[I] = 0

denolf (changes requested):
Thanks for this update, this is an easier to grasp example at this stage of the programming guide.
I propose to just initialize the local buffer to 0 in the worker, it feels strange to add 1 to random and potentially rubbish values residing in the uninitialized local buffer.

denolf (approved):

PR #2305, opened 2025-05-12
Vector Reduce Max single column designs

pvasireddy-amd:
Updated single core designs to support both bf16 and i32 datatypes.

Added 3 single-column designs under the `single_col_designs` folder, each supporting both bf16 and i32:
- [x] `vector_reduce_max_shared.py`: Uses shared memory of the neighboring core tiles
- [x] `vector_reduce_max_cascade.py`: Cascades partial results across cores of the column
- [x] `vector_reduce_max_memtile.py`: Uses Memtile to aggregate the intermediate results

This PR includes only placed designs.


aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
+  V third = max(second, second_shift);
+  V third_shift = shift_bytes(third, third, 8U);
+  V fourth = max(third, third_shift);
+  V fourth_shift = shift_bytes(fourth, fourth, 4U);
+  V fifth = max(fourth, fourth_shift);
+  if constexpr (std::is_same<V, aie::vector<bfloat16, 32>>::value) {
+    V fifth_shift = shift_bytes(fifth, fifth, 2U);
+    fifth = max(fifth, fifth_shift);
+  }
+  auto last = extract_elem(fifth, 0U);
```
andrej: Would AIE API's [`aie::reduce_max()`](https://download.amd.com/docnav/aiengine/xilinx2025_1/aiengine_api/aie_api/doc/group__group__reduce.html#gaa76b524ab8ffa3070184336d8a8806d7) be more clear?

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
+  V after_vector;
+  V running_max = tiny;
+
   AIE_PREPARE_FOR_PIPELINING
   AIE_LOOP_MIN_ITERATION_COUNT(8)
-  for (int32_t i = 0; i < input_size; i += vector_size) {
-    v16int32 next = *(v16int32 *)(in + i);
-    v16int32 test = max(running_max, next);
+  for (int32_t i = 0; i < input_size; i += VECTOR_SIZE) {
+    V next = *(V *)(in + i);
```
andrej: Would AIE API's `aie::load_v()` be more clear than this cast?

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
 #include "../aie_kernel_utils.h"
 #include <aie_api/aie.hpp>
 
-void _reduce_max_vector(int32_t *restrict in, int32_t *restrict out,
+template <typename T, typename V, int VECTOR_SIZE>
+void _reduce_max_vector(T *restrict in, T *restrict out,
                         const int32_t input_size) {
```
andrej:
```suggestion
template <typename T, unsigned VECTOR_SIZE>
void _reduce_max_vector(T *restrict in, T *restrict out,
                        const int32_t input_size) {
    using V = aie::vector<T, VECTOR_SIZE>;
```

Even more clear would be, i.m.o., to use `aie::vector<T, VECTOR_SIZE>` explicitly everywhere but that may be a matter of taste.

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
+    tiny = broadcast_to_v32bfloat16(
+        (bfloat16)std::numeric_limits<bfloat16>::lowest());
+  } else {
+    static_assert("Unsupported vector type");
+  }
+  V after_vector;
+  V running_max = tiny;
+
   AIE_PREPARE_FOR_PIPELINING
   AIE_LOOP_MIN_ITERATION_COUNT(8)
```
andrej, line 31: `assert(input_size / VECTOR_SIZE >= 8);`?

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
-  for (int32_t i = 0; i < input_size; i += vector_size) {
-    v16int32 next = *(v16int32 *)(in + i);
-    v16int32 test = max(running_max, next);
+  for (int32_t i = 0; i < input_size; i += VECTOR_SIZE) {
```
andrej, line 32: You have a template `VECTOR_SIZE`, suggesting any vector size is supported, but then vector sizes 16 for int32 and 32 for bfloat16 are hard-coded above (but not respected here). I suggest removing `VECTOR_SIZE` from the template list and fixing it as a constexpr in the if/else above, or use `VECTOR_SIZE`  in the if/else above as well.

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
+  V third_shift = shift_bytes(third, third, 8U);
+  V fourth = max(third, third_shift);
+  V fourth_shift = shift_bytes(fourth, fourth, 4U);
+  V fifth = max(fourth, fourth_shift);
+  if constexpr (std::is_same<V, aie::vector<bfloat16, 32>>::value) {
+    V fifth_shift = shift_bytes(fifth, fifth, 2U);
+    fifth = max(fifth, fifth_shift);
+  }
+  auto last = extract_elem(fifth, 0U);
+  *(T *)out = last;
```
andrej, line 52: Would `aie::store_v()` be more clear?

aie_kernels/aie2/reduce_max.cc
```diff
@@ -16,58 +16,100 @@
...
-  v16int32 running_max = tiny;
+  V tiny;
+  if constexpr (std::is_same<V, aie::vector<int32_t, 16>>::value) {
+    tiny = broadcast_to_v16int32((int32_t)INT32_MIN);
+  } else if constexpr (std::is_same<V, aie::vector<bfloat16, 32>>::value) {
+    tiny = broadcast_to_v32bfloat16(
+        (bfloat16)std::numeric_limits<bfloat16>::lowest());
+  } else {
+    static_assert("Unsupported vector type");
+  }
```
andrej:
```suggestion
  V tiny = aie::broadcast<T, VECTOR_SIZE>(std::numeric_limits<T>::lowest());
```

programming_examples/basic/vector_reduce_max/Makefile
```diff
@@ -9,24 +9,39 @@
...
 aie_py_src=${targetname}.py
 use_placed?=0
 
 ifeq (${use_placed}, 1)
 aie_py_src=${targetname}_placed.py
+dtype=i32
+else
+aie_py_src=${targetname}.py
+dtype=i32
+endif
```
andrej:
```suggestion
ifeq (${use_placed}, 1)
aie_py_src=${targetname}_placed.py
else
aie_py_src=${targetname}.py
endif

dtype=i32
```

programming_examples/basic/vector_reduce_max/single_col_designs/Makefile
```diff
@@ -0,0 +1,107 @@
...
+
+dtype?=i32
+
+aie_py_src=${targetname}.py
+use_shared?=0
+use_memtile?=0
+
+ifeq (${use_shared}, 1)
+aie_py_src=${targetname}_shared.py
+dtype=bf16
```
andrej:
```suggestion
```

programming_examples/basic/vector_reduce_max/single_col_designs/Makefile
```diff
@@ -0,0 +1,107 @@
...
+else ifeq (${use_memtile}, 1)
+aie_py_src=${targetname}_memtile.py
+dtype=bf16
```
andrej:
```suggestion
```

programming_examples/basic/vector_reduce_max/single_col_designs/Makefile
```diff
@@ -0,0 +1,107 @@
...
+else 
+aie_py_src=${targetname}_cascade.py
+dtype=bf16
```
andrej:
```suggestion
```

programming_examples/basic/vector_reduce_max/single_col_designs/Makefile
```diff
@@ -0,0 +1,107 @@
...
+endif
+
```
andrej, line 36:
```suggestion

dtype=bf16
```

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+                "reduce_max_vector_bfloat16", inputs=[op_ty, out_ty, np.int32]
+            )
+            compute_max = external_func(
+                "compute_max_bfloat16", inputs=[out_ty, out_ty, out_ty]
+            )
+        else:
+            reduce_max_vector = external_func(
+                "reduce_max_vector", inputs=[op_ty, out_ty, np.int32]
+            )
+            compute_max = external_func("compute_max", inputs=[out_ty, out_ty, out_ty])
```
andrej: Nit: You could just `external_func(f"reduce_max_vector_{dtype_str}")` and rename the functions accordingly

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+
+        # Tile declarations
+        ShimTile = tile(0, 0)
+        MemTile = tile(0, 1)
+
+        cores = [tile(0, 2 + i) for i in range(n_cores)]
+
+        inA_fifos = []
+        outC_fifos = []
```
andrej, line 70: Nit: Why are these named A and C (and no B)? Could just be input and output

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+dtype_map = {
+    "bf16": bfloat16,
+    "i32": np.int32,
+}
+
+
+def my_reduce_max(dev, in1_size, out_size, dtype_str, trace_size):
+    n_cores = 4
+    in_dtype = dtype_map[dtype_str]
+    out_dtype = dtype_map[dtype_str]
```
andrej: Why not just `dtype` if the data type is the same for both input and output to make it easier for the reader?

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+
+        object_fifo_link(of_in, inA_fifos, [], of_a_offsets)
+
+        # Set up a packet-switched flow from core to shim for tracing information
+        tiles_to_trace = [cores[0]]
+        if trace_size > 0:
+            trace_utils.configure_packet_tracing_flow(tiles_to_trace, ShimTile)
+
+        if dtype_str == "bf16":
+            min_val = np.array([bfloat16(float(-4.0))], dtype=bfloat16)
```
andrej: This will break for real-world values (your test bench does not test anything smaller than `-4.0` but in other use cases inputs may be smaller).

programming_examples/basic/vector_reduce_max/single_col_designs/Makefile
```diff
@@ -0,0 +1,107 @@
...
+srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+
+include ${srcdir}/../../../makefile-common
+
+VPATH := ${srcdir}/../../../../aie_kernels/aie2
+
+devicename ?= $(if $(filter 1,$(NPU2)),npu2,npu)
+targetname = vector_reduce_max
+in1_size =  524288# in bytes
+out_size = 4 # in bytes, should always be equal to in1_size
```
andrej: I think the comment here is misleading/wrong.

If the max is always just a single scalar value, why have this parameter at all? I.m.o. it would make the Python code easier to understand not to have this parameter.

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+                datatype=np.ndarray[(O,), np.dtype[out_dtype]],
+                name=f"elem_out_{i}",
+                initial_value=min_val,
+            )
+            tmp_buffer = buffer(
+                tile=cores[i],
+                datatype=np.ndarray[(O,), np.dtype[out_dtype]],
+                name=f"tmp_buffer_{i}",
+                initial_value=min_val,
+            )
```
andrej: These buffers each hold a single scalar value, correct? Why the indirection through memory? You may get a slight bump in performance if you let these be in registers (i.e., functions return the max value, do not store in memory).

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
```diff
@@ -0,0 +1,219 @@
...
+                @core(cores[i], "reduce_max.cc.o")
+                def core_body():
+                    elem_out = outC_fifos[i].acquire(ObjectFifoPort.Produce, 1)
+                    for _ in range_(num_iter):
+                        elem_in = inA_fifos[i].acquire(ObjectFifoPort.Consume, 1)
+                        reduce_max_vector(elem_in, tmp_buffer, elems_per_core)
+                        compute_max(nextC_buffer, tmp_buffer, nextC_buffer)
+                        inA_fifos[i].release(ObjectFifoPort.Consume, 1)
+                    elem_out[0] = nextC_buffer[0]
+                    outC_fifos[i].release(ObjectFifoPort.Produce, 1)
```
andrej, line 131: As discussed above, you can get rid of the indirection through memory. You'd have to change the kernel functions to return, rather than store, the max. Then this would become something like:

```suggestion
                    elem_out = outC_fifos[i].acquire(ObjectFifoPort.Produce, 1)
                    running_max = arith.constant(T.i32(), min_val)  # <- todo switch T.bf16() for the other case
                    for _, _, final_max in range_(num_iter, iter_args=[running_max], insert_yield=False):
                        elem_in = inA_fifos[i].acquire(ObjectFifoPort.Consume, 1)
                        local_max = reduce_max_vector(elem_in, elems_per_core)
                        running_max = compute_max(local_max, running_max)
                        inA_fifos[i].release(ObjectFifoPort.Consume, 1)
                        yield_([running_max])
                    elem_out[0] = final_max
                    outC_fifos[i].release(ObjectFifoPort.Produce, 1)
```

You'd need:
```
from aie.dialects.scf import yield_
from aie.dialects.aie import T
import aie.dialects.arith as arith
```

It would be nice to have an example like this with a little more advanced loop in the repository so we could point to this, and this would be a good opportunity to add it. (And again, it may be slightly faster to keep the result in registers.)

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_memtile.py
```diff
@@ -0,0 +1,243 @@
...
+                @core(cores[i], "reduce_max.cc.o")
+                def core_body():
+                    elem_out = outC_fifos[i].acquire(ObjectFifoPort.Produce, 1)
+                    for _ in range_(num_iter):
+                        elem_in = inA_fifos[i].acquire(ObjectFifoPort.Consume, 1)
+                        reduce_max_vector(elem_in, tmp_buffer, elems_per_core)
+                        compute_max(nextC_buffer, tmp_buffer, nextC_buffer)
+                        inA_fifos[i].release(ObjectFifoPort.Consume, 1)
+                    elem_out[0] = nextC_buffer[0]
+                    outC_fifos[i].release(ObjectFifoPort.Produce, 1)
```
andrej: Instead of duplicating all this from above, why not put the `if i != 0` statement inside of the core and only emit the additional code below if `i == 0`?

andrej (commented):
Looks mostly good to me. Most of my comments are nitpicks to make the code shorter and easier to read. I don't know if there was a requirement/desire (for the purpose of exhibition) not to use the AIE API in this code, since everything is at the intrinsic level; if so, please disregard my comments regarding that. 

My two most important critiques would be (1) removing unused/duplicate parameters (e.g. output size is always one scalar value; input and output data types are always the same, so don't need separate variables) and (2) avoiding the indirection through memory for results.

Other than that looks good! Let me know what you think of these points.

aie_kernels/aie2/reduce_max.cc
pvasireddy-amd: `last` is a scalar value

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
pvasireddy-amd: Agreed. This is a data-bound design, so the performance impact may be limited. I’ll include this change in a follow-up PR focused on kernel performance improvements, as it requires changing the function definition to return the result instead of storing it in memory.

programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_cascade.py
pvasireddy-amd: Will include in the follow-up PR.

AndraBisca (approved):
LGTM! :)

PR #2311, opened 2025-05-14
fixed vector_reduce_add example

userVector:
original programming example didn't work with test.cpp from other examples. the issue apparently stemmed from distinct kernel syntax requirements associated with the older way of managing host-NPU buffers.

now works with the newer host-NPU buffer method already present in most other examples.
furthermore: supports tiled computation. kernel code uses templates.

### DIFF
--- a/aie_kernels/aie2/reduce_add.cc
+++ b/aie_kernels/aie2/reduce_add.cc
@@ -13,35 +13,32 @@
 #include <stdlib.h>
 #include <type_traits>
 
-#include "../aie_kernel_utils.h"
 #include <aie_api/aie.hpp>
 
-static void _reduce_add_scalar(int32_t *restrict in, int32_t *restrict out,
-                               const int32_t input_size) {
+template <typename T_in, typename T_out, const int N>
+void reduce_add_scalar_int32(T_in *a, T_out *c) {
   event0();
-  int32_t running_total = 0;
-  for (int32_t i = 0; i < input_size; i++) {
-    running_total = running_total + in[i];
+  int32_t sum=0;
+  for (int i = 0; i < N; i++) {
+    sum = sum + a[i];
   }
-  *out = running_total;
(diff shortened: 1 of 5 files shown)


aie_kernels/aie2/reduce_add.cc
```diff
@@ -13,35 +13,32 @@
...
-                               const int32_t input_size) {
+template <typename T_in, typename T_out, const int N>
+void reduce_add_vector_int32(T_in *in, T_out *out) {
   event0();
   v16int32 zero = broadcast_to_v16int32((int32_t)0);
   const int32_t vector_size = 16;
   v16int32 after_vector;
   v16int32 running_total = zero;
-  AIE_PREPARE_FOR_PIPELINING
-  AIE_LOOP_MIN_ITERATION_COUNT(8)
```
fifield, line 39: This appears to revert the changes from https://github.com/Xilinx/mlir-aie/pull/2248

programming_examples/basic/vector_reduce_add/CMakeLists.txt
fifield, line 1: Are any of the changes to this file required? It appears to be reverting recent changes such as https://github.com/Xilinx/mlir-aie/pull/2234?

fifield (changes requested):
This PR appears to revert some recent changes, please clean it up to ensure only the necessary changes are included.

hunhoffe:
Thanks for putting this together, and apologies it sat so long without a response. I'm doing a pass over older PRs to keep the queue current, and I'm going to close this one out for now since it's been inactive for a while and main has moved on quite a bit (the surrounding examples have changed, so this would need a rebase to apply cleanly).

This isn't a judgment on the change itself — closing is completely reversible. If the fix is still relevant and you'd like to pick it back up, please rebase on the latest main and reopen, and I'll take a look. Thanks again for taking the time to contribute.

PR #2319, opened 2025-05-16
Fix hiding progress bar

ypapadop-amd:
This PR hides the progress bar fully if `aiecc --progress=False` to avoid extra output during JIT.


fifield (approved):
I didn't even realize that `progress=False` is the default because it was broken. A few status message from child processes still get through after this PR, but it's an improvement.

PR #2334, opened 2025-05-22
Add New Transpose Example to Support Varying Sizes and Data Types, Use AIE API

andrej:
From the README:

* Supports matrices of different sizes `M`, `N`.
* Supports transposing matrices of different data types, even data types that are smaller than the DMA access granularity (e.g. `i16`).
* The input matrix is tiled into sub-tiles of size `m`, `n` for processing on the compute core.
* After the transpose kernel has executed on on the compute core, the `m`x`n` tiles must be completely transposed, i.e., no transpose of the `m`x`n`-sized subtiles using the output DMAs may occur. This allows this design to be modified and combined with other computations on the core after the transpose took place ("kernel fusion").


HashimSharifAMD:
The new documentation looks good! The transform example is flexible to choices of M and N.

jgmelber (approved):

PR #2337, opened 2025-05-23
draft: Single CT read control_packet

joeldushouyu:
@erwei-xilinx 

This is the PR that I was hoping to use as an example for the control-packet feature.

## Background:
In this passthrough program, I am trying to send a control packet to read a certain register and send the result of that register value back to host.

Currently, the program does not work correctly due to the potential issue I discussed earlier.

I defined [packet_flow](https://github.com/joeldushouyu/mlir-aie/blob/c9660a52ec705a13f876a475afca6b33d3decbec/programming_examples/basic/passthrough_control_packet/passthrough_aie.py#L133) path for  
1. Control packet from CT(Compute Tile) -> TileControl(packet_id = 13)
2. TileControl -> send result back to CT (packet_id = 14).

## ISSUE 
As of right now, the kernel stuck at this [line](https://github.com/joeldushouyu/mlir-aie/blob/c9660a52ec705a13f876a475afca6b33d3decbec/programming_examples/basic/passthrough_control_packet/kernel/kernel.cc#L86).
After doing some analysis, I found something strange with how the packet_flow is being translated, and I suspect that could be the reason for why the read result of control packet not being send back.

Looking at the switchbox dialect in "build/input_physical.mlir", I found the section below to be strange.

```mlir
    %tile_0_2 = aie.tile(0, 2) {allocation_scheme = "basic-sequential", controlled_id = #aie.packet_info<pkt_type = 0, pkt_id = 26>, controller_id = #aie.packet_info<pkt_type = 0, pkt_id = 27>}
    %switchbox_0_2 = aie.switchbox(%tile_0_2) {
      %0 = aie.amsel<0> (0)
      %1 = aie.amsel<1> (0)
      %2 = aie.amsel<2> (0)
      %3 = aie.amsel<3> (0)
      %4 = aie.amsel<4> (0)
      %5 = aie.masterset(DMA : 0, %0)
      %6 = aie.masterset(DMA : 1, %3)
      %7 = aie.masterset(South : 1, %2)
      %8 = aie.masterset(South : 3, %1)
      %9 = aie.masterset(TileControl : 0, %4)
      aie.packet_rules(Core : 0) {
        aie.rule(31, 14, %3)
      }
      aie.packet_rules(DMA : 1) {
        aie.rule(31, 13, %4)
      }
      aie.packet_rules(DMA : 0) {
        aie.rule(31, 9, %2)
      }
      aie.packet_rules(South : 2) {
        aie.rule(31, 6, %0)
      }
      aie.packet_rules(Trace : 0) {
        aie.rule(31, 1, %1)
      }
    }

```
While it make sense for packet_flow 13 being send from DMA-1 - > %2 -> %9(TileControl 0)
But I expected to see packet_id 14 under TileControl(as shown below) instead of Core. If this were indeed an bug, 
then this explains why the kernel is stuck at this case. 
```mlir
aie.packet_rules(TileControl:0){
  aie.rule(31, 14, %3)
}
```

### DIFF
--- a/programming_examples/basic/passthrough_control_packet/CT_0_2.py
+++ b/programming_examples/basic/passthrough_control_packet/CT_0_2.py
@@ -1,0 +1,155 @@
+from re import L
+from struct import pack
+import numpy as np
+import sys
+
+from aie.dialects.aie import *
+from aie.dialects.aiex import *
+from aie.extras.context import mlir_mod_ctx
+from aie.helpers.dialects.ext.scf import _for as range_
+import math
+
+import numpy as np
+import sys
+
+from ml_dtypes import bfloat16
+from aie.extras.context import mlir_mod_ctx
+from aie.dialects.aie import *
+from aie.dialects.aiex import *
+from aie.helpers.dialects.ext.scf import _for as range_
+
+import numpy as np
+import sys
+
+from aie.dialects.aie import *
+from aie.dialects.aiex import *
+from aie.extras.context import mlir_mod_ctx
+from aie.helpers.dialects.ext.scf import _for as range_
+
+from aie.dialects import memref
+
+from aie.dialects._aie_ops_gen import buffer as buffer_raw
+from aie.helpers.util import try_convert_np_type_to_mlir_type
+import numpy as np
+import sys
+import aie.utils.trace as trace_utils
+from aie.utils.trace import PortEvent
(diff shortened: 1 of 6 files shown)


joeldushouyu:
Thanks for the clarification! That makes sense in this case.

Here is my plan for the example(if we still want to move on adding  a control packet example into the repo)
I will configure 2 CT with Shimtile.

1. CT 0 will send read/write control packet to CT 1 and to the shim tile
2. The result of the read control packet will be sent back to the host for verification.

Any ideas?

PR #2332, opened 2025-05-22
update build_driver.sh due to driver bump

joeldushouyu:
In commit : 5a67804a8458dfb0afc227ee2f38c8dff7fff62f

It used a new driver version. However, that commit did not change the corresponding deb file for XRT.

```bash
(ironenv) shouyud@ser11-NucBox-EVO-X1:~/develope_mlir/mlir-aie/xdna-driver/xrt/build/Release$ ls -la
.....
-rw-rw-r--  1 shouyud shouyud 6276324 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-base.deb
-rw-rw-r--  1 shouyud shouyud 4085634 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-base-dev.deb
-rw-rw-r--  1 shouyud shouyud 4082501 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-base-dev.tar.gz
-rw-rw-r--  1 shouyud shouyud 6274474 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-base.tar.gz
-rw-rw-r--  1 shouyud shouyud 1728944 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-npu.deb
-rw-rw-r--  1 shouyud shouyud 1728071 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-npu.tar.gz
-rw-rw-r--  1 shouyud shouyud     578 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-runtime.deb
-rw-rw-r--  1 shouyud shouyud 1334632 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-Runtime.deb
-rw-rw-r--  1 shouyud shouyud      29 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-runtime.tar.gz
-rw-rw-r--  1 shouyud shouyud 1333518 May 22 14:54 xrt_202520.2.20.0_24.04-amd64-Runtime.tar.gz
-rw-rw-r--  1 shouyud shouyud    2609 May 22 14:53 xrt-config.cmake
```

```bash
(ironenv) shouyud@ser11-NucBox-EVO-X1:~/develope_mlir/mlir-aie/xdna-driver/build/Release$ ls -la
...
drwxrwxr-x  4 shouyud shouyud     4096 May 22 14:55 xrt
[...]

PR #2341, opened 2025-05-27
Fix file names in `build_drivers.sh`

andrej:
#2324 changed the commit SHA that we check out from the XDNA repository in the `./utils/build_drivers.sh` script without updating the version numbers of the `.deb` files that get built and installed. That version number changed in the upstream XRT repository [PR #8896](https://github.com/Xilinx/XRT/pull/8896).


fifield:
see https://github.com/Xilinx/mlir-aie/pull/2332

andrej:
Ah jeez, sorry. Thanks for pointing me to that

PR #2344, opened 2025-05-28
Add runtime sequence section to mini-tutorial

AndraBisca:


programming_guide/mini_tutorial/README.md
```diff
@@ -167,6 +167,37 @@ More on the Object FIFO data movement patterns in [Section 2b](../section-2/sect
...
+rt = Runtime()
+with rt.sequence(tile_ty, tile_ty) as (a_in, c_out):
+    rt.start(my_worker)
+    rt.fill(of_in.prod(), a_in)
+    rt.drain(of_out.cons(), c_out, wait=True)
+```
+
+Up to five buffers are supported in the runtime sequence, where the fifth is typically used for trace support. This is further described in [Section 4b](../section-4/section-4b/README.md) of the programming guide.
+
+Runtime sequence commands are submitted to and executed by a dedicated command processor in order. Commands that are set to `wait` will block the dedicated command processor until a token associated with their completion is generated. The command processor waits for the runtime sequence to complete before returning by interrupting the host processor.
```
fifield: > The command processor waits for the runtime sequence to complete before returning by interrupting the host processor.

I'm not sure this is entirely true. The command processor will wait if it's running a wait op. The host code will wait if it's executing an `xrt::wait` call. But there is no requirement that either must wait. A host program could fire off multiple runtime sequences, without waiting, and sync by some other means (e.g. polling on a memory location). One example of this is [an xrt runlist](https://github.com/Xilinx/mlir-aie/blob/main/test/npu-xrt/add_one_two_txn/test.cpp#L143-L179). Are the semantics of an "iron" runtime sequence different?

jgmelber: Jeff is correct in his assessment. The current wording is a bit to strong on the absolutes of the synchronization. There are more nuances to consider.

AndraBisca: Thank you for the detailed feedback. I adjusted the README section to a less absolute explanation.

programming_guide/mini_tutorial/README.md
```diff
@@ -169,6 +169,37 @@ More on the Object FIFO data movement patterns in [Section 2b](../section-2/sect
...
+rt = Runtime()
+with rt.sequence(tile_ty, tile_ty) as (a_in, c_out):
+    rt.start(my_worker)
+    rt.fill(of_in.prod(), a_in)
+    rt.drain(of_out.cons(), c_out, wait=True)
+```
+
+Up to five buffers are supported in the runtime sequence, where the fifth is typically used for trace support. This is further described in [Section 4b](../section-4/section-4b/README.md) of the programming guide.
+
+Runtime sequence commands are submitted to and executed by a dedicated command processor in order. The command processor will wait on commands that are set to `wait` until a token associated with their completion is generated. When all the commands in the runtime sequence have been executed the command processor sends an interrupt to the host processor.
```
jgmelber, line 193: It might be good to mention `task_groups` here to optimize the grouping of waits for parallel tasks: https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/basic/memcpy/README.md

AndraBisca: Done! Could you please review the changes?

Issue #2092, opened 2025-03-04
NPU2 Programming Examples


jgmelber:
Work started in #1986. Continued in #2163.

vision:
- [ ] fix color_detect (currently not run on Strix)
- [ ] add virtual NPU2 device to fix color_threshold
- [ ] add virtual NPU2 device to edge_detect

basic:
- [ ] fix tiling_exploration 
- [ ] matrix_multiplication:
   - [ ] add virtual NPU2 device to cascade
   - [ ] add virtual NPU2 device to matrix_vector
   - [ ] add virtual NPU2 device to single_core
- [ ] add virtual NPU2 device to matrix_scalar_add
- [ ] add virtual NPU2 device to passthrough_kernel
- [ ] add virtual NPU2 device to passthrough_pykernel
- [ ] add virtual NPU2 device to row_wise_bias_add
- [x] fix to vector_exp (currently not run on Strix)

ml:
- [x] fix conv2d
- [x] fix conv2d_fused_relu
- [x] fix bottleneck (currently not run on Strix)
- [x] fix resnet (currently not run on Strix)

jgmelber:
Add an environment variable to set NPU2 in Makefiles

jgmelber:
> * [ ]  Add an environment variable to set NPU2 in Makefiles

#2103

AndraBisca:
vector_exp updated by #2163

jgmelber:
#2189 helps any LUT kernels for AIE2P

jgmelber:
Many of these examples need single core kernels for aie2p in addition to the currently available aie2 versions. This requires slight adaptations specific to the architecture still using the [aie_api](https://xilinx.github.io/aie_api/) C++ library.  The kernels are generally located here: https://github.com/Xilinx/mlir-aie/tree/main/aie_kernels

AndraBisca:
#2345 

[x] fix bottleneck
[x] fix resnet

PR #2345, opened 2025-05-29
Increase stack size for conv skip to enable resnet and bottleneck designs on Strix

mariodruiz:
Increase the stack size of the `conv2dk1_skip` kernel to enable resnet and bottleneck designs on Strix

Part of #2092


AndraBisca:
Thank you for looking into this! Could you please also turn on the strix .lit files (add a `RUN:` before the last commands) so the tests are run in the CI?

mariodruiz:
@AndraBisca lit tests enabled

AndraBisca (approved):
LGTM!

PR #2358, opened 2025-06-02
Update get_trace_summary for multiple cores

jackl-xilinx:


fifield:
Is there a test to go with this feature?

PR #2360, opened 2025-06-03
Support aiebu elf format for instruction binaries

fifield:
* add aiebu as a submodule
* aiecc.py: generate insts.bin.elf alongside insts.bin (during `--aie-generate-npu-insts`)
* add test which uses `xrt::elf` in test.cpp instead of loading insts.bin into a buffer


fifield:
replaced by #2395

PR #2368, opened 2025-06-04
Update test after #2361

fifield:
Fix test after #2361


jackl-xilinx:
@fifield How strange. The changes to test.cpp should already be part of PR#2361 but either way, it should be added. Thanks!

PR #2375, opened 2025-06-04
Bump windows workflows to version windows-2022

fifield:
Update due to `windows-2019` deprecation noticed by @pvasireddy-amd 

Windows workflows are printing the following notice:
"`The windows-2019 runner image is being deprecated, consider switching to windows-2022(windows-latest) or windows-2025 instead. For more details see https://github.com/actions/runner-images/issues/12045.`"

### DIFF
--- a/utils/mlir_aie_wheels/pyproject.toml
+++ b/utils/mlir_aie_wheels/pyproject.toml
@@ -48,4 +48,4 @@ before-build = [
 ]
 
 [build-system]
-requires = ["nanobind"]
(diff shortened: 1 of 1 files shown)


fifield:
Failing with error:
```
ninja: error: 'C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/DIA SDK/lib/amd64/diaguids.lib', needed by 'bin/aie-opt.exe', missing and no known rule to make it
```

PR #2382, opened 2025-06-06
Remove colshift from parse_trace.py in programming examples

fifield:
Remove `colshift` as a widely used command line argument for the programming_examples `parse_trace.py` script.

Although marked as optional, `--colshift` is often required for correct operation of the `parse_trace.py` script. The issue is that `parse_trace.py` parses mlir files to find out about expected trace outputs, then uses that information when parsing the trace data. One of the things `parse_trace.py` learns about from the mlir are the expected columns. But because some devices are virtual devices, column `N` in the mlir file is not always column `N` in the trace data. Most often the shift is by one, to avoid the shim-less column. But starting with https://github.com/Xilinx/mlir-aie/pull/2370, tests no longer have exclusive access to XRT, and the runtime might start a design anywhere on the array if it is partially occupied. This leads to failures in the mentioned PR because the column shift amounts are hard coded. This is also likely to happen in the wild on systems with other things running on NPU (e.g. Windows or multi-user Linux).

This PR leaves the `--colshift` as an optional command line argument but removes its use everywhere and replaces it with a helper function to figure out the starting column shift amount. This is computed as the difference between the expected and observed columns with trace activity.

This PR also removes the use of regular expressions to walk over the mlir, and uses the dialect python bindings instead.


jackl-xilinx (approved):
Looks good. Thanks for simplifying the  mlir parse code and for scanning for the smallest used tile column to set the colshift automatically.

Issue #2378, opened 2025-06-05
PathFinder fails to find legal routing


pvasireddy-amd:
When running the single column based [example](https://github.com/Xilinx/mlir-aie/blob/routing-prob/programming_examples/basic/passthrough_kernel/single_col.py
), the build fails during the pathfinding step with the error:

`Unable to find a legal routing`

This issue occurs even when using more columns of the device by changing `NPU1Col1` to `NPU1Col4`.

In placed style (low-level IRON), the problem can be avoided by defining a dummy tile (a tile that is not used for computation and has no connections), which allows the pathfinder to find a circuit flow, like in vector_reduce_max [example](https://github.com/Xilinx/mlir-aie/blob/max-on-tensor/programming_examples/basic/vector_reduce_max/single_col_designs/vector_reduce_max_memtile.py).

PR #2384, opened 2025-06-09
Fixes issue #2378

pvasireddy-amd:


lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
```diff
@@ -25,12 +25,8 @@ using namespace xilinx::AIE;
 LogicalResult DynamicTileAnalysis::runAnalysis(DeviceOp &device) {
   LLVM_DEBUG(llvm::dbgs() << "\t---Begin DynamicTileAnalysis Constructor---\n");
   // find the maxCol and maxRow
-  maxCol = 0;
-  maxRow = 0;
-  for (TileOp tileOp : device.getOps<TileOp>()) {
-    maxCol = std::max(maxCol, tileOp.colIndex());
-    maxRow = std::max(maxRow, tileOp.rowIndex());
-  }
```
fifield, line 33: There are several other places in this file with code like:
```
  maxCol = std::max(maxCol, col);
  maxRow = std::max(maxRow, row);
```
Can we fix and/or remove those too?

erwei-xilinx:
`create-flows/vecmul_4x4.mlir` is a routing stress test for a xcvc1902 target device. I wonder if it can eventually pass, if given more time.

programming_examples/basic/vector_reduce_max/Makefile
```diff
@@ -9,24 +9,46 @@
...
 devicename ?= $(if $(filter 1,$(NPU2)),npu2,npu)
 targetname = vector_reduce_max
-in1_size = 4096 # in bytes
+in1_size =  8192# in bytes
 out_size = 4 # in bytes, should always be equal to in1_size
 trace_size = 1024
 
+dtype?=i32
+
+colshift ?= $(if $(filter npu,$(devicename)),1,0)
```
fifield: Does this accidentally revert the removal of `colshift` which was part of #2382 ?

pvasireddy-amd: This was added along with the test but since the test will be added in #2305, removed from this PR.

fifield:
> > Even after doubling the testing time, `create-flows/vecmul_4x4.mlir` test fails.
> 
> @fifield @pvasireddy-amd Would it be okay to filter out this test in all GitHub hosted runners like this? [51bafca](https://github.com/Xilinx/mlir-aie/commit/51bafcae0726a91983f487fca0234fabe458de2f)
> 
> I think the test should have no issue with our self hosted runners. The issue seems to be happening only on the GitHub hosted runners.

We really shouldn't be increasing the timeout. It's fine to filter it out if it runs in some places. IIRC the router is generally only slow for debug builds. If it's a problem outside of debug builds we should take a closer look at the code to see why it's slow.

programming_examples/basic/vector_reduce_max/Makefile
fifield, line 1: nit: whitespace change to unrelated file

jgmelber (approved):

PR #2385, opened 2025-06-10
GEMM diet

andrej:
- Merge `mm.cc` and `mm_b_col_maj.cc` kernels, which were mostly duplicated code. The differences are handled by C++ templates.
- Remove `Makefile.chess` which is also mostly duplicated from `Makefile`. Differences can be triggered using environment variable `use_chess=1`.
- Add `AIE_LOOP_FLATTEN` macro which was forgotten in #2248


aie_kernels/aie2/mm.cc
jgmelber, line 1: These kernels could use more comments in general

aie_kernels/aie2p/mm.cc
jgmelber, line 1: These kernels could use more comments in general

jgmelber (approved):
It would be good to comment and link to aie_api as a reference for the GEMM modes per architecture. But overall this LGTM

PR #2386, opened 2025-06-11
Update mini tutorial

AndraBisca:
Replace uses of GlobalBuffer with LocalBuffer in exercise_1.
Add answers to each exercise.


programming_guide/mini_tutorial/exercise_1/answer.py
jgmelber, line 1: It would be nice to have answer_1, _2 etc as it makes sense where there are simple testable milestones in multi-objective tutorials.

programming_guide/mini_tutorial/exercise_3/answer.py
jgmelber, line 1: It would be nice to have answer_1, _2 etc as it makes sense where there are simple testable milestones in multi-objective tutorials.

programming_guide/mini_tutorial/exercise_2/answer.py
jgmelber, line 1: It would be nice to have answer_1, _2 etc as it makes sense where there are simple testable milestones in multi-objective tutorials.

programming_guide/mini_tutorial/exercise_1/answer.py
AndraBisca: Done! I also modified the numbering of the tasks in each exercise so that they always start with 1. Since the first task in each exercise is to run the program, most often I started with answer_2 to match the code with the task number. Does that make sense?

jgmelber (approved):

PR #2387, opened 2025-06-11
[GEMM] Add option for column-major C output

andrej:


aie_kernels/aie2/mm.cc
```diff
@@ -4,7 +4,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// Copyright (C) 2024, Advanced Micro Devices, Inc.
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
```
ypapadop-amd:
```suggestion
// Copyright (C) 2024-2025, Advanced Micro Devices, Inc.
```

aie_kernels/aie2/mm.cc
```diff
@@ -23,16 +23,29 @@
 
 #include "zero.cc"
 
-template <typename T_in, typename T_out, int rowA, int colA, int colB>
+template <typename T_in, typename T_out, int rowA, int colA, int colB, bool b_row_maj = true, bool c_row_maj = true>
```
ypapadop-amd: Some kind of an accessor class would make it more succinct. We don't get mdspan until C++23 (https://en.cppreference.com/w/cpp/container/mdspan.html) but the layout parametrization is what I was thinking.

ypapadop-amd:
What else is needed for this kernel?

andrej:
Not sure. I created this to help unblock your work, is there something else you needed? I'm not sure if there is interest in merging this. If you'd like to see this merged, someone should probably spend some time to make sure performance of this is adequate, which it might not be.

aie_kernels/aie2/mm.cc
andrej: This is a good suggestion. I'm going to keep it in mind for a future refactor. I don't have a lot of time to work on this currently and it seems like multiple people would like to see this merged, so I'll leave it as-is for now.

jgmelber (approved):

ypapadop-amd:
Thanks for working on this!

PR #2390, opened 2025-06-11
Fix `bf16` GEMM on Strix using Peano

### LINKED ISSUE #2383: Programming Example Basic Matrix Multiplication Whole Array: Kernal did not complete for float dtype
I am testing the Programming example on commit 8904ec1eeb1a0eb7d63b004c3db9c3f339ea01f1. When I changed the input dtype to bf16 and the output dtype to f32, it gives the following error.

Image

The change I made in the Makefile-common is: 
dtype_in?=bf16
dtype_out?=f32

The command I am runing is: 
make -f ../Makefile clean && env use_placed=1 n_aie_cols=8 make -f ../Makefile devicename=npu2 && env use_placed=1 n_aie_cols=8 make -f ../Makefile run devicename=npu2

andrej:
Fixes #2383. The AIE API supposedly supports an `8x8x8` GEMM for `bf16`, but the way this is achieved is through emulation via bfp16. See footnote (e) in the table [here](https://xilinx.github.io/aie_api/group__group__mmul.html).

Something in this emulation code seems to break in peano (chess is fine). As a fix for now, let's choose a different intrinsic size -- `4x8x4` -- that does not rely on this emulation. This restores the GEMM to functional. Drawback: may come with a performance penalty.

@L-roro mostly investigated and came up with this fix.

Related: PR #2385 adds the required `-DAIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16` flag to peano as well (currently only chess is called with this flag). With that PR merged, when the issue in peano is fixed, we can consider switching the AIE API call back to the `8x8x8` variant.

### DIFF
--- a/aie_kernels/aie2p/mm.cc
+++ b/aie_kernels/aie2p/mm.cc
@@ -214,9 +214,9 @@ matmul_vectorized_8x8x8_bf16_bf16(const bfloat16 *__restrict pA,
 
   // After extensive experimentation, the 8x8x8 aie::mmul size was found to be
   // optimal for AIE2P, in combination with the 2x2 mmul expanded kernel
-  constexpr int r = 8;
+  constexpr int r = 4;
   constexpr int s = 8;
-  constexpr int t = 8;
+  constexpr int t = 4;
 
   // Since the kernel has been expanded 2 times for both A ('m' dimension) and B
   // ('n' dimension), the following assertions veirify this even division for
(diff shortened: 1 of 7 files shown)


L-roro:
Yeah, this is the exact issue I was trying to fix. I would like to do a more versatile fix though, to be able to choose between emulation or not. I am going to try to extend your second pr for this #2385. Additionally, tests need to be added. In particular, I believe these changes break chess now for example.

andrej:
Another, potentially better, workaround for this (for now) is to downgrade peano to before the breaking change, which is commit `31d2aa6e9f865e4e2a567252244e033a381a1ba1`:

```
pip uninstall llvm-aie
pip install llvm-aie==19.0.0.2025052701+31d2aa6e -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
```

PR #2394, opened 2025-06-12
Flag slow tests in test name

erwei-xilinx:
... so that CI tests building with Debug mode shall avoid those tests which could timeout.


erwei-xilinx:
Cc: @pvasireddy-amd @fifield 
Could this PR fix the issue in https://github.com/Xilinx/mlir-aie/pull/2384?

pvasireddy-amd:
> Cc: @pvasireddy-amd @fifield Could this PR fix the issue in #2384?

Thanks @erwei-xilinx, this helps with the timeout issue.

PR #2399, opened 2025-06-16
Update aiecc to build elfs for control packet payload and runtime sequence

fifield:
Add more functionality to `aiecc --aie-generate-ctrlpkt` so that:

1. The control packet data payload is generated by aiecc (`ctrlpkt.bin` in the tests)
1. The control packet data payload is packaged into elf file during aiebu-asm assembly step
1. The control sequence binary and elfs are generated by aiecc (`ctrlpkt_dma_seq.bin` and `ctrlpkt_dma_seq.elf` in the tests)

Previously these were generated by the lit scripts. The above filenames are hard coded the same in aiecc as they were in the lit scripts. This is a temporary solution.

Generating the elf for the control packet configuration runtime sequence is new functionality and a new test `control_packet_reconfig_elf` is added where elfs are used instead of transaction binaries. Like the transaction binary version `control_packet_reconfig` there are two elfs; one for kernel configuration and one for kernel execution.

The intent is that the packaging by the compiler and aiecc will get [refined further](https://github.com/fifield/mlir-aie/pull/10) to combine the two elfs.

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -613,37 +613,7 @@ class FlowRunner:
                 print(f"copy {tmp} to {opts.txn_name}")
             shutil.copy(tmp, opts.txn_name)
 
-    async def process_ctrlpkt(self, module_str):
-        with Context(), Location.unknown():
-            run_passes(
-                "builtin.module(aie.device(convert-aie-to-control-packets{elf-dir="
-                + self.tmpdirname
-                + "}))",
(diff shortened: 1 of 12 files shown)


fifield:
@erwei-xilinx this change will impact mlir-air use cases like [this one](https://github.com/Xilinx/mlir-air/blob/main/test/xrt/23_ctrlpkt_config/run_npu1_chess.lit).

PR #2407, opened 2025-06-18
control packet lowering for aiex.configure

fifield:
This PR implements a control packet lowering of `aiex.configure`:

* Modify control packet lowering (`aie-opt --convert-aie-to-control-packets`) to insert configuration control packet ops using `aiex.configure` as insertion point.
* Modify control packet control sequence generation (`aie-opt --aie-ctrl-packet-to-dma`) to be "in-place", instead of creating new control sequence.

The effect of the above is to combine the control packet configuration and user control runtime sequences into one transaction binary. When used with `aiecc --aie-generate-ctrlpkt --aie-generate-elf`, the output elf will contain the combined transaction binary as well as the control packet payload.

The control packet reconfiguration tests are updated to use the combined runtime sequences (instruction binaries).

Test `ctrl_packet_reconfig_elf` is updated to use the combined elf for configuration and control. The xclbin contains only the routing necessary to enable configuration with packets and all control packet configuration details are hidden from the host program:

```c++
  // Skeleton xclbin (pdi) containing only the control packet network
  auto xclbin = xrt::xclbin(std::string("aie1.xclbin"));
  ...
  // elf containing the design's configuration (as packet data) and control (as txn binary)
  xrt::elf elf0("aie2.elf");
  xrt::module mod0{elf0};
  ...
  unsigned int opcode = 3;
  auto kernel0 = xrt::ext::kernel(context, mod0, kernelName);
  ...
  // Arguments 1, 2 (both zero) are for the xrt managed runtime sequence and control packet payload.
  // The control packet payload buffer, managed by the runtime, is the arg after the last user argument. 
  // These are loaded from the elf, managed by xrt, and do not appear in the host code.
  kernel0(opcode, 0, 0, bo_in, bo_out).wait2();
```

### DIFF
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -994,8 +994,6 @@ class FlowRunner:
         file_ctrlpkt_dma_seq_mlir = self.prepend_tmp(
             f"{device_name}_ctrlpkt_dma_seq.mlir"
         )
-        # file_ctrlpkt_dma_seq_bin = opts.ctrlpkt_dma_seq_name.format(device_name)
-        # file_ctrlpkt_elf = opts.ctrlpkt_elf_name.format(device_name)
         run_passes(
             f"builtin.module(aie.device(convert-aie-to-control-packets{{"
             + f"device-name={device_name} elf-dir={self.tmpdirname}}}))",


fifield:
rebased onto / depends on #2532

Issue #2411, opened 2025-06-24
GeMM Failing Shapes on Strix for bf16


L-roro:
I have been testing a number of different shapes for the GeMM programming example with bf16 on Strix. Some of them are failing with somewhat unhelpful errors. I am unsure of where these errors should be reported.

When using Peano (Chess does not have this same issue), generating the bins (aie_cdo_elfs.bin) consistently triggers the following error with some non power of two shapes (not all of them though, I could not find a pattern!).

`[AIE ERROR] _XAie_LoadProgMemSection():231: Overflow of program memory`

The size of the insts does not look unreasonable to me (these inputs to GeMM are triggering the error):
```
512x2560x512 64x64x64 8cols
8.0K    build/insts_512x2560x512_64x64x64_8c.txt

1280x1280x1280 64x64x64 4cols
8.0K    build/insts_1280x1280x1280_64x64x64_4c.txt

1280x2304x3328 64x64x64 4cols
8.0K    build/insts_1280x2304x3328_64x64x64_4c.txt

2304x1280x1280 64x64x64 4cols
12K     build/insts_2304x1280x1280_64x64x64_4c.txt
```

There is also one power of two shape that is failing with the same error:
```
256x1024x256 32x32x32 8cols
8.0K    build/insts_256x1024x256_32x32x32_8c.txt
```

Here are some randomly picked examples of working shapes:
```
420K    build/final_512x4096x512_64x64x64_8c.xclbin
8.0K    build/insts_512x4096x512_64x64x64_8c.txt
or
224K    build/final_256x1024x256_32x32x32_4c.xclbin
4.0K    build/insts_256x1024x256_32x32x32_4c.txt
or
348K    build/final_4096x4096x2560_64x64x64_8c.xclbin
36K     build/insts_4096x4096x2560_64x64x64_8c.txt
```

Here goes a list of other failing MxKxN shapes with mxkxn=64x64x64 and using 4 columns on Peano:
```
1280_2304_3328
256_1280_1280
256_1280_4352
256_1280_2304
256_2304_2304
256_2304_3328
256_2304_4352
256_1280_3328
1280_1280_256
1280_1280_1280
1280_1280_3328
1280_1280_2304
1280_1280_4352
1280_2304_256
1280_2304_1280
256_2304_1280
1280_2304_2304
1280_2304_4352
2304_1280_1280
```

[...]

PR #2391, opened 2025-06-12
Extend bfp emulation

L-roro:
This pr is built on top of #2385. Its aim is to allow users to select the between bf16 emulation with bfp16 or not on Strix. 

These commits fix the currently broken programming example for GeMM single and whole array when using bf16 as input.

Additionally, adds test to the CI for these datatypes which are currently untested. 
There are still many untested combinations for these kernels (datatypes, shapes, chess or peano compiler, ...). How many other combinations should be added is something to be discussed, but I believe that one test with bf16 as input and another with bf16 as input and bfp16 emulation are important to have.

PR #2414, opened 2025-06-24
Change microkernel shape

L-roro:
This small pr addresses this [previous pr](https://github.com/Xilinx/mlir-aie/pull/2391#issuecomment-2976013831) about separating calls to the emulated bf16 hardware or not. In that pr, a random shape was chosen for the bf16 microkernel rst values. Here are some results from a small benchmark related to this performed on the whole_array programming example without using tracing, but only total execution time:
Untitled
Untitled

These results have been obtained under the following conditions:
Tested commit: 
[Extend bfp emulation (#2391) · Xilinx/mlir-aie@bd85924](https://github.com/Xilinx/mlir-aie/commit/bd85924700f85f8c772abc6b36a416afa4910c57) 

Warmup iterations before each datapoint: 10

Iterations for each datapoint (using average): 20

Tested matrix shapes: All combinations of values for M, K and N in the set (512 1024 2560 4096)

Compiler: Peano; clang version 19.0.0 (https://github.com/Xilinx/llvm-aie 77e395c8db5c08f230884b9980ec2647eb182695)

Tested tile expansion: 2x2

This small comparison has also spawned issue #2411


jgmelber (approved):

joeldushouyu:
Hello @L-roro 

Thank you for the performance profile above. I  was wondering if there is any available performance profile for the 8x8x8 configuration in comparison with the other benchmarks mentioned above? I understand there may be a potential [precision issue](https://xilinx.github.io/aie_api/group__group__mmul.html).  But I’d appreciate any additional insights you can share.

Thank you in advance!

L-roro:
Hi @joeldushouyu 

I did this exact same benchmarks for the 8x8x8 configuration. As you mention, this configuration uses bfp16 hardware to emulate the bf16 and therefore the precision loss is higher. Nevertheless, I have not stored any results based on the precision loss that I might share with you unfortunately, since I was just focused on performance.

I also feel forced to inform you that these results have not yet been confirmed by the rest of the team, and I am fully aware of the possibility of getting higher performance out of the hardware. This was just focused on choosing a default configuration for people trying out the matrix multiplication and does not exploit the hardware to its maximum (which can be achieved with, for example, different tilings than those that I am testing here, among other things).

With these caveats in mind, I'll share this too:
Only bfp16 emulation here
image

Here you have them together, with the 4 column sets removed for clarity
image
(note that the one marked f32 is using f32 as output instead of bf16, but the input still is bf16)

andrej (approved):

PR #2419, opened 2025-06-30
Disable tracing if the trace flag is set to 0 for workers

pvasireddy-amd:
Current setup is true for any value except `None`, so tracing happens even when the flag is disabled for the worker.


AndraBisca:
When the flag is disabled, what value does it have? Something like False or 0?

Issue #1959, opened 2024-12-05
Sliding_window_conditional python version not working


pvasireddy-amd:
In #1907, it was found that if-else conditional statements in the Python code are not lowered properly and requires fixes.

PR #2423, opened 2025-07-07
Example using if-else from aie.extras.dialects.ext.scf

pvasireddy-amd:
Adds `if-else` wrapper.

Replaces the MLIR version of the sliding window conditional example with Python version using `if_`, `else_` 

Issue #1959


fifield (commented):
I'm curious why we can't reuse something from `aie.extras.dialects.ext.scf` instead?  e.g. https://github.com/makslevental/mlir-python-extras/blob/main/tests/test_scf.py

pvasireddy-amd:
> I'm curious why we can't reuse something from `aie.extras.dialects.ext.scf` instead? e.g. https://github.com/makslevental/mlir-python-extras/blob/main/tests/test_scf.py

I didn't know about this. Updated the code accordingly.

test/npu-xrt/dynamic_object_fifo/sliding_window_conditional/aie2.py
```diff
@@ -0,0 +1,79 @@
...
+#
+# (c) Copyright 2024 AMD Inc.
+
+# REQUIRES: ryzen_ai, valid_xchess_license
+#
+# RUN: xchesscc_wrapper aie2 -I %aietools/include -c %S/kernel.cc -o ./kernel.o
+# RUN: %python %S/aie2.py > ./aie2.mlir
+# RUN: %python aiecc.py --no-aiesim --aie-generate-npu-insts --aie-generate-xclbin --no-compile-host --dynamic-objFifos --xclbin-name=final.xclbin --npu-insts-name=insts.bin ./aie2.mlir
+# RUN: clang %S/test.cpp -o test.exe -std=c++17 -Wall %xrt_flags -lrt -lstdc++ %test_utils_flags
+# RUN: %run_on_npu1% ./test.exe
```
jgmelber, line 14: Is it possible to support this test on both npu1 and npu2?

pvasireddy-amd: I tried running the `sliding_window_conditional` example on both npu1 and npu2. After doing this, not only did that test fail, but the `packet_flow_fanout` test started failing too, even though they’re supposed to be independent. Now both of them try to run on npu2, and while `packet_flow_fanout` has been working fine before, it fails when `sliding_window_conditional` also runs on npu2.

AndraBisca: I think it's fine to run the design in this PR only on NPU1 for now. We should make a separate PR to enable more of the tests in test/npu-xrt/ to run on NPU2.

python/helpers/dialects/ext/scf.py
```diff
@@ -49,3 +51,26 @@ def _for(
...
+@contextmanager
+def if_(cond, hasElse=True, insert_yield=True, loc=None, ip=None):
+    if loc is None:
+        loc = get_user_code_loc()
+    if_op = IfOp(cond, hasElse=hasElse, loc=loc, ip=ip)
+    with InsertionPoint(if_op.thenRegion.blocks[0]):
+        yield if_op
+        if insert_yield:
+            yield_([])
+    # TODO: Default, empty else region.
```
AndraBisca: Can this TODO be deleted?

AndraBisca (approved):
Aside from the small comment, this looks good to me!

PR #2437, opened 2025-07-17
Use a new release for nightly wheels

jgmelber:
Temporary fix that adds a new release `latest-wheels-2` for nightly wheels


jgmelber:
> LGTM as a short-term solution. For long term, I wonder if it's a good place to start using minor versions to name the release names, like "latest-wheels-v0.0.2". We have been in version 0.0.1 for quite a while:
> 
> https://github.com/Xilinx/mlir-aie/blob/3751be76098952d58e1f412c1da8611c24a9931f/utils/mlir_aie_wheels/setup.py#L236

We should align these with the true releases indeed.

PR #2449, opened 2025-07-28
Numerically Stable Softmax

Victor-Jung:
This PR makes the Softmax kernel for aie2p numerically stable by subtracting the maximum before exponentiating. This change prevent overflow when the input value is getting large. For instance, the previous Softmax implementation returns NaNs as soon as the input are above a few hundreds.

For the formula and rational behind a numerically stable Softmax, see [this blogpost](https://jaykmody.com/blog/stable-softmax/).

I compared the performance of the original version with two other approaches: calculating the maximum value using a scalar method and using a vectorized method. The vectorized approach resulted in only a slight increase in runtime, approximately 10%.

Original Softmax, 1000 runs:
	Average NPU time: 1315.63 us.
	Standard Deviation: 279.85 us
Stable Softmax (scalar max), 100 runs:
	Average NPU time: 3417.95 us.
	Standard Deviation: 429.21 us.
Stable Softmax (vectorized max), 1000 runs:
	Average NPU time: 1449.88 us.
	Standard Deviation: 258.79 us.

Additionally I've made the `env_setup.sh` script POSIX friendly.

### DIFF
--- a/programming_examples/ml/softmax/Makefile
+++ b/programming_examples/ml/softmax/Makefile
@@ -91,7 +91,7 @@ ${targetname}.exe: ${srcdir}/test.cpp
 	rm -rf _build
 	mkdir -p _build
 	cd _build && ${powershell} cmake `${getwslpath} ${srcdir}` -DTARGET_NAME=${targetname}
-	cd _build && ${powershell} cmake --build . --config Release --verbose
+	cd _build && ${powershell} cmake --build . --config Release
 ifeq "${powershell}" "powershell.exe"
 	cp _build/${targetname}.exe $@
 else


utils/quick_setup.sh
```diff
@@ -61,7 +61,7 @@ python3 -m pip install --upgrade pip
 python3 -m pip install mlir_aie -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels-2/ 
 export MLIR_AIE_INSTALL_DIR="$(pip show mlir_aie | grep ^Location: | awk '{print $2}')/mlir_aie"
 
-python3 -m pip install llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
+python3 -m pip install llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/download/nightly/llvm_aie-19.0.0.2025071101+b3cd09d3-py3-none-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
```
jgmelber: This should be in its own PR if the nightly Peano from tonight doesn't fix the issues

programming_examples/ml/softmax/Makefile
```diff
@@ -91,15 +91,20 @@ ${targetname}.exe: ${srcdir}/test.cpp
 	rm -rf _build
 	mkdir -p _build
 	cd _build && ${powershell} cmake `${getwslpath} ${srcdir}` -DTARGET_NAME=${targetname}
-	cd _build && ${powershell} cmake --build . --config Release
+	cd _build && ${powershell} cmake --build . --config Release --verbose
```
jgmelber:
```suggestion
	cd _build && ${powershell} cmake --build . --config Release
```

aie_kernels/aie2p/softmax.cc
```diff
@@ -12,7 +12,7 @@
 #include <aie_api/aie.hpp>
 #include <stdint.h>
 
-#define SM_VEC_LEN 16   // 32
+#define SM_VEC_LEN 32   // 32
```
jgmelber, line 15: Is this really 32?

Victor-Jung: I though we could choose between 8/16/32/64 for bfloat16 (from this [page of the documentation](http://cervino-doc/aie2p/r1p13/intrinsics/group__group__datatype__vector.html))

PR #2451, opened 2025-07-29
Remove unused python dependencies

ypapadop-amd:
I couldn't find any uses of `pandas` or `psutil` in the codebase, so I'm checking if it's safe to remove them.

### DIFF
--- a/.github/workflows/generateDocs.yml
+++ b/.github/workflows/generateDocs.yml
@@ -32,7 +32,7 @@ jobs:
         run: sudo apt-get install -y graphviz lld doxygen
 
       - name: Install Python packages
-        run: sudo pip install psutil rich numpy pybind11 nanobind
+        run: sudo pip install rich numpy pybind11 nanobind
 
       - name: Install Ninja
         # https://github.com/llvm/actions/tree/main/install-ninja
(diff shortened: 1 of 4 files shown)


.github/workflows/generateDocs.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: sudo apt-get install -y graphviz lld doxygen
 
       - name: Install Python packages
-        run: sudo pip install psutil rich numpy pybind11 nanobind
+        run: sudo pip install rich numpy pybind11 nanobind
```
ypapadop-amd: Any reason not to use `requirements.txt` here?

.github/workflows/lintAndFormat.yml
```diff
@@ -40,26 +40,26 @@ jobs:
 
       - name: Install Python packages
         run: |
-          
-          pip install cmake==3.27.9 numpy psutil pybind11 rich pkginfo lit PyYAML requests
+
+          pip install cmake==3.27.9 numpy pybind11 rich pkginfo lit PyYAML requests
```
ypapadop-amd: This can probably also just use `requirements.txt`.

.github/workflows/generateDocs.yml
jgmelber: Not as far as I know

.github/workflows/lintAndFormat.yml
jgmelber: Likely, yeah

PR #2453, opened 2025-07-29
Run the CI Nightly at Midnight

Victor-Jung:
Several of our dependencies are nightly builds (for instance Peano). To naturally track if one of our dependency is breaking IRON, we can run the CI every night at midnight.

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -10,6 +10,9 @@ on:
         required: false
         default: ''
   merge_group:
+  schedule:
+    # Runs at midnight (00:00) every day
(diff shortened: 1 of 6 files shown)


jgmelber:
This looks good but should be merged when we can revert #2456

.github/workflows/buildRyzenWheels.yml
```diff
@@ -13,8 +13,9 @@ on:
     tags:
       - 'v*.*.*'
   merge_group:
-
   schedule:
+    # Runs at midnight (00:00) every day
+    - cron: '0 0 * * *'
```
jgmelber:
```suggestion
```

PR #2460, opened 2025-07-31
Bump llvm to 77914c96

erwei-xilinx:


fifield:
@jsetoain fyi aievec failures with llvm bump

jsetoain:
> > @jsetoain fyi aievec failures with llvm bump
> 
> I found the issue and fixed it. The tests are re-enabled.

Thanks, Erwei. I need to catch up with discussions upstream, I don't quite understand why the pass-through values have to be poison now.

Cheers!

PR #2461, opened 2025-07-31
Remove Jupyter notebook dependency

ypapadop-amd:
This PR moves the Jupyter notebook dependency as an optional. `pip install requirements_notebook.txt` was added to the actions that call `ninja check-reference-designs`.

### DIFF
--- a/.github/workflows/buildAndTestAieTools.yml
+++ b/.github/workflows/buildAndTestAieTools.yml
@@ -87,6 +87,7 @@ jobs:
             source aie-venv/bin/activate
             pip install -r python/requirements.txt
             HOST_MLIR_PYTHON_PACKAGE_PREFIX=aie pip install -r python/requirements_extras.txt
+            pip install -r python/requirements_notebook.txt
 
             VERSION=$(utils/clone-llvm.sh --get-wheel-version)
             pip -q download mlir==$VERSION \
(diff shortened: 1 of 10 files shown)


fifield:
> @ypapadop-amd, @jgmelber, I think lit is still needed in requirements.txt. Future PRs are failing due to failing to find llvm-lit: https://github.com/Xilinx/mlir-aie/actions/runs/16676816882/job/47219809873#step:4:1017
> 
> I suspect that this PR itself is passing due to quick_setup.sh pulling a previously latest mlir-aie wheel, which still has lit in its wheel's dependency (i.e. its own requirements.txt), and pip installed lit. Installing future mlir-aie wheels no longer pip installs lit, leading to the errors.

I think that workflow needs `pip install -r python/requirements_test.txt` after sourcing quick_setup.sh. It is probably not needed in quick_setut.sh itself.

erwei-xilinx:
> I think that workflow needs `pip install -r python/requirements_test.txt` after sourcing quick_setup.sh. It is probably not needed in quick_setup.sh itself.

That's a good point. Yeah, that should work without needing to add lit back to requirements.txt.

ypapadop-amd:
It's probably needed in quick_setup.sh to maintain the previous behavior. I didn't realize that workflow file had two workflows, my bad. Added in https://github.com/Xilinx/mlir-aie/pull/2464 to allow it to pass.

PR #2469, opened 2025-08-04
Separating development-related packages to requirements_dev.txt

ypapadop-amd:
This PR separates development packages to a dedicated `requirements_dev.txt` file.

This is done to
1. reduce the overall dependencies of the wheels (basic venv is now at around `128MB`) and
1. not impose specific packages (e.g., cmake, setuptools) that could clobber the environment of a user of IRON.

The PR changes the workflow, e.g., `pre-commit` is now only available via `requirements_dev.txt` so I'm opening it for discussion before any official review.

### DIFF
--- a/programming_examples/basic/combined_transpose/combined_transpose.py
+++ b/programming_examples/basic/combined_transpose/combined_transpose.py
@@ -5,10 +5,9 @@
 #
 # (c) Copyright 2025 Advanced Micro Devices, Inc. or its affiliates
 import numpy as np
-import sys
 import argparse
 
-from aie.iron import Kernel, ObjectFifo, Program, Runtime, Worker
+from aie.iron import Kernel, ObjectFifo, Program, Runtime, Worker, str_to_dtype
 from aie.iron.placers import SequentialPlacer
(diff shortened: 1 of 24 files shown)


.github/workflows/generateDocs.yml
```diff
@@ -32,7 +32,9 @@ jobs:
         run: sudo apt-get install -y graphviz lld doxygen
 
       - name: Install Python packages
-        run: sudo pip install -r python/requirements.txt
+        run: |
+          sudo pip install -r python/requirements.txt
+          sudo pip install -r python/requirements_dev.txt
```
fifield: I wonder if `sudo` actually required here

ypapadop-amd: Probably not, removed.

erwei-xilinx (approved):
Thanks for working on this! LGTM.

The wheel-building CIs are failing because those yaml scripts only work with the origin/* branches instead of your fork, in case if you want to test your changes with those workflows.

ypapadop-amd:
Putting this in normal review.

jgmelber (approved):

PR #2471, opened 2025-08-04
Update README for sw stack and downloads

jgmelber:


ypapadop-amd:
There's also https://github.com/Xilinx/mlir-aie/blob/main/docs/README.md, does that need to be updated too?

jgmelber:
> There's also https://github.com/Xilinx/mlir-aie/blob/main/docs/README.md, does that need to be updated too?

No badges on the doxygen site

PR #2474, opened 2025-08-04
Add diagnostic handling to aiecc run_passes

fifield:
Add an explicit diagnostic handler to the `run_passes` utility in aiecc `main.py`. This lets the diagnostics be captured and printed in a controlled way for verbose mode (e.g. warnings) or for when an error occurs.


fifield:
For some reason this Works For Me flawlessly...I cannot get it to fail locally...but only partially works in CI.

andrej:
Just to double-check, does this take care of the ominous "not all errors were captured" error?

And, would the same also be needed for the in-memory passes in [`run_passes_module`](https://github.com/Xilinx/mlir-aie/blob/77aff8b0faa6f28f6e42b8eaca2ddda5db7b7587/python/compiler/aiecc/main.py#L373)?

fifield:
> Just to double-check, does this take care of the ominous "not all errors were captured" error?
> 
> And, would the same also be needed for the in-memory passes in [`run_passes_module`](https://github.com/Xilinx/mlir-aie/blob/77aff8b0faa6f28f6e42b8eaca2ddda5db7b7587/python/compiler/aiecc/main.py#L373)?

I think so, but it's been so long since I made the patch that I don't recall the exact error message. I was only holding it back because the more informational version, which is to print the warnings for `aiecc -v`, wasn't working in CI and I hadn't spent time to debug why. But now it seems to be failing for different reasons after my last rebase.

PR #2475, opened 2025-08-05
Add external C functions support for JIT

mawad-amd:
This PR adds:
1. External function support for JIT-compiled designs
2. Tests for external functions
3. Helper `tensor.fill_` function and test for it

The syntax looks like:
```python
add_value = ExternalKernel(
     "add_value",
     source_string="""extern "C" {
         void add_value(int* input, int* output, int tile_size) {
             for (int i = 0; i < tile_size; i++) {
                 output[i] = input[i] + ADD_VALUE;
             }
         }
     }""",
     arg_types=[
         np.ndarray[(16,), np.dtype[np.int32]],
         np.ndarray[(16,), np.dtype[np.int32]],
         np.int32,
     ],
     compile_flags=[f"-DADD_VALUE=2"],
 )
transform(input_tensor, output_tensor, add_value)
```


fifield:
Looks good, I have some questions:
* Can there be multiple top level functions in one source string, which go into a single object file? Or will multiple `ExternalKernel` be combined into one `.o`?
* At the risk of bikeshedding a bit, why `ExternalKernel` instead of `ExternalFunction` (if a single function)? or `ExternalModule` (if multiple functions per source string)?
* Does C++ mangling work? i.e. could I drop the `extern "C"`?

mawad-amd:
For first and last comment, the implementation in the PR inherits the capabilities of the Kernel and Worker class.

> * Can there be multiple top level functions in one source string, which go into a single object file? Or will multiple `ExternalKernel` be combined into one `.o`?

Technically, yes, you could have multiple kernels but you only declare one function that you will use at the moment. We could turn the `name` into a list of names (my changes should still work) but the [Worker class supports only one single external function call](https://github.com/Xilinx/mlir-aie/blob/49cf4e395f6e97af20712adae664a082ee9319c0/python/iron/worker.py#L110-L113). Seems like we should be able to support multiple functions in the Worker abstraction but please correct me if not (I didn't author the worker class).

> * At the risk of bikeshedding a bit, why `ExternalKernel` instead of `ExternalFunction` (if a single function)? or `ExternalModule` (if multiple functions per source string)?

Yeah I like `ExternalFunction` too. Will make the change (I think we have 3 votes).

> * Does C++ mangling work? i.e. could I drop the `extern "C"`?

Looks like yes! I have just tried what is below. Not suggesting that the user go find the mangled name but not sure yet what would be a nice way to resolve that or even better lookup the function when dealing generating the MLIR (not sure if it is straight forward). Any ideas? (Would love to discuss this more but prefer to keep it to a different PR).
```python
    add_one_templated = ExternalKernel(
        "_Z11add_one_intIiEvPT_S1_i",
        source_string="""
            template<typename T>
            void add_one_int(T* input, T* output, int tile_size) {
                for (int i = 0; i < tile_size; i++) {
                    output[i] = input[i] + 1;
                }
            }

            template void add_one_int<int>(int* input, int* output, int tile_size);
        """,
        arg_types=[
            np.ndarray[(16,), np.dtype[np.int32]],
            np.ndarray[(16,), np.dtype[np.int32]],
            np.int32,
        ],
    )
```

python/iron/jit.py
```diff
@@ -161,13 +190,28 @@ def decorator(*args, **kwargs):
...
+                    inst_path=inst_path,
+                    xclbin_path=xclbin_path,
+                    work_dir=kernel_dir,
+                )
+            except Exception as e:
+                # Clean up cache directory on any compilation failure
+                if os.path.exists(kernel_dir):
+                    import shutil
+
+                    shutil.rmtree(kernel_dir)
```
ypapadop-amd, line 235: Wouldn't we want the artifacts of a failed compilation? Also importing under exception can create issues, unless importing is always exception-free.

python/iron/jit.py
```diff
@@ -177,9 +221,112 @@ def decorator(*args, **kwargs):
...
+        "-Wno-attributes",
+        "-Wno-macro-redefined",
+        "-Wno-empty-body",
+        "-Wno-missing-template-arg-list-after-template-kw",
+    ]
+
+    common_flags = ["-O2", "-std=c++20", "-DNDEBUG"] + warning_flags
+
+    # Build compilation command
+    cmd = [f"{os.environ.get('PEANO_INSTALL_DIR', '')}/bin/clang++"]
```
ypapadop-amd: There's already a function to retrieve that.

test/python/jit_extern_functions.py
```diff
@@ -0,0 +1,844 @@
...
+    """Test basic ExternalKernel with simple add_one operation."""
+    # Create input and output tensors
+    input_tensor = iron.randint(0, 100, (1024,), dtype=np.int32, device="npu")
+    output_tensor = iron.zeros((1024,), dtype=np.int32, device="npu")
+    initial_tensor = input_tensor.numpy().copy()
+
+    # Create ExternalKernel for adding one
+    add_one = ExternalKernel(
+        "add_one",
+        source_string="""extern "C" {
```
ypapadop-amd, line 98: This is an inline kernel, not an external. Unless I'm misunderstanding the meaning of external here.

test/python/jit_extern_functions.py
```diff
@@ -0,0 +1,844 @@
...
+            void add_one(int* input, int* output, int tile_size) {
```
ypapadop-amd, line 99: How are we going to treat different datatypes? Can templates be supported?

ypapadop-amd (commented):
Some comments on the code. The biggest one is that I don't see how this can be used outside `iron.jit` and relatively simple core functions.

Things that I'd be looking as a developer:
- Ability to separate the C and Python codes for ease of development and testability.
- Ability to compile an external function / kernel outside `iron.jit`.
- Maintaining the flexibility of compiling the C code manually.

python/iron/jit.py
```diff
@@ -177,9 +221,112 @@ def decorator(*args, **kwargs):
...
+            shutil.copy2(func._source_file, source_file)
+        else:
+            return
+    else:
+        raise ValueError("Neither source_string nor source_file is provided")
+
+    # Add device-specific flags based on actual device detection
+    current_device = get_current_device()
+
+    warning_flags = [
```
ypapadop-amd: All of warning flags and options are already used in https://github.com/Xilinx/mlir-aie/blob/abfd78b95940658e67bd90112962dcd4c79d0c2a/python/iron/compile/compile.py#L14

I suggest that we either expand that function to handle the cases that are missing here or at least factoring out the flags / options to avoid repetition.

mawad-amd: Reused the function. Much better. Thanks https://github.com/Xilinx/mlir-aie/pull/2475/commits/4feb272ab54d5a40d8e061ddfb1d1674a7fbc282

python/iron/jit.py
mawad-amd: Moved the import. I don’t want corrupt build artifacts in the cache. The user gets an error message that they can address.

test/python/jit_extern_functions.py
mawad-amd: We are discussing the naming in comments. External as in External to Python.

test/python/jit_extern_functions.py
mawad-amd: Discussed in comments.

ypapadop-amd:
> For first and last comment, the implementation in the PR inherits the capabilities of the Kernel and Worker class.
> 
> > * Can there be multiple top level functions in one source string, which go into a single object file? Or will multiple `ExternalKernel` be combined into one `.o`?
> 
> Technically, yes, you could have multiple kernels but you only declare one function that you will use at the moment. We could turn the `name` into a list of names (my changes should still work) but the [Worker class supports only one single external function call](https://github.com/Xilinx/mlir-aie/blob/49cf4e395f6e97af20712adae664a082ee9319c0/python/iron/worker.py#L110-L113). Seems like we should be able to support multiple functions in the Worker abstraction but please correct me if not (I didn't author the worker class).

A `.o` can have multiple exported symbols and we should be able to call them if they are store as a key-value pair (key being a a "templated" function type and value the strongly typed one).

> > * Does C++ mangling work? i.e. could I drop the `extern "C"`?
> 
> Looks like yes! I have just tried what is below. Not suggesting that the user go find the mangled name but not sure yet what would be a nice way to resolve that (i.e., what would the user type for name) or even better lookup the function when generating the MLIR (not sure if it is straight forward). Any ideas? (Would love to discuss this more but prefer to keep it to a different PR).
> 
> ```python
>     add_one_templated = ExternalKernel(
>         "_Z7add_oneIiEvPT_S1_i",
>         source_string="""
>             template<typename T>
>             void add_one(T* input, T* output, int tile_size) {
>                 for (int i = 0; i < tile_size; i++) {
>                     output[i] = input[i] + 1;
>                 }
>             }
> 
>             template void add_one<int>(int* input, int* output, int tile_size);
>         """,
>         arg_types=[
>             np.ndarray[(16,), np.dtype[np.int32]],
>             np.ndarray[(16,), np.dtype[np.int32]],
>             np.int32,
>         ],
>     )
> ```

We can get a list of exported functions from the `.o` and demangle them using cxxfilt (https://pypi.org/project/cxxfilt/). Then at the call site, you can give the demangled name and match it to one in the list. E.g., 
```python
>     add_one_templated = ExternalKernel(
>         "void add_one<int>(int*, int*, int)",
>         source_string="""
>             template<typename T>
>             void add_one(T* input, T* output, int tile_size) {
>                 for (int i = 0; i < tile_size; i++) {
>                     output[i] = input[i] + 1;
>                 }
>             }
> 
>             template void add_one<int>(int* input, int* output, int tile_size);
>         """,
```

python/iron/jit.py
```diff
@@ -177,9 +219,85 @@ def decorator(*args, **kwargs):
...
+        else:
+            return
+    else:
+        raise ValueError("Neither source_string nor source_file is provided")
+
+    # Add device-specific flags based on actual device detection
+    current_device = get_current_device()
+
+    # Determine target architecture based on device type
+    if isinstance(current_device, NPU2):
```
ypapadop-amd: Will this work with `NPU2Col1`?

mawad-amd: I hope so, @jgmelber?

mawad-amd: I see what you mean. Fixed.

mawad-amd:
OK I got CI working finally. NPU2 runner was hitting a cached NPU1 code somehow (is there a shared storage?). The hash now contains enough information to disambiguate archs. Please review the PR again.

Regarding the suggested improvements (ordered by what I think is important):
1. Backend improvements: There are some ownership issues at the moment (even without considering the JIT module.  I think what we need to do is to restructure the entire backend (for unplaced and JIT) so that we have a Program that owns Kernel/Function objects and provide APIs for compilation and iterators over objects. I can discuss more and happy to help with this, but this will require decent amount of work, more than one person but is the most important thing IMO.
2. Support for multiple functions within the same object file: once worker class support multiple functions (1), we can pursue this.
3. C++ support would be nice improvement. Open an issue for this.

python/iron/jit.py
```diff
@@ -133,18 +143,61 @@ def jit(function=None, is_placed=True, use_cache=True):
...
+                not hasattr(func, "_compiled") or not func._compiled
+            ):  # Don't compile if already compiled
+                external_kernels.append(func)
+
+        # Determine target architecture based on device type
+        try:
+            current_device = get_current_device()
+
+            # Determine target architecture based on device type
+            if isinstance(current_device, (NPU2, NPU2Col1)):
```
ypapadop-amd, line 186: Since the arch/col combination is a type, then we'll need to handle all different combinations (NPU[1-2]Col[1-3])). Maybe we should make this a utility function, it seems like a good candidate.

python/iron/jit.py
```diff
@@ -161,25 +214,120 @@ def decorator(*args, **kwargs):
...
+        return
+
+    # Create source file in kernel directory
+    source_file = os.path.join(kernel_dir, f"{func._name}.cc")
+
+    # Handle both source_string and source_file cases
+    if func._source_string is not None:
+        # Use source_string (write to file)
+        try:
+            with open(source_file, "w") as f:
```
ypapadop-amd, line 274: You can avoid writing to a file and then compiling by passing in the source code directly using something like `clang++ -x c++ -o output.o -`

We can extend `compile_cxx_core_function` to handle character streams instead of files.

python/iron/jit.py
mawad-amd: Yes, was thinking the same! I really dislike that we can set the `current_device` to both an enum and a class. Would this be good? Would like to hear @jgmelber opinion too:

```python
# In config.py
def get_target_architecture():
    current_device = get_current_device()

    # Handle class-based devices
    if isinstance(current_device, (NPU2, NPU2Col1)):
        return "aie2p"
    elif isinstance(current_device, (NPU1, NPU1Col1)):
        return "aie2"

    # Handle enums by name
    if hasattr(current_device, "name"):
        name = current_device.name.lower()
        if name.startswith("npu1"):
            return "aie2"
        elif name.startswith("npu2"):
            return "aie2p"

    raise RuntimeError(f"Unsupported device type: {current_device} (type: {type(current_device)})")
```

python/iron/jit.py
mawad-amd: Writing the file is intentional to have it in the cache.

python/iron/jit.py
fifield: I'd propose using the target model which already has this information to support C++ RTTI. Then it's a simple wrapper like getting the number of rows or columns:
```python
    # in iron/device.py
    @property
    def arch(self):
        return get_target_model(self._device).get_target_arch()
```

I guess it would return a python enum of the C++:
```c++
enum class AIEArch : uint32_t {
  AIE1 = 1,
  AIE2 = 2,
  AIE2p = 3,
};
```
The TODO here is to add `get_target_arch()` wrapping `AIETargetModel::getTargetArch` in `python/AIEMLIRModule.cpp`

ypapadop-amd:
@fifield @mawad-amd I made a prototype to support mangled names in https://github.com/Xilinx/mlir-aie/pull/2480 I need to figure out why I need explicit paths but it seems to work.

python/iron/jit.py
jgmelber: I am all for simplifications. I think the more we move to supporting IRON and use the class based names is best, but we should lean on the compiler infrastructure as much as possible like @fifield suggests.

PR #2480, opened 2025-08-06
C++ mangled name support in Kernel

ypapadop-amd:


python/iron/kernel.py
```diff
@@ -30,7 +61,12 @@ def __init__(
             bin_name (str): The name of the binary (used for linking to a compute core)
             arg_types (list[type[np.ndarray]  |  np.dtype], optional): The type signature of the function. Defaults to [].
         """
-        self._name = name
+
+        symbol_name = find_mangled_symbol(f"build/{bin_name}", name)
```
ypapadop-amd: I need to figure out how to get the actual .o path here, it's just the name and then it happens to work during linking.

mawad-amd: I think there is a general implicit assumption that the `programming_examples/basic/vector_reduce_min/vector_reduce_min.py` file is called within the build dir.

ypapadop-amd: Good call. `aiecc.py` is called in the build directory which does the linking, but the MLIR generation happens earlier. I currently need the object file earlier, unless I defer mangling until linking (which it would be the preferred approach).

Generating MLIR while in the `build` directory allows it to work correctly.

ypapadop-amd:
Maybe this would be better as an MLIR pass?

PR #2500, opened 2025-08-12
WSL Compatibility: ironenv config examines Windows xrt-smi

thomthehound:
Old behavior: script always fails on WSL, making builds impossible if XRT is also installed.
New behavior: script detects if it is running in WSL and looks at the Windows xrt-smi.exe instead of the Linux version. All logic is strictly identical to the original file.

Frankly, I question the need for this script in the first place. It is redundant to other checks, it ignores the environment, and it flagrantly disrespect arguments explicitly passed to the makes. If I weren't a newcomer, I would propose eliminating it entirely. At minimum, it could probably use another pass to force it to respect arguments and the environment.

### DIFF
--- a/python/iron/config.py
+++ b/python/iron/config.py
@@ -10,6 +10,19 @@ import subprocess
 
 from .device import NPU1, NPU2
 
+# Detect WSL
+def is_wsl() -> bool:
+    try:
+        with open("/proc/sys/kernel/osrelease", "r", encoding="utf-8") as kernel:
+            return "microsoft" in kernel.read().lower()
+    except OSError:
+        return False
+    
+# Prefer Windows xrt-smi when in WSL. Linux native otherwise.
+def xrt_smi_path() -> str:
+    if is_wsl():
+        return "/mnt/c/Windows/System32/AMD/xrt-smi.exe"
+    return "/opt/xilinx/xrt/bin/xrt-smi"
 
 def detect_npu_device():
     """Detects the current device in the system.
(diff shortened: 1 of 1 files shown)


thomthehound:
> I'd expect `xrt-smi` to be in the PATH somewhere, Linux, Windows or WSL.

I agree 100%; that is the way to go. I think that is something we should work towards in the future. There is a lot of hard-pathing going on in the codebase (as in this example). It also might be possible to skip XRT compile entirely for WSL users if we put xclbinutil on the PATH as well (although finding it in Windows can be painful).

jgmelber (approved):

fifield:
> > I'd expect `xrt-smi` to be in the PATH somewhere, Linux, Windows or WSL.
> 
> I agree 100%; that is the way to go. I think that is something we should work towards in the future. There is a lot of hard-pathing going on in the codebase (as in this example). It also might be possible to skip XRT compile entirely for WSL users if we put xclbinutil on the PATH as well (although finding it in Windows can be painful).

Don't folks usually source `/opt/xilinx/xrt/setup.sh` as part of their env setup, putting these both in their path? Maybe I'm the only one who does this?

jgmelber:
> What I was saying about xclbinutil was just a tangential thought about whether or not WSL users even need to have XRT to begin with, or if Ryzen AI Software already comes with everything we need.

XRT is required for the runtime to launch kernels on the NPU. Ryzen AI Software is not a requirement of this flow

jgmelber:
> I agree with both of those things. I'm just trying to make the WSL experience easier for the user. That's all.

Your efforts are GREATLY appreciated! 🙌

PR #2506, opened 2025-08-13
Override for the object file name of ExternalFunction

ypapadop-amd:
This adds an optional parameter for the object file name when we wish to override the default.

### DIFF
--- a/test/python/jit_extern_functions.py
+++ b/test/python/jit_extern_functions.py
@@ -109,17 +109,11 @@ def test_simple_add_one():
     )
 
     # Apply the transform
-    print("Applying transform")
-    print(f"Input tensor: {input_tensor}")
-    print(f"Output tensor: {output_tensor}")
-    print(f"Initial tensor: {initial_tensor}")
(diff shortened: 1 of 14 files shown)


jgmelber (commented):
Test case? Otherwise LGTM

test/python/jit_extern_functions.py
```diff
@@ -844,3 +843,43 @@ def test_compiler_flag_combinations(compile_flags, expected_value):
...
+        }""",
+        arg_types=[
+            np.ndarray[(16,), np.dtype[np.int32]],
+            np.ndarray[(16,), np.dtype[np.int32]],
+            np.int32,
+        ],
+    )
+
+    # Apply the transform
+    print("Applying transform")
```
mawad-amd: Are these `prints` needed?

mawad-amd (approved):

test/python/jit_extern_functions.py
ypapadop-amd: Probably not, I copy-pasted code :)

PR #2508, opened 2025-08-15
Improved packet_switch example

AndraBisca:
In this PR:
- additional comments added in the IRON design files for clarity,
- variables are renamed to be more representative of their function, for example `objFifo_core02_cons_buff_0` is renamed to `core02_buff_in`,
- in the runtime sequence, replaced `shim_dma_single_bd_task()` operations with `dma_configure_task()` ones as they regroup all the information which was previously separated between the `memref.GlobalOp`s and `ShimDMAAllocationOp`s (which also enables us to delete those operations).


python/dialects/aiex.py
```diff
@@ -249,7 +249,7 @@ def shim_dma_single_bd_task(
     transfer_len: int | None = None,
     issue_token: bool = False,
     burst_length: int = 0,
-    packet: tuple[int]|None = None
+    packet: tuple[int] | None = None,
```
fifield, line 252: not really the fault of this PR, but `packet` is missing in the docstring

jgmelber (approved):
Good cleanup!

PR #2509, opened 2025-08-17
Add a code object cache for external functions

mawad-amd:
Today if we write:
```python
source_code = """extern "C" {
        void add_constant_1(int* input, int* output, int tile_size) {
            for (int i = 0; i < tile_size; i++) {
                output[i] = input[i] + 1;
            }
        }
        void add_constant_2(int* input, int* output, int tile_size) {
            for (int i = 0; i < tile_size; i++) {
                output[i] = input[i] + 2;
            }
        }
    }"""
# Create ExternalFunction for adding one (first stage)
tile_size = 16
add_one_function = ExternalFunction(
    "add_constant_1",
    source_string=source_code,
    arg_types=[
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.int32,
    ],
)
# Create ExternalFunction for adding two (second stage)
add_two_function = ExternalFunction(
    "add_constant_2",
    source_string=source_code,
    arg_types=[
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.int32,
    ],
)
# Apply the transform: input -> add_constant_1 -> add_constant_2 -> output
# This will apply: input + 1 + 2 = input + 3
apply_pipeline_transform(
    input_tensor, output_tensor, add_one_function, add_two_function
)
```
The two `ExternalFunction`s will be associated with two different objects which doesn't make sense. The source code is the same, the compiler flags are the same, and hence we should be smart enough to detect that we need a single code object. This PR adds a cache so we can support multiple kernels from the same object without much overhead or extra lines of code.


ypapadop-amd:
> @ypapadop-amd, any update?

It appears that initializing an `ExternalFunction` expects to find the source file from where the initialization happens. We will also pay the cost of reading the source file, even if the key is not useful to a non-JIT flow.

Can we put the cache inside the compile function instead of being a singleton of the class?

mawad-amd:
You are welcome to change the JIT backend to support managing the code objects. As long as the abstraction is the same, that’s fine by me.

The behavior that this PR enable should just work. It doesn’t make sense to have same code, same compiler flags and you end up with two different code objects for JIT. Archive idea is fine but doesn’t solve the problem.

ypapadop-amd:
> You are welcome to change the JIT backend to support managing the code objects. As long as the abstraction is the same, that’s fine by me.
> 
> The behavior that this PR enable should just work. It doesn’t make sense to have same code, same compiler flags and you end up with two different code objects for JIT. Archive idea is fine but doesn’t solve the problem.

I never suggested that you should have different code objects, which is not supported anyway.

I will debate programmability though. This PR requires duplication of compile options at the creation of `ExternalFunction` even if they don't apply for that particular compiled function but still requires centralized function definition. I cannot think of another system with a similar behavior.

I don't think that following is valid, but it could be if you'd defer all caching and related until compilation:
```python
tile_size = 16
add_one_function = ExternalFunction(
    "add_constant_1",
    source_string="""extern "C" {
        void add_constant_1(T1* input, T1* output, int tile_size) {
            for (int i = 0; i < tile_size; i++) {
                output[i] = input[i] + 1;
            }
        }""",
    arg_types=[
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.int32,
    ],
    compilation_flags = ["-DT1=int"],
)

add_two_function = ExternalFunction(
    "add_constant_2",
    source_string="""extern "C" {
        void add_constant_2(T2* input, T2* output, int tile_size) {
            for (int i = 0; i < tile_size; i++) {
                output[i] = input[i] + 2;
            }
        }
    }""",
    arg_types=[
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.ndarray[(tile_size,), np.dtype[np.int32]],
        np.int32,
    ],
    compilation_flags = ["-DT2=int"],
)

apply_pipeline_transform(
    input_tensor, output_tensor, add_one_function, add_two_function
)
```
It'll require stitching the source code strings together and probably detect duplicate macro definitions.

An archive would be on the other side of the spectrum as shown in https://github.com/Xilinx/mlir-aie/pull/2509#issuecomment-3197048259

fifield:
> I don't think that following is valid, but it could be if you'd defer all caching and related until compilation:
> 
> ```python
> tile_size = 16
> add_one_function = ExternalFunction(
>     "add_constant_1",
>     source_string="""extern "C" {
>         void add_constant_1(T1* input, T1* output, int tile_size) {
>             for (int i = 0; i < tile_size; i++) {
>                 output[i] = input[i] + 1;
>             }
>         }""",
>     arg_types=[
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.int32,
>     ],
>     compilation_flags = ["-DT1=int"],
> )
> 
> add_two_function = ExternalFunction(
>     "add_constant_2",
>     source_string="""extern "C" {
>         void add_constant_2(T2* input, T2* output, int tile_size) {
>             for (int i = 0; i < tile_size; i++) {
>                 output[i] = input[i] + 2;
>             }
>         }
>     }""",
>     arg_types=[
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.int32,
>     ],
>     compilation_flags = ["-DT2=int"],
> )
> 
> apply_pipeline_transform(
>     input_tensor, output_tensor, add_one_function, add_two_function
> )
> ```
> 

It's counter intuitive (broken?) that this doesn't already work.

> It'll require stitching the source code strings together and probably detect duplicate macro definitions.

Why? I would expect each `ExternalFunction` to create an object file from its source string + flags and that those object files might get linked together at the end depending on the requirements (e.g. they are on the same core tile). What is mlir-aie doing instead?

ypapadop-amd:
> > It'll require stitching the source code strings together and probably detect duplicate macro definitions.
> 
> Why? I would expect each `ExternalFunction` to create an object file from its source string + flags and that those object files might get linked together at the end depending on the requirements (e.g. they are on the same core tile). What is mlir-aie doing instead?

I thought we can only link against a single `.o`. If it's indeed a limitation, `ld- --relocatable` can combine `.o` files.

mawad-amd:
> ```python
> tile_size = 16
> add_one_function = ExternalFunction(
>     "add_constant_1",
>     source_string="""extern "C" {
>         void add_constant_1(T1* input, T1* output, int tile_size) {
>             for (int i = 0; i < tile_size; i++) {
>                 output[i] = input[i] + 1;
>             }
>         }""",
>     arg_types=[
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.int32,
>     ],
>     compilation_flags = ["-DT1=int"],
> )
> 
> add_two_function = ExternalFunction(
>     "add_constant_2",
>     source_string="""extern "C" {
>         void add_constant_2(T2* input, T2* output, int tile_size) {
>             for (int i = 0; i < tile_size; i++) {
>                 output[i] = input[i] + 2;
>             }
>         }
>     }""",
>     arg_types=[
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.ndarray[(tile_size,), np.dtype[np.int32]],
>         np.int32,
>     ],
>     compilation_flags = ["-DT2=int"],
> )
> 
> apply_pipeline_transform(
>     input_tensor, output_tensor, add_one_function, add_two_function
> )
> ```
> 
> It'll require stitching the source code strings together and probably detect duplicate macro definitions.

This PR is not addressing this case. This looks like a good motivation for what you are proposing (an actual use case would be even better). Again, you are free to implement it but This PR addresses a different use case per the PR description. I would prefer if we move the archive discussion to a different issue/PR. This PR is not trying to solve that problem.

mawad-amd:
> Have you considered two `ExternalFunction`s passed to the same `Worker`?

Do you have an example? The example I saw (and wrote the test after) was using two workers.

jgmelber:
> > Have you considered two `ExternalFunction`s passed to the same `Worker`?
> 
> Do you have an example? The example I saw (and wrote the test after) was using two workers.

This is the case I am thinking about: https://github.com/Xilinx/mlir-aie/blob/19a0a209c83f4b856d085b5b6dce515723b394f6/programming_examples/ml/silu/Makefile#L33-L57

ypapadop-amd:
> > It'll require stitching the source code strings together and probably detect duplicate macro definitions.
> 
> Why? I would expect each `ExternalFunction` to create an object file from its source string + flags and that those object files might get linked together at the end depending on the requirements (e.g. they are on the same core tile). What is mlir-aie doing instead?

@fifield I made a janky pass of this in https://github.com/Xilinx/mlir-aie/pull/2530

jgmelber:
Is this redundant to #2611

andrej:
> If we consider different build systems, SCons or some other Python-based build systems may be easier to integrate.

Yes, I'd be on board with considering other build systems as well. I'm not well versed in the space but just see a lot of overlap.

> I can't envision how CMake can help with this. It's not great to begin with if you want to mix multiple languages let alone integrating it as a JIT build system.

What we've done elsewhere is create a CMake function like `add_aie_gemm()` that allows you to specify the particular parameters for the operator you want  to generate. That allows for a clean interface between user and producer, since behind the function, you can do whatever you need to generate your targets (e.g., create archives), while only exposing the parameters to the user (i.e., the JIT user would only need to specify the required GEMM parameters). If then you need that operator elsewhere (outside of JIT), you can use the same CMake function, without having to worry about how to build it. My impression is that the build instructions of the operators are currently more intertwined with the JIT implementation, so extracting an operator out of the JIT flow when you need it elsewhere might be harder (?). I'll admit this is just from a superficial view of how the JIT flow is built up, so please feel free to correct me.

jgmelber:
Bumping this as well. What's the latest status on this? @mawad-amd @andrej @ypapadop-amd

hunhoffe:
Closing in favor of https://github.com/Xilinx/mlir-aie/pull/3025

Issue #2515, opened 2025-08-19
Remove PYTHONPATH additions


ypapadop-amd:
Removing additions to PYTHONPATH (e.g., https://github.com/Xilinx/mlir-aie/blob/0dd0f191ead9174f2e054437582925527af49d22/utils/env_setup.sh#L67) will remove both the need for the set-up script and manually creating `.pth` files when IRON is installed from a wheel.

Ideally, the installation would be structured as
```
aie/
   compiler/
   dialects/
   ...
   non-python-stuff/
```
but this would require a total restructuring.

Instead, it appears that there may be a way to override the default behaviour during setup.py and copy our own .pth in `venv/lib/pythonX.Y/site-packages/` as described in https://niteo.co/blog/setuptools-run-custom-code-in-setup-py/

PR #2516, opened 2025-08-19
Remove PYTHONPATH manipulation during wheel installation

ypapadop-amd:
This PR adds a pth file by default in site-packages to avoid adding `mlir_aie/python` to the PYTHONPATH. The pth way was chosen because restructuring the wheel package would be a much bigger undertaking which can happen at a later time. More information in https://github.com/Xilinx/mlir-aie/issues/2515


jgmelber (approved):

erwei-xilinx:
@ypapadop-amd thanks for working on this. There are two variants to the wheels that mlir-aie releases: `mlir_aie` and `mlir_aie_no_rtti`. Would the aie.pth be able to support both? Triton for example is built with rtti set to "off", so in triton-air, we would have to use `mlir_aie_no_rtti` instead.

erwei-xilinx:
With `mlir_aie_no_rtti`, the path under site-packages looks like `aie_venv/lib/python3.10/site-packages/mlir_aie_no_rtti/python`.

ypapadop-amd:
> With `mlir_aie_no_rtti`, the path under site-packages looks like `aie_venv/lib/python3.10/site-packages/mlir_aie_no_rtti/python`.

Are they both installed side-by-side or only one at a time?

erwei-xilinx:
> > With `mlir_aie_no_rtti`, the path under site-packages looks like `aie_venv/lib/python3.10/site-packages/mlir_aie_no_rtti/python`.
> 
> Are they both installed side-by-side or only one at a time?

They are intended to be installed only one at a time.

erwei-xilinx (approved):

ypapadop-amd:
@erwei-xilinx @jgmelber the latest changes add a `-no_rtti` at the end of the version. Let me know if this is an acceptable solution.

erwei-xilinx (approved):
Looks good to me, thanks!

PR #2518, opened 2025-08-20
Peano linker support

ypapadop-amd:
This PR adds a function to combine object files from Python using Peano.

### DIFF
--- a/python/iron/compile/__init__.py
+++ b/python/iron/compile/__init__.py
@@ -1,1 +1,10 @@
+# __init__.py -*- Python -*-
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
+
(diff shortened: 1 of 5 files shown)


programming_examples/vision/edge_detect/Makefile
```diff
@@ -47,15 +47,15 @@ else
 	echo "Device type not supported"
 endif
 
-build/combined_gray2rgba_addWeighted.a: build/gray2rgba.cc.o build/addWeighted.cc.o
+build/combined_gray2rgba_addWeighted.o: build/gray2rgba.cc.o build/addWeighted.cc.o
 	mkdir -p ${@D}
-	ar rvs $@ $< $(word 2,$^)
+	${shell python3 -c "import aie.iron.compile; aie.iron.compile.merge_object_files(output_file = '${@D}/combined_gray2rgba_addWeighted.o', object_files = ['${@D}/gray2rgba.cc.o', '${@D}/addWeighted.cc.o'])"}
```
fifield: I'm not sure how this is better than `ar rvs` here.

ypapadop-amd: I meant it as the test. Do we assume that `ar` is always present on a system?

fifield (approved):

PR #2519, opened 2025-08-20
allow specifying install directory of AIE-RT in build-mlir-aie.sh

andrej:
The runtime lib (with the test utils) [only gets built](https://github.com/Xilinx/mlir-aie/blob/19a0a209c83f4b856d085b5b6dce515723b394f6/runtime_lib/CMakeLists.txt#L63) if CMake finds the AIE-RT driver (a.k.a. libXAIE). The only places our CMake scripts looks for libXAIE is in a Vitis installation or in a path specified during build. This is probably part of the reason [a lot of the programming examples build the test_utils.cpp themselves](https://github.com/Xilinx/mlir-aie/blob/19a0a209c83f4b856d085b5b6dce515723b394f6/programming_examples/basic/shuffle_transpose/Makefile#L24) instead of linking against the prebuilt library -- it's probably missing from a lot of installations. Since we are already shipping aie-rt as a submodule, it might be good to at some point add something to these build scripts that just builds it and uses it for the test lib.

For now, this enables the convenience of using `./utils/build-mlir-aie.sh` and specifying where AIE-RT is built. And while I'm at it, I'd like to sneak in the option of changing to a Debug build as well since I often find myself editing the build script just for that.


fifield (approved):

andrej:
Documenting in case I run into this again in the future; setting this environment variable appears to no longer be needed, instead this works:

```
# clone MLIR-AIE including all submodules
cd mlir-aie/third_party/aie-rt/driver
mkdir build
cd build
CMAKE_INSTALL_PREFIX=/path/to/aie-rt-install/ cmake ..
make
make install
cd mlir-aie
CMAKE_PREFIX_PATH=/path/to/aie-rt-install ./utils/build-mlir-aie-from-wheels.sh
```

Jeff's related PR: #2545

Issue #2485, opened 2025-08-08
Introduce `get_target_architecture` helper function


mawad-amd:
This issue proposes adding a `get_target_architecture` function that returns something like "aie2p" or "aie2". Jeff's suggestion which Joe and I recommend is below.

_Originally posted by @fifield in https://github.com/Xilinx/mlir-aie/pull/2475#discussion_r2257830162_

I'd propose using the target model which already has this information to support C++ RTTI. Then it's a simple wrapper like getting the number of rows or columns:
```python
    # in iron/device.py
    @property
    def arch(self):
        return get_target_model(self._device).get_target_arch()
```

I guess it would return a python enum of the C++:
```c++
enum class AIEArch : uint32_t {
  AIE1 = 1,
  AIE2 = 2,
  AIE2p = 3,
};
```
The TODO here is to add `get_target_arch()` wrapping `AIETargetModel::getTargetArch` in `python/AIEMLIRModule.cpp`

PR #2524, opened 2025-08-22
VectorToAIEVec: get device arch from IR, instead of pass option

erwei-xilinx:
- The conversion pass from `vector` to `aievec` needs to be placed fairly early in the `aiecc` pipeline, when it doesn't yet parse the device model; it's more convenient if the conversion pass reads the `aie.device` op instead.
- Registers conversion passes with `vector` dialect and `aievec` dialect into `aiecc` pipeline.

### DIFF
--- a/lib/Dialect/AIEVec/Utils/Utils.cpp
+++ b/lib/Dialect/AIEVec/Utils/Utils.cpp
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "aie/Dialect/AIEVec/Utils/Utils.h"
+#include "aie/Dialect/AIE/IR/AIEDialect.h"
 #include "mlir/Dialect/Affine/IR/AffineOps.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Vector/IR/VectorOps.h"
@@ -103,6 +104,17 @@ VectorType getFlattenedVectorType(VectorType vecTy) {
   return VectorType::get(
       {std::accumulate(shape.begin(), shape.end(), 1, std::multiplies<>())},
       vecTy.getElementType());
+}
+
+std::optional<xilinx::AIE::AIEArch> getAIEVersionFromModule(ModuleOp moduleOp) {
+  llvm::SmallVector<xilinx::AIE::DeviceOp> deviceOps;
+  moduleOp.walk([&](xilinx::AIE::DeviceOp d) { deviceOps.push_back(d); });
+
+  if (deviceOps.empty())
+    return std::nullopt;
+
+  const auto &targetModel = deviceOps.front().getTargetModel();
+  return targetModel.getTargetArch();
 }
 
 } // namespace xilinx::aievec
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -36,6 +36,7 @@ from aie.passmanager import PassManager
 
 INPUT_WITH_ADDRESSES_PIPELINE = lambda scheme, dynamic_objFifos, ctrl_pkt_overlay: (
     Pipeline()
+    .add_pass("convert-vector-to-aievec{target-backend=llvmir}")
     .lower_affine()
     .add_pass("aie-canonicalize-device")
     .Nested(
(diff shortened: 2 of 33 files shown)


lib/Dialect/AIEVec/Utils/Utils.cpp
```diff
@@ -105,4 +106,15 @@ VectorType getFlattenedVectorType(VectorType vecTy) {
       vecTy.getElementType());
 }
 
+std::optional<xilinx::AIE::AIEArch> getAIEVersionFromModule(ModuleOp moduleOp) {
```
stephenneuendorffer: The downside of this is that it makes the AIEVec lowerings useless outside of a device.   Perhaps they should *also* allow passing from the command line?

python/compiler/aiecc/main.py
```diff
@@ -397,6 +399,12 @@ def downgrade_ir_for_peano(llvmir):
     return llvmir
 
 
+def drop_alignment_for_peano(llvmir):
```
stephenneuendorffer, line 418: Is this still needed now that peano is updated close to HEAD?

stephenneuendorffer (commented):
While I think this could be simplified in the code, I don't really think this is a net win.  I think this shouldn't need to be tied so closely to the rest of the AIE dialect.

lib/Dialect/AIEVec/Utils/Utils.cpp
stephenneuendorffer: Could we use this: https://mlir.llvm.org/docs/DataLayout/#targetdevicespecinterface-targetdevicespecinterface?

fifield:
> * The conversion pass from `vector` to `aievec` needs to be placed fairly early in the `aiecc` pipeline, when it doesn't yet parse the device model; it's more convenient if the conversion pass reads the `aie.device` op instead.

It looks like the `aie-translate` call decoding the architecture in aiecc could be moved up ~15 lines of code so that `aie_target` is known when you run the aievec pass

You could also do something like [this](https://github.com/Xilinx/mlir-aie/blob/main/python/compiler/aiecc/main.py#L215):
```python
    device = find_ops(
        module.operation,
        lambda o: isinstance(o.operation.opview, aiedialect.DeviceOp),
    )
    device = aiedialect.AIEDevice(int(device[0].device))
    tm = aiedialect.get_target_model(device)
    ...
```
and implement https://github.com/Xilinx/mlir-aie/issues/2485 to get the arch.

python/compiler/aiecc/main.py
erwei-xilinx: Yeah I ran some local checks and it seems to be no longer needed. Let me run this through the CI to make sure.

https://github.com/Xilinx/mlir-aie/pull/2524/commits/5f3caceab08bf4621f085a927f3d12cf2be40762

erwei-xilinx: No looks like this peano hack is still needed for the current peano. CI log: https://github.com/Xilinx/mlir-aie/actions/runs/17162960610/job/48696490531?pr=2524

lib/Dialect/AIEVec/Utils/Utils.cpp
erwei-xilinx: Makes great sense to me. I realized that those changes I made to the pass are not necessary to make the workflow work. Those commits have been reverted.

erwei-xilinx:
> > * The conversion pass from `vector` to `aievec` needs to be placed fairly early in the `aiecc` pipeline, when it doesn't yet parse the device model; it's more convenient if the conversion pass reads the `aie.device` op instead.
> 
> It looks like the `aie-translate` call decoding the architecture in aiecc could be moved up ~15 lines of code so that `aie_target` is known when you run the aievec pass
> 
> You could also do something like [this](https://github.com/Xilinx/mlir-aie/blob/main/python/compiler/aiecc/main.py#L215):
> 
> ```python
>     device = find_ops(
>         module.operation,
>         lambda o: isinstance(o.operation.opview, aiedialect.DeviceOp),
>     )
>     device = aiedialect.AIEDevice(int(device[0].device))
>     tm = aiedialect.get_target_model(device)
>     ...
> ```
> 
> and implement #2485 to get the arch.

Great point! I moved it exactly 15 lines above and the workflow seems to still work. The changes made to the `convert-vector-to-aievec` have been reverted, and PR changed to adopt the first solution. Looking forward to seeing https://github.com/Xilinx/mlir-aie/issues/2485 landed in an PR, so that I can switch to that instead.

PR #2526, opened 2025-08-22
Remove `Pure` trait for `LockOp`

erwei-xilinx:
Locks in AIE are hardware resources and `LockOp` sets its initial value, so they have hardware side effects. They shouldn't get canonicalized away even if they have no users.


fifield (approved):
I recently had an issue with "unused" locks getting canonicalized away in simple unit tests. Maybe this will fix it!

PR #2530, opened 2025-08-22
Combining multiple object files in JIT

ypapadop-amd:


python/iron/jit.py
```diff
@@ -147,6 +145,7 @@ def decorator(*args, **kwargs):
 
         # Clear any instances from previous runs to make sure if the user provided any broken code we don't try to recompile it
         ExternalFunction._instances.clear()
+        ExternalFunction._bin_name = function.__name__ + ".o"
```
mawad-amd, line 202: Can we add a comment explaining why we need this line?

ypapadop-amd:
> Thanks for writing this. This approach looks fine to me and tying the code object name to the kernel itself is good enough for me for now. The non-ideal thing here is that a user must have one function per source file. It's not ideal but a non-issue for me. I will let Jeff and Joe decide if that's an issue.
> 
> I ported the tests I wrote to have each function in different file here if you wanted to reuse them https://gist.github.com/mawad-amd/56fed24599343f22ab4f51d0c679a4a9

That is true, you can't have more than one function referenced from an ExternalFunction in a single compilation unit as the current ExternalFunction design has 1:1 map between external function and compilation units. This is only a prototype to allow the use-case that was brought up.

I don't have a good solution for having multiple ExternalFunctions to bind to the same object file. Maybe we allow one ExternalFunction to bind to another one? Or create an array of ExternalFunctions?

PR #2532, opened 2025-08-22
Allow multiple devices in one design and add device configuration op (`aiex.configure`)

andrej:
- allow devices to have symbol names and add compile-time flags to decide which devices/sequences should be compiled
- for files that are specific to a certain device, `aiecc.py` will generate separate files for each device
- during core ELF generation, write paths back to MLIR instead of relying on a fixed path in aiecc.py; to this end, add a "set elf for core" pass
- factor out some of the address generation for write32/blockwrites that was duplicated code (will also be reused in future reconfig PR)
- make AIECreatePathFindFlows work in multi-threaded environment with multiple devices
- factor out getOrCreateDataMemref (with deduplication) into a utility function (this will become useful for future reconfig PR)
- add a `aiex.configure` op which for now has no lowering yet (this will be useful for future PRs)

Examples of new syntax:
- [good](https://github.com/andrej/mlir-aie/blob/reconfig-clean/test/dialect/AIEX/sequence_with_configures.mlir)
- [bad](https://github.com/andrej/mlir-aie/blob/reconfig-clean/test/dialect/AIEX/sequence_with_configures_bad.mlir)


andrej:
Looks like tests finally pass. After review, I'll squash the commits before merging and get rid of the meaningless commit messages.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -325,6 +334,11 @@ def AIE_CoreOp: AIE_Op<"core", [
     for each core.  The name of this file can be be specified using the `elf_file`
     attribute.
 
+    If the `elf_file` attribute is present, no MLIR besides a terminator may be
+    present in the core; in that case, the binary file linked dictates what 
+    will run in the core. The path specified should is relative to the MLIR
+    file.
```
fifield, line 341: Why does the presence of the attribute require that the core body is deleted? It seems to hard code things for the current aiecc use case.

fifield (commented):
I get some failures when running with Vitis.

A number of them are like this:
```
********************
FAIL: AIE_TEST :: unit_tests/aie/05_tiledma/aie.mlir (15 of 189)
******************** TEST 'AIE_TEST :: unit_tests/aie/05_tiledma/aie.mlir' FAILED ********************
Exit Code: 1

Command Output (stdout):
--
error: input MLIR must contain at least one aie.device
```

This one manually uses `elf_file` attribute and failed:
```
/work/acdc/build/aie/test/unit_tests/chess_compiler_tests_aie2/02_precompiled_kernel/aie.mlir.prj/input_with_addresses.mlir:9:17: error: 'aie.core' op When `elf_file` attribute is specified, core body must be empty (consist of exactly one `aie.end` op).
    %core_1_3 = aie.core(%tile_1_3) {
```

python/compiler/aiecc/main.py
```diff
@@ -1327,75 +1430,116 @@ async def run_flow(self):
...
+                self.gen_sim(pb.task, aie_target, input_physical, device_name)
+            )
 
-            await asyncio.gather(*processes)
+        input_physical_with_elfs_str = await read_file_async(input_physical_with_elfs)
+
+        if (opts.cdo or opts.xcl or opts.pdi) and opts.execute:
+            await self.process_cdo(input_physical_with_elfs_str, device_name)
+
+        processes = []
```
fifield: This erases the `process.append` of line 1519, resulting in:
```
/work/vitis-venv/bin/python3.12 /work/acdc/build/aie/bin/aiecc.py --aiesim --xchesscc --xbridge --no-compile-host /work/acdc/aie/test/unit_tests/chess_compiler_tests_aie2/03_simple/aie.mlir -I/work/acdc/build/aie/runtime_lib/x86_64/test_lib/include -L/work/acdc/build/aie/runtime_lib/x86_64/test_lib/lib -ltest_lib /work/acdc/aie/test/unit_tests/chess_compiler_tests_aie2/03_simple/test.cpp
/work/acdc/build/aie/python/aie/compiler/aiecc/main.py:1528: RuntimeWarning: coroutine 'FlowRunner.gen_sim' was never awaited
```

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: I thought it would be more clear this way, removing the ambiguity of which code gets executed if both the attribute is there and there is code inside the body. If this complicates some use case I'm not aware of, I can remove this restriction.

andrej:
> I get some failures when running with Vitis.

Oh, that's not good. Looks like none of the tests in `unit_tests/aie` ran, I didn't realize this. I'll run those manually.

Would it be fine to refactor those to require the code be wrapped in an `aie.device`? That would make things easier on my end, but if for backwards-compatibility we need to support tiles etc directly under `module`, I can make that possible too.

fifield:
> > I get some failures when running with Vitis.
> 
> Oh, that's not good. Looks like none of the tests in `unit_tests/aie` ran, I didn't realize this. I'll run those manually.
> 
> Would it be fine to refactor those to require the code be wrapped in an `aie.device`? That would make things easier on my end, but if for backwards-compatibility we need to support tiles etc directly under `module`, I can make that possible too.

The current state of things was a convenience for backward compatibility. IIRC VC1902 is assumed if `aie.device` is missing. In my opinion we can refactor the older tests to include `aie.device` and remove the backward compatibility feature.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -537,13 +542,61 @@ def AIE_RuntimeSequenceOp : AIEX_Op<"runtime_sequence", [NoTerminator, HasParent
...
+  }];
+
+  let regions = (region
+    AnyRegion:$body
+  );
+
+  let hasVerifier = 1;
+}
+
+def AIE_RunOp: AIEX_Op<"run", [HasParent<"ConfigureOp">]> {
```
AndraBisca, line 587: Could you please add a short summary for this operation?

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -62,6 +62,9 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
   matchAndRewrite(FlowOp flowOp, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     Operation *Op = flowOp.getOperation();
+    DeviceOp d = flowOp->getParentOfType<DeviceOp>();
+    assert(d);
```
AndraBisca: Does this assert check if a device was found? Since asserts can be disabled in certain compilation modes it might be better to use an if statement.

lib/Targets/AIETargetLdScript.cpp
```diff
@@ -63,14 +63,17 @@ static void writeLDScriptMap(raw_ostream &output, BufferOp buf, int offset) {
...
 LogicalResult xilinx::AIE::AIETranslateToLdScript(ModuleOp module,
                                                   raw_ostream &output,
-                                                  int tileCol, int tileRow) {
+                                                  int tileCol, 
+                                                  int tileRow,
+                                                  llvm::StringRef deviceName) {
   DenseMap<TileID, Operation *> tiles;
   DenseMap<Operation *, SmallVector<BufferOp, 4>> buffers;
 
+  DeviceOp targetOp = AIE::DeviceOp::getForSymbolInModule(module, deviceName);
```
AndraBisca: Does this need a check?

test/aiecc/simple.mlir
```diff
@@ -32,12 +32,14 @@
...
 
 module {
-  %12 = aie.tile(1, 2)
-  %buf = aie.buffer(%12) : memref<256xi32>
-  %4 = aie.core(%12)  {
-    %0 = arith.constant 0 : i32
-    %1 = arith.constant 0 : index
-    memref.store %0, %buf[%1] : memref<256xi32>
-    aie.end
+  aie.device(xcvc1902) {
```
AndraBisca, line 35: Is this not added by the device canonicalization pass?

jgmelber: IIRC it is. However this is something I wanted to revisit at some point. Maybe vck190 isn't the best default. And perhaps devices could be inserted/modified by flags to aiecc.py at compile time rather than at MLIR generation from Python or design time.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -272,4 +273,24 @@ def AIEGenerateColumnControlOverlay : Pass<"aie-generate-column-control-overlay"
   ];
 }
 
+def AIESetELFforCore : Pass<"aie-set-elf-for-core", "DeviceOp"> {
```
fifield: could do this one in memory in aiecc main.py:
```python
def set_elf_file_on_tile(module, row, col, path):
    for core in find_ops(
        module.operation,
        lambda o: isinstance(o.operation.opview, CoreOp)
        and o.tile.owner.opview.col.value == col
        and o.tile.owner.opview.row.value == row,
    ):
        core.attributes['elf_file'] = StringAttr.get(path)
    return module

```

andrej: Oh yes that's much better. I knew as I was doing it that a whole pass for this was overkill but I didn't think of this obvious solution

test/aiecc/simple.mlir
andrej: Copying my discussion with @fifield with regards to not specifying an `aie.device`:

> > Would it be fine to refactor those to require the code be wrapped in an aie.device? That would make things easier on my end, but if for backwards-compatibility we need to support tiles etc directly under module, I can make that possible too.
> 
> The current state of things was a convenience for backward compatibility. IIRC VC1902 is assumed if aie.device is missing. In my opinion we can refactor the older tests to include aie.device and remove the backward compatibility feature.

python/compiler/aiecc/main.py
```diff
@@ -1327,75 +1430,116 @@ async def run_flow(self):
...
-                processes.append(
-                    self.process_core(
-                        core,
-                        aie_target,
-                        aie_peano_target,
-                        file_with_addresses,
+            # 2.5) Targets that require the cores to be lowered but apply across all devices
+            if opts.txn and opts.execute:
+                input_physical_with_elfs_str = await read_file_async(input_physical_with_elfs)
+                input_physical_with_elfs = await self.process_txn(input_physical_with_elfs_str)
```
fifield: If there are multiple `aie.device` in the input file, and the device is specified with `--device-name`, then only `aie.core` in that device will have elfs made for them by `self.process_cores` (line 1442).  But all devices' cores are expected to have elfs in `self.process_txn` (line 1448) because it calls `convert-aie-to-transaction` on every `aie.device` in `input_physical_with_elfs.mlir`. If a `aie.core` does not have an elf in `convert-aie-to-transaction` then the error `"'aie.core' op Expected lowered ELF file to be given as attribute `elf_file` for this core. Compile cores first."` is thrown. This happens in `test/npu-xrt/add_one_two_txn` when the mlir files are combined. I think the fix will be to filter by device and/or sequence name in `self.process_txn`

andrej: Indeed. Thanks for catching that, I will implement your suggested fix.

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -648,48 +607,23 @@ struct WriteBdToBlockWritePattern : OpConversionPattern<NpuWriteBdOp> {
...
       rewriter.setInsertionPoint(
           op->getParentOfType<AIEX::RuntimeSequenceOp>());
-      std::string name = "blockwrite_data_";
-      while (dev.lookupSymbol(name + std::to_string(cachedId)))
-        cachedId++;
-      name += std::to_string(cachedId);
-      global = rewriter.create<memref::GlobalOp>(
-          op->getLoc(), name, rewriter.getStringAttr("private"), memrefType,
-          initVal, true, nullptr);
+      global = getOrCreateDataMemref(rewriter, dev, op.getLoc(), words);
```
fifield, line 615: I implemented the optimization you've refactored here, but I had been wondering if it should be removed instead. The reason is that I had to disable it for some cases in a downstream branch because unlike the current scheme, some runtime schemes patch the address in the data _before_ writing it to the shim dma (e.g. cert, maybe "new elf" too). So the data can't be shared in those cases.

andrej: I just factored it out because I ended up using it multiple places and didn't want to have code duplication. Hopefully that should make removing it (or somehow adapting it to the cert case) easier down the line too.

python/compiler/aiecc/main.py
andrej: fixed

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: I have left this check in place for now. Let me know if you want to discuss it further.

include/aie/Dialect/AIEX/IR/AIEX.td
andrej: Done!

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: Indeed, replaced the assert with a proper error.

lib/Targets/AIETargetLdScript.cpp
andrej: It does. Added

test/aiecc/simple.mlir
andrej: Added the "implicit" `aie.device` in all tests and going forward, a device declaration will be required. Since now multiple devices will be possible in the same file, it is required to specify which device operations belong to. We can add functionality to make the device _type_ a command-line configurable parameter so we can reuse the same design for multiple architectures, as suggested by @jgmelber .

include/aie/Dialect/AIE/Transforms/AIEPasses.h
```diff
@@ -47,6 +47,7 @@ createAIEAssignBufferDescriptorIDsPass();
 std::unique_ptr<mlir::OperationPass<DeviceOp>>
 createAIEGenerateColumnControlOverlayPass();
 std::unique_ptr<mlir::OperationPass<DeviceOp>> createAIEAssignTileCtrlIDsPass();
+std::unique_ptr<mlir::OperationPass<DeviceOp>> createAIESetELFforCorePass();
```
fifield:
```suggestion
```
I think this was removed?

include/aie/Dialect/AIEX/IR/AIEXDialect.h
```diff
@@ -13,6 +13,12 @@
 
 #include "aie/Dialect/AIE/IR/AIEDialect.h"
 
+#include "mlir/IR/BuiltinAttributes.h"
+
+namespace xilinx::AIEX {
+const char defaultRuntimeSequenceName[] = "sequence";
```
fifield: Could this be in the op like `RuntimeSequenceOp::defaultName` and still work in tablegen?
Or better,
```c++
let extraClassDeclaration = [{
     llvm::StringRef getDefaultName() { return "sequence"; }
}];
```

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -437,7 +437,8 @@ xilinx::AIE::convertTransactionBinaryToMLIR(mlir::MLIRContext *ctx,
   // create aie.device
   std::vector<AIEDevice> devices{AIEDevice::npu1_1col, AIEDevice::npu1_2col,
                                  AIEDevice::npu1_3col, AIEDevice::npu1};
-  auto device = builder.create<DeviceOp>(loc, devices[columns - 1]);
+  auto device = builder.create<DeviceOp>(
+      loc, devices[columns - 1], StringAttr::get(builder.getContext(), "main"));
```
fifield: Is the intent to use the default name here? I know the `getDefaultName()` doesn't exist, but maybe like,
```suggestion
      loc, devices[columns - 1], StringAttr::get(builder.getContext(), DeviceOp::getDefaultName()));
```
 Or have a constructor that fills in the default,
```suggestion
      loc, devices[columns - 1]);
```

lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
```diff
@@ -39,8 +39,8 @@ struct AIECanonicalizeDevicePass
 
     Location location = builder.getUnknownLoc();
     auto deviceOp = builder.create<DeviceOp>(
-        location,
-        AIEDeviceAttr::get(builder.getContext(), AIEDevice::xcvc1902));
+        location, AIEDeviceAttr::get(builder.getContext(), AIEDevice::xcvc1902),
+        StringAttr::get(builder.getContext(), "main"));
```
fifield: Replace "main" with default getter?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -549,13 +550,127 @@ LogicalResult AIEX::NpuWriteBdOp::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// NpuWrite32Op
+//===----------------------------------------------------------------------===//
+
+template <typename T>
+std::optional<uint32_t> getAbsoluteAddress(T *op) {
```
fifield:
```suggestion
static std::optional<uint32_t> getAbsoluteAddress(T *op) {
```

lib/Dialect/AIEX/Transforms/CMakeLists.txt
```diff
@@ -36,4 +36,5 @@ add_mlir_dialect_library(AIEXTransforms
   MLIRPass
   MLIRSupport
   MLIRTransformUtils
+  AIEXUtils
```
fifield: I think LINK_LIBS was alphabetical.

lib/Dialect/AIEX/Utils/AIEUtils.cpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace mlir;
+using namespace xilinx;
+using namespace xilinx::AIEX;
+
+static unsigned cachedId = 0;
+
+namespace xilinx {
+namespace AIEX {
+
+memref::GlobalOp getOrCreateDataMemref(OpBuilder &builder, AIE::DeviceOp dev,
```
fifield:
```suggestion

static unsigned cachedId = 0;

memref::GlobalOp AIEX::getOrCreateDataMemref(OpBuilder &builder, AIE::DeviceOp dev,
```

programming_examples/basic/matrix_multiplication/makefile-common
```diff
@@ -162,7 +162,7 @@ ${trace_mlir_target}: ${srcdir}/${aie_py_src}
 
 ${xclbin_target}: ${mlir_target} ${kernels:%=build/%.o}
 	mkdir -p ${@D}
-	cd ${@D} && aiecc.py --alloc-scheme=${buffer_aloc_flag} --aie-generate-xclbin --no-compile-host --xclbin-name=${@F} \
+	cd ${@D} && aiecc.py --alloc-scheme=${buffer_aloc_flag} --aie-generate-xclbin --no-compile-host --xclbin-name=${@F} --progress \
```
fifield: was this one intended?

python/compiler/aiecc/cl_arguments.py
```diff
@@ -270,6 +270,24 @@ def parse_args(args=None):
...
+        dest="ctrlpkt_dma_seq_name",
+        default="{0}_ctrlpkt_dma_seq.bin",
+        help="Output filename for control packet DMA sequence. `{0}` is replaced with the device name.",
+    )
+    parser.add_argument(
+        "--ctrlpkt-elf-name",
+        dest="ctrlpkt_elf_name",
+        default="{0}_ctrlpkt.elf",
+        help="Output filename for control packet DMA sequence and control packet binary data in combined ELF file. `{0}` is replaced with the device name.",
+    )
```
fifield, line 290: Why the new elf name option? One already exists. I don't think it need differentiation based on what's inside, esp in the future.

python/compiler/aiecc/cl_arguments.py
```diff
@@ -270,6 +270,24 @@ def parse_args(args=None):
...
+        "--ctrlpkt-name",
+        dest="ctrlpkt_name",
+        default="{0}_ctrlpkt.bin",
+        help="Output filename for control packet binary data. `{0}` is replaced with the device name.",
+    )
+    parser.add_argument(
+        "--ctrlpkt-dma-seq-name",
+        dest="ctrlpkt_dma_seq_name",
+        default="{0}_ctrlpkt_dma_seq.bin",
+        help="Output filename for control packet DMA sequence. `{0}` is replaced with the device name.",
```
fifield, line 283: This will get deleted with https://github.com/Xilinx/mlir-aie/pull/2407

test/dialect/AIEX/sequence_with_configures_bad.mlir
```diff
@@ -0,0 +1,82 @@
+//===- sequence_with_configures.mlir ---------------------------*- MLIR -*-===//
```
fifield:
```suggestion
//===- sequence_with_configures_bad.mlir -----------------------*- MLIR -*-===//
```

test/npu-xrt/add_one_two/aie.mlir
fifield, line 1: I have more of this test merging in:
https://github.com/Xilinx/mlir-aie/pull/2513/commits/2e5cdc18bf361e8241670df948fb4f9f5c793691
https://github.com/Xilinx/mlir-aie/pull/2513/commits/249dc0741a78402741f0f25001e2722eff9edd74
https://github.com/Xilinx/mlir-aie/pull/2513/commits/5a4672de1f968d8ec7bd054a8c83fbf643c6d5e1

third_party/bootgen
fifield, line 1: Is the change to `bootgen` intended?

utils/build-llvm.sh
fifield, line 1: I think the changes to this file are unrelated.

include/aie/Dialect/AIE/Transforms/AIEPasses.h
andrej: Removed, thanks for catching this

include/aie/Dialect/AIEX/IR/AIEXDialect.h
andrej: I think I can do it but it would have to be a static method, since the `getDefaultRuntimeSequenceName()` gets called when the builder is instantiated but before the Op gets created.

lib/Dialect/AIE/Transforms/AIECanonicalizeDevice.cpp
andrej: Replaced with `DeviceOp::getDefaultDeviceName()`. But the other question is, do we still need this pass at all since we decided all valid MLIR should have a device region and all the tests that didn't are updated. If we want to keep the pass around, it might need to be refactored; if there's any device, the `runOnOperation()` will exit, even if there may be operations outside of a device. But then that raises the question, do we put the operations not in a device inside of another, new device, or inside one of the ones already there, ...? I vote to just remove this pass.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
andrej: Indeed, this function is only used as a helper within the same compilation unit. Changed to static

lib/Dialect/AIEX/Transforms/CMakeLists.txt
andrej: alphabetized

python/compiler/aiecc/cl_arguments.py
andrej: Currently for the control packet flow, the ELF file name seems to be hardcoded here:
https://github.com/Xilinx/mlir-aie/blob/69ffd7a153de7eab5b397216057ba04bb420dad4/python/compiler/aiecc/main.py#L731
The elf name option is only used together with transaction sequences:
https://github.com/Xilinx/mlir-aie/blob/69ffd7a153de7eab5b397216057ba04bb420dad4/python/compiler/aiecc/main.py#L754

Probably it makes little sense that the user would invoke the compiler with options to generate both the transaction sequence and the control packets, but if she did, I believe if we used the same command line option it would lead to one file overwriting the other?

Not sure though since there's a lot going on in aiecc.py.

python/compiler/aiecc/cl_arguments.py
andrej: Looking forward to #2407, seems much cleaner for control packet lowering

test/npu-xrt/add_one_two/aie.mlir
andrej: discussed that it's fine to leave this one merged in this PR

third_party/bootgen
andrej: no, thanks for catching

utils/build-llvm.sh
andrej: indeed, thanks for catching

python/iron/program.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         self._device = device
         self._rt = rt
 
-    def resolve_program(self, placer: Placer | None = None):
+    def resolve_program(self, placer: Placer | None = None, device_name="main"):
```
hunhoffe, line 46: @andrej would you consider a follow-up PR to update the documentation for what this argument means, and also consider a different name for it? The term "device' is already used in the forward facing API.

PR #2541, opened 2025-08-27
Add trace with JIT example

mawad-amd:
Initial version of trace integration with JIT. Here is the output:

In code, it looks like this:
https://github.com/Xilinx/mlir-aie/blob/612af4961539f1c02b9a14728e061e68e0589f45/programming_examples/basic/vector_scalar_mul/vector_scalar_mul_jit.py#L179-L190

And this:

https://github.com/Xilinx/mlir-aie/blob/612af4961539f1c02b9a14728e061e68e0589f45/programming_examples/basic/vector_scalar_mul/vector_scalar_mul_jit.py#L115-L126

I would have liked to hide this away too but it's not straight forward at the moment because of how the default workflow works (i.e., I can't simply call `enable_trace` for the runtime sequence if tracing is enabled in the backend). There are also other complications for tracing that Jack has brought up.

Output trace visualization: 
image


mawad-amd:
@jackl-xilinx @AndraBisca @jgmelber This is an initial draft for trace support in JIT.

hunhoffe:
Closing in favor of https://github.com/Xilinx/mlir-aie/pull/3025

PR #2544, opened 2025-08-28
Core file path set to tmpdirname

ypapadop-amd:
This PR fixes the path where core files should go. Before that, `tmpdir` was ignored for those.


ypapadop-amd:
> This might break code using the libxaiev2 backend (`aie-translate --aie-generate-xaie`) because it generates C++ to load the elf. @andrej is that still an issue after #2532 ?

Happy to try an example. Do we have one?

andrej:
I think #2532 should probably fix this. I have tried to remove the reliance on any sort of path structure for where the core files are located, instead using the "elf_file" attribute in MLIR. #2532 passes tests, so if there is a test that tests this aspect of the XAIEV2 flow, then we should be robust to using any path.

ypapadop-amd:
> I think #2532 should probably fix this. I have tried to remove the reliance on any sort of path structure for where the core files are located, instead using the "elf_file" attribute in MLIR. #2532 passes tests, so if there is a test that tests this aspect of the XAIEV2 flow, then we should be robust to using any path.

Are core files still generated? Since I call `compile_mlir_module()` directly in my own build scripts, the core files are generated to the current directory, which is not expected.

ypapadop-amd:
@andrej @fifield would this be good to go?

andrej:
Actually I see you already did merge main. I'd have a look at the generated MLIR and what the value of the elf file path is, and if those files actually exist.

mawad-amd:
Is this PR also needed to make sure concurrent builds are correct?

andrej:
No, this is unrelated to that.

andrej:
Yes, and each test runs in its own working directory. The other tests were failing because they put things in the home directory, the same home directory for all tests. That's at least my understanding. I think this was just a cleanup effort from Yiannis to make it possible to put these files inside the tmpdir created by aiecc.py.

PR #2545, opened 2025-08-28
Remove use of Vitis aie-rt

fifield:
This removes the dependence on aie-rt (`libxaiengine.so`) from Vitis for simulation targets. Instead the aie-rt submodule in `third_party/aie-rt` is used. This the same as the `libxaienginecdo` library that a few mlir-aie components statically link against compiled with the additional simulation backend. It also refactors `aie_libxaie_ctx_t` struct used by libxaie target to avoid some strange misalignment/corruption issues I was seeing between compilers. Tested with Vitis 2025.1 and it passes aie1 and aie2 simulation.


fifield:
> currently failing with
> 
> ```
> /home/runner/work/mlir-aie/mlir-aie/third_party/aie-rt/driver/src/io_backend/ext/xaie_sim.c:40:10: fatal error: main_rts.h: No such file or directory
>    40 | #include "main_rts.h"
>       |          ^~~~~~~~~~~~
> ```
> 
> due to missing aie-rt patch which is needed to avoid using Vitis xioutils [stephenneuendorffer/aie-rt#14](https://github.com/stephenneuendorffer/aie-rt/pull/14)

resolved.

Issue #2441, opened 2025-07-23
Investigate if `--dynamic-ObjFifos` would be a reasonable default


andrej:
Several issues [have](https://github.com/Xilinx/mlir-aie/issues/2426) [popped](https://github.com/Xilinx/mlir-aie/issues/2411) [up](https://github.com/Xilinx/mlir-aie/issues/1773) over time caused by a program memory overflow due to the ObjectFIFO loop unrolling. This is solved by enabling the `--dynamic-ObjFifos` flag, which disables the unrolling.

I think we should:

 - [ ] Investigate if there is a performance degradation when using `--dynamic-ObjFifos`
 - [ ] If not, make `--dynamic-ObjFifos` the default
 - [ ] Perhaps remove the unrolling code from the ObjectFIFO altogether and see if an built-in LLVM/MLIR pass can do the same when needed for performance

PR #2549, opened 2025-09-02
Enable dynamic objectfifo lowering by default

AndraBisca:
Addresses some points from #2441


andrej:
Andra, just to let you know I have one of Endri's GEMM designs with the dynamic object FIFO and did not measure any performance degradation vs. the unrolled ObjectFIFO.

hunhoffe (approved):

PR #2558, opened 2025-09-04
Pin action-suggester version to 1.22 until a bug is fixed on reviewdog

erwei-xilinx:
Info: https://github.com/reviewdog/action-suggester/issues/95


fifield (approved):
Thanks for the quick fix

PR #2560, opened 2025-09-05
Bump llvm to 064f02da

erwei-xilinx:


erwei-xilinx:
> Hi @andrej, one of the test for bd-chains-and-dma-tasks was disabled after llvm bump, due to changes in the canonicalizer behavior that the last block is now canonicalized away.

So I think this error message shall never get triggered, as any disconnected block shall get canonicalized away, before triggering this error message.

andrej:
That makes sense. Thanks for the update!

PR #2571, opened 2025-09-10
Rename mlir_tutorials to mlir_exercises

AndraBisca:
Proposal to rename this folder as the name 'tutorials' catches the eye of new users that are often looking for the IRON programming guide.


fifield (commented):
Good idea to rename these to avoid confusion. But the examples currently default to running on a development board, not simulation, so the name is still misleading. If the intent is to demonstrate AIE1 simulation then ideally they would get changed so that the default is simulation (which would also let the examples get tested in CI).

PR #2584, opened 2025-09-17
Add `parse_trace.py` to mlir-aie wheel

erwei-xilinx:
... so that trace can be parsed using the wheel only, without requiring a clone of mlir-aie.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -41,6 +41,7 @@ declare_mlir_python_sources(AIEPythonSources.Utils
     utils/trace.py
     utils/trace_events_enum.py
     utils/config.py
+    ../programming_examples/utils/parse_trace.py
 )
 
 declare_mlir_python_sources(AIEPythonSources.Helpers


erwei-xilinx:
Cc: @jackl-xilinx

programming_examples/CMakeLists.txt
```diff
@@ -138,3 +138,10 @@ add_lit_testsuite(check-reference-designs "Running the aie reference designs"
   ARGS "-sv --timeout 600"
 )
 set_target_properties(check-reference-designs PROPERTIES FOLDER "Tutorials")
+
+# Add programming_examples/utils to installed project.
+declare_mlir_python_sources(AIEPythonSources.ProgrammingExampleUtils
+  ADD_TO_PARENT AIEPythonSources
+  SOURCES
+    utils/parse_trace.py
+)
```
fifield: Should it go in `python/utils` if it's generally useful, instead of living in `programming_examples`?

erwei-xilinx: Yeah that makes sense to me. Happy to work on it. Copying more colleagues into the discussion in case if it might impact their work: @AndraBisca, @jackl-xilinx, @mawad-amd, @jgmelber.

ypapadop-amd:
It seems it's a command-line tool, right? If you want it to be available in the path, you need to do something similar to https://github.com/Xilinx/mlir-aie/blob/59a4a00de97ad4faccf0224a76e74b70bc459db8/utils/mlir_aie_wheels/pyproject.toml#L68

programming_examples/CMakeLists.txt
erwei-xilinx: Feedbacks, comments and reviews are greatly appreciated. I haven't received any objections on this change so far, so I think I will land it some time tomorrow, if I do not receive any objections by then.

andrej: I think this is a good change, please go ahead 👍

PR #2589, opened 2025-09-18
[aievec] Add `arith.addf` to llvmir lowering

erwei-xilinx:
- Add `AddAccFloatIntrOp` to xllvm dialect, which lowers to `aie2.add.accfloat` intrinsic.
- Add `arith.addf` to llvmir lowering pattern to `-vector-to-aievec` pipeline.
~~- TODO: Add mlir ir test for lowering to xllvm~~

### DIFF
--- a/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
+++ b/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
@@ -242,7 +242,7 @@ public:
       // Float types
       if (lhsBitWidth == 32) {
         // FP32 add_elem
-        return {DecodedAddElemOp::Kind::FP32_FP32_FP32_16x1x1x1, -1};
+        return {DecodedAddElemOp::Kind::FP32_FP32_FP32_16x1x1x1, /*conf*/ 0};
       }
     }
     return {DecodedAddElemOp::Kind::UNSUPPORTED, -1};
--- a/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
+++ b/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
@@ -205,6 +205,19 @@ void encodeConf(uint32_t conf[2], const BufferParams &x, const BufferParams &z,
   conf[1] |= sub << 17;
 }
 
+class AddOpConversion
+    : public mlir::ConvertOpToLLVMPattern<aievec::aie1::AddOp> {
+public:
+  using ConvertOpToLLVMPattern<aievec::aie1::AddOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(aievec::aie1::AddOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    op.emitWarning() << "aie.add conversion is not implemented\n";
+    return failure();
+  }
+};
+
 class AddElemOpConversion
     : public mlir::ConvertOpToLLVMPattern<aievec::AddElemOp> {
 public:
@@ -2381,7 +2394,8 @@ void populateAIEVecToLLVMConversionPatterns(
     mlir::LLVMTypeConverter &converter, mlir::RewritePatternSet &patterns,
     Aie2Fp32Emulation aie2Fp32EmulationOption) {
   // clang-format off
-  patterns.add<AddElemOpConversion,
+  patterns.add<AddOpConversion,
+               AddElemOpConversion,
                SubOpConversion,
                FMAOpConversion,
                MulOpConversion,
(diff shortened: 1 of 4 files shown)


jsetoain:
@erwei-xilinx Why is this lowering necessary? As far as I remember, Peano could handle `fadd` just fine. The natural lowering from `arith.addf` to LLVM IR `fadd` should just work.

I've also noticed that you're removing AIE1 patterns and replacing them with AIE2 patterns, what's the rationale behind this?

erwei-xilinx:
> @erwei-xilinx Why is this lowering necessary? As far as I remember, Peano could handle `fadd` just fine. The natural lowering from `arith.addf` to LLVM IR `fadd` should just work.
> 
> I've also noticed that you're removing AIE1 patterns and replacing them with AIE2 patterns, what's the rationale behind this?

@jsetoain, I wasn't able to get `arith.addf` to work with peano, as I am getting this error `LLVM ERROR: unable to legalize instruction: %29:_(<16 x s16>) = G_FADD %26:_, %28:_ (in function: core_0_2)`. But I think it is really useful to know that they plan to support fadd as part of their scope. I can create another issue in the llvm-aie about this, if this is what they plan to support.

Until then, I'm afraid that I have no choice but to stick with what I am doing now, to unblock the higher-level abstraction lowerings that I am working on.

It wasn't my intention to replace the `aie1.add` error emission. Let me bring it back. I do not plan to break any aie1 backward compatibility.

jsetoain:
> > @erwei-xilinx Why is this lowering necessary? As far as I remember, Peano could handle `fadd` just fine. The natural lowering from `arith.addf` to LLVM IR `fadd` should just work.
> > I've also noticed that you're removing AIE1 patterns and replacing them with AIE2 patterns, what's the rationale behind this?
> 
> @jsetoain, I wasn't able to get `arith.addf` to work with peano, as I am getting this error `LLVM ERROR: unable to legalize instruction: %29:_(<16 x s16>) = G_FADD %26:_, %28:_ (in function: core_0_2)`. But I think it is really useful to know that they plan to support fadd as part of their scope. I can create another issue in the llvm-aie about this, if this is what they plan to support.
> 
> Until then, I'm afraid that I have no choice but to stick with what I am doing now, to unblock the higher-level abstraction lowerings that I am working on.
> 
> It wasn't my intention to replace the `aie1.add` error emission. Let me bring it back. I do not plan to break any aie1 backward compatibility.

I'm not too concerned with breaking aie1, support is partial anyway, and there are no plans to complete it. My main concern was related to where the new intrinsic would live.

I'm surprised it's failing to legalize `fadd`, I thought we were actively using it somewhere else, but maybe I'm mixing it with `add`. In any case, I have no issue with adding another legalization, just making sure the work was necessary 🙂

Thanks! 😁

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -218,6 +218,75 @@ class AddOpConversion
...
+  LogicalResult
+  matchAndRewrite(aievec::AddElemOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    auto decodedAddElemOp = decodeAddElemOp(adaptor);
+
+    if (decodedAddElemOp.kind == DecodedAddElemOp::Kind::UNSUPPORTED) {
+      op.emitWarning() << "aievec.add_elem conversion is not supported.\n";
+      return failure();
+    }
```
jsetoain, line 260: This should probably be an error, not a warning. In any case, error/warning messages have to be tested as well.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -218,6 +218,75 @@ class AddOpConversion
...
+  static DecodedAddElemOp decodeAddElemOp(OpAdaptor op) {
+    auto lhs = op.getLhs();
+    auto lhsVecTy = cast<VectorType>(lhs.getType());
+    auto lhsScaTy = lhsVecTy.getElementType();
+    unsigned lhsBitWidth = lhsScaTy.getIntOrFloatBitWidth();
+
+    // Integer types
+    if (llvm::isa<IntegerType>(lhsScaTy)) {
+      return {DecodedAddElemOp::Kind::UNSUPPORTED, -1};
+    } else {
```
jsetoain, line 241: Is this needed? AddElemOp is never built for integer types when the target backend is LLVM IR.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -218,6 +218,75 @@ class AddOpConversion
...
+          loc, VectorType::get({8}, rewriter.getI64Type()),
+          forceCastOperandsToSignature(
+              rewriter, loc, operands,
+              {VectorType::get({8}, rewriter.getI64Type()),
+               VectorType::get({8}, rewriter.getI64Type()),
+               rewriter.getI32Type()}));
+    } else {
+      op.emitWarning() << "aievec.add_elem conversion is not supported.\n";
+      return failure();
+    }
```
jsetoain, line 281: Wouldn't this be redundant with the earlier check?

Issue #2562, opened 2025-09-05
Implement a MNIST with an MLP Guide Example


mawad-amd:
Implement a simple "getting started" example that runs MNIST inference using the JIT graph prototype with the runlist backend. The example should load a pre-trained MLP model (flatten → dense → dense → dense), compile it with the JIT graph targeting runlist, run predictions on sample images, and print the top predicted class.

```py
model.add(Flatten(input_shape=(28,28)))
model.add(Dense(128, relu))
model.add(Dense(32, relu))
model.add(Dense(10, softmax))
```

jgmelber:
#2611 related towards performance

hunhoffe:
@mawad-amd Out of curiosity, what is the status of this example?

hunhoffe:
A lot has changed since this was first proposed, but I think it would make a good first issue to have an MLP implementation for any contributors looking for work. The solution may look different than initially proposed.

PR #2596, opened 2025-09-19
Enable aie_kernels includes + cleanup cmake scripts

AndraBisca:
The goal is to avoid copying pragmas into the C kernels, as well as avoid include paths like this:
`#include "../aie_kernel_utils.h"`

This PR introduces an `install_headers.cmake` script that installs given header files to the `install/include` path and copies them from source to the `build/include` path. This script was applied to both aie_kernels and aie_api header files.


aie_kernels/aiekernels.cmake
```diff
@@ -0,0 +1,12 @@
+function(add_aie_kernels_headers SRCPATH BUILDPATH INSTALLPATH)
+  message("Installing aie_kernels includes from ${SRCPATH} in ${BUILDPATH}")
+
+  # copy header files into build area
+  install(DIRECTORY ${SRCPATH}/ DESTINATION ${BUILDPATH}/aie_kernels)
```
fifield: This generates an installation rule, but that's not quite what we want here. The goal is to have the headers in the build area (e.g. for testing) _without_ installing. I see that the aieapi.cmake follows the [same pattern](https://github.com/Xilinx/mlir-aie/blob/main/runtime_lib/aie_api/aieapi.cmake#L6) @jgmelber

ypapadop-amd: You can copy files with something like the following
```cmake
        add_custom_command(
            OUTPUT ${DESTINATION_FILE}
            COMMAND ${CMAKE_COMMAND} -E copy_if_different ${FILE} ${DESTINATION_FILE}
            DEPENDS ${FILE}
            COMMENT "Copying ${FILE} to build directory ${arg_DESTINATION}"
            )
```

AndraBisca: Applied this to both aie_kernels and aie_api. Interestingly lit testing passes, however, I don't see the files in the build/include folder.

ypapadop-amd: It doesn't work for directories, you need to do file-by-file or use file globs.

AndraBisca: So there are apparently commands that support copying directories, however, when I try to copy the files from the install path to the build path I get a error because the files are not yet found in the install path. I got around this by copying from the source directories instead, but is there a better way to do this?

ypapadop-amd: `file(COPY` happens during configuration, so unless the configuration changes, it won't re-copy the file, i.e., when the file changes. It also does not have a target, so you can't add a dependency on it.

fifield: There are some examples of the file-by-file approach suggested by @ypapadop-amd in the repository:
https://github.com/Xilinx/mlir-aie/blob/066da63c1febe0177250a7b7c69464d87a605344/runtime_lib/test_lib/CMakeLists.txt#L59-L68

AndraBisca: Thank you for the explanations and example! I implemented the same structure for aie_kernels and aie_api, with two main differences:
- I used file glob_recurse to retrieve the header files to copy recursively,
- I used part of the path of each header file to create the target name as there were many duplicate targets, in aie_api especially.

One last point related to the example given by @fifield: here we also copy the header files from source directory ${CMAKE_CURRENT_SOURCE_DIR} to the build area, and not from the install area. My understanding is that the latter would be preferable.

fifield: > One last point related to the example given by @fifield: here we also copy the header files from source directory ${CMAKE_CURRENT_SOURCE_DIR} to the build area, and not from the install area. My understanding is that the latter would be preferable.

Source to build is what we want. The build target cannot depend on the install target.

mawad-amd:
@AndraBisca This PR will be very useful for #2562 and #2607. Would love to see it merged soon!

aie_kernels/aiekernels.cmake
AndraBisca: Ah I must've misunderstood when we initially talked about it. Thank you!

aie_kernels/aiekernels.cmake
```diff
@@ -0,0 +1,27 @@
+function(add_aie_kernels_headers SRCPATH BUILDPATH INSTALLPATH)
```
ypapadop-amd: I think this needs copyright.

runtime_lib/aie_api/aieapi.cmake
```diff
@@ -1,13 +1,28 @@
 
 function(add_aie_api_headers SRCPATH BUILDPATH INSTALLPATH)
```
ypapadop-amd: Copyright here too.

aie_kernels/aiekernels.cmake
ypapadop-amd: I think this needs copyright.

CMakeLists.txt
```diff
@@ -285,6 +285,12 @@ if(NOT LLVM_INSTALL_TOOLCHAIN_ONLY)
     PATTERN "CMakeFiles" EXCLUDE
     PATTERN "config.h" EXCLUDE)
 
+  include("aie_kernels/aiekernels.cmake")
```
ypapadop-amd: It's better to have all CMake modules in one directory instead with the source code, e.g., `mlir-aie/cmake`

ypapadop-amd (changes requested):
You can probably combine the two CMake functions into something that copies/install header files, they seem to be the same.

runtime_lib/CMakeLists.txt
```diff
@@ -141,10 +141,10 @@ install(TARGETS xaienginecdo_static DESTINATION lib EXPORT xaienginecdo_static)
 install(EXPORT xaienginecdo_static DESTINATION lib/cmake/aie)
 target_include_directories(xaienginecdo_static SYSTEM PRIVATE ${BOOTGEN_SOURCE_DIR}/cdo-driver)
 
-include("aie_api/aieapi.cmake")
-set(AIEAPI_SOURCE_DIR "../third_party/aie_api/include/aie_api")
-add_aie_api_headers(${AIEAPI_SOURCE_DIR}
+include("${AIE_SOURCE_DIR}/cmake/install_headers.cmake")
```
ypapadop-amd, line 144: You can do relative include here (i.e., ../cmake/) but it's ok like this too.

ypapadop-amd (approved):
LGTM

fifield (approved):

PR #2599, opened 2025-09-20
Fix LLVM IR compatibility issue in `aiecc.py` by updating `downgrade_ir_for_peano`

Fangtangtang:
When compiling with `aiecc.py`, I encountered errors such as:

```
error: expected ')' at end of argument list
declare void @llvm.memcpy.p0.p0.i64(ptr noalias writeonly captures(none), ptr noalias readonly captures(none), i64, i1 immarg) #1
                                                          ^
```

This was caused by the presence of `captures(none)` in the IR.
To fix this, I updated `compiler/aiecc/main.py` so that `downgrade_ir_for_peano` now replaces `captures(none)`:

```python
def downgrade_ir_for_peano(llvmir):
    llvmir = llvmir.replace("getelementptr inbounds nuw", "getelementptr inbounds")
    llvmir = llvmir.replace("captures(none)", "nocapture")
    return llvmir
```

This resolves the compilation error and allows the IR to be correctly lowered.

Note: I noticed a similar issue was mentioned earlier in PR #2038


jgmelber:
Is this fix worth including for users who would like to use a particular older version of LLVM-AIE? @fifield

PR #2604, opened 2025-09-23
Batch control packets by tile for aie2p

fifield:
This PR adds combining of control packet data copies for aie2p.

Instead of sending and then syncing each 1-4 word control packet as individual shim dma transfers, batch them so that consecutive control packets to the same (row, col) are combined into single shim dma transfers.

For the test `npu-xrt/ctrl_packet_reconfig_4x1_cores` this reduces the number of overall `aiex.npu.dma_memcpy_nd` ops in the control packet runtime sequence from 567 to 36.

Somewhat surprisingly, the `Ctrl_Pkt_Tlast_Error_Enable` configuration bit in the tiles didn't need to be cleared for this to work.

~~TODO: this feature probably needs a standalone unit test.~~

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIECtrlPacketToDma.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIECtrlPacketToDma.cpp
@@ -176,6 +176,8 @@ struct AIECtrlPacketToDmaPass : AIECtrlPacketToDmaBase<AIECtrlPacketToDmaPass> {
         builder.create<AIEX::NpuSyncOp>(loc, shimCol, shimRow, dir, chan,
                                         col_num, row_num);
         ++batchIt;
+        if (batchIt == batches.end())
+          break;
       }
 
       erased.push_back(f);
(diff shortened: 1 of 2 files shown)


lib/Dialect/AIEX/Transforms/AIECtrlPacketToDma.cpp
```diff
@@ -82,56 +82,102 @@ struct AIECtrlPacketToDmaPass : AIECtrlPacketToDmaBase<AIECtrlPacketToDmaPass> {
...
+        auto shimRow = builder.getI32IntegerAttr(0);
+        auto shimCol = builder.getI32IntegerAttr(col);
+        auto dir = builder.getI32IntegerAttr(1); // MM2S
+        auto chan = builder.getI32IntegerAttr(batchIt->shimChan);
+        auto col_num = builder.getI32IntegerAttr(1);
+        auto row_num = builder.getI32IntegerAttr(1);
+        builder.create<AIEX::NpuSyncOp>(loc, shimCol, shimRow, dir, chan,
+                                        col_num, row_num);
+        ++batchIt;
+        if (batchIt == batches.end())
```
fifield, line 179: ugh, this is a bit of slop I ignored. could just iterate over batches instead...

PR #2612, opened 2025-09-25
Revert accidental edit on main.py

jackl-xilinx:


andrej (approved):
Just noticed this and here's already the reversion, thanks!

fifield:
Sadly nobody reviewed the original PR which might have caught the problem :(

andrej:
I confirmed that this is a clean reversion:

```
git diff 2c78875241 origin/revert-aieccpy-change python/compiler/aiecc/main.py
```

prints nothing (2c78875241 is the last commit to main.py before the accidental change).

Since I'll need this and it looks clean I hit merge.

PR #2617, opened 2025-09-29
Disable lowering `arith.subf` on 16-lane `vector`s to `aievec.sub_elem`

erwei-xilinx:
… as peano can lower `arith.subf` on vector of that shape, for aie2.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
@@ -3660,6 +3660,8 @@ static void configureAIEVecV2Legalizations(ConversionTarget &target,
     return laneSize != 16;
   });
 
+  target.addLegalOp<arith::SubFOp>();
+
   target.addDynamicallyLegalOp<arith::MulIOp>([](arith::MulIOp op) {
     auto resultType = dyn_cast<VectorType>(op.getType());
     if (!resultType)


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -3626,7 +3626,7 @@ static void configureAIEVecV2Legalizations(ConversionTarget &target,
       return true;
 
     unsigned laneSize = getVectorLaneSize(resultType);
-    return laneSize != 16;
+    return laneSize == 16;
```
fifield: This makes 16-lane vectors legal, but doesn't it also flip non-16-lane vectors from legal to illegal? Does that have any implications? Why doesn't the same apply to similar operations like AddFOp? Sorry if these are basic questions, I'm not familiar with this code.

erwei-xilinx: So peano supports lowering arith ops on vectors to AIE vector intrinsics. I have experimented with their coverage, and noticed that for `arith.subf`, `vector<16xbf16>` can successfully compile, but not other vector shapes. So this legality check is fundamentally us checking for what peano can support. But it probably makes more sense that we remove this legality check, and leave it to peano to return its error message (cc: @jsetoain).

Re `arith.addf`, I have tried the same experiments with peano, but haven't been able to get it to work, which is why for `arith.addf`, we still lower to `aievec.add_elem` for now, to unblock us in mlir-air in the short term.

jsetoain (approved):
lgtm

PR #2620, opened 2025-09-30
Make sure CI tests do not use previous run's JIT cache

andrej:
The `IRON_CACHE_HOME` environment variable is unique for each CI run and is supposed to ensure that the JIT-compiled binaries don't get reused/cached between CI runs. However, `lit`, does not by default copy all environment variables from where it's called, meaning the `IRON_CACHE_HOME` environment variable gets dropped and cached files get generated in the `~/.iron/cache` directory instead. This should fix that, making sure the JIT targets get regenerated/tested on each CI run. 

This is likely what caused the latest test failures in #2532.

cc @mawad-amd


fifield (approved):
I'm tempted to `rm -rf ~/.iron/* && chmod a-w ~/.iron` on the runners after this lands.

mawad-amd (approved):
This looks good. Thanks Andre.

mawad-amd:
Merging to test the rest of the fix in https://github.com/Xilinx/mlir-aie/pull/2622

PR #2622, opened 2025-09-30
Add lock over cache dir

### LINKED ISSUE #2621: JIT cache is not atomic
During lit testing or other parallel execution it can happen that two kernels with the same cache key race to fill the cache. When this happens the errors are strange and confusing because the files are corrupted during compilation. For example:
```
[ERROR]  : BIF file is empty - design.bif 

Error encountered while running: bootgen -arch versal -image /home/jfifield/.iron/cache/f5628ea683592ee1/design.bif -o /home/jfifield/.iron/cache/f5628ea683592ee1/design.pdi -w
```
```
ld.lld: error: /work/ryzen-ai-venv/lib/python3.10/site-packages/llvm-aie/bin/../lib/aie2-none-unknown-elf/libc.a(memcpy.cpp.o):(.comment) is being placed in '.comment'
ld.lld: error: /work/ryzen-ai-venv/lib/python3.10/site-packages/llvm-aie/bin/../lib/aie2-none-unknown-elf/crt1.o:(.comment) is being placed in '.comment'
ld.lld: error: <internal>:(.comment) is being placed in '.comment'
ld.lld: error: <internal>:(.symtab) is being placed in '.symtab'
ld.lld: error: <internal>:(.shstrtab) is being placed in '.shstrtab'
ld.lld: error: <internal>:(.strtab) is being placed in '.strtab'
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)
```
I can reliably reproduce the problem like this:
```
while true; do rm -rf ~/.iron; lit programming_guide/mini_tutorial; done
```

The JIT should hold a `flock` or use a similar synchronization mechanism while filling the cache, or the cache should be per thread or per process.

mawad-amd:
Closes #2621

### DIFF
--- a/python/iron/jit.py
+++ b/python/iron/jit.py
@@ -12,8 +12,9 @@ import hashlib
 import numpy as np
 import pyxrt as xrt
 import shutil
-import sys
-import traceback
+import fcntl
+import contextlib
+import time
 
 from aie.extras.context import mlir_mod_ctx
 from ..utils.xrt import read_insts_binary
@@ -27,6 +28,44 @@ from aie.dialects.aie import AIEDevice
 # Kernels are cached based on their hash value of the MLIR module string. If during compilation,
 # we hit in the cache, the `iron.jit` will load the xclbin and instruction binary files from the cache.
 IRON_CACHE_HOME = os.environ.get("IRON_CACHE_HOME", os.path.expanduser("~/.iron/cache"))
+
+
+@contextlib.contextmanager
+def file_lock(lock_file_path, timeout_seconds=60):
+    """
+    Context manager for file locking using flock to prevent race conditions.
+
+    Args:
+        lock_file_path (str): Path to the lock file
+        timeout_seconds (int): Maximum time to wait for lock acquisition in seconds
+    """
+    lock_file = None
+    try:
+        # Create lock file if it doesn't exist
+        os.makedirs(os.path.dirname(lock_file_path), exist_ok=True)
+        lock_file = open(lock_file_path, 'w')
+
+        # Try to acquire exclusive lock with timeout
+        start_time = time.time()
+        while True:
+            try:
+                fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
+                break
+            except OSError:
+                # Lock is held by another process
+                if time.time() - start_time > timeout_seconds:
+                    raise TimeoutError(f"Could not acquire lock on {lock_file_path} within {timeout_seconds} seconds")
+                time.sleep(0.1)
+
+        yield lock_file
+
+    finally:
+        if lock_file is not None:
+            try:
+                fcntl.flock(lock_file.fileno(), fcntl.LOCK_UN)
+            except OSError:
+                pass  # Ignore errors when releasing lock
+            lock_file.close()
 
 
 class CircularCache:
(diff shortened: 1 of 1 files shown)


python/iron/jit.py
```diff
@@ -29,6 +30,61 @@
...
+
+    Args:
+        lock_file_path (str): Path to the lock file
+        timeout_seconds (int): Maximum time to wait for lock acquisition in seconds
+    """
+    lock_file = None
+    try:
+        # Create lock file if it doesn't exist
+        os.makedirs(os.path.dirname(lock_file_path), exist_ok=True)
+        lock_file = open(lock_file_path, "w")
```
fifield:
```suggestion
        try:
            f = os.open(lock_file_path, os.O_CREAT | os.O_EXCL)
            os.close(f)
        except FileExistsError:
            pass  # File already exists
        lock_file = open(lock_file_path, "a")
```
I'm suspicious the lock file creation is not atomic, and that is why errors remain. I'm testing a version of this locally and haven't seen any errors...so far.

mawad-amd: Yeah I tested locally and can't reproduce it yet. Was just about to push some debug printfs.

fifield: I saw another error twice running tests in a loop:
```
FAILED ../../../../aie/test/python/cache_functionality.py::test_cache_lambda_functions
```
*seems* less frequent, but hard to know for sure.

mawad-amd: So we did serialize compilation which I think rules out https://github.com/Xilinx/mlir-aie/pull/2544 and anything related to the compiler. I will test locally. I would like to resolve any weird concurrency issues.

mawad-amd: I think they maybe a driver/runtime issue:
```
               Failed: Only 3/5 processes succeeded
E               
E               Process details:
E               
E               Process 0: FAILED (return code: 1)
E                 STDOUT: ERROR: RuntimeError: DRM_IOCTL_AMDXDNA_CREATE_HWCTX IOCTL failed (err=-2): No such file or directory
E               
E               Process 1: SUCCESS (return code: 0)
E                 STDOUT: SUCCESS
E               
E               Process 2: SUCCESS (return code: 0)
E                 STDOUT: SUCCESS
E               
E               Process 3: SUCCESS (return code: 0)
E                 STDOUT: SUCCESS
E               
E               Process 4: FAILED (return code: 1)
E                 STDOUT: ERROR: RuntimeError: DRM_IOCTL_AMDXDNA_CREATE_HWCTX IOCTL failed (err=-2): No such file or directory

```

mawad-amd: I was running into a similar issue the other day. This actually explains the weird behavior I was seeing. There seems to be some limit on how many things we can run in parallel. In https://github.com/Xilinx/mlir-aie/pull/2611 I had to keep lowering down the number of cached kernels to 1 and didn’t have time to debug it back then. I now think we should be doing something similar to `make run -JMAX_CONCURRENT_HW_CONTEXTS` or something similar.

maawad: Not sure how to fix this in a robust way at the moment but I am pretty sure we are running out of hardware contexts. Eddie said the max is 6 hardware contexts on Phoenix. 
* Setting parallel jobs to a lower numbers should fix this but it doesn't make sense to lower all tests concurrency to get one test passing
* Some tests may have multiple kernels in flight which means multiple contexts in flight so even if we remove this test, other tests will not be always reliable. Making tests run one kernel at a time doesn't make sense either.
* Treating the `DRM_IOCTL_AMDXDNA_CREATE_HWCTX` as a non error is not a solution either

Let me know your thoughts.

mawad-amd: Not sure how to fix this in a robust way at the moment but I am pretty sure we are running out of hardware contexts. Eddie said the max is 6 hardware contexts on Phoenix. 
* Setting parallel jobs to a lower numbers should fix this but it doesn't make sense to lower all tests concurrency to get one test passing and will hide other concurrency bugs 
* Some tests may have multiple kernels in flight which means multiple contexts in flight so even if we remove this test, other tests will not be always reliable. Making tests run one kernel at a time doesn't make sense either.
* Treating the `DRM_IOCTL_AMDXDNA_CREATE_HWCTX` as a non error is not a solution either

Let me know your thoughts.

python/iron/jit.py
```diff
@@ -29,6 +30,61 @@
...
+                time.sleep(0.1)
+
+        yield lock_file
+
+    finally:
+        if lock_file is not None:
+            try:
+                fcntl.flock(lock_file.fileno(), fcntl.LOCK_UN)
+            except OSError:
+                pass  # Ignore errors when releasing lock
```
andrej, line 89: Why? I think we'd rather know if it fails and why.

From man(2) flock:
> Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate file descriptors, or when all such file descriptors have been closed.

Keeping in mind `lock_file` is presumably the only file descriptor we have in the current process of the lock file, the explicit unlock here is redundant with the `lock_file.close()` in the next line, which will also release the lock.

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -111,8 +111,11 @@ jobs:
 
           # -j here to reduce the number of parallel chess jobs.
           # -j4 for 32GB RAM, -j12 for 64GB RAM
+          # On Phoenix, we can support 6 concurrent contexts          
           if [ x"${{ matrix.runner_type }}" == x"amdhx370" ]; then
             LIT_OPTS="-j4 $LIT_OPTS"
+          elif [ x"${{ matrix.runner_type }}" == x"amd7940hs" ]; then
+            LIT_OPTS="-j6 $LIT_OPTS"
```
fifield: I'd rather not to do this. It adds an artificial bottleneck to something that's already a bottleneck. In practice we don't see concurrency failures with `-j12`. Testing in mlir-aie is setup to control concurrency at the top level and tests are supposed to be good citizens by using 1 thread or process to not blow things up.

I suggest pulling the concurrency tests into a new top level target with no inter-test concurrency, so that those tests can control it for themselves.

mawad-amd: Addressed here https://github.com/Xilinx/mlir-aie/pull/2622/commits/67dbfc3e8f153314e4ef0cf07751627344491220

fifield (approved):

jgmelber (approved):

PR #2626, opened 2025-10-02
Update bf16_softmax.mlir

fifield:
Test is expected to fail because legalization rules changed and would need further updates to re-enable C++ backend.


fifield:
https://github.com/Xilinx/mlir-aie/pull/2617#issuecomment-3361916959

PR #2630, opened 2025-10-03
BD chaining on Memtile using objectFifo

pvasireddy-amd:


python/iron/dataflow/objectfifo.py
```diff
@@ -119,6 +120,26 @@ def obj_type(self) -> type[np.ndarray]:
         """The tensor type of each buffer belonging to the ObjectFifo"""
         return self._obj_type
 
+    def use_bd_chain(self, iter_count: int = None):
```
andrej: Could we name this `set_iteration_count` instead? BD Chains are used in both cases.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1722,7 +1722,8 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [HasParent<"DeviceOp">, Symbol]
         // repeat_count==1 means "do it once"
         OptionalAttr<ConfinedAttr<AIEI32Attr, [IntMinValue<1>]>>:$repeat_count,
         InitValuesArrayAttr:$initValues,
-        OptionalAttr<BDPadLayoutArrayAttr>:$padDimensions
+        OptionalAttr<BDPadLayoutArrayAttr>:$padDimensions,
+        OptionalAttr<AIEI32Attr>:$bd_chain_iter_count
```
andrej: Could we just name this `iter_count` instead? As I understand, BD chains are used in both cases.

test/npu-xrt/bd_chain_repeat_on_memtile/test.cpp
```diff
@@ -0,0 +1,90 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+#include "../../../runtime_lib/test_lib/xrt_test_wrapper.h"
```
andrej:
```suggestion
#include "xrt_test_wrapper.h"
```

That's what the `%test_utils_flags` in this are for:
```
# RUN: clang %S/test.cpp -o test.exe -std=c++17 -Wall %xrt_flags -lrt -lstdc++ %test_utils_flags
```

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1726,7 +1726,9 @@ def AIE_ObjectFifoCreateOp: AIE_Op<"objectfifo", [HasParent<"DeviceOp">, Symbol]
         // repeat_count==1 means "do it once"
         OptionalAttr<ConfinedAttr<AIEI32Attr, [IntMinValue<1>]>>:$repeat_count,
         InitValuesArrayAttr:$initValues,
-        OptionalAttr<BDPadLayoutArrayAttr>:$padDimensions
+        OptionalAttr<BDPadLayoutArrayAttr>:$padDimensions,
+        // iter_count*depth number of objects will be transferred
```
AndraBisca: I think this comment could be a little more detailed. For example, a few words to describe iter_count as the number of iterations of the bd_chain generated at a lower-level. And then iter_count * elemNumber = number of objects that will be transferred (assuming repeat_count == 1). It might be good to improve the comment for repeat_count as well, something like: number of times each individual object is repeated.

test/objectFifo-stateful-transform/bd_chain_on_memtile/chain_with_split_and_join.mlir
```diff
@@ -0,0 +1,186 @@
+//===- bd_chain_on_memtile/ichain_with_split_and_join.mlir ----------------*- MLIR -*-===//
```
AndraBisca: For maintainability and readability, I think it would be good to also add two other tests for each pattern separately, one for distribute and one for join.

It would also be good to add a simple MLIR test with one objectfifo that uses both iter_count and repeat_count.

test/objectFifo-stateful-transform/bd_chain_on_memtile/split_and_join_with_iterations.mlir
```diff
@@ -0,0 +1,186 @@
...
+    %shim_noc_tile_0_0 = aie.tile(0, 0)
+    %mem_tile_0_1 = aie.tile(0, 1)
+    aie.objectfifo @in(%shim_noc_tile_0_0, {%mem_tile_0_1}, 2 : i32) : !aie.objectfifo<memref<1024xui8>> 
+    aie.objectfifo @split_0(%mem_tile_0_1, {%tile_0_2}, 2 : i32) {iter_count = 8 : i32} : !aie.objectfifo<memref<512xui8>> 
+    aie.objectfifo @split_1(%mem_tile_0_1, {%tile_0_3}, 2 : i32) {iter_count = 8 : i32} : !aie.objectfifo<memref<512xui8>> 
+    aie.objectfifo.link [@in] -> [@split_0, @split_1]([] [0, 512])
+    aie.objectfifo @join_0(%tile_0_2, {%mem_tile_0_1}, 2 : i32) {iter_count = 8 : i32} : !aie.objectfifo<memref<512xui8>> 
+    aie.objectfifo @join_1(%tile_0_3, {%mem_tile_0_1}, 2 : i32) {iter_count = 8 : i32} : !aie.objectfifo<memref<512xui8>> 
+    aie.objectfifo @out(%mem_tile_0_1, {%shim_noc_tile_0_0}, 2 : i32) : !aie.objectfifo<memref<1024xui8>> 
+    aie.objectfifo.link [@join_0, @join_1] -> [@out]([0, 512] [])
```
AndraBisca, line 184: What would happen if the iter_count for each individual objectfifo is different?

test/objectFifo-stateful-transform/bd_chain_on_memtile/infinite_loop.mlir
```diff
@@ -0,0 +1,59 @@
+//===- bd_chain_on_memtile/infinite_loop.mlir ----------------*- MLIR -*-===//
```
AndraBisca: I think I may be misunderstanding but just to confirm, isn't the infinite loop the standard behavior? If that is the case I don't think you need to add a test for that since it is captured in the base tests of the objectfifo.

pvasireddy-amd: Deleted

include/aie/Dialect/AIE/IR/AIEOps.td
pvasireddy-amd: I added a bit more detailed comment to the description part for both `repeat_count` and `iter_count`.

test/objectFifo-stateful-transform/bd_chain_on_memtile/split_and_join_with_iterations.mlir
pvasireddy-amd: Added this to the `split_with_iterations.mlir` and `join_with_iterations.mlir` tests.

test/objectFifo-stateful-transform/bd_chain_on_memtile/chain_with_split_and_join.mlir
pvasireddy-amd: Added separate tests for split and join and also a MLIR test that uses both `iter_count` and `repeat_count` features of `objectFifo`.

python/iron/dataflow/objectfifo.py
```diff
@@ -119,6 +120,26 @@ def obj_type(self) -> type[np.ndarray]:
...
+        """Set iteration count for DMA BD (Buffer Descriptor) chaining on MemTile for the ObjectFifo.
+
+        Args:
+            iter_count (int): Number of forward chain iterations.
+                - Must be in range [1, 256]: Forward chain with specified number of iterations
+
+        Raises:
+            ValueError: If iter_count is not provided or is outside the valid range [1, 256]
+        """
+        if iter_count is None:
```
hunhoffe: I would suggest not giving the argument a default if you disallow that default. So the function signature would be:
```python
def set_iter_count(self, iter_count: int):
```

And then you can do the two checks as one conditional:
```python
if not iter_count or iter_count < 1 or iter_count > 256:
```
If it's None or 0 the first check will fail, but since we require ```iter_count``` to be at least 1, this is ok

python/iron/dataflow/objectfifo.py
```diff
@@ -254,16 +275,31 @@ def resolve(
...
+                self._op = object_fifo(
+                    self.name,
+                    self._prod_tile_op(),
+                    self._cons_tiles_ops(),
+                    self._get_depths(),
+                    np_ndarray_type_to_memref_type(self._obj_type),
+                    dimensionsToStream=self._dims_to_stream,
+                    dimensionsFromStreamPerConsumer=dims_from_stream_per_cons,
+                    plio=self._plio,
+                    iter_count=self._iter_count,
```
hunhoffe: Can you do this without the if/else? Just send in the ```iter_count``` value of None (or whatever the default value should be) and let the lower level validation figure it out. I don't like repeating so many lines of code for a small change.

test/python/objFifo_bd_chain_repeat.py
```diff
@@ -0,0 +1,82 @@
...
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 AMD Inc.
+
+# RUN: %python %s | FileCheck %s --check-prefix=NORMAL
+# RUN: not %python %s zero 2>&1 | FileCheck %s --check-prefix=ERROR_ZERO
+# RUN: not %python %s high 2>&1 | FileCheck %s --check-prefix=ERROR_HIGH
+# RUN: not %python %s args 2>&1 | FileCheck %s --check-prefix=ERROR_ARGS
+
```
hunhoffe: Small stylistic nit: It's sometime nice to add the expected output lines right above each individual test, so then it's easy to tell what belongs to what.

PR #2642, opened 2025-10-09
Remove include of xrt in test_utils.h, use compiled test_lib in programming_examples

fifield:
Use a forward declaration of `xrt::device` and `xrt::kernel` instead of `#include` so that test_utils.h can be `#included` in environments without xrt in the include path.

Change programming examples cmake to use compiled test_lib in programming_examples instead of test_utils.cpp

### DIFF
--- a/programming_guide/section-2/section-2f/01_single_double_buffer/test.cpp
+++ b/programming_guide/section-2/section-2f/01_single_double_buffer/test.cpp
@@ -16,6 +16,8 @@
 #include "cxxopts.hpp"
 #include "test_utils.h"
 #include "xrt/xrt_bo.h"
+#include "xrt/xrt_device.h"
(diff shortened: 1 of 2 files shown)


fifield:
Currently this is broken on programming examples because they include test_utils.cpp directly on the gcc command line instead of linking it as a library. So they need to be updated to get the -DTEST_UTILS_USE_XRT flag (to enable xrt) or to use the built libtest_utils instead.

fifield:
> Currently this is broken on programming examples because they include test_utils.cpp directly on the gcc command line instead of linking it as a library. So they need to be updated to get the -DTEST_UTILS_USE_XRT flag (to enable xrt) or to use the built libtest_utils instead.

I updated the examples to use the pre-compiled test_utils from the build or install.

PR #2649, opened 2025-10-13
Add control packet payload binary disassembly

fifield:
Add control packet disassembly for debug purposes. Like `txn2mlir.py` but for control packet payload binaries.


lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -731,6 +858,90 @@ xilinx::AIE::convertTransactionBinaryToMLIR(mlir::MLIRContext *ctx,
...
+
+  // create a new ModuleOp and set the insertion point
+  auto module = ModuleOp::create(loc);
+  OpBuilder builder(module.getBodyRegion());
+  builder.setInsertionPointToStart(module.getBody());
+
+  // Create aie.device - default to npu1_1col
+  // Note: Control packets don't have device info in the binary format
+  auto device = builder.create<DeviceOp>(loc, AIEDevice::npu1_1col,
+                                         StringAttr::get(builder.getContext()));
```
fifield: FIXME: get the arch from the command line. we could also figure out the max column, to make a reasonable guess at the target.

PR #2651, opened 2025-10-14
Avoid default device in example

ypapadop-amd:
I started using a Strix box and the default device being `npu` doesn't let me run the examples.

### DIFF
--- a/programming_examples/basic/vector_vector_add/vector_vector_add.py
+++ b/programming_examples/basic/vector_vector_add/vector_vector_add.py
@@ -98,7 +98,6 @@ def main():
         "-d",
         "--device",
         choices=["npu", "npu2"],
-        default="npu",
         help="Target device",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


andrej (approved):
LGTM. For my own understanding, this change works because we call [`detect_npu_device()`](https://github.com/Xilinx/mlir-aie/blob/6b68e7da5d076b6f1e26bcf7cc39c058de70ef8e/python/iron/config.py#L30`) upon [including](https://github.com/Xilinx/mlir-aie/blob/6b68e7da5d076b6f1e26bcf7cc39c058de70ef8e/python/iron/config.py#L87) iron to set the device to a sensible value.

Side note: Our `env_setup.sh` script sets an environment variable to distinguish NPU1 and NPU2. The `detect_npu_device()` opens a subprocess calling `xrt-smi` each time that `iron.config` is included. Is that a little heavy-handed, i.e. should we check for the environment variable first before calling `xrt-smi`? Or maybe only call `xrt-smi` lazily when `config["device"]` is first needed, rather than directly upon import?

PR #2656, opened 2025-10-20
Add support for reading AIE registers via ctrl packet configuration

jackl-xilinx:
Add python wrappers to enable generating and configuring control packets to read AIE registers such as core status and trace status. The primary goal is to provide designers feedback when trace overflow has occurred thereby invalidating trace results and necessitating a larger trace buffer or a segmented trace routing topology.

This is accomplished via the following additional python functions in the top level placed python code (`vector_scalar_mul_placed.py`) as well as additions in the host code (which is defined in `python/utils/xrt.py`). 

In the the top level placed python, we call:
`trace_utils.configure_packet_ctrl_flow([ComputeTile2], CtrlShimTile)`
to configure the packet flow for the control packets both from the host and back to the host.
 
Then at the end of runtime sequence, we call:
`trace_utils.config_ctrl_pkts_aie([ComputeTile2], CtrlShimTile, output_offset=trace_size, num_pkts=2)`
to configure the control packet configuration to receive the 2 control packet commands and route the response back to an xrt output buffer. 

The host code addition that generates the control packets is found in `python/utils/xrt.py` under the `setup_buffer_data` function where we write the control packet commands to an input xrt buffer. The results are then read in the `setup_and_run_aie` function as the core status and trace status registers are printed out , along with a checking of the trace overflow status.

In our example python host code wrappers, we continue to use xrt buffer 7 for output trace data but now we use xrt buffer 6 for input control packet data and share the main data output buffer (xrt buffer 5) for output control packet data which is offset from the main output data. 

This changes the default trace config to use the same packet ID for all trace routes but currently, there is a packet routing bug when this is applied to the conv2d 14x14 24 core design. 

*Update 10/22/25* - No connections valid from trace other than southbound and s2mm0. The error that was created however should not have occurred. Need to investigate further so correct error is flagged and that this connection is not allowed in the future. But my 24 core test case needs to be redesigned since freeing up a column doesn't help trace. Can try packet flow objfifos instead.

*Update 11/3/25* - Push out pathfinder bug fix into separate PR to be resolved separately. Current bug treats dma nodes for Djikstra's algorithm as a single node whereas it should be separated as s2mm and mm2s. Making it the same node allows pathfinder to route it to itself (trace -> s2mm -> mm2s -> east). This should not be allowed.

### DIFF
--- a/python/utils/trace.py
+++ b/python/utils/trace.py
@@ -1324,6 +1324,7 @@ def configure_packet_tracing_aie2(
         MemEvent.EDGE_DETECTION_EVENT_1,
     ],
     packet_id=16,
+    enable_packet_id=False,
     shim_burst_length=64,
 ):
 
@@ -1386,6 +1387,12 @@ def configure_packet_tracing_aie2(
 
     exist_core_tile_traces = []
     for i in range(len(tiles_to_trace)):
+
+        if enable_packet_id:
+            p_id = packet_id
+        else:
+            p_id = i + 1
+
         if isShimTile(tiles_to_trace[i]):
             if tiles_to_trace[i] == shim:
                 configure_shimtile_tracing_aie2(
@@ -1394,8 +1401,7 @@ def configure_packet_tracing_aie2(
                     stop=stop_user_event,
                     events=shimtile_events,
                     enable_packet=1,
-                    # packet_id=i + 1, # DEPRECATED - auto incrementing packet_id assignment
-                    packet_id=packet_id,
+                    packet_id=p_id,
                     packet_type=PacketType.SHIMTILE,
                 )
                 configure_timer_ctrl_shimtile_aie2(tiles_to_trace[i], start_user_event)
@@ -1406,8 +1412,7 @@ def configure_packet_tracing_aie2(
                     stop=stop_shimtile_broadcast_event,
                     events=shimtile_events,
                     enable_packet=1,
-                    # packet_id=i + 1, # DEPRECATED - auto incrementing packet_id assignment
-                    packet_id=packet_id,
+                    packet_id=p_id,
                     packet_type=PacketType.SHIMTILE,
                 )
                 configure_timer_ctrl_shimtile_aie2(
@@ -1420,8 +1425,7 @@ def configure_packet_tracing_aie2(
                 stop=stop_memtile_broadcast_event,
                 events=memtile_events,
                 enable_packet=1,
-                # packet_id=i + 1, # DEPRECATED - auto incrementing packet_id assignment
-                packet_id=packet_id,
+                packet_id=p_id,
                 packet_type=PacketType.MEMTILE,
             )
             configure_timer_ctrl_memtile_aie2(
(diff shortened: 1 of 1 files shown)


programming_examples/basic/vector_scalar_mul/vector_scalar_mul_placed.py
```diff
@@ -81,15 +84,19 @@ def core_body():
         tiles_to_trace = [ComputeTile2, ComputeTile2]
         if trace_size > 0:
             trace_utils.configure_packet_tracing_flow(tiles_to_trace, ShimTile)
+            trace_utils.configure_packet_ctrl_flow([ComputeTile2], CtrlShimTile)
+
+        trace_size_int32 = trace_size // 4
```
hunhoffe: Can you do math on dtypes instead of magic number 4?

python/utils/trace.py
```diff
@@ -874,6 +874,125 @@ def configure_shimtile_dma_tracing_aie2(
     )
 
 
+def push_bd_to_task_queue(
+    shim, bd_id, direction, channel, enable_token  # 0: s2mm, 1: mm2s
+):
+    task_queue_addr = 0x0
+    if direction == 0:  # s2mm
```
hunhoffe: Can you use constant variable? The comment you added is nice, but reusable constants with descriptive names seem nicer (at least, to me) :)

python/utils/trace.py
```diff
@@ -1387,15 +1524,16 @@ def configure_packet_tracing_aie2(
...
+    configure_shimtile_dma_aie2(
         shim=shim,
         channel=1,
         bd_id=15,
         ddr_id=ddr_id,
         size=trace_size,
         offset=trace_offset,
         enable_token=enable_token,
-        enable_packet=1,
+        enable_packet=1,  # not used
```
hunhoffe: this comment is confusing to me.... why is it not used? Why is it specified if not needed?

python/utils/xrt.py
```diff
@@ -204,11 +204,21 @@ def setup_aie(
 
     if in_0_shape and in_0_dtype:
         if verbosity >= 1:
-            print("register 1st input to group_id 3")
+            print(
```
hunhoffe, line 207: Consider using syntax:
print(f"my format string {my var}")

hunhoffe (commented):
I left some nit-picky python style comments, implement as you see fit! Nothing is urgent.

python/utils/trace.py
jackl-xilinx: Great suggestion. I'll add that.

python/utils/trace.py
jackl-xilinx: Good point. Cleaned up.

python/utils/xrt.py
jackl-xilinx: thanks for the suggestion.

programming_examples/basic/vector_scalar_mul/vector_scalar_mul_placed.py
jackl-xilinx: Good suggestion.

PR #2658, opened 2025-10-20
[aievec] Add support for aie2p vector ops in `-convert-aievec-to-llvm`

erwei-xilinx:
~~- TODO: mlir ir tests.~~

### DIFF
--- a/test/Conversion/AIEVecToLLVM/test-srs-aie2p.mlir
+++ b/test/Conversion/AIEVecToLLVM/test-srs-aie2p.mlir
@@ -1,0 +1,39 @@
+//===- test-srs-aie2p.mlir -------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt %s -split-input-file --convert-aievec-to-llvm="aie-target=aie2p" | FileCheck %s
+
+func.func @v64bf16_srs_v64f32(%arg0 : vector<64xf32>) {
+  %c0 = arith.constant 0 : i32
+  %0 = aievec.srs %arg0, %c0 : vector<64xf32>, i32, vector<64xbf16>
+  return
+}
+
+// CHECK-LABEL: @v64bf16_srs_v64f32
+// CHECK-SAME: %[[ARG0:.*]]: vector<64xf32>
+// CHECK-NEXT: %[[SHIFT0:.*]] = arith.constant 0 : i32
+// CHECK-DAG: %[[INDEX0:.*]] = llvm.mlir.constant(0 : i32) : i32
+// CHECK-DAG: %[[INDEX1:.*]] = llvm.mlir.constant(1 : i32) : i32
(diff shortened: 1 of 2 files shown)


lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -950,23 +1020,41 @@ class UPSOpConversion : public mlir::ConvertOpToLLVMPattern<aievec::UPSOp> {
           rewriter.getI32IntegerAttr(op.getShift()));
 
       SmallVector<Value> operands({opSrcVal, shiftCst, signCst});
+      bool isAIE2p = (aieTarget == "aie2p");
```
jsetoain: It's better to just have a different pattern for AIE2, and AIE2P. Keep them separated. If everything is the same except the external intrinsic, you can probably use a template and instantiate one or the other when configuring legalizations.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -1092,6 +1239,7 @@ class SRSOpConversion : public mlir::ConvertOpToLLVMPattern<aievec::SRSOp> {
 
     // Integer types
     Value srsIntrOp = nullptr;
+    bool isAIE2p = (aieTarget == "aie2p");
```
jsetoain: Same here. As a general rule, we want a different set of patterns for each legalization. It's okay if the patterns are parametrized, and you instantiate with different parameters for different architectures. But those large "if" statements with slight variations here and there, depending on the architecture, spell doom.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -2412,15 +2629,12 @@ class ShuffleOpConversion
 
 void populateAIEVecToLLVMConversionPatterns(
     mlir::LLVMTypeConverter &converter, mlir::RewritePatternSet &patterns,
-    Aie2Fp32Emulation aie2Fp32EmulationOption) {
+    Aie2Fp32Emulation aie2Fp32EmulationOption, StringRef aieTarget) {
```
jsetoain: We don't want to deal with this like so. We want "populate" functions per backend, so that each can be maintained in an isolated way.

You can have a look at how it's done for the [Vector to AIEVec conversion pass](https://github.com/Xilinx/mlir-aie/blob/2cb8aedb6135353227bc1bfda7f41b77c70f3895/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp#L3098-L3186). We have `populateCommon`, `populateV1`, and `populateV2`.

jsetoain (changes requested):
Just some minor adjustments, for maintainability.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
erwei-xilinx: Sure, updated to use templates.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
erwei-xilinx: Updated.

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
erwei-xilinx: Makes sense. Codebase restructured to manage passes into pipelines: `CommonConversionPatterns`, `AIE2ConversionPatterns` and `AIE2pConversionPatterns`.

jsetoain (approved):
lgtm

PR #2665, opened 2025-10-23
[aievec] Fixup a condition in trunc/ext op lowering

erwei-xilinx:
- Previously, trunc/ext for i16->i64 will be incorrectly lowered to i32->i64 due to i32 being returned as acc type (should be i64).
- Ensure trunc/ext always lower to aievec, as arith dialect's direct lowerings to llvm behave unexpectedly.

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
@@ -3636,7 +3636,7 @@ static void configureAIEVecV1Legalizations(ConversionTarget &target,
 
 static void configureAIEVecV2PLegalizations(ConversionTarget &target,
                                             TargetBackend backend) {
-  // AIE2P-specific legalization: ExtFOp is always illegal
+  // AIE2P-specific legalization: ExtFOp on vector is always illegal
   target.addDynamicallyLegalOp<arith::ExtFOp>([](arith::ExtFOp extfOp) {
     auto srcType = dyn_cast<VectorType>(extfOp.getIn().getType());
     auto dstType = dyn_cast<VectorType>(extfOp.getOut().getType());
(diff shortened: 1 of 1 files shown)


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -3306,12 +3312,8 @@ static void configureAIEVecCommonLegalizations(ConversionTarget &target,
     if (!isa<IntegerType>(srcScalarType) || !isa<IntegerType>(dstScalarType))
       return true;
 
-    unsigned srcLaneSize = getVectorLaneSize(srcType);
-    unsigned dstLaneSize = getVectorLaneSize(dstType);
-    unsigned srcElWidth = srcScalarType.getIntOrFloatBitWidth();
-    unsigned dstElWidth = dstScalarType.getIntOrFloatBitWidth();
-    return srcLaneSize != 32 || (dstElWidth <= srcElWidth) ||
-           (dstLaneSize != srcLaneSize);
+    // For aie2, ExtSIOp is always illegal
```
jsetoain: If `ExtSIOp` is always illegal, then it's better to mark it as illegal (as opposed to dynamically legal). If it's only illegal for AIE2, illegality should be marked in the AIE2 legalizations, as opposed to the common ones.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -3344,13 +3346,8 @@ static void configureAIEVecCommonLegalizations(ConversionTarget &target,
...
       return true;
 
-    unsigned srcLaneSize = getVectorLaneSize(srcType);
-    unsigned dstLaneSize = getVectorLaneSize(dstType);
-    unsigned srcElWidth = srcScalarType.getIntOrFloatBitWidth();
-    unsigned dstElWidth = dstScalarType.getIntOrFloatBitWidth();
-
-    return srcLaneSize != 32 || (dstElWidth >= srcElWidth) ||
-           (dstLaneSize != srcLaneSize);
+    // For aie2, TruncIOp is always illegal
```
jsetoain: Same here for `TruncIOp`

jsetoain (commented):
Just a couple of nitpicks.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
erwei-xilinx: Makes sense. I should move those legalizer changes to aie2 pipeline instead. Updated.

lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
erwei-xilinx: Updated.

PR #2673, opened 2025-10-29
Unplaced GEMM for Large Workloads

kurtis-b-1:
GEMM using the IRON tiler currently doesn't work for large workloads, e.g. K >=1024. This is due to an optimization in the IRON tiling logic to combine tiles if the tiles across each step are next to each other. For large workloads, this results in a TAP that could have a size value exceeding the maximum allowed for that size dimension. 

A workaround for this is allowing for this optimization to be disabled. This workaround is a band-aid for getting unplaced GEMM working for large workloads. A better solution is discussed in Issue https://github.com/Xilinx/mlir-aie/issues/2674.


python/helpers/taplib/tensortiler2d.py
```diff
@@ -29,6 +29,7 @@ def simple_tiler(
         tile_col_major: bool = False,
         iter_col_major: bool = False,
         pattern_repeat: int = 1,
+        prune_step: Optional[bool] = True,
```
hunhoffe: Looks good but please update docstring below too!

python/helpers/taplib/tensortiler2d.py
```diff
@@ -52,6 +53,7 @@ def simple_tiler(
             tile_col_major=tile_col_major,
             iter_col_major=iter_col_major,
             pattern_repeat=pattern_repeat,
+            prune_step=prune_step,
```
hunhoffe, line 57: Docstring?

python/helpers/taplib/tensortiler2d.py
```diff
@@ -65,6 +67,7 @@ def group_tiler(
         iter_col_major: bool = False,
         pattern_repeat: int = 1,
         allow_partial: bool = False,
+        prune_step: Optional[bool] = True,
```
hunhoffe: Docstring?

python/helpers/taplib/tensortiler2d.py
```diff
@@ -110,6 +114,7 @@ def step_tiler(
         iter_col_major: bool = False,
         allow_partial: bool = False,
         pattern_repeat: int = 1,
+        prune_step: Optional[bool] = True,
```
hunhoffe: Docstring?

python/helpers/taplib/tensortiler2d.py
hunhoffe: Also -- I don't think you need an optional here. It can just be a bool? And that can be true elsewhere as well? If not, preferred type hint syntax is bool | None (I think the Optional is going away in best practice type hints)

python/helpers/taplib/tensortiler2d.py
```diff
@@ -336,6 +342,7 @@ def __sizes_strides_for_step_tile_group(
         tile_col_major: bool,
         tile_group_col_major: bool,
         pattern_repeat: int,
+        prune_step: Optional[bool] = True,
```
hunhoffe: Type, docstring?

hunhoffe (changes requested):
Looks good! Please double check types and docstrings

python/helpers/taplib/tensortiler2d.py
kurtis-b-1: Should I add a docstring for this function call?

python/helpers/taplib/tensortiler2d.py
kurtis-b-1: Updated! Thanks for the review.

hunhoffe (approved):
Looks good!

PR #2677, opened 2025-10-31
[TEST] Build wheels for Windows

jgmelber:


jgmelber:
> Having the option to grab a pre-compiled `xrtbinutil` might be convenient.

Heard. 

> I know we direct people to turn off SecureBoot, but I'm really not comfortable with that being a universal requirement. 

This was from early days when the kernel driver wasn't included in Linux. At this point we can likely move on from that direction. And soon once XRT and xdna-driver are upstream _in Linux_ we can definitely. 

> I'll also try to put together a list of things that need to be fixed in llvm-aie, so maybe somebody with more credibility than I on that project can get some movement there. That would save a few hundred lines of code.

Please let us know on this side too. 

_**Thanks for the work and looking forward to your PR on ~Monday**_

thomthehound:
Also, because I'm sure you are curious, the 'magic bullet' on Windows is to make sure `Xilinx_XRT` is never set. `XRT_ROOT` is fine, but `Xilinix_XRT` seems to be a poison to stop consuming system paths, and that nukes the whole thing. Clearly that is meant for internal driver builds. And, because the Windows install of XRT blithely encourages the user to execute the packaged `setup.bat` the same way they would `source` from the Linux install, they are sabotaged before they even start. So annoying, but there is nothing I can do about that but instruct.

PR #2681, opened 2025-11-04
Fix case bug

fifield:
fix bug where tests aren't actually running in CI:

BEFORE:
```
Found Ryzen AI device: [0000:c5:00.1]
	model: 'Phoenix'
NPU1 detected but aietools for aie2 not available
```
```
Found Ryzen AI device: [0000:c6:00.1]
	model: 'Strix'
NPU2 detected but aietools for aie2p not available
```
[AFTER:
](https://github.com/Xilinx/mlir-aie/actions/runs/19078726153?pr=2681)
```
Found Ryzen AI device: [0000:c5:00.1]
	model: 'Phoenix'
Running tests on NPU1 with command line: /home/github/actions-runner/_work/mlir-aie/mlir-aie/utils/run_on_npu.sh
```
```
Found Ryzen AI device: [0000:c6:00.1]
	model: 'Strix'
Running tests on NPU2 with command line: /home/github/actions-runner/_work/mlir-aie/mlir-aie/utils/run_on_npu.sh
```

PR #2684, opened 2025-11-04
Integrate python test lit config with helper classes

hunhoffe:
This is meant to be a small follow-on to #2661


hunhoffe:
Waiting on https://github.com/Xilinx/mlir-aie/pull/2681 before review

Issue #2653, opened 2025-10-16
[Runtime Sequence] Wrong placement of wait statements


Victor-Jung:
@hunhoffe 
Whenever no task group is provided in a `rt.drain`, `wait` and `free` statements will be emitted right after the first `rt.drain` call. If the user has two `rt.drain` in a row, and the second drain depends on some `rt.fill`, then the `rt.fill` will be freed and the application will deadlock.

The desirable behavior is to insert the wait and free after the last consecutive `drain`. A warning could be emitted to prevent the user that they should explicitly use the task groups to handle runtime sequence dependency.

For instance:
```
rt.drain(memO.cons(), O, tap=O_tiles[division_factor*head_idx],
                        wait=True,
                        placement = Tile(col = 7, row = 0),
                    )
rt.drain(memO2.cons(), O, tap=O_tiles[division_factor*head_idx + 1],
            wait=True,
            placement = Tile(col = 7, row = 0),
        )
```
Generates:
```
%4 = aiex.dma_configure_task_for @memO {
        aie.dma_bd(%arg3 : memref<1x512x64xbf16>, 0, 16384, [<size = 1, stride = 0>, <size = 1, stride = 32768>, <size = 256, stride = 64>, <size = 64, stride = 1>]) {burst_length = 0 : i32}
        aie.end
} {issue_token = true}
aiex.dma_start_task(%4)
aiex.dma_await_task(%4)
%5 = aiex.dma_configure_task_for @memO2 {
  aie.dma_bd(%arg3 : memref<1x512x64xbf16>, 16384, 16384, [<size = 1, stride = 0>, <size = 1, stride = 32768>, <size = 256, stride = 64>, <size = 64, stride = 1>]) {burst_length = 0 : i32}
  aie.end
} {issue_token = true}
aiex.dma_start_task(%5)
aiex.dma_await_task(%5)
```
But it should generate:
```
%4 = aiex.dma_configure_task_for @memO {
        aie.dma_bd(%arg3 : memref<1x512x64xbf16>, 0, 16384, [<size = 1, stride = 0>, <size = 1, stride = 32768>, <size = 256, stride = 64>, <size = 64, stride = 1>]) {burst_length = 0 : i32}
        aie.end
 } {issue_token = true}
aiex.dma_start_task(%4)
%5 = aiex.dma_configure_task_for @memO2 {
[...]

PR #2687, opened 2025-11-06
Refine default Python runtime sequence behavior; remove experimental code.

hunhoffe:
Previously, the default behavior for runtime sequences that don't use task groups was to insert dma_frees after each operation that waited. This can lead to weird behavior, even for simple cases such as two drains in a row, as noted in https://github.com/Xilinx/mlir-aie/issues/2653. 

This PR changes the default behavior to use a default task group when a task group is not specified. This introduces a nice property this ensures runtime sequences using a single task group run the same was as runtime sequences that don't explicitly use a task group. 

I think this is easier to reason about, but would welcome feedback!

The only code that really relied on the default behavior was some experimental code I had added some time back. I've deleted that code; it should be easy enough to fetch in the git history if it is needed in the future.

TODOs:
- [x] Create tests to verify the default behavior
- [x] Modify the default behavior if necessary
- [x] Ensure the default behavior is documented in the programming guide

### DIFF
--- a/python/helpers/dialects/ext/scf.py
+++ b/python/helpers/dialects/ext/scf.py
@@ -1,3 +1,4 @@
+import numpy as np
 from typing import Sequence
 
 from ....ir import InsertionPoint, Value
@@ -24,7 +25,18 @@ def _for(
     The insert_yield parameter defaults to True; if left as True, the user no longer needs to manually insert
     yield operations (```yield_([])```). If the user wishes to specify yield directly (such as if there is
     a return value from the loop body), insert_yield should be set to False.
+
+    I also added some handling of numpy data types for the loop variable
     """
+    if not (
+        isinstance(start, int)
+        or isinstance(start, np.integer)
+        or isinstance(start, Value)
+    ):
+        raise ValueError(
+            f"MLIR SCF for loops can only use integer types, not {type(start)}"
+        )
+
     if step is None:
         step = 1
     if stop is None:
@@ -32,9 +44,11 @@ def _for(
         start = 0
     params = [start, stop, step]
     for i, p in enumerate(params):
-        if isinstance(p, int):
+        if isinstance(p, np.integer):
+            p = p.item()
+        if isinstance(p, np.integer) or isinstance(p, int):
             p = constant(p, index=True)
-        if not _is_index_type(p.type):
+        elif not _is_index_type(p.type):
             p = index_cast(p, to=T.index())
         params[i] = p
 
--- a/python/helpers/util.py
+++ b/python/helpers/util.py
@@ -1,8 +1,6 @@
 from collections import defaultdict
 import numpy as np
-import sys
-from typing import Sequence, get_args, get_origin, TypeVar
-
+from typing import Sequence, get_args, get_origin
 from aie._mlir_libs import _aie as CustomTypes
 
 from ..extras import types as T
--- a/test/python/control_flow.py
+++ b/test/python/control_flow.py
@@ -1,0 +1,77 @@
+import numpy as np
+
+from aie.iron import ObjectFifo, Program, Runtime, Worker
+from aie.iron.placers import SequentialPlacer
+from aie.iron.device import NPU2
+from aie.iron.controlflow import range_
+from aie.helpers.dialects.ext.func import func
+
+# RUN: %python %s | FileCheck %s
+
+
+def custom_loop_type(loop_dtype):
+    # Define tensor types
+    vector_size = 1024
+    loop_iter = 4
+    line_size = vector_size // loop_iter
+    line_type = np.ndarray[(line_size,), np.dtype[np.uint8]]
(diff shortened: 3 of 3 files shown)


programming_guide/section-2/section-2d/RuntimeTasks.md
```diff
@@ -174,15 +174,16 @@ Currently, a `WorkerRuntimeBarrier` may take any value between 0 and 63. This is
...
+`RuntimeTask`s can be added to a task group by specifying their `task_group` input. Tasks in the same group will be appended to the runtime sequence and executed in order. The `finish_task_group()` operation is used to mark the end of a task group, i.e., after this operation all of the tasks in the group will be waited on for completion after which they will be freed at the same time.
+If a `RuntimeTask` group is not explicitly defined for DMA tasks defined in a `Runtime`'s `sequence`, then a single default task group is used.
 
 > **NOTE:**  Because of their ability to wait on runtime tasks until completion and free all the resources at the same time, task groups are well-placed to handle the asynchronous nature of runtime data movement tasks.
```
andrej:
```suggestion
> **NOTE:**  A call to  `finish_task_group()` blocks the runtime sequence until all of the group's tasks annotated with `wait=True`  ("awaited tasks") have completed. After waiting, all resources of the task group -- including those _not_ annotated with `wait=True` ("unawaited tasks") -- will be freed and reused for subsequent tasks. 
> 
> To avoid race conditions, any unawaited tasks in the group should form a dependency of an awaited task.
> It is only safe to remove a `wait=True` if you can reason that another, awaited task in the same group can only complete if the awaited task also completed.
> For example, you may choose to set `wait=False` on an input fill if you can guarantee that a later (awaited) output drain depends on the input and completes only if the input fill completed as well.
>
> If you suspect a race condition, the safest (but possibly slower) solution is to annotated _all_ tasks (including inputs) with `wait=True`.
```

programming_guide/section-2/section-2d/RuntimeTasks.md
```diff
@@ -174,15 +174,16 @@ Currently, a `WorkerRuntimeBarrier` may take any value between 0 and 63. This is
...
 
 To facilitate this reconfiguration step, IRON introduces `RuntimeTaskGroup`s which can be created using the `task_group()` function as defined in [runtime.py](../../../python/iron/runtime/runtime.py).
 
-`RuntimeTask`s can be added to a task group by specifying their `task_group` input. Tasks in the same group will be appended to the runtime sequence and executed in order. The `finish_task_group()` operation is used to mark the end of a task group, i.e., after this operation all of the tasks in the group will be waited on for completion after which they will be freed at the same time and the runtime sequence will be reconfigured by the next task group.
+`RuntimeTask`s can be added to a task group by specifying their `task_group` input. Tasks in the same group will be appended to the runtime sequence and executed in order. The `finish_task_group()` operation is used to mark the end of a task group, i.e., after this operation all of the tasks in the group will be waited on for completion after which they will be freed at the same time.
```
andrej:
```suggestion
`RuntimeTask`s can be added to a task group by specifying their `task_group` input. Tasks in the same group will be appended to the runtime sequence and executed in order. The `finish_task_group()` operation is used to mark the end of a task group. This call waits for tasks in the group annotated with `wait=True` to complete, and then frees _all_ resources used by the task.
```

andrej: Not sure I like the "in order" in this description. I think tasks within a group can run in parallel or other orders, but the idea of task groups is to order tasks _across_ groups, right?

python/iron/runtime/runtime.py
```diff
@@ -319,36 +319,36 @@ def sequence(*args):
...
 
-            no_waits = []
+            def finish_task_group(tg, task_group_actions):
+                actions = task_group_actions[tg]
+                for fn, args in actions:
+                    if fn == dma_await_task:
+                        fn(*args)
+                for fn, args in actions:
+                    if fn != dma_await_task:
+                        fn(*args)
```
andrej: Not familiar with the code so please correct me if I'm misunderstanding something. Is case specifically for the the `dma_free_task` action, or are there others here? That would make sense for me (wait first, then free everything). If it's only `free`, can we make it explicit and raise an Error if any other actions other than `await` and `free` are encountered in the list? I think it would make the intent clearer.

hunhoffe: Yes, I can rewrite the code to be more easily readable. Pushed in recent commit!

hunhoffe:
@andrej I quite like your suggestion, and have enabled it be default, but I did leave an escape hatch so "expert users" may mix default/not default if they wish through a flag to the ```Runtime()``` constructor.

I also updated the tests to test this flag. I then found a fair amount of programming examples that mixed default/explicit task groups, and updated them with a more detailed comment.

jgmelber (approved):

PR #2690, opened 2025-11-07
[aiecc.py] Switch to using bundled `convert-to-llvm` pass

erwei-xilinx:
- Deprecate the previous lowering-to-llvm pipeline which calls individual conversions to llvm.
- Enables `dynamic=True` option, to allow the upstream llvm lowering passes to query any DLTI specs specified on `aie.device`.
- Adds a pass `aie-vector-transfer-lowering` which is simply a wrapper around the upstream `vector::populateVectorTransferLoweringPatterns`; this step is currently excluded from the upstream `convert-to-llvm` pass.

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEVectorTransferLowering.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEVectorTransferLowering.cpp
@@ -1,5 +1,4 @@
-//===- AIEVectorTransferLowering.cpp ------------------------------*- C++
-//-*-===//
+//===- AIEVectorTransferLowering.cpp -----------------------------*- C++-*-===//
 //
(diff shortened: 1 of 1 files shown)


lib/Dialect/AIE/Transforms/AIEVectorTransferLowering.cpp
```diff
@@ -0,0 +1,72 @@
+//===- AIEVectorTransferLowering.cpp ------------------------------*- C++
+//-*-===//
```
fifield: nit: comment line wrap

erwei-xilinx: Thanks for pointing it out! Updated.

PR #2691, opened 2025-11-07
Python Device class extension and cleanup

hunhoffe:
There has been a lot of neat progress on using targetmodel information from C++ in the Python Device classes. This PR continues this trend, and also makes code a bit more succinct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@ Turn off SecureBoot (Allows for unsigned drivers to be installed):
    >  [0000:66:00.1]  :  NPU Strix
    >  ```
 
+   IRON requires that `xrt-smi` be in your path.
+
 ### Install IRON and MLIR-AIE Prerequisites
 
 1. Install the following packages needed for MLIR-AIE:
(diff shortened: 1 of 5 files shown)


python/iron/device/device.py
```diff
@@ -71,169 +73,114 @@ def __init__(self, device: AIEDevice) -> None:
...
                 self._tiles[c].append(Device.__DeviceTile(c, r))
 
+    def tile_iterator(self) -> Generator[Tile, None, None]:
+        """
+        Iterates over the device tiles deterministically
+        """
+        for c in range(self._tm.columns()):
+            for r in range(self._tm.rows()):
+                yield self._tiles[c][r]
+        return None
```
andrej, line 89: Nit: I think this `return None` can be omitted.

andrej (approved):
Nice streamlining! Always nice to see a PR that deletes more lines than it adds. :)

python/iron/device/device.py
hunhoffe: Whoops, good catch! I will need to do a follow on PR related to this, and I'll make it note to fix it then!

Issue #2606, opened 2025-09-24
Python bindings silently fail if XRT is not found.


stephenneuendorffer:
If the XRT setup script is not found, then many iron bindings are silently not injected into the iron namespace.  The resulting message looks like a python setup problem and is misleading.   It would be better to either fail early, or to inject dummy names that fail later with a good error message.

hunhoffe:
https://github.com/Xilinx/mlir-aie/pull/2692 addresses part of this issue. XRT no longer causes silent failures, but the failures aren't early either and there is not a good failure message. I'll leave this open because it's only been partially addressed.

Issue #2252, opened 2025-04-30
Permit tensor creation without backing storage


ypapadop-amd:
We need some kind to communicate shape and dtype for tensors that are not backed by data when we do compilation only, no dispatch. The use case is the following:
```python

@iron.jit
def npu_function(input0, input1, output)
    ...

input0 = dummy_tensor(shape=(128, ), dtype=np.float16)
input1 = dummy_tensor(shape=(128, ), dtype=np.float16)
output = dummy_tensor(shape=(128, ), dtype=np.float16)
npu_function(input0, input1, output) # this only compiles, does not dispatch
```

`dummy_tensor` can return a `iron.Tensor` or some other type to not overburden the interface of `iron.Tensor`.

hunhoffe:
Reference for the future: https://docs.pytorch.org/docs/stable/torch.compiler_fake_tensor.html

hunhoffe:
The spirit of this issue is addressed through two existing mechanisms:

Compile without tensor allocation:
- `design.compile(xclbin_path=..., inst_path=...)` on a `CallableDesign` / `CompilableDesign` compiles to an xclbin with no runtime tensors required at all — shape/dtype information flows entirely through `CompileTime[T]` parameters.
- `design.as_mlir(None, None, None)` returns the resolved MLIR text without compiling or allocating tensors, explicitly documented to accept `None` for tensor args when all shape/dtype info comes from `CompileTime[T]` params.

Internal fake-tensor machinery:

For the algorithms layer (transform, for_each, reduce, etc.), a set of internal lightweight descriptors (`_make_fake_tensor`, `_TypeDescriptor`, `_ParamDescriptor`) in `python/iron/algorithms/_transform.py` provide shape+dtype objects without backing storage. These are used internally to drive MLIR generation without real NPU memory. They are not yet exposed as a public API, as I have limited reasons to, but the underlying capability exists.

@ypapadop-amd do you agree this issue/capability is adequately addressed?

hunhoffe:
I'm going to close this issue for now, but do feel free to file new issues if you have any difficulties!

PR #2692, opened 2025-11-07
Decouple IRON Tensor from XRT Host Runtime

hunhoffe:
This PR:
- Creates an abstract Tensor class
- Creates an XRTTensor class, which is used when PyXRT is present
- Creates a CPUOnlyTensor for when PyXRT is not present
- Imports PyXRT on-the-fly in the JIT code. If PyXRT is not found, this will raise an exception (https://github.com/Xilinx/mlir-aie/issues/2606)
- Extends tensor tests a little for more data types and to test both implementations (XRTTensor and CPUOnlyTensor)

This PR leaves several related tasks for future work:
* This PR does not untangle PyXRT from a generic runtime abstraction
* This PR does not seek to optimize the placements of syncs within the XRTTensor; it leaves existing syncing behavior largely as-is.
* This PR does not attempt to fully standardize the interface to either torch.Tensor or np.array

### DIFF
--- a/python/iron/hostruntime/tensor.py
+++ b/python/iron/hostruntime/tensor.py
@@ -125,7 +125,7 @@ class Tensor(ABC):
         """
         Number of bytes consumed by elements in the tensor
         """
-        return self.numel * self.element_size
+        return self.numel() * self.element_size
 
     @cached_property
     def element_size(self) -> int:
@@ -223,7 +223,6 @@ class Tensor(ABC):
         if self.device == "npu":
             self._sync_to_device()
 
-    @cached_property
     def numel(self):
         """
         Calculates the number of elements in the tensor.


python/iron/hostruntime/xrtruntime/__init__.py
```diff
@@ -0,0 +1,4 @@
+try:
```
ypapadop-amd, line 9: It needs copyright blurb here.

python/iron/hostruntime/tensor.py
```diff
@@ -43,42 +36,31 @@ def __init__(self, shape_or_data, dtype=np.uint32, device="npu"):
...
             raise ValueError(f"Unsupported device: {device}")
 
         self.device = device
 
         if isinstance(shape_or_data, tuple):
             self.shape = shape_or_data
             self.dtype = dtype
-            self.data = np.zeros(self.shape, dtype=self.dtype)
         else:
             np_data = np.array(shape_or_data, dtype=dtype)
```
ypapadop-amd: Can we avoid creating `np.array` just to figure out shape_or_data?

python/iron/hostruntime/tensor.py
```diff
@@ -43,42 +36,31 @@ def __init__(self, shape_or_data, dtype=np.uint32, device="npu"):
...
-                self.__sync_to_device()
+    def __repr__(self):
+        """
+        Return a string representation of the tensor.
+
+        Note: This method may implicitly trigger data synchronization to devices.
+        """
+        if self.device == NPU_DEVICE:
+            self._sync_from_device()
+        array_str = np.array2string(self.data, separator=",")
```
ypapadop-amd, line 63: `data` is an attribute of the derived classes. Maybe leave it to the derived classes or use `hasattr()`?

python/iron/hostruntime/tensor.py
```diff
@@ -43,42 +36,31 @@ def __init__(self, shape_or_data, dtype=np.uint32, device="npu"):
...
-            if self.device == "npu":
-                self.__sync_to_device()
+    def __repr__(self):
+        """
+        Return a string representation of the tensor.
+
+        Note: This method may implicitly trigger data synchronization to devices.
+        """
+        if self.device == NPU_DEVICE:
+            self._sync_from_device()
```
ypapadop-amd: This should be only for XRTTensor.

hunhoffe:
@ypapadop-amd I am glad you see this as a step forward and thank you for your review and feedback! I think some of your comments come from a difference of view on scope of the ```iron.Tensor``` class. I agree with your comments on tensor vs. backing store, but I was seeing the ```iron.Tensor``` as having a different role.

A generic tensor class is already implemented in ```torch.Tensor``` and I do not have an interest in reinventing it here. We have a loose goal of interoperability with other ```torch.Tensor```-like things, which is not explored in this PR for reasons we've discussed offline, and that similarity is how I see us gaining more generic tensor functionality. So my goal for ```iron.Tensor``` was to create a ```torch.Tensor```-like thing that handles tensor operations on an NPU/CPU system without relying on runtime-specific details of managing the backing store. So most of the things you found surprising were deliberate choices on my part.

I am open to changing my point of view on what the ```iron.Tensor``` should be, but right now I don't think I have a concrete example of a technical flaw or limitation of what I proposed here. Care to offer a counter-example to my plan? Does HSA have some notion of sync-to-device and sync-from-device? If so, I think what I have here is flexible enough for the purposes I've described.

python/iron/hostruntime/tensor.py
hunhoffe: I will delay addressing this comment pending the top-level discussion

ypapadop-amd:
The proposed design leads to something like `CPUTensor._sync_to_device()` which IMO is confusing (which device?). This is also why https://github.com/Xilinx/mlir-aie/issues/2252 is not satisfied: the base class (which is a fake tensor) still relies on the existence of a backing store.

> Does HSA have some notion of sync-to-device and sync-from-device? If so, I think what I have here is flexible enough for the purposes I've described.

It has a combination of different types of memories, e.g., coarse and fine-grain, and different fences support on a per packet basis. Think of stream synchronization in GPUs vs explicit sync as with the IRON tensor model.

As an aside, unrelated to reviewing this PR, I view explicit data synchronization on a high-level programming paradigm as problematic. IRON knows what is input and what is output, it should be able to annotate which tensors need sync after the invocation of a function, with the additional option of deferring sync when you have a sequence of tasks that reuse buffers. The user should not have to manually manage syncs in most cases.

python/iron/hostruntime/config.py
```diff
@@ -8,7 +8,10 @@
 import shutil
 import subprocess
 
-from .device import NPU1, NPU2
+from ..device import NPU1, NPU2
+
+NPU_DEVICE = "npu"
```
mawad-amd: Is there an actual need for introducing these global variables? Ok to use them internally but not exposing them to users. If not, I recommend keeping the old syntax of default values of `'npu'` and `'cpu'` which aligns what people write for PyTorch. If you want to introduce something, consider `iron.npu` and `iron.cpu` (or `iron.device.npu` and `iron.device.cpu`) that would look slightly better but still unnecessary.

hunhoffe:
Ok, thanks for elaborating! A few notes regarding your last comment, which will (hopefully) make my thought process more clear:

Regarding calling `iron.tensor._sync_to_device()` and `iron.tensor._sync_from_device()`:
* Function names denoted by a leading underscore imply they are Python's version of "private" functions. So it is the intent -- by convention -- that this function only be used internally by ```iron.Tensor``` and subclasses. 

Regarding "which device":
* This is a notion taken from ```torch```, that the tensor IS aware of the current device and the user does not call manual syncs but instead expresses their intent for the device/backing store of a tensor via calls to the ```tensor.to(device)``` function, in either ```torch``` or ```iron``` (this is one way the current tensor design is ```torch.Tensor```-like).

Regarding tensors being aware of backing store:
* A tensor cannot be a tensor without data. So it has to have some notion of a backing store.
* This awareness is extra important in a multi-device setting. The user can express to the ```Tensor``` their intent for the backing store, again via the ```tensor.to(device)``` function; this is important in avoiding unneeded syncs. So, the tensor here understands the difference between devices, and the abstract functions `iron.tensor._sync_to_device()` and `iron.tensor._sync_from_device()` provide the basis for custom device management via various runtimes, like PyXRT.
* This also has benefits for code maintenance, because if this tensor has knowledge of CPU and NPU devices -- which it does now -- then my hypothesis is that there may be common patterns. I've tried to capture some of those common patterns in the top-level ```iron.Tensor``` class, but they can easily be overridable in a child class if desired. If we move to other devices, I think we can extend/refine the interface between tensor/device/backing store.


Anyways, while I understand this does not match your thoughts on a solution to https://github.com/Xilinx/mlir-aie/issues/2252, I still believe it is *a* solution because I can't think of an example where the current structure is insufficient to express a programming pattern.

Also -- this is a good point that the concepts I was trying to represent by these classes is unclear. At the very least, I can improve documentation.

ypapadop-amd:
> Regarding tensors being aware of backing store:
> 
> * A tensor cannot be a tensor without data. So it has to have some notion of a backing store.

This is not true when it comes to compilation and that's why PyTorch supports it: https://docs.pytorch.org/docs/stable/torch.compiler_fake_tensor.html (the documentation shares the same concerns as me).

> Anyways, while I understand this does not match your thoughts on a solution to #2252, I still believe it is _a_ solution because I can't think of an example where the current structure is insufficient to express a programming pattern.

I disagree. The idea behind #2252 is to be able to compile a design without unnecessary memory overhead for a tensor that we are only using for the description of the data it offers.

hunhoffe:
@ypapadop-amd @mawad-amd Thank you both for your thoughts! I definitely managed to misinterpret the main issue of #2252 and I can see now that you are both correct. I have removed the primary reference to it in this PR. I will leave further thoughts on #2252 for another time and place.

python/iron/hostruntime/xrtruntime/__init__.py
hunhoffe: Fixed in recent commit.

python/iron/hostruntime/config.py
hunhoffe: I am generally against sprinkling string constants everywhere throughout code, but I will concede to your point about PyTorch. I reverted this change in a new commit.

python/iron/hostruntime/xrtruntime/tensor.py
```diff
@@ -0,0 +1,73 @@
...
+
+import numpy as np
+import pyxrt as xrt
+
+from ..tensor import Tensor
+
+
+class XRTTensor(Tensor):
+    """
+    Tensor object backed by NPU or CPU memory, fulfilled using PyXRT runtime operations.
```
ypapadop-amd: Should this say only NPU memory?

python/iron/hostruntime/tensor.py
hunhoffe: I've add a change in a recent commit so this is only done when creating the array is also useful for enforcing properties of arrays (all elements single type, etc.)

python/iron/hostruntime/tensor.py
hunhoffe: In a recent commit, I've changed it to an abstract property to enforce subclass implementation of both shape and data. Good call!

python/iron/hostruntime/xrtruntime/tensor.py
hunhoffe: Well, it's accessible from both.... I've fixed the documentation to say this. Another good catch!

PR #2696, opened 2025-11-10
Generate trace events enum package with CMake instead of manual generation

fifield:
## Summary
Automates the generation of trace event enumerations during the build process and adds support for multiple AIE architectures (AIE1, AIE2/AIEML, AIE2P) instead of only AIE2.

## Changes
- **Package structure**: Renamed `trace_events_enum.py` to `trace_events` package with per-architecture modules (`aie`, `aie2`, `aie2p`)
- **Import pattern**: Changed from `from aie.utils.trace_events_enum import ...` to `from aie.utils.trace_events.aie2 import ...` for explicit architecture selection
- **Class naming**: Renamed `PLEvent` to `ShimTileEvent` to match expected usage in trace utilities
- **Build integration**: 
  - Generate Python enum files and JSON database during CMake build into `build/regdb/` directory
  - Copy generated files to Python package directory with POST_BUILD commands
  - Install generated files alongside checked-in `__init__.py`
- **Generator improvements**:
  - Map `aieml` architecture to `aie2` Python module name
  - Conditionally emit `UCEvent` enum only when events are defined
  - Generate unified `events.json` database for future compiler integration
- **Updated imports**: Modified `trace.py`, `parse_trace.py`, and `parse_eventIR.py` to use new package structure

## Benefits
- **Multi-architecture support**: Trace utilities now work with AIE1, AIE2/AIEML, and AIE2P instead of only AIE2
- Trace events automatically stay in sync with aie-rt header files
- No manual regeneration needed when aie-rt is updated
- Architecture-specific imports make code more explicit and maintainable
- Reduced git repository size (generated files no longer checked in)

## Testing
- All existing Python tests pass
- Verified import compatibility across all architectures (AIE1: 123 events, AIE2: 127 events, AIE2P: 127 events)
- Confirmed JSON database generation

### DIFF
--- a/programming_examples/basic/vector_scalar_mul/vector_scalar_mul_placed.py
+++ b/programming_examples/basic/vector_scalar_mul/vector_scalar_mul_placed.py
@@ -16,7 +16,7 @@ from aie.iron.controlflow import range_
 
 import aie.utils.trace as trace_utils
 from aie.utils.trace import PortEvent
-from aie.utils.trace_events_enum import CoreEvent, MemEvent, ShimTileEvent, MemTileEvent
+from aie.utils.trace_events.aie2 import CoreEvent, MemEvent
 
 
 def my_vector_scalar_mul(dev, in1_size, in2_size, out_size, int_bit_width, trace_size):
--- a/programming_examples/ml/conv2d_14x14/conv2dk14_32core_placed.py
+++ b/programming_examples/ml/conv2d_14x14/conv2dk14_32core_placed.py
@@ -13,7 +13,7 @@ from aie.extras.context import mlir_mod_ctx
 from aie.iron.controlflow import range_
 import aie.utils.trace as trace_utils
 from aie.utils.trace import PortEvent
-from aie.utils.trace_events_enum import CoreEvent, MemEvent, ShimTileEvent, MemTileEvent
+from aie.utils.trace_events.aie2 import CoreEvent, MemEvent, ShimTileEvent, MemTileEvent
 from aie.helpers.taplib import TensorTiler2D, TensorAccessSequence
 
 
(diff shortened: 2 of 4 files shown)


fifield:
> ## Summary
> Automates the generation of trace event enumerations during the build process and adds support for multiple AIE architectures (AIE1, AIE2/AIEML, AIE2P) instead of only AIE2.
> 
There's only one minor difference between aie2 and aie2p, so the differences don't matter in practice. The main goal here is to enable aie1 and aie2ps/aie4 in future.

jackl-xilinx:
@fifield Can you do a diff between the generated events and the previous list of events in `trace_events_enum.py`. There are gaps where some values are not defined and this causes an issue with the numbering so I jut want to make sure all values are defined in the enum list.

fifield:
> @fifield Can you do a diff between the generated events and the list of events. There are gaps where some values are not defined and this causes an issue with the numbering so I jut want to make sure all values are defined in the enum list.

You're right there were gaps. I didn't check vs. the original file. Fixed in https://github.com/Xilinx/mlir-aie/pull/2696/commits/113327ffad4a54815d04149627e67d2868197541

utils/generate_events_enum.py
```diff
@@ -25,65 +25,259 @@ class MemEvent(Enum):
...
-    match = re.match(regex, line)
+    """Parse a single event declaration line."""
+    match = regex.match(line)
     if not match:
         return
     name, num = match.group(1), int(match.group(2))
-    if match.group(2) in dict:
-        sys.stderr.write("Duplicate event number {}.".format(num))
+    if num in dict:
+        sys.stderr.write(f"Warning: Duplicate event number {num} for {name}\n")
```
andrej: I wonder if this should error?

fifield: @copilot make this an error

PR #2710, opened 2025-11-13
[aiecc.py] Enable some recent passes that optimizes vectorized code's schedule

erwei-xilinx:
Add a new option, "optimization level" (-O) to `aiecc.py` to control the level of aggressiveness on schedule optimization passes.

All recent passes optimizing the vector code's schedule go to `-O3`, whereas `-O2` (default) gives the same behavior as the existing workflow.

This option now also drives the `-O` flags in peano `opt`, `llc` and `clang`.


fifield (approved):
Is there any reason to put these behind a flag (e.g. disable with `-O0`)? Or are these always beneficial / no downside?

PR #2727, opened 2025-11-21
Use `aie.tile` reference in `ShimDMAAllocationOp`

andrej:
# This PR

Before:
```mlir
%shim_noc_tile_0_0 = aie.tile(0, 0)
aie.shim_dma_allocation @alloc0 (MM2S, 0, 0)  // (unclear link between the two)
```

After:
```mlir
%shim_noc_tile_0_0 = aie.tile(0, 0)
aie.shim_dma_allocation @alloc0 (%shim_noc_tile_0_0, MM2S, 0)
```

 1. Refactor the `ShimDMAAllocationOp` to take in a SSA tile (`aie.tile`) instead of an integer for the column. Validation fails if the tile is not a shim tile.
 2. Remove automatic creation of the `aie.tile` in the later lowering passes if it does not exist. Instead, create the `aie.tile` together with the `ShimDMAAllocationOp` if it does not already exist. Error in later passes if the tile does not exist.

# Background

We use the `ShimDMAAllocationOp` to refer to the shim-side end of an 'open-ended' ObjectFIFO, i.e. an ObjectFIFO going to the host's main memory. The `ShimDMAAllocationOp` is a handle that allows us to pipe data from the runtime sequence into an ObjectFIFO. At lower levels (non-ObjectFIFO), the op is used to to represent one end of a flow (DMA channel + shim column) that can be used from shim DMA tasks or `dma_memcpy_nd` to pipe data into/from that flow. (Please inform me if there are additional uses of this op that I am not aware of.)

Currently, this ShimDMAAllocationOp has three attributes: DMA direction, column (as an integer) and channel (as an integer).

# Why?

 1. `aie.tiles` are the existing abstraction for representing tiles. I believe we should prefer it over integer values (e.g., harder to validate). When you look at lowered MLIR, this makes the link between the tile and the ObjectFIFO end clearer.
 2. The earlier `assign-tile-controller-ids` pass assigns necessary metadata with the shim tiles. Creating them without replicating those attributes breaks designs, so the current implementation is broken -- the part of the code that creates the shim `aie.tiles` just isn't exercised currently. I ran into this with multi-device configs.

# Future

Even though I view this as simply cleaning up an existing op we already have and making it more sturdy, this could potentially also be seen as a step towards a "half-ObjectFIFO" that we've discussed previously (to enable this, (1) the verifier should allow other tiles than shim tiles, (2) lowering passes of the ObjectFIFO would likely need to be adapted to support this, (3) the op should probably be renamed).

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -2352,7 +2352,7 @@ LogicalResult ShimDMAAllocationOp::verify() {
   int row = tileOp.getRow();
 
   if (!targetModel.isShimNOCorPLTile(col, row)) {
-    return emitOpError("tile must be a shim tile (row 0), but got tile(")
+    return emitOpError("tile must be a shim tile, but got tile(")
            << col << ", " << row << ")";
   }
 
--- a/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
@@ -219,7 +219,7 @@ public:
 
     auto channelDir = infoOp.getChannelDir();
     bool isMM2S = channelDir == AIE::DMAChannelDir::MM2S;
-    int col = shimTile.getCol();
+    int tileCol = shimTile.getCol();
     int tileRow = shimTile.getRow();
 
     // initialize fields to zero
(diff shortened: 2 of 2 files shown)


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2341,6 +2341,28 @@ void BDChainOp::print(OpAsmPrinter &printer) {
...
+  if (!tileOp) {
+    return emitOpError("tile operand must be a TileOp");
+  }
+
+  const auto &targetModel = getTargetModel(*this);
+  int col = tileOp.getCol();
+  int row = tileOp.getRow();
+
+  if (!targetModel.isShimNOCorPLTile(col, row)) {
+    return emitOpError("tile must be a shim tile (row 0), but got tile(")
```
hunhoffe: do you want to hard-code the row into the error message?

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -211,9 +211,16 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
...
+    if (!shimTile) {
+      return op->emitOpError(
+          "shim_dma_allocation op must reference a valid TileOp.");
+    }
+
     auto channelDir = infoOp.getChannelDir();
     bool isMM2S = channelDir == AIE::DMAChannelDir::MM2S;
-    int col = infoOp.getCol();
+    int col = shimTile.getCol();
+    int tileRow = shimTile.getRow();
```
hunhoffe, line 223: naming convention is messy with `col` and `tileRow`. Should be `col` and `row` or `tileCol` and `tileRow`?

PR #2730, opened 2025-11-22
Parallel/algorithms

hunhoffe:
I am creating this PR so the code does not get left behind.

#### Update by James:
This PR adds an `algorithm` library to simplify creating a design to run some tensor transformation on the NPU. These are compatible with `iron.jit`. Implemented initial support for C++ kernels and require specific function signatures as shown below. Follow-up work is needed to standardize kernel signatures and expand algorithm compatibility with kernels.

#### Using C++ Kernels (ExternalFunction)

The kernel signature and `ExternalFunction` currently must match the algorithm's expected format. We further assume array-like parameters in `params` are inputs that will require ObjectFifos and scalar-type parameters are inputs that will be encoded as MLIR constants.

| Algorithm | C++ Kernel Signature | ExternalFunction arg_types |
|-----------|---------------------|---------------------------|
| `transform` | `void kernel(T* in, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, *param_types, np.int32]` |
| `transform_binary` | `void kernel(T* in1, T* in2, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, tile_ty, *param_types, np.int32]` |
| `for_each` | `void kernel(T* in, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, *param_types, np.int32]` |
| `transform_parallel` | `void kernel(T* in, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, *param_types, np.int32]` |
| `transform_parallel_binary` | `void kernel(T* in1, T* in2, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, tile_ty, *param_types, np.int32]` |

### DIFF
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -127,7 +127,7 @@ runs:
           rm -rf "$WORKSPACE_ROOT"
         fi 
         
-        git clone --recursive -b ${{ github.head_ref || github.ref_name }} https://github.com/${{ github.repository }}.git "$WORKSPACE_ROOT"
+        git clone --recursive -b ${{ github.event.pull_request.head.ref || github.ref_name }} https://github.com/${{ github.event.pull_request.head.repo.full_name || github.repository }}.git "$WORKSPACE_ROOT"
         ls "$WORKSPACE_ROOT"
         
         if [ x"${{ inputs.MATRIX_OS }}" == x"windows-2022" ]; then
--- a/.github/workflows/mlirAIEDistro.yml
+++ b/.github/workflows/mlirAIEDistro.yml
@@ -51,7 +51,12 @@ jobs:
         run: |
           if [ x"${{ inputs.AIE_COMMIT }}" == x"" ]; then
             sudo apt install jq
-            AIE_PROJECT_COMMIT=$(curl -s https://api.github.com/repos/Xilinx/mlir-aie/commits/${{ github.head_ref || github.ref_name }} | jq -r '.sha[:8]')
+            if [ x"${{ github.event_name }}" == x"pull_request" ]; then
+              AIE_PROJECT_COMMIT="${{ github.event.pull_request.head.sha }}"
+              AIE_PROJECT_COMMIT="${AIE_PROJECT_COMMIT:0:8}"
+            else
+              AIE_PROJECT_COMMIT=$(curl -s https://api.github.com/repos/Xilinx/mlir-aie/commits/${{ github.ref_name }} | jq -r '.sha[:8]')
+            fi
           else
             AIE_PROJECT_COMMIT="${{ inputs.AIE_COMMIT }}"
           fi
@@ -190,7 +195,8 @@ jobs:
       shell: bash
       run: |
 
-        git clone --recursive https://github.com/${{ github.repository }}.git
+        AIE_REPO="${{ github.event.pull_request.head.repo.full_name || github.repository }}"
+        git clone --recursive https://github.com/${AIE_REPO}.git
         pushd mlir-aie
         git reset --hard ${{ needs.get_aie_project_commit.outputs.AIE_PROJECT_COMMIT }}
         git submodule update
--- a/include/aie/Dialect/AIE/IR/AIEDialect.h
+++ b/include/aie/Dialect/AIE/IR/AIEDialect.h
@@ -46,6 +46,12 @@ template <typename ConcreteType>
 struct SkipAccessibilityCheckTrait
     : mlir::OpTrait::TraitBase<ConcreteType, SkipAccessibilityCheckTrait> {};
 
+// Marker trait for operations that can be flow endpoints (e.g., TileOp, CoreOp,
+// MemOp)
+template <typename ConcreteType>
+struct IsFlowEndPoint : mlir::OpTrait::TraitBase<ConcreteType, IsFlowEndPoint> {
+};
+
 class TileOp;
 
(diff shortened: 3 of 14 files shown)


python/iron/kernel.py
```diff
@@ -8,6 +8,10 @@
 
 import hashlib
 import numpy as np
+import os
```
yenjames: TODO: clean this

programming_examples/algorithms/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Available Algorithms
+
+| Algorithm | Description | ExternalFunction Support |
+|-----------|-------------|--------------------------|
+| `transform` | Unary transformation with tiled processing | Yes |
+| `transform_binary` | Binary transformation with tiled processing | Yes |
+| `for_each` | In-place transformation (same tensor for input/output) | Yes |
+| `transform_parallel` | Parallel `transform` across multiple AIE tiles | No |
+| `transform_parallel_binary` | Parallel `transform_binary` across multiple AIE tiles |  No |
```
jgmelber: Is anything blocking extending this to support `ExternalFunctions`?

jgmelber (approved):

programming_examples/algorithms/README.md
yenjames: No not necessarily. It's just the current way to support `ExternalFunctions` is not final nor ideal imo and I figured showing a "flavour" of it in the non-parallel transforms. If you prefer I can make the same set of assumptions for the `ExternalFunction` format and apply them to parallel ones.

jgmelber: I think it is a good idea to apply the assumptions uniformly and document them clearly.

programming_examples/algorithms/transform_parallel_binary.py
```diff
@@ -0,0 +1,69 @@
+# transform_parallel_binary.py -*- Python -*-
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2024-2026 Advanced Micro Devices, Inc.
```
jgmelber:
```suggestion
# (c) Copyright 2026 Advanced Micro Devices, Inc.
```

programming_examples/algorithms/README.md
```diff
@@ -0,0 +1,68 @@
...
+| `transform_parallel_binary` | `void kernel(T* in1, T* in2, T* out, params..., int32_t tile_size)` | `[tile_ty, tile_ty, tile_ty, *param_types, np.int32]` |
+
+### Example: transform with ExternalFunction
+```python
+TILE_SIZE = 16
+tile_ty = np.ndarray[(TILE_SIZE,), np.dtype[np.int16]]
+scalar_ty = np.ndarray[(1,), np.dtype[np.int32]]
+
+# Input and output should still be declared as tiled shapes in arg_types
+# TODO: remove this discrepancy
```
hunhoffe: Is this TODO still a TODO? If it is, I think some thought should be given on where to document it and a longer comment might be warranted.

python/iron/algorithms/transform.py
```diff
@@ -0,0 +1,401 @@
...
+            raise ValueError(
+                f"Tensor {i} dtype {t.dtype} doesn't match expected {ref_dtype}"
+            )
+
+    num_inputs = len(inputs)
+    num_elements = np.size(inputs[0])
+    dtype = ref_dtype
+
+    # Determine number of columns based on device
+    num_columns = 4
```
hunhoffe: Using the device() object, you can fetch the number of available columns. I think this is better than hard-coding these values.

python/utils/jit.py
```diff
@@ -150,11 +153,41 @@ def decorator(*args, **kwargs):
...
+    Filter out non-tensor arguments from args. Required for algorithms because
+    they pass ExternalFunction and scalar values in their signature that should
+    not be interpreted as runtime sequence arguments.
+
+    Removes:
+    - ExternalFunction instances
+    - Scalar values (int, float, np.integer, np.floating), embedded as MLIR constants
+    """
+    import numpy as np
+    from aie.iron.kernel import ExternalFunction
```
hunhoffe: If this is to avoid circular imports, I understand. But to avoid import spaghetti I really try to keep all imports at the top of the module.

test/python/npu-xrt/test_algorithms.py
```diff
@@ -0,0 +1,446 @@
...
+# =============================================================================
+
+
+def test_transform_add():
+    """Test transform algorithm with simple add_one operation"""
+    input = iron.randint(0, 100, (1024,), dtype=np.int32, device="npu")
+    output = iron.zeros_like(input)
+    original = input.numpy().copy()
+    iron.jit(is_placed=False)(transform)(
+        lambda a: a + 1, input, output, tile_size=TILE_SIZE
```
hunhoffe, line 40: Pretty much all of these are +1 only. Would it make sense to parameterize the addend, e.g., test with a + n where n might be [1, 2, 3]

hunhoffe: I also wonder about whether signs are handled well, e.g., can you do n = -1 and have it work correctly? Might be good to try arange for a test like that, so make sure 0 -> -1 correctly, or something like that.

PR #2733, opened 2025-11-22
Device and ObjectFifo tests

hunhoffe:
This PR adds some helper functions that will be useful for channel counting in the SequentialPlacer. One of the helper functions already existed but what untested. I also realized there were no general objectfifo tests only in Python, so I added a few to provide light testing.

### DIFF
--- a/programming_guide/iron_configuration.md
+++ b/programming_guide/iron_configuration.md
@@ -8,7 +8,7 @@
 // 
 //===----------------------------------------------------------------------===//-->
 
-# IRON Python Binding Configurations
+# IRON Python Configurations
 
 There are several options that exist to configure the IRON Python programming environment.
 
@@ -42,4 +42,6 @@ If the IRON device is not set, many designs will try it fetch it on demand using
 <abc.NPU1 object at 0x722a65903a10>
 ```
 
+## IRON Cache Location
 
+The IRON jit feature caches compiled objects in a directory defined by ```IRON_CACHE_DIR```. By default this value is the user's home directory.
--- a/python/iron/compile/__init__.py
+++ b/python/iron/compile/__init__.py
@@ -6,5 +6,17 @@
 #
 # (c) Copyright 2025 Advanced Micro Devices, Inc.
 
-from .compile import compile_cxx_core_function, compile_mlir_module
+import os
+from pathlib import Path
+
 from .link import merge_object_files
+from .utils import (
+    compile_cxx_core_function,
+    compile_mlir_module,
+    compile_external_kernel,
+)
+
+# The `iron.compiledesign` decorator below caches compiled kenrels inside the `IRON_CACHE_HOME` directory.
+# Kernels are cached based on their hash value of the MLIR module string. If during compilation,
+# we hit in the cache, the `iron.jit` will load the xclbin and instruction binary files from the cache.
+IRON_CACHE_HOME = os.environ.get("IRON_CACHE_HOME", Path.home() / ".iron" / "cache")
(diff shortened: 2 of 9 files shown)


python/iron/device/device.py
```diff
@@ -191,14 +189,40 @@ def get_num_connections(self, tile: Tile, output: bool) -> int:
         else:
             return self.get_num_dest_switchbox_connections(tile)
 
-    def is_legal_mem_affinity(self, src_tile: Tile, dst_tile: Tile) -> bool:
+    def is_legal_mem_affinity(self, *tiles: Tile) -> bool:
```
fifield: How is this the same/different than the target model API of the same name which it wraps? Is there something missing in the C++ version which could be added there to make this a simple(r) wrapper?

hunhoffe: Andra particularly mentioned that this function may not do what we want it to do because it's used in very particular ways within the compiler. If the sequential placer is going to use this Python wrapper for channel counting, I needed to make sure the function did what I thought it did.

I think maybe this makes sense for now, but could be revisited if the SequentialPlacer is moved to the compiler?

python/iron/device/device.py
```diff
@@ -191,14 +189,40 @@ def get_num_connections(self, tile: Tile, output: bool) -> int:
...
         """Returns whether memory on a destination can be accessed by a source.
         Returns:
             int: Number of connections (channels) available on the tile
```
andrej: Please update this documentation. I don't understand what it means when this function takes multiple tiles. They must all be able to access each other's memory? Also returns a bool not int.

python/iron/device/device.py
```diff
@@ -191,14 +189,40 @@ def get_num_connections(self, tile: Tile, output: bool) -> int:
...
+        if len(tiles) == 1:
+            return True
+
+        first_tile = tiles[0]
+        is_first_compute = self._tm.is_core_tile(first_tile.col, first_tile.row)
+        is_first_mem = self._tm.is_mem_tile(first_tile.col, first_tile.row)
+
+        for i in range(len(tiles) - 1):
+            src_tile = tiles[i]
+            dst_tile = tiles[i + 1]
```
andrej: Trying to understand the semantics. If we give tiles A, B, C, D as input, is the function supposed to check if A can access B's memory, then B can access C's memory, and so forth, as a chain?

Maybe some context where this is used would help clear things up for me.

python/iron/device/device.py
hunhoffe: I wanted to keep consistency with the target model, but really I should rename this, "are neighbors" or "have common neighbor" or something. Anyways, the documentation is updated :)

python/iron/device/device.py
hunhoffe: No, it's if there is one tile from which all other tiles can access.  Basically, this function exists to more programmatically help with estimating channel counts for object fifos in the placer. The integration is for a subsequent PR.

Currently, the SequentialPlacer tries to estimate channels without understanding how neighbors work, which is not great. Even if this isn't perfect, it should lead to at least a slightly more disciplined heuristic than what is currently used.

andrej: Thanks, that makes a lot of sense.

andrej (approved):

python/iron/device/device.py
fifield: > this function may not do what we want it to do because it's used in very particular ways within the compiler

That doesn't really answer my question. What's different in the python version vs. the C++? I would like to understand why they can't be combined. Can we please at least document the python to explain the duplication?

hunhoffe: I do think your point about code deduplication and diverging implementations (C++ impl vs Python Impl) is good and should not be forgotten. I've written more on my methodology in approaching this work and also included the deduplicating task as a TODO in this issue: https://github.com/Xilinx/mlir-aie/issues/2738

Does this address the concern?

Issue #2572, opened 2025-09-10
Make `IRON_CACHE_DIR` an environment variable


AndraBisca:
[This](https://github.com/Xilinx/mlir-aie/blob/main/python/iron/jit.py#L29) line in iron/jit.py sets the path for the JIT cache:
`IRON_CACHE_DIR = os.path.expanduser("~/.iron/cache")`

It would be nice to have the ability to change this without having to modify the code directly.

mawad-amd:
This was added in #2611. Any suggestions for where to document environment variables? Once documented, I will close this. @AndraBisca @jgmelber 

https://github.com/Xilinx/mlir-aie/blob/69ffd7a153de7eab5b397216057ba04bb420dad4/python/iron/jit.py#L26-L29

hunhoffe:
@mawad-amd To add my two cents, I think it's ok that this is an environment variable. As for documentation, I am thinking that we need to add an "IRON Configuration" section to the README and lists environment variables.

I don't see mentions of jit in the programming guide, so I don't think it needs to be added there unless we add a jit section there too. 

Thoughts @AndraBisca and @jgmelber ?

hunhoffe:
Configuration documentation was added and merged in (found here: https://github.com/Xilinx/mlir-aie/blob/main/programming_guide/iron_configuration.md).

PR #2736, opened 2025-11-24
IRON Environment Var Documentation; minor code restructure.

hunhoffe:
This PR introduces no new functionality, but instead just moves code around into a slightly more coherent structure. This is one small part of a series of PRs that are planned; the goal is to break them into small pieces.

In addition to moving the code around, this code also makes the ```IRON_CACHE_HOME``` variable default location more platform independent and also documents the existence of this variable in ```programming_guide/iron_configuration.md```. It is my hope this addresses the last remaining TODO of https://github.com/Xilinx/mlir-aie/issues/2572.

### DIFF
--- a/python/iron/hostruntime/jit.py
+++ b/python/iron/hostruntime/jit.py
@@ -9,7 +9,6 @@
 import os
 import functools
 import hashlib
-import shutil
 import fcntl
 import contextlib
 import time


python/iron/compile/__init__.py
```diff
@@ -6,5 +6,13 @@
...
+import os
+from pathlib import Path
+
 from .link import merge_object_files
+from .utils import compile_cxx_core_function, compile_mlir_module
+
+# The `iron.compiledesign` decorator below caches compiled kenrels inside the `IRON_CACHE_HOME` directory.
+# Kernels are cached based on their hash value of the MLIR module string. If during compilation,
+# we hit in the cache, the `iron.jit` will load the xclbin and instruction binary files from the cache.
+IRON_CACHE_HOME = os.environ.get("IRON_CACHE_HOME", Path.home() / ".iron" / "cache")
```
andrej, line 22: I believe when the environment variable is set, the type of this will be `str`. If it's not, the default value has  type `Path`. If you haven't already, please double-check that this won't cause issues.

andrej (commented):
The changeset shows a lot of lines/classes/functions added (e.g. `CircularBuffer`, `NPUKernel`). Perhaps you forgot to check in some files you deleted?

Edit: Looks like you just fixed that, having a look now.

python/iron/compile/utils.py
```diff
@@ -22,11 +24,9 @@ def compile_cxx_core_function(
 ):
     """
     Compile a C++ core function.
-
     This function supports only the Peano compiler.
-
     Parameters:
-        source_path (str): Path to C++ source.
+        source_path (list[str]): Path to C++ source.
```
andrej: Is it really a list?

python/iron/compile/__init__.py
hunhoffe: I believe this is only used by `os.path.join`, which does understand PosixPath objects -- but good call to double check!

From my local test:
```
>>> IRON_CACHE_HOME2 = os.environ.get("NOT_REAL", Path.home() / ".iron" / "cache")
>>> type(IRON_CACHE_HOME2)
<class 'pathlib.PosixPath'>
>>> IRON_CACHE_HOME2
PosixPath('/home/ehunhoff/.iron/cache')
>>> os.path.join(IRON_CACHE_HOME2, "hello")
'/home/ehunhoff/.iron/cache/hello'
```

python/iron/compile/utils.py
hunhoffe: Fixed, thanks!

PR #2737, opened 2025-11-25
IRON host runtime abstraction

hunhoffe:
# Summary
This PR begins the process of consolidating runtime support code into one implementation that supports tracing, JIT, programming examples, and tests.

Apologies that the PR is so large, but many components of the repo were good candidates to use shared runtime library functions!

# Details

## The primary goals of this work are to:
1. Deduplicate logic between the JIT runtime code (e.g., the NPUKernel class in IRON), the XRT helper code (e.g., the ```AIE_Application``` class in the ```aie.utils.xrt``` module) and the amd/IRON runtime code (e..g, the  AIEDeviceManager class in https://github.com/amd/IRON/blob/devel/applications/llama_3.2_1b/src/aie_device_manager.py). This will help maintainability and ensure improvements made for efficiency (e.g., in pre-loading, buffer handling, etc.) are consistently used.
2. Continue abstracting specifics of XRT away from the conceptual use of a runtime. This is a follow on to a previous PR or two on ```iron.Tensors```. 
3. Bring tracing capabilities into JIT
4. Caching for insts buffer objects and XRT contexts. This is not fully optimized.

## Other miscellaneous things included in this PR:
* Rename some pytest files in test/python to follow the traditional ```test_*.py``` naming convention. Make sure lit test files are not named with this convention. This allows one to run ```pytest test/python``` and successfully run all the pytests in that directory. This runs locally much faster than trying to use lit.
* Some small steps towards consolidating information on tracing into a ```TraceConfig``` object. Hopefully this will complement and not conflict with https://github.com/Xilinx/mlir-aie/pull/2705
* Generally, take some small steps towards having python/utils be a separate package (including runtime support) with separate dependencies from python/iron. This process is not complete, but started.
* Truncate some output to only show 100 errors, to avoid overloading the CI runners with output
* Reduce the lit test concurrency by introducing a new concurrency group for tests that run on the npu using xrt in test/npu-xrt and test/python/npu-xrt.
* If xrt is not present, you can still generate MLIR. A nice warning is printed to stderr, addressing: https://github.com/Xilinx/mlir-aie/issues/2606

## What this PR does NOT do:
* The XRT runtime is not particularly optimized. Correctness was emphasized first in this PR. Later work towards optimization should analyze the syncing behavior of the runtime + tensors to ensure no extra overheads are incurred.
* Make changes to compilation, configuration, etc. But, hopefully, with one runtime entrypoint, these changes will be easier to make in future PRs. Similar to refactoring for making a more cohesive tracing API.
* Fully integrate tracing utilities into a library format with transparent support for difference architectures.
* Add a good test suite for tracing utilities.

### DIFF
--- a/python/utils/hostruntime/hostruntime.py
+++ b/python/utils/hostruntime/hostruntime.py
@@ -4,6 +4,7 @@
 from abc import ABC, abstractmethod
 import numpy as np
 from pathlib import Path
+import sys
 from typing import TYPE_CHECKING
 
 from .. import tensor
@@ -93,6 +94,21 @@ class KernelResult(ABC):
 
 class HostRuntime(ABC):
     """An abstract class for a generic host runtime"""
+
+    def check_device_consistency(self):
+        """
+        Check if the overridden device matches the runtime device.
+        """
+        mod = sys.modules[__package__]
+        override = getattr(mod, "_CURRENT_DEVICE", None)
+        if override:
+            runtime_device = self.device()
+            if getattr(override, "_device", None) != getattr(
+                runtime_device, "_device", None
+            ):
+                raise RuntimeError(
+                    f"Overridden device {override} does not match runtime device {runtime_device}"
+                )
 
     @abstractmethod
     def load(self, npu_kernel: NPUKernel, **kwargs) -> KernelHandle:
--- a/python/utils/hostruntime/xrtruntime/hostruntime.py
+++ b/python/utils/hostruntime/xrtruntime/hostruntime.py
@@ -126,6 +126,7 @@ class XRTHostRuntime(HostRuntime):
         Raises:
             HostRuntimeError: If xclbin or insts files do not exist, or if kernel is not found.
         """
+        self.check_device_consistency()
         xclbin_path = Path(npu_kernel.xclbin_path).resolve()
         insts_path = Path(npu_kernel.insts_path).resolve()
         kernel_name = npu_kernel.kernel_name
@@ -188,6 +189,7 @@ class XRTHostRuntime(HostRuntime):
         Raises:
             HostRuntimeError: If arguments are invalid or kernel execution fails (and fail_on_error is True).
         """
+        self.check_device_consistency()
         # Filter out callable functions and check arg types
         args = [a for a in args if not callable(a)]
         if not all([isinstance(a, self._tensor_class) for a in args]):
@@ -423,6 +425,7 @@ class CachedXRTRuntime(XRTHostRuntime):
         Raises:
             HostRuntimeError: If xclbin or insts files do not exist, or if kernel is not found.
         """
+        self.check_device_consistency()
         xclbin_path = Path(npu_kernel.xclbin_path).resolve()
         insts_path = Path(npu_kernel.insts_path).resolve()
         kernel_name = npu_kernel.kernel_name
--- a/test/python/npu-xrt/test_device_override.py
+++ b/test/python/npu-xrt/test_device_override.py
@@ -36,3 +36,31 @@ def test_device_override():
     # If not, it returns None.
     # In both cases, it should match original_device (assuming runtime didn't change/disappear)
     assert type(utils.get_current_device()) == type(original_device)
+
+
+def test_device_consistency():
+    # Define a mock runtime class
+    class MockRuntime(utils.HostRuntime):
+        def load(self, *args, **kwargs):
+            pass
+
+        def run(self, *args, **kwargs):
+            pass
+
+        def device(self):
+            return NPU1()
+
+    runtime = MockRuntime()
+
+    # Set compatible override
+    utils.set_current_device(NPU1())
+    # Should not raise
+    runtime.check_device_consistency()
+
+    # Set incompatible override
+    utils.set_current_device(NPU2())
+    with pytest.raises(RuntimeError, match="does not match runtime device"):
+        runtime.check_device_consistency()
+
+    # Reset
+    utils.set_current_device(None)


python/utils/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,158 @@
...
+            self._device_type = NPU2
+        elif any(
+            keyword in device_type_str
+            for keyword in [
+                "NPU",
+                "NPU Phoenix",
+                "RyzenAI-npu1",
+            ]
+        ):
+            self._device_type = NPU1
```
ypapadop-amd: Maybe have this as a stand-alone function?

python/iron/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,158 @@
...
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
+    def __init__(self):
+        # TODO: what is there is more than one device?
+        self._device = pyxrt.device(0)
+        device_type_str = self._device.get_info(pyxrt.xrt_info_device.name)
+
+        # Fetch the device type by matching strings for NPU2 or NPU1
+        # TODO: how to use only a portion of the device rather than whole array?
```
ypapadop-amd: You could return a tuple of device and max columns to let users know what is the limit.

hunhoffe: In my mind, you can ask the device object returned by ```runtime.device()``` for things like columns and rows, e.g., https://github.com/Xilinx/mlir-aie/blob/a81f6822022bbf526bdb1a99f319418563b66468/python/iron/device/device.py#L94

So maybe this doesn't need to be a TODO. Thoughts?

python/utils/xrtruntime/hostruntime.py
hunhoffe: I believe this function will go away before this PR is finished because this is information we can fetch from the compiler about the device/targetmodel.

I think the enum AIEArch is the key, but I haven't fully finished the implementation of this yet. Thoughts @fifield ?

fifield: > I believe this function will go away before this PR is finished because this is information we can fetch from the compiler about the device/targetmodel.
> 
> I think the enum AIEArch is the key, but I haven't fully finished the implementation of this yet. Thoughts @fifield ?

You could certainly use `AIEArch` or `AIEDevice` enums here. I don't have much else to add here other than this basic logic is already is in iron device.py and in aie_lit_utils. Hopefully the string names of things are stable by now, but the fewer places it's parsed the better.

python/iron/hostruntime/xrtruntime/hostruntime.py
ypapadop-amd: Do I need to instantiate XRT to ask that?

python/utils/xrtruntime/hostruntime.py
ypapadop-amd: How does the compiler know the device, esp. if there is no device present?

python/iron/hostruntime/xrtruntime/hostruntime.py
hunhoffe: The device target model is defined in the aiecc compiler code, I believe. I do not think it is dependent on XRT, although I have never tried it -- if you have a different experience, let me know!

python/utils/xrtruntime/hostruntime.py
hunhoffe: At the level of an iron program, you can instantiate a device object in Python based on the aie C++ target model code and use it for generating the MLIR. My current understanding is that the device type actually needs to be present for the MLIR generation, not just the compilation process; aiecc just parses the device name from the mlir-aie device operation. So, for the compiler, the device type is embedded in the input file and not something parsed from the system.

I haven't fully untangled how I want to set defaults in the JIT infrastructure yet as this is just a work in progress and I am not yet introducing a "Compilable" or "Runnable" abstractions yet (which will eventually fit beneath the JIT frontend).

I'm open to suggestions!

python/iron/hostruntime/xrtruntime/hostruntime.py
ypapadop-amd: I'm not sure. Right now, I have hardcoded the number of columns based on the device type but I'd prefer if IRON has that knowledge.

python/utils/xrtruntime/hostruntime.py
ypapadop-amd: OK, that makes sense.

python/iron/hostruntime/xrtruntime/hostruntime.py
fifield: The target model bindings only depend on the dialect bindings, not xrt. 

```python
>>> from aie.dialects.aie import get_target_model, AIEDevice
>>> get_target_model(AIEDevice.npu1).columns()
4
>>> get_target_model(AIEDevice.npu1_1col).columns()
1
>>> 
```

hunhoffe:
If I don't want to accidentally destroy the tracing utils with my consolidation, I think I need to do this PR first: https://github.com/Xilinx/mlir-aie/pull/2743

python/iron/hostruntime/xrtruntime/hostruntime.py
hunhoffe: For the record -- in case someone uses this PR as a reference in the future -- this can also be done in the IRON bindings in a more object-oriented fashion with:
```python
>>> from aie.iron.device import NPU1, NPU2
>>> dev1 = NPU1()
>>> dev1.cols
4
>>> dev1.rows
6
>>> dev2 = NPU2()
>>> dev2.cols
8
>>> dev2.rows
6
```

python/utils/xrtruntime/hostruntime.py
hunhoffe: I didn't end up trying to remove this completely, since some of the duplication of this logic (scripts, python lit config, python utils) have some complex dependencies. I wasn't sure the best place to place shared logic, but I'm happy to revisit this in a future PR.

programming_examples/basic/vector_vector_add_BDs_init_values/test.cpp
```diff
@@ -140,12 +140,14 @@ int main(int argc, const char *argv[]) {
...
-                << std::endl;
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << *(bufOut + i)
+                  << " != " << *(bufInA + i) << " + " << *(bufInB + i)
+                  << std::endl;
+      }
       errors++;
     } else {
-      if (verbosity > 1)
+      if (verbosity >= 1)
```
andrej, line 150: Might produce a lot of output for passing runs?

programming_examples/basic/vector_vector_modulo/test.cpp
```diff
@@ -140,12 +140,14 @@ int main(int argc, const char *argv[]) {
...
-                << std::endl;
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << *(bufOut + i)
+                  << " != " << *(bufInA + i) << " + " << *(bufInB + i)
+                  << std::endl;
+      }
       errors++;
     } else {
-      if (verbosity > 1)
+      if (verbosity >= 1)
```
andrej, line 150: Might produce a lot of output for passing runs?

programming_examples/basic/vector_vector_mul/test.cpp
```diff
@@ -140,14 +140,15 @@ int main(int argc, const char *argv[]) {
...
+        std::cout << "Error in output " << *(bufOut + i)
+                  << " != " << *(bufInA + i) << " * " << *(bufInB + i)
+                  << std::endl;
+      }
       errors++;
-    } else {
-      if (verbosity > 1)
-        std::cout << "Correct output " << *(bufOut + i)
-                  << " == " << *(bufInA + i) * *(bufInB + i) << std::endl;
+    } else if (verbosity >= 1) {
```
andrej, line 149: Might produce a lot of output for passing runs vs. verbosity > 1?

programming_examples/ml/block_datatypes/bfp_conversion/test.cpp
```diff
@@ -209,11 +209,13 @@ int main(int argc, const char *argv[]) {
...
-      std::cout << "Error in output " << outputTransformed[i]
-                << " != " << expectedResultVec[i] << std::endl;
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << outputTransformed[i]
+                  << " != " << expectedResultVec[i] << std::endl;
+      }
       errors++;
     } else {
-      if (verbosity > 1)
+      if (verbosity >= 1)
```
andrej, line 218: Might produce a lot of output for passing runs?

programming_examples/ml/block_datatypes/vector_passthrough/test.cpp
```diff
@@ -162,11 +162,13 @@ int main(int argc, const char *argv[]) {
...
-      std::cout << "Error in output " << int(*(bufOut + i))
-                << " != " << int(*(bufInA + i)) << std::endl;
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << int(*(bufOut + i))
+                  << " != " << int(*(bufInA + i)) << std::endl;
+      }
       errors++;
     } else {
-      if (verbosity > 1)
+      if (verbosity >= 1)
```
andrej, line 171: Might produce a lot of output for passing runs?

programming_examples/ml/eltwise_add/test.cpp
```diff
@@ -40,8 +40,10 @@ int verify(int size, std::vector<T> A, std::vector<T> B, std::vector<T> C,
   for (uint32_t i = 0; i < size; i++) {
     T ref = A[i] + B[i];
     if (!test_utils::nearly_equal(ref, C[i], 0.00390625)) {
-      std::cout << "Error in output " << C[i] << " != " << ref << " from "
-                << A[i] << " + " << B[i] << std::endl;
+      if (errors < 100) {
+        std::cout << "Error in output " << C[i] << " != " << ref << " from "
+                  << A[i] << " * " << B[i] << std::endl;
+      }
```
andrej, line 49: Might want to add "... [output truncated]" if errors >= 100 at the end of the loop.

programming_examples/ml/scale_shift/test.cpp
```diff
@@ -41,11 +41,13 @@ int verify(int size, std::vector<T> A, std::vector<T> B, std::vector<T> C,
...
-      std::cout << "Error in output " << D[i] << " != " << ref << " from "
-                << A[i] << " * " << B[i] << " + " << C[i] << std::endl;
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << D[i] << " != " << ref << " from "
+                  << A[i] << " * " << B[i] << " + " << C[i] << std::endl;
+      }
       errors++;
     } else {
-      if (verbosity > 1)
+      if (verbosity >= 1)
```
andrej, line 50: Might produce a lot of output for passing runs?

programming_examples/ml/softmax/test.cpp
```diff
@@ -64,13 +64,14 @@ int verify(int size, int tile_size, std::vector<T> A, std::vector<T> B,
...
-      errors++;
-    } else {
-      if (verbosity > 1)
-        std::cout << "Correct output " << B[i] << " == " << RefVec[i]
+      if (verbosity >= 1) {
+        std::cout << "Error in output " << B[i] << " != " << RefVec[i]
                   << std::endl;
+      }
+      errors++;
+    } else if (verbosity >= 1) {
```
andrej, line 72: Might produce a lot of output for passing runs?

python/utils/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    return DEFAULT_TENSOR_CLASS.rand(*args, **kwargs)
+
+
+def arange(*args, **kwargs):
+    return DEFAULT_TENSOR_CLASS.arange(*args, **kwargs)
+
+
+def zeros_like(*args, **kwargs):
+    return DEFAULT_TENSOR_CLASS.zeros_like(*args, **kwargs)
+
```
andrej, line 128: Combinatorial explosion when we start to support more methods on tensors and potentially more tensor types? Don't know if there is a solution to this.

python/utils/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+if has_xrt:
+    from .hostruntime.xrtruntime.hostruntime import CachedXRTRuntime
+
+    DEFAULT_NPU_RUNTIME = CachedXRTRuntime()
+else:
+    DEFAULT_NPU_RUNTIME = None
+
+
+def get_current_device():
+    return DEFAULT_NPU_RUNTIME.device()
```
andrej:
```suggestion
def get_current_device():
    return DEFAULT_NPU_RUNTIME.device() if DEFAULT_NPU_RUNTIME else None
```

python/utils/compile/__init__.py
```diff
@@ -16,7 +16,9 @@
     compile_external_kernel,
 )
 
-# The `iron.compiledesign` decorator below caches compiled kenrels inside the `IRON_CACHE_HOME` directory.
+# Compiled kenrels are cached inside the `IRON_CACHE_HOME` directory.
 # Kernels are cached based on their hash value of the MLIR module string. If during compilation,
```
andrej: Is this comment still accurate? I see code below that hashes Python code but it might be unrelated.

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+class HostRuntimeError(Exception):
+    """
+    Error raised when a NPU kernel encounters an error during runtime operations.
+    """
+
+    pass
+
+
+class KernelHandle(ABC):
+    """KernelHandles may be used a cache keys, and so should implement these methods."""
```
andrej:
```suggestion
    """KernelHandles may be used as cache keys, and so should implement these methods."""
```

andrej: "so should implement these methods" but the class has no required methods?

andrej: Also could you state which cache. I don't think it's the JIT artifact cache but the XRT context cache

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+    @abstractmethod
+    def device(self) -> "Device":
+        pass
+
+    @classmethod
+    def read_insts_sequence(cls, insts_path: Path):
+        """Reads instructions from a text file (hex numbers, one per line)."""
+        with open(insts_path, "r") as f:
+            insts_text = f.readlines()
+        insts_text = [l for l in insts_text if l != ""]
```
andrej: Method appears to not return anything and has no visible side effects as far as I can tell. I think you want to parse the hex values `data = [int(l,16) for l in insts_text]` and `return np.frombuffer(data ...)` or remove the method altogether if unneeded.

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+                        f"Error: Reference index {idx} out of bounds for {len(io_args)} IO buffers"
+                    )
+                    return 1
+                io_args[idx].to("cpu")
+                o = io_args[idx].numpy()
+                e = np.equal(r, o)
+                errors += np.size(e) - np.count_nonzero(e)
+            else:
+                print("Error: Reference data must be a list of (index, data) tuples")
+                return 1
```
andrej: Making `ref` a dictionary {idx: reference_buffer} makes this check redundant and prevents users from specifying multiple reference buffers from the same index. In my opinion also looks sleeker.

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+
+    @classmethod
+    def verify_results(cls, io_args, ref, verbosity=0):
+        errors = 0
+        if verbosity >= 1:
+            print("Verifying results ...")
+
+        # Handle ref being list or single
+        if not isinstance(ref, list):
+            ref = [ref]
```
andrej: With a dict this won't be necessary and is just as short (`(0, ref)` vs `{0: ref}`)

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+            if isinstance(item, tuple) and len(item) == 2:
+                idx, r = item
+                if idx >= len(io_args):
+                    print(
+                        f"Error: Reference index {idx} out of bounds for {len(io_args)} IO buffers"
+                    )
+                    return 1
+                io_args[idx].to("cpu")
+                o = io_args[idx].numpy()
+                e = np.equal(r, o)
```
andrej: Might want to use `np.allclose` for float data types here (or allow user to pass in an equality comparator function/lambda for full flexibility?)

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+            return cls.read_insts_sequence(insts_path)
+        else:
+            raise HostRuntimeError(
+                "Unsupported file extension for instruction file: expected .bin or .txt"
+            )
+
+    @classmethod
+    def prepare_args_for_trace(
+        cls, args: list[Tensor], trace_config: TraceConfig
+    ) -> list[Tensor]:
```
andrej, line 240: Would it make sense to have `trace_config` be part of the `NPUKernel`? Whether the trace buffer is expected as the last argument, with dummy arguments in between, or suffixed after the output buffer, etc., is a property of the kernel itself. Once it's compiled that way, you can't call it with another config. Having to pass it to each of these methods seems like a potential source for error (give a different `trace_config` at kernel-generation time vs. at runtime). Or am I misunderstanding?

More generally, do we have an abstraction that tells us what kind of buffers an `NPUKernel` expects as input (number of buffers, sizes, data types) that would allow us to type-check at call-time whether all required inputs have been supplied?

python/utils/hostruntime/hostruntime.py
```diff
@@ -0,0 +1,310 @@
...
+            else:
+                print("Error: Reference data must be a list of (index, data) tuples")
+                return 1
+        return errors
+
+    def run_test(
+        self,
+        io_args,
+        ref,
+        npu_kernel,
```
andrej: nit: I think I'd visually prefer having `npu_kernel` as first argument (specify what you call/run first, then its arguments).

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+            )
+        if not insts_path.exists() or not insts_path.is_file():
+            raise HostRuntimeError(
+                f"insts {insts_path} does not exist or is not a file."
+            )
+
+        xclbin = pyxrt.xclbin(str(xclbin_path))
+        self._device.register_xclbin(xclbin)
+        xclbin_uuid = xclbin.get_uuid()
+        context = pyxrt.hw_context(self._device, xclbin_uuid)
```
andrej: I'm not sure about exact performance implications but I know there's a limited number of hardware contexts, so it would be nice to only load the xclbin and set up the hardware context once if we use multiple kernels out of one xclbin, like [here](https://github.com/amd/IRON/blob/a4b6ffeac0c12260cedc74104e95e81271f78608/applications/llama_3.2_1b/src/aie_device_manager.py#L40).

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+    """
+    A cached version of XRTHostRuntime that caches up to n contexts,
+    depending on the type of NPU.
+    It reuses contexts for the same xclbin (identified by path and mtime).
+    """
+
+    def __init__(self):
+        super().__init__()
+        self._context_cache = OrderedDict()
+        self._insts_cache = OrderedDict()
```
andrej, line 321: Just so you know since Python 3.7 all (regular) dicts are ordered too

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+        if self.npu_str == "npu1":
+            self._cache_size = NPU1_CACHE_SIZE
+        else:
+            self._cache_size = NPU2_CACHE_SIZE
+
+        env_cache_size = os.environ.get("XRT_CONTEXT_CACHE_SIZE")
+        if env_cache_size is not None:
+            self._cache_size = min(self._cache_size, int(env_cache_size))
```
andrej: Not sure taking the min is the most intuitive from a user perspective. Maybe the environment variable should just take precedence?

```suggestion
        self._cache_size = os.environ.get("XRT_CONTEXT_CACHE_SIZE") or self._cache_size
```

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+        kernel_name = npu_kernel.kernel_name
+
+        if not xclbin_path.exists() or not xclbin_path.is_file():
+            raise HostRuntimeError(
+                f"xclbin {xclbin_path} does not exist or is not a file."
+            )
+        if not insts_path.exists() or not insts_path.is_file():
+            raise HostRuntimeError(
+                f"insts {insts_path} does not exist or is not a file."
+            )
```
andrej, line 440: nit: TOCTOU and I don't really know if these checks add much over the exception that would be thrown just from trying to open the file when it doesn't exist?

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+
+        xclbin_mtime = xclbin_path.stat().st_mtime
+        insts_mtime = insts_path.stat().st_mtime
+
+        # Context Cache Lookup
+        context_key = (str(xclbin_path), xclbin_mtime, str(insts_path), insts_mtime)
```
andrej: I don't think the `insts_path` needs to be part of the context key. You're not caching the insts.bin and I think it should be fine to share the same context for multiple invocations of the same xclbin with different instruction sequences.

python/utils/npukernel.py
```diff
@@ -0,0 +1,50 @@
...
+        return self._kernel_name
+
+    # Blocking call.
+    def __call__(self, *args):
+        from . import DEFAULT_NPU_RUNTIME
+
+        return DEFAULT_NPU_RUNTIME.load_and_run(
+            self,
+            list(args),
+            trace_config=self._trace_config,
```
andrej: Any reason load_and_run shouldn't just access trace config from the `self` we're passing in as the first arg?

python/utils/npukernel.py
```diff
@@ -0,0 +1,50 @@
...
+
+class NPUKernel:
+    def __init__(
+        self,
+        xclbin_path,
+        insts_path,
+        device_index=0,
+        kernel_name="MLIR_AIE",
+        trace_config: TraceConfig | None = None,
+    ):
```
andrej, line 24: Doesn't need to be addressed in this PR, but just wondering out loud, is the intent for this when I want to call the same xclbin with different runtime sequences (or, even a runtime-generated runtime sequence) to just create multiple `NPUKernel` instances with the same `xclbin_path` and let the caching of the XRT host runtime take care of only loading the xclbin once and reusing the same context? Related, have you thought about how this abstraction will work for the full ELF flow (which unifies xclbin, with potentitally multiple kernels, and potentially multiple instruction sequences)?

python/utils/trace/utils.py
andrej, line 1: Did you really add all this or is the diff just unhelpful?

andrej (approved):
Thank you for this, it'll be good to have just one canonical way to invoke NPU kernels. Please feel free to ignore almost all of my comments, I just stream-of-consciousness commented what I noticed when I saw it. Some clarifying questions in there for my understanding as well.

python/utils/trace/utils.py
hunhoffe: Diff is unhelpful. I definitely didn't add much of anything new for tracing!

python/utils/npukernel.py
hunhoffe: Good questions! My intent for this is:
* It already existing in JIT, and I didn't see the point of removing it
*  I see it as a less-sophisticated shadow of what we consider to be the AIE_Base/Operation in amd/IRON. 
* I expect this will be deprecated in favor of some sore of operation class.

I didn't specifically account for all of the capabilities of the ELF flow, but I did integrate some features into the underlying host runtime that will help:
* Instructions and xclbins are cached in separate caches, so multiple instruction sequences is not a problem from that level
* Similarly, the kernel name is not a part of the cache key, so multiple kernels per cached xclbin is not a problem

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: Error errors are a good habit.

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: Ah yes, you are right! I cached the instructions separately, so I'll try to change this.

python/utils/npukernel.py
hunhoffe: Oh yeah. That would make sense -- nice catch! I will submit a change.

programming_examples/ml/block_datatypes/vector_passthrough/test.cpp
hunhoffe: The test.cpp files are completely inconsistent about using >= 1 or > 1. I arbitrarily unified them to one representation. This is add-hoc anyways, if  you want to control logging in C++ programmatically with different levels we should be using a logging library anyways.

programming_examples/ml/eltwise_add/test.cpp
hunhoffe: Will do!

python/utils/__init__.py
hunhoffe: Your point is quite true, but this is from a previous PR, and so is a bit out of scope. I think we can programmatically do overrides using python namespace hackery but this is more readable for now. The goal is also not to completely reimplement `torch.Tensor` so there is some motivation to avoid scope creep here.

python/utils/compile/__init__.py
hunhoffe: Yes. I didn't make any changes to JIT compilation (this is a completely different cache to the runtime context cache). Those are hashes to python lambda functions, which I think we chatted about elsewhere.

python/utils/hostruntime/hostruntime.py
hunhoffe: Whoops, this comment was out of date. Nice catch! I will fix it.

python/utils/hostruntime/hostruntime.py
hunhoffe: This is a good thing to think about. The link between how tracing is expressed in a design, the compiled artifact, and the runtime configuration at this point does not exist. Hopefully it will be easier to define this in the future, but right now improvements to the tracing API are not priority.

To your question: nope! Future work for sure!

python/utils/hostruntime/hostruntime.py
hunhoffe: This is a good idea, I will do this.

python/utils/hostruntime/hostruntime.py
hunhoffe: Dict is a good idea, will do!

python/utils/hostruntime/hostruntime.py
hunhoffe: Fair point!

python/utils/compile/__init__.py
andrej: But if they're hashes to Python lambda functions, they're not hashes of the MLIR module string as stated in the comment, no? Or is that also hashed elsewhere and I'm just mixing things up? Just to be clear, I think hashing Python is good because we can do it early without having to go through the Python-bindings-to-MLIR chain, just want to make sure I'm understanding correctly that that is what is actually happening.

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -133,6 +133,9 @@ jobs:
           rm -rf $IRON_CACHE_HOME
           mkdir $IRON_CACHE_HOME
 
+          # Set number of contexts to maintain in cache per process
+          export XRT_CONTEXT_CACHE_SIZE=3
```
mawad-amd, line 137: I suggest prefixing the environment variables with `IRON_` similar to `IRON_CACHE_DIR`. Maybe `IRON_XRT_CONTEXT_CACHE_SIZE` or just `IRON_CONTEXT_CACHE_SIZE`?

programming_examples/basic/passthrough_kernel/test.py
```diff
@@ -7,13 +7,13 @@
 # (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 import numpy as np
 import sys
-import aie.utils.xrt as xrt_utils
 import aie.utils.test as test_utils
+import aie.iron as iron
+from aie.utils import DEFAULT_NPU_RUNTIME
```
mawad-amd: why the upercase?

python/utils/compile/__init__.py
mawad-amd: The hash should include both MLIR and the lambda (and other things too and is still missing a preprocessor output to handle nested headers -- there is an issue for that). I wrote this comment but it might make sense to just remove the explanation of how we hash since it will become outdated and we don't need to explain it here.

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+        kernel_handle: XRTKernelHandle,
+        args,
+        trace_config=None,
+        fail_on_error: bool = True,
+    ) -> XRTKernelResult:
+        # Filter out callable functions and check arg types
+        args = [a for a in args if not callable(a)]
+        if not all([isinstance(a, self._tensor_class) for a in args]):
+            raise HostRuntimeError(
+                f"The {self.__class__.__name__} can only take {self._tensor_class.__name__} as arguments, but got: {args}"
```
mawad-amd, line 197: Tensors or callables?

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+import pyxrt
+
+from ..hostruntime import HostRuntime, HostRuntimeError, KernelHandle, KernelResult
+
+if TYPE_CHECKING:
+    from aie.iron.device import Device
+from .tensor import XRTTensor
+
+# I got these values through experimentation on two machines
+NPU1_CACHE_SIZE = 6
```
mawad-amd: This deserves a better comment to explain what this number is. FYI, even these numbers may cause issues for parallel builds (I saw tbe new env varaible above) since the number of contexts for all processes.
See chain of comments here https://github.com/Xilinx/mlir-aie/pull/2622#discussion_r2400292581

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+
+                # Try to create context, evicting if necessary
+                context = None
+                retries = 0
+                max_retries = len(self._context_cache)
+                while context is None:
+                    try:
+                        context = pyxrt.hw_context(self._device, xclbin_uuid)
+                    except RuntimeError as e:
+                        # If we hit a resource limit (err=-2 usually means EMFILE/ENFILE or similar resource exhaustion)
```
mawad-amd, line 475: Same comment as above. The cache could be empty and you would hit this error since the number of contexts is global per device.

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,388 @@
...
+                    )
+
+            insts = self.read_insts(insts_path)
+            insts_bo = None
+            if hasattr(pyxrt, "module") and isinstance(insts, pyxrt.module):
+                kernel = pyxrt.ext.kernel(context, insts, kernel_name)
+            else:
+                kernel = pyxrt.kernel(context, kernel_name)
+                # Handle insts caching
+                group_id = kernel.group_id(1)
```
mawad-amd, line 516: Add comment for magic number 1. IIRC, I wrote one explaining what it in the NPU kernel class.

test/python/npu-xrt/test_cached_xrt_runtime.py
```diff
@@ -0,0 +1,374 @@
...
+
+        # Run first kernel
+        transform(input_tensor, input_tensor, lambda x: x + 1)
+        assert len(runtime._context_cache) == 1
+        key1 = list(runtime._context_cache.keys())[0]
+
+        # Run second kernel (different lambda -> different xclbin)
+        transform(input_tensor, input_tensor, lambda x: x * 2)
+
+        assert len(runtime._context_cache) == 1
```
mawad-amd, line 169: The cache kicks out kernels if it runs into an out of context errors which may happen because of parallel runs. Wouldn't this test be flaky because of that?

test/python/npu-xrt/test_cached_xrt_runtime.py
```diff
@@ -0,0 +1,374 @@
...
+
+    for i in range(limit + 1):
+        transform(input_tensor, input_tensor, lambda x, val=i: x + val)
+
+        if i == 0:
+            first_key = list(runtime._context_cache.keys())[0]
+
+        # Check size
+        expected_size = min(i + 1, limit)
+        assert len(runtime._context_cache) == expected_size
```
mawad-amd, line 199: Same comment on cache size expectations.

test/python/npu-xrt/test_cached_xrt_runtime_insts.py
```diff
@@ -0,0 +1,209 @@
...
+    # Note: We can't easily check object identity of BOs if they are wrapped,
+    # but we can check if the entry is the same object.
+    assert entry1 is entry2
+
+
+def test_insts_initialization(runtime):
+    """Test that insts_bo is initialized during load."""
+
+    input_tensor = iron.tensor((32,), dtype=np.int32)
+    input_tensor[:] = np.arange(32, dtype=np.int32)
```
mawad-amd: `iron.arange` instead of the two lines? Other tests have the same pattern.

mawad-amd (approved):
The cache tests that check sizes will likely be flaky due to the parallel runs (see comment above). Aside from that, this looks great. Thanks, Erika!

python/utils/hostruntime/hostruntime.py
hunhoffe: Ah yes I think this function is no longer needed.

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: Heh! I just dated myself pretty good. Thanks!

.github/workflows/buildAndTestRyzenAI.yml
hunhoffe: I am removing IRON naming convention from utils, since I think these capabilities may/may not be more general.

programming_examples/basic/passthrough_kernel/test.py
hunhoffe: Fixed!

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: Tensors. There's no function chaining at this point (that I know of) so what passes between compiled kernels is all tensors.

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: I added the npu-xrt tests and the python tests that run things via xrt to a new lit test group which limits parallelism for just those tests. In addition, I limit the size of the cache.

There could be some issues with flakiness if very many github actions are run concurrently, but I believe what I have is enough to control the flakiness at the moment. Thoughts?

python/utils/hostruntime/xrtruntime/hostruntime.py
mawad-amd: Sorry, I meant the comment should be "Tensor or callable".

python/utils/hostruntime/xrtruntime/hostruntime.py
mawad-amd: I think CI has been stable from what I have seen. I was just suggesting adding a comment to explain the number/issue since it took us hours to understand the failures.

jgmelber (approved):
This looks great to me. No additional comments, other than I can see the time you put into this work!

.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -130,18 +130,21 @@ jobs:
...
           ninja check-aie
           ninja check-aie-concurrency
           popd
 
-      - name: Cleanup IRON_CACHE_HOME
+      - name: Cleanup NPU_CACHE_HOME
         if: always()
         run: |
-          rm -rf $IRON_CACHE_HOME
+          rm -rf $NPU_CACHE_HOME
```
ypapadop-amd, line 147: I think the cache is cleared above?

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,389 @@
...
+                del insts_bo
+
+        return XRTKernelResult(r, stop - start)
+
+    def device(self) -> "Device":
+        from aie.iron.device import NPU1, NPU2
+
+        if self.npu_str == "npu1":
+            return NPU1()
+        else:
```
ypapadop-amd, line 252: Should we assert if `npu_str` is not `npu2`?

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def __init__(self):
+        super().__init__()
+        # We use OrderedDict so that we can use Fifo behavior for LRU eviction policies
+        self._context_cache = OrderedDict()
+        self._insts_cache = OrderedDict()
+
+        if self.npu_str == "npu1":
+            self._cache_size = NPU1_CACHE_SIZE
+        else:
```
ypapadop-amd: Same here.

python/utils/hostruntime/__init__.py
```diff
@@ -0,0 +1,23 @@
+# __init__.py -*- Python -*-
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
```
ypapadop-amd:
```suggestion
# (c) Copyright 2026 Advanced Micro Devices, Inc.
```

python/utils/jit.py
```diff
@@ -130,28 +85,25 @@ def decorator(*args, **kwargs):
...
-            elif current_device in (AIEDevice.npu1, AIEDevice.npu1_1col):
-                target_arch = "aie2"
-            else:
-                raise RuntimeError(f"Unsupported device type: {type(current_device)}")
-        except Exception as e:
-            raise
+        current_device = DefaultNPURuntime.device()
+
+        # Determine target architecture based on device type
+        if isinstance(current_device, (NPU2, NPU2Col1)):
```
ypapadop-amd, line 95: For future additions, I vote that the arch becomes part of those `NPU*` objects.

test/npu-xrt/add_one_objFifo/test.py
```diff
@@ -12,143 +12,38 @@
...
-        return 0
+def main(opts):
+    ref_data = np.arange(1, IN_SIZE + 1, dtype=np.uint32)
+    inA = iron.tensor(ref_data, dtype=np.uint32)
+    inB = iron.tensor(ref_data, dtype=np.uint32)
+    out = iron.zeros((OUT_SIZE,), dtype=np.uint32)
+    ref_data = ref_data + 41
+
+    npu_opts = test_utils.create_npu_kernel(opts)
+    if not DefaultNPURuntime.run_test(
```
ypapadop-amd, line 34: I had to do a double take here in the check for `not` and then saying pass.

.github/workflows/buildAndTestRyzenAI.yml
hunhoffe: It's a bit over cautious but we clean up both before and after we run the tests, just to ensure we don't allow state to build up on the CI runners. If we move to docker CI system, this will likely be unnecessary.

python/utils/hostruntime/__init__.py
hunhoffe: Nice catch, thanks!

python/utils/jit.py
hunhoffe: Good suggestion!

programming_examples/basic/vector_scalar_mul/test.py
```diff
@@ -7,8 +7,9 @@
 # (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
```
ypapadop-amd: Copyright needs to be updated to a range in those.

ypapadop-amd:
I saw that the set_current_device test was removed. Is the function removed too? For some kernels, that's the only way to set the device.

ypapadop-amd:
I think that's undesirable for user code. Can we have a dummy runtime checked in as part of this PR that does that?

hunhoffe:
It is undesirable. If you want to create a dummy host runtime or other host runtime code to meet additional needs beyond what is used/tested in this repo, I think that would be an excellent follow on PR for you to make :)

hunhoffe:
@ypapadop-amd I re-added `set_current_device`. Let me know if you have any problems with the changes after this PR is merged in!

PR #2742, opened 2025-11-26
Use upstream eudsl-python-extras

hunhoffe:
`mlir-python-extras` has been upstreamed to `llvm/eudsl`. This PR updates the Python bindings to use those new packages, and use analogous package paths for dialect extensions for consistency based on some changes between the original `mlir-python-extras` organization and the new `eudsl-python-extras` structure.

This doesn't yet change the install method or package extras into the wheels.

It seems there were some weird errors I got in the MLIR-AIE Distro workflow. I can't think of any changes I made in this PR that would have broken it, but I did what was necessary to get it working again but I'm not sure of the root cause.

### DIFF
--- a/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
+++ b/lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
@@ -130,6 +130,68 @@ static SmallVector<Value> forceCastOperandsToSignature(OpBuilder &builder,
       }));
 }
 
+// Utility function to get or create a noinline scalar helper function.
+// This is used to create optimization barriers that prevent LLVM from
+// re-vectorizing unrolled scalar operations.
+//
+// Parameters:
+//   - module: The parent module to insert the function into
+//   - rewriter: The pattern rewriter
+//   - opName: Base name of the operation (e.g., "fdiv", "addf", "mulf")
+//   - device: Target device ("aie2", "aie2p", etc.)
+//   - argTypes: Input argument types
+//   - resultType: Return type
+//   - bodyBuilder: Lambda that builds the function body given (builder, loc,
+//   args)
+//
+// Returns: The helper function (created or existing)
+//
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -142,7 +142,7 @@ Turn off SecureBoot (Allows for unsigned drivers to be installed):
    python3 -m pip install llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
 
    # Install MLIR Python Extras
-   HOST_MLIR_PYTHON_PACKAGE_PREFIX=aie python3 -m pip install -r python/requirements_extras.txt
+   EUDSL_PYTHON_EXTRAS_HOST_PACKAGE_PREFIX=aie python3 -m pip install -r python/requirements_extras.txt
```
hunhoffe, line 145: @andrej I think it is still in the README?

andrej: You're right, my bad. I was looking at `docs/README.md` but you changed it there too, I was just looking at the wrong thing. So Maks's upstream change to be able to get rid of the prefix using a config flag in the requirements.txt is still outstanding?

hunhoffe: Good question! I should have explained better earlier. This PR does use a version of the upstream repo that includes Maks' upstream changes for the flag. However, the version of pip we use in all the workflows does not yet support that flag, resulting in a cascade of dependency update errors when I updated pip to be able to use the flag (hence why this small PR had like 60+ commits to it.... ). Since Erwei was still needing to commit hotfixes, I'd figure I'd split this task into three more bite-sized pieces:
* Use the new upstream repo so Erwei no longer needs to maintain a fork (this PR)
* Package the extras into the wheels directly so that users of mlir-aie do not have to know about the extras (in-progress - https://github.com/Xilinx/mlir-aie/pull/2786).
* Update actions and instructions to use the new flag (should only affect mlir-aie and not IRON, https://github.com/Xilinx/mlir-aie/pull/2787)

PR #2751, opened 2025-12-03
Update Object FIFO data layout transformations

AndraBisca:
This PR aims to improve existing documentation for the `toStream` and `fromStream` objectfifo attributes in the programming guide. It also introduces two new examples in section-2c. Finally, it adds new verification checks to the `ObjectFifoLinkOp` to identify when provided data layout transformations access beyond the length of the transfer in the cases of a split or join.


programming_guide/section-2/section-2c/README.md
```diff
@@ -97,6 +101,8 @@ class object_fifo:
 
 Our compiler directly lowers Object FIFOs that make use of the aforementioned data layout transformations to `AIE_DMABDOp`. You can use the `dimensionsToStream` input to describe in which order the `producerTile`'s DMA should push the objects onto the stream. Similarly, the `dimensionsFromStreamPerConsumer` input describes to the DMAs of each individual tile in the `consumerTiles` in what layout to retrieve the objects from the stream.
 
+> **NOTE:**  Data layout transformations in the Object FIFO are applied at object granularity.
```
andrej: This seems ambiguous to me. One could also interpret this as "the granularity of data layout transformations is one ObjectFIFO object (e.g. stride 1 == 1 ObjectFIFO object)". Maybe something like "Data layout transformations are applied to individual ObjectFIFO objects and cannot act across object boundaries."

programming_guide/section-2/section-2c/README.md
```diff
@@ -61,6 +61,10 @@ for(int i = 0; i < size_2; i++)
             //                                   + k * stride_0]
 ```
 
+It is important to note that data layout transformations are interpreted differently depending on whether data is pushed onto or read from the AXI stream:
+- when data is pushed to the AXI stream from local memory, the tile DMA can directly access the data following the desired access pattern.
+- when data is read from the AXI stream, the tile DMA writes the data into local memory according to the desired access pattern, however, it does not have control over the order in which the data arrives.
```
andrej: I understand the point you are making, but I don't know if it would be clear to somebody without the necessary background. Maybe something like this would be clearer?

- when data are _pushed_ to the AXI stream, the layout describes from where in memory the DMA _reads_ the elements to push to the stream (where to ``get items from''); 
- when data are read from the AXI stream, the data layout describes where in memory the DMA _writes_ the elements that are arriving over the stream in-sequence (where to ``put arriving items'').

programming_guide/section-2/section-2c/README.md
AndraBisca: Yes, I see the confusion. This is much better.

programming_guide/section-2/section-2c/README.md
AndraBisca: Yes, I like this better as well. I updated the doc.

PR #2764, opened 2025-12-13
[aievec] Use `llvm-link` to provide `aie_runtime_lib` implementations for vector ops

erwei-xilinx:
- Extend the existing `chess_intrinsic_wrapper` workflow to peano (`peano_intrinsic_wrapper`).
- Change `aievec` implementations of `exp` and `rsqrt` to point to `aie_runtime_lib` impls.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAISw.yml
+++ b/.github/workflows/buildAndTestRyzenAISw.yml
@@ -95,6 +95,10 @@ jobs:
             HOST_MLIR_PYTHON_PACKAGE_PREFIX=aie pip install -r python/requirements_extras.txt
             pip install -r python/requirements_dev.txt
 
+            # Install llvm-aie (Peano)
+            pip install llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/download/nightly/llvm_aie-19.0.0.2025071101+b3cd09d3-py3-none-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
+            export PEANO_INSTALL_DIR="$(pip show llvm-aie | grep ^Location: | awk '{print $2}')/llvm-aie"
+
             VERSION=$(utils/clone-llvm.sh --get-wheel-version)
 
             cd /workspace


aie_runtime_lib/AIE2/lut_based_ops.h
```diff
@@ -22,7 +22,8 @@ alignas(aie::vector_decl_align) extern int16 exp_flut_ab[512];
 alignas(aie::vector_decl_align) extern int16 exp_flut_cd[512];
 alignas(aie::vector_decl_align) extern unsigned char m_inv_lut[128];
 
-__attribute__((always_inline)) v16accfloat getExpBf16(v16bfloat16 x) {
+static inline __attribute__((always_inline)) v16accfloat
```
stephenneuendorffer, line 25: not a good idea to put static functions into a header file, even if they are marked inline

aie_runtime_lib/AIE2/peano_intrinsic_wrapper.cpp
```diff
@@ -0,0 +1,82 @@
...
+// Include the C++ implementations
+#include "lut_based_ops.h"
+#include "vec_math.h"
+
+// Provide extern "C" wrapper functions that call the C++ implementations
+// These have C linkage (no name mangling) for LLVM backend linking
+extern "C" {
+
+// LUT-based operations
+v16accfloat getExpBf16_wrapper(v16bfloat16 in) { return getExpBf16(in); }
```
stephenneuendorffer, line 29: um, why do this?  why not just define the original functions to be extern "C"?

stephenneuendorffer: And the chess_intrinsic_wrapper is serving a very different purpose from this...I wouldn't confuse the roles of the two.

erwei-xilinx:
Ok let me switch to a less intrusive approach: https://github.com/Xilinx/mlir-aie/pull/2767 In the new PR, I switch to mapping to peano's scalar implementation of aie2p invsqrt.

Issue #2765, opened 2025-12-15
provide wheels for Python 3.14


simeonschaub:
It would be great if wheels could be provided for Python 3.14, as that's already the default on Fedora and I couldn't figure out how to install `pyxrt` for older Python versions. I don't know what the process for this is, is it just a matter of bumping the version somewhere or is there more work involved?

PR #2766, opened 2025-12-15
Generate Python 3.14 wheels

hunhoffe:
In response to https://github.com/Xilinx/mlir-aie/issues/2765, this PR creates wheels for Python 3.14. However, this is not fully supported in our test environment, so I add this with the caveat that it is not fully supported or tested.

I also updated the pybind11 version.

This PR also removes a pip purge command that often fails on the CI Runners. Instead, each job uses an ephemeral pip cache directory that is cleared at the end of each job.

### DIFF
--- a/python/helpers/dialects/ext/func.py
+++ b/python/helpers/dialects/ext/func.py
@@ -198,7 +198,7 @@ class FuncBase:
             self.emit()
 
     def _is_decl(self):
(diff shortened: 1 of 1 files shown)


simeonschaub:
I can confirm that this works locally for compiling some simple examples

fifield (approved):

PR #2768, opened 2025-12-16
correctly quote arguments to `grep`

simeonschaub:
My system grep was doing a recursive search of the entire directory
returning nonsense here. Fix this by correctly quoting the argument to
`grep`.


andrej (approved):

hunhoffe (approved):
Thanks for the fix!

PR #2775, opened 2025-12-18
Enable reconfiguration through lowerings for `aiex.configure` and `aiex.run`

andrej:
Sorry for the hefty changeset. I ripped out all unrelated/experimental changes, but it still ended up a bit big.

This PR:
- Allows you to 'call' other runtime sequences by implementing an inlining-based lowering for `aiex.run` (which up to this point was just a stub).
- Allows you to reconfigure between multiple designs by adding a lowering pass for `aiex.configure` using `LOADPDI` instructions.
- Adds an `--expand-load-pdis` option which replaces `LOADPDI` instructions with a sequence of `write32s` and reset instructions. This can be more performant than a simple `LOADPDI`.

All together, this makes it possible to run multiple designs within the same runtime sequence.

A very simple example is in [`test/npu-xrt/reconfigure_loadpdi`](https://github.com/Xilinx/mlir-aie/blob/8d06d68815f68924f0568e0756e4feea69333a71/test/npu-xrt/reconfigure_loadpdi/aie.mlir). 

Here is a screenshot from some results from a more sophisticated example for SwiGLU, showing the performance benefits of this last bullet point (expanding load PDIs into write32s and manually inserting reset instructions):
benchmark_results_7

Next steps:
 - Harness the `convert-aie-to-control-packets` pass as an alternative "backend" for the "materialize runtime sequences" pass (probably in combination with an empty load PDI for resetting) for potential performance gains. Re-configuration (i.e., swapping out configurations between devices) using control packets is as of yet untested.
 - I have a write32/blockwrite-coalescing pass that could potentially reduce the number of write32s.
 - Add Python abstractions to make this usable from a higher level and so we don't have to write the cumbersome `memref.subview` and `memref.reinterpret_cast`s.

### DIFF
--- a/test/Passes/expand-load-pdi/empty_device.mlir
+++ b/test/Passes/expand-load-pdi/empty_device.mlir
@@ -1,0 +1,36 @@
+//===- empty_device.mlir ---------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt --aie-expand-load-pdi %s | FileCheck %s
+
+// This tests how the load_pdi instruction for an empty device are expanded.
+// All that should be generated is an empty device load_pdi, which is used to
+// reset the device. There should not follow any write32s/blockwrites, as there
+// is nothing to configure in an empty device.
+
+module {
+
+    // CHECK: aie.device(npu2_1col) @empty_0 {
+    // CHECK: }
+
+    aie.device(npu2_1col) @my_empty_device {
+        aie.end
+    }
+
+    aie.device(npu2_1col) @main {
+        // CHECK: aie.runtime_sequence(%arg0: memref<1xi32>) {
+        aie.runtime_sequence (%arg0: memref<1xi32>) {
+            // CHECK: aiex.npu.load_pdi {device_ref = @empty_0}
+            aiex.npu.load_pdi { device_ref = @my_empty_device }
+        }
+        // CHECK: }
+    }
+
+}
--- a/test/Passes/expand-load-pdi/error_require_lowered_core.mlir
+++ b/test/Passes/expand-load-pdi/error_require_lowered_core.mlir
@@ -1,0 +1,35 @@
+//===- error_require_lowered_core.mlir -------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt --verify-diagnostics --aie-expand-load-pdi %s
+
+// This test should error as load_pdi operations can only be inlined after the
+// code in each core has been lowered to an ELF file and linked to the core
+// with the `elf_file` attribute.
+
+module {
+
+    aie.device(npu2_1col) @my_core_device {
+        %tile = aie.tile(0, 2)
+        // expected-error @+1 {{Expected lowered ELF file}}
+        aie.core(%tile) {
+            aie.end
+        }
+    }
+
+    aie.device(npu2_1col) @main {
+        aie.runtime_sequence (%arg0: memref<1xi32>) {
+            // expected-error @+1 {{Failed to generate configuration operations}}
+            aiex.npu.load_pdi { device_ref = @my_core_device }
+        }
+    }
+
+
+}
--- a/test/Passes/expand-load-pdi/switchbox_config.mlir
+++ b/test/Passes/expand-load-pdi/switchbox_config.mlir
@@ -1,0 +1,43 @@
+//===- empty_device.mlir ---------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt --aie-expand-load-pdi %s | FileCheck %s
+
+// This tests how some configuration register writes are expanded from a
+// load_pdi instructions. The expand-load-pdi pass for this file should generate
+// an empty load_pdi instruction (for resetting the device) followed by 
+// register writes that configure the switchbox connections specified.
+
+
+module {
+
+    // CHECK: aie.device(npu2_1col) @empty_0 {
+    // CHECK: }
+
+    aie.device(npu2_1col) @my_switchbox_device {
+        %tile = aie.tile(0, 2)
+        aie.switchbox(%tile) {
+            aie.connect<South : 0, Core : 0>
+        }
+    }
+
+    aie.device(npu2_1col) @main {
+        // CHECK: aie.runtime_sequence(%arg0: memref<1xi32>) {
+        aie.runtime_sequence (%arg0: memref<1xi32>) {
+            // CHECK: aiex.npu.load_pdi {device_ref = @empty_0}
+            // CHECK: aiex.npu.write32 {address = 2355200 : ui32, value = 2147483653 : ui32}
+            // CHECK: aiex.npu.write32 {address = 2355476 : ui32, value = 2147483648 : ui32}
+            aiex.npu.load_pdi { device_ref = @my_switchbox_device }
+        }
+        // CHECK: }
+    }
+
+
+}
--- a/test/Passes/expand-load-pdi/switchbox_config_multiple.mlir
+++ b/test/Passes/expand-load-pdi/switchbox_config_multiple.mlir
@@ -1,0 +1,70 @@
+//===- empty_device.mlir ---------------------------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt --aie-expand-load-pdi %s | FileCheck %s
+
+// This tests how multiple load_pdi instructions are expanded. There should be
+// empty load_pdi instructions generated between configurations for the device
+// reset. Since the firmware 'optimizes' away multiple load_pdi instructions
+// with the same address, these should be different empty devices so they 
+// don't get skipped.
+
+
+module {
+
+    // CHECK: aie.device(npu2_1col) @empty_2 {
+    // CHECK: }
+    // CHECK: aie.device(npu2_1col) @empty_1 {
+    // CHECK: }
+    // CHECK: aie.device(npu2_1col) @empty_0 {
+    // CHECK: }
+
+    aie.device(npu2_1col) @my_switchbox_device_1 {
+        %tile = aie.tile(0, 2)
+        aie.switchbox(%tile) {
+            aie.connect<South : 0, Core : 0>
+        }
+    }
+
+    aie.device(npu2_1col) @my_switchbox_device_2 {
+        %tile = aie.tile(0, 2)
+        aie.switchbox(%tile) {
+            aie.connect<North : 0, Core : 0>
+        }
+    }
+
+    aie.device(npu2_1col) @my_switchbox_device_3 {
+        %tile = aie.tile(0, 2)
+        aie.switchbox(%tile) {
+            aie.connect<South : 0, DMA : 0>
+        }
+    }
+
+    aie.device(npu2_1col) @main {
+        // CHECK: aie.runtime_sequence(%arg0: memref<1xi32>) {
+        aie.runtime_sequence (%arg0: memref<1xi32>) {
+            // CHECK: aiex.npu.load_pdi {device_ref = @empty_0}
+            // CHECK: aiex.npu.write32 {address = 2355200 : ui32, value = 2147483653 : ui32}
+            // CHECK: aiex.npu.write32 {address = 2355476 : ui32, value = 2147483648 : ui32}
+            aiex.npu.load_pdi { device_ref = @my_switchbox_device_1 }
+            // CHECK: aiex.npu.load_pdi {device_ref = @empty_1}
+            // CHECK: aiex.npu.write32 {address = 2355200 : ui32, value = 2147483663 : ui32}
+            // CHECK: aiex.npu.write32 {address = 2355516 : ui32, value = 2147483648 : ui32}
+            aiex.npu.load_pdi { device_ref = @my_switchbox_device_2 }
+            // CHECK: aiex.npu.load_pdi {device_ref = @empty_2}
+            // CHECK: aiex.npu.write32 {address = 2355204 : ui32, value = 2147483653 : ui32}
+            // CHECK: aiex.npu.write32 {address = 2355476 : ui32, value = 2147483648 : ui32}
+            aiex.npu.load_pdi { device_ref = @my_switchbox_device_3 }
+        }
+        // CHECK:     }
+    }
+
+
+}


python/compiler/aiecc/main.py
```diff
@@ -979,9 +1005,11 @@ async def generate_full_elf_config_json(
                 "PDIs": [],
             }
 
-            pdi_id = device_to_id_mapping[device_name]
-            pdi_filename = self.pdi_file_name(device_name)
-            kernel_entry["PDIs"].append({"id": pdi_id, "PDI_file": pdi_filename})
+            for other_device_name, other_pdi_id in device_to_id_mapping.items():
+                pdi_filename = self.pdi_file_name(other_device_name)
+                kernel_entry["PDIs"].append(
+                    {"id": other_pdi_id, "PDI_file": pdi_filename}
```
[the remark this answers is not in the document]
andrej: For posterity: Every kernel needs all PDIs because it might configure other kernels using `LOADPDI`.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -184,14 +186,30 @@ struct AIEDMATasksToNPUPass : AIEDMATasksToNPUBase<AIEDMATasksToNPUPass> {
...
+
+    if (buf_arg) {
       if (!target_model.isShimNOCTile(tile.getCol(), tile.getRow())) {
         return bd_op->emitOpError("DDR memory (runtime input arguments) can "
                                   "only be referred to on shim tiles.");
       }
+
       unsigned arg_idx = buf_arg.getArgNumber();
-      int64_t offset = bd_op.getOffsetInBytes();
+      offset += bd_op.getOffsetInBytes();
```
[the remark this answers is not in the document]
andrej: Fixed typo. For posterity: This logic is correct. The BD offset should be added on top of any other offsets that might have been calculated from the `memref.subviews`.

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -850,3 +857,408 @@ std::unique_ptr<mlir::OperationPass<xilinx::AIE::DeviceOp>>
...
+    return dma1 == dma2;
+  if (auto mem1 = dyn_cast<MemOp>(dma1)) {
+    return mem1.isEquivalent(dma2);
+  } else if (auto memTile1 = dyn_cast<MemTileDMAOp>(dma1)) {
+    return memTile1.isEquivalent(dma2);
+  } else if (auto shim1 = dyn_cast<ShimDMAOp>(dma1)) {
+    return shim1.isEquivalent(dma2);
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
andrej: For posterity: This is not duplicated code, the code for `areDMAsEquivalent` in AIEDialect.cpp is different and implements an actual equivalence check -- whereas this function simply dispatches to the correct equivalence checker based on the DMA tile type.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -980,8 +978,47 @@ AIE::RuntimeSequenceOp AIEX::RunOp::getCalleeRuntimeSequenceOp() {
...
+    // For memref types, check compatibility (same shape and element type)
+    auto argType = dyn_cast<MemRefType>(arg.getType());
+    auto valType = dyn_cast<MemRefType>(val.getType());
+
+    if (argType && valType) {
+      if (argType.getShape() != valType.getShape() ||
+          argType.getElementType() != valType.getElementType()) {
+        return emitOpError()
+               << "argument " << i << " type mismatch: " << "expected "
+               << argType << " but got " << valType;
```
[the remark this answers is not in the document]
andrej: replaced with exact check

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -629,57 +629,47 @@ enum OutputType {
...
+      device = builder.getBlock()->getParentOp()->getParentOfType<DeviceOp>();
+    }
+    OpBuilder::InsertionGuard guard(builder);
+    builder.setInsertionPointToStart(device.getBody());
+    for (auto &op : operations) {
+      if (op.cmd.Opcode != XAIE_IO_BLOCKWRITE) {
+        global_data.push_back(nullptr);
+        continue;
+      }
+      uint32_t size = op.cmd.Size / 4;
```
hunhoffe, line 650: Out of an abundance of caution, do you need to check divisibility here or is it guaranteed?

python/compiler/aiecc/cl_arguments.py
```diff
@@ -387,8 +447,44 @@ def parse_args(args=None):
...
+        action="store_true",
+        help="Expand load_pdi operations into explicit device reset and configuration sequences",
+    )
+    parser.add_argument(
+        "--no-materialize",
+        dest="materialize_runtime_sequence",
+        default=True,
+        action="store_false",
+        help="Do not 'materialize' the runtime sequence (lower aiex.configure and aiex.run operations)",
+    )
```
hunhoffe, line 403: I think there's a bit more you could do here to correctly parse and validate the new options and, rather than allowing overriding behavior, error on conflicting options. Maybe you need a subparser? Maybe not, but regardless, here is a snippet generated by Google for comma delineated lists with a set number of valid options:
```python
import argparse
import sys

def csv_choices_type(value, allowed_choices):
    """
    Custom type function to split a comma-separated string and validate choices.
    """
    items = [item.strip() for item in value.split(',')]
    for item in items:
        if item not in allowed_choices:
            # Raise an error that argparse will catch and display nicely
            raise argparse.ArgumentTypeError(
                f"Invalid choice: '{item}' (choose from {', '.join(allowed_choices)})"
            )
    return items

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description="Parse comma-delimited arguments with choices.")
    
    # Define the allowed options
    VALID_OPTIONS = ['read', 'write', 'delete', 'list']

    parser.add_argument(
        '--permissions',
        type=lambda v: csv_choices_type(v, VALID_OPTIONS), # Pass choices to the custom type function
        help=f"A comma-separated list of permissions (options: {', '.join(VALID_OPTIONS)})"
    )

    args = parser.parse_args()
    
    print(f"Provided permissions: {args.permissions}")

    # Example of how to use the parsed list later
    if 'read' in args.permissions:
        print("User has read permission.")
```

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
andrej: I didn't author this line, this is just moved. I do think it is guaranteed. @fifield do you know?

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -212,6 +217,7 @@ def AIE_SwitchboxOp: AIE_Op<"switchbox", [
     int colIndex();
     int rowIndex();
     TileOp getTileOp();
+    bool isEquivalent(SwitchboxOp other);
```
fifield: There's a mix of `bool isEquivalent(Operation *)` and `bool isEquivalent(OpType)` here. Can they all be the former?
```suggestion
    bool isEquivalent(Operation *other);
```

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -389,6 +396,9 @@ def AIE_CoreOp: AIE_Op<"core", [
     bool isMemWest() { return ((rowIndex() % 2) == 0); };
     bool isEmpty();
     TileOp getTileOp();
+    // Check whether another core has equivalent program memory.
+    // Note that this currently only checks the ELF attribute, so only produces a meaningful result after compiling the program memory.
+    bool isEquivalent(CoreOp other);
```
fifield:
```suggestion
    bool isEquivalent(Operation *other);
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1376,6 +1469,20 @@ bool CoreOp::isEmpty() {
           llvm::isa<AIE::EndOp>(body.front().front()));
 }
 
+bool CoreOp::isEquivalent(CoreOp other) {
+  if (!other)
```
fifield:
```suggestion
bool CoreOp::isEquivalent(Operation *op) {
  CoreOp other = llvm::dyn_cast_if_present<CoreOp>(op);
  if (!other)
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2114,6 +2229,47 @@ int SwitchboxOp::colIndex() { return getTileOp().colIndex(); }
 
 int SwitchboxOp::rowIndex() { return getTileOp().rowIndex(); }
 
+bool SwitchboxOp::isEquivalent(SwitchboxOp other) {
+  if (!other)
```
fifield:
```suggestion
bool SwitchboxOp::isEquivalent(Operation *op) {
  SwitchboxOp other = llvm::dyn_cast_if_present<SwitchboxOp>(op);
  if (!other)
```

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -254,4 +271,54 @@ def AIELegalizeControlPacket : Pass<"aie-legalize-ctrl-packet", "AIE::DeviceOp">
...
+    Example command-line usage:
+    ```
+      aie-opt --aie-expand-load-pdi \
+        --reset-switches-tiles=shim,mem,core \
+        --reset-switches-mode=ifused \
+        --reset-locks-tiles=mem,core \
+        --reset-locks-mode=always \
+        input.mlir
+    ```
+  }];
```
fifield, line 284: I can't find any unit tests for this pass to see what it does.

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -850,3 +863,402 @@ std::unique_ptr<mlir::OperationPass<xilinx::AIE::DeviceOp>>
 xilinx::AIE::createConvertAIEToControlPacketsPass() {
   return std::make_unique<ConvertAIEToControlPacketsPass>();
 }
+
+mlir::LogicalResult
+xilinx::AIE::generateAndInsertConfigOps(OpBuilder &builder,
+                                        xilinx::AIE::DeviceOp device,
+                                        llvm::StringRef clElfDir) {
```
fifield: This is nearly identical to `convertAIEToConfiguration` before the changes in this PR. I think I see why it evolved this way, but it feels like there's still extra duplication here, although I don't have any suggestions.

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -850,3 +863,402 @@ std::unique_ptr<mlir::OperationPass<xilinx::AIE::DeviceOp>>
...
+            WireBundle::East, WireBundle::PLIO, WireBundle::NOC,
+            WireBundle::Trace, WireBundle::TileControl}) {
+        uint32_t numDestChans =
+            targetModel.getNumDestSwitchboxConnections(col, row, bundle);
+        if (numDestChans > 0) {
+          possibleDests.push_back(std::make_pair(bundle, numDestChans));
+        }
+      }
+
+      // Iterate over all sources of connections in the current device
```
fifield:
```suggestion
      // Iterate over all sources of connections in the current tile
```

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
```diff
@@ -850,3 +863,402 @@ std::unique_ptr<mlir::OperationPass<xilinx::AIE::DeviceOp>>
...
+  }
+
+  return success();
+}
+
+// Reset fine-grained switch connections based on the provided configuration
+static LogicalResult resetSwitchConnectionsFineGrained(DeviceOp device,
+                                                       DeviceOp previousDevice,
+                                                       ResetConfig switchConfig,
+                                                       AIERTControl &ctl) {
```
fifield: This doesn't account for any packet flow differences?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1298,6 +1337,56 @@ int ShimMuxOp::colIndex() { return getTileOp().colIndex(); }
...
+  auto &region2 = dma2->getRegion(0);
+
+  if (region1.empty() != region2.empty())
+    return false;
+
+  if (region1.empty())
+    return true;
+
+  auto &block1 = region1.front();
+  auto &block2 = region2.front();
```
fifield: Don't dma regions have multiple blocks?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1298,6 +1337,56 @@ int ShimMuxOp::colIndex() { return getTileOp().colIndex(); }
...
+      return false;
+
+    for (auto attr1 : attrs1) {
+      auto attr2 = it2->getAttr(attr1.getName());
+      if (!attr2 || attr1.getValue() != attr2)
+        return false;
+    }
+
+    ++it1;
+    ++it2;
```
fifield: Should operands also be checked here? For example locks and buffers? Did you look at `mlir::OperationEquivalence`?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2114,6 +2229,47 @@ int SwitchboxOp::colIndex() { return getTileOp().colIndex(); }
 
 int SwitchboxOp::rowIndex() { return getTileOp().rowIndex(); }
 
+bool SwitchboxOp::isEquivalent(SwitchboxOp other) {
```
fifield: This does not consider switchbox ops with packet flow information:
```mlir
    %switchbox_4_0 = aie.switchbox(%shim_noc_tile_4_0) {
      aie.connect<North : 1, South : 2>
      aie.connect<South : 3, North : 1>
      %0 = aie.amsel<0> (0)
      %1 = aie.amsel<5> (3)
      %2 = aie.masterset(South : 3, %0) {keep_pkt_header = true}
      %3 = aie.masterset(South : 0, %1) {keep_pkt_header = true}
      aie.packet_rules(TileControl : 0) {
        aie.rule(31, 15, %1)
      }
      aie.packet_rules(Trace : 0) {
        aie.rule(31, 1, %0)
      }
    }
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2479,6 +2635,55 @@ RuntimeSequenceOp::getForSymbolInDeviceOrError(DeviceOp deviceOp,
...
+          if (!llvm::isa<ShimDMAAllocationOp>(symbolDefOp) &&
+              !llvm::isa<DeviceOp>(symbolDefOp) &&
+              !llvm::isa<RuntimeSequenceOp>(symbolDefOp) &&
+              !llvm::isa<BufferOp>(symbolDefOp) &&
+              !llvm::isa<memref::GlobalOp>(symbolDefOp)) {
+            op->emitOpError()
+                << "references symbol '"
+                << symbolRef.getRootReference().getValue()
+                << "' which must be either a ShimDMAAllocationOp, DeviceOp, "
+                   "RuntimeSequenceOp or BufferOp, but got: "
```
fifield:
```suggestion
                   "RuntimeSequenceOp, BufferOp or GlobalOp, but got: "
```

python/compiler/aiecc/cl_arguments.py
fifield: Is this (or something similar enough) not possible with the [llvm command line library](https://llvm.org/docs/CommandLine.html)?

test/dialect/AIEX/run_error_arg_type_mismatch.mlir
```diff
@@ -0,0 +1,35 @@
...
+// Test detection of argument type mismatches when calling runtime sequences
+
+module {
+  aie.device(npu2) @main {
+    %tile00 = aie.tile(0, 0)
+    
+    aie.runtime_sequence @main_seq(%arg0: memref<16xi32>) {
+      aiex.configure @config_wrong_type {
+        // expected-error @+1 {{argument 0 type mismatch}}
+        aiex.run @seq_expects_i16(%arg0) : (memref<16xi32>)
```
fifield, line 22: Can this just be a static type error?  i.e. it should fail to parse with just `aie-opt %s` ?

fifield (commented):
The aie-expand-load-pdi pass needs unit tests.

I wonder if you considered representing the reset logic more explicitly in the IR as a step in the lowering? i.e. could a user craft their own reset sequence via mlir instead of via aiecc flags?

python/compiler/aiecc/cl_arguments.py
andrej: @fifield does that have Python bindings?

fifield: Sorry, I intended for my comment to reference the custom arg parsing in AIEExpandLoadPdi.cpp, for:
```
        --reset-switches-tiles=shim,mem,core
        --reset-switches-mode=ifused
        --reset-locks-tiles=mem,core
        --reset-locks-mode=always
```
but I don't think it's an issue to leave in what you have.

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: no longer relevant, see comment below

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
andrej: working on this

lib/Conversion/AIEToConfiguration/AIEToConfiguration.cpp
andrej: I needed to insert ops at the current insertion point of the builder, where `convertAIEToConfiguration` finds its own insertion point. I changed this so that `convertAIEToConfiguration` first finds the insertion point and then calls this method to remove the duplication.

test/dialect/AIEX/run_error_arg_type_mismatch.mlir
andrej: It actually was already part of the verifier and errored without the `--aie-materialize-runtime-sequences` pass. I moved it to `tests/dialect/AIEX`.

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
andrej: Added to `/tests/Passes/expand-load-pdi`

PR #2792, opened 2025-12-29
Updates and fixes to mlirDistro workflow w/ pip upgrade

hunhoffe:
The pyyaml version in the requirements isn't synced correctly to the pyyaml required by eudsl python extras. This doesn't always throw an error with the normal install sequence (so it doesn't show in the runner) but if you try to re-install the ```requirements.txt``` or install ```requirements_extras.txt``` before ```requirements.txt```, you end up with an import error due to pyyaml version constraint conflicts. I realized that pyyaml isn't really used/needed -- except in a singular helper script -- so I went ahead and moved that dependency to a development dependency with less versioning restrictions.

Next, I realized that the wheels tags don't always appear to be set correctly. The tags are particularly important because newer versions of pip (v25.x, I think) enforce that the filename matches the internal tags of the wheel, and currently that is not always the case. This comes into play for, example, the ```utils/build-mlir-aie-from-wheels.sh``` script, because ```pip download``` stopped working for me.

I also removed some faulty validation code that I had accidentally merged in in https://github.com/Xilinx/mlir-aie/pull/2789

### DIFF
--- a/python/iron/compile/cache/utils.py
+++ b/python/iron/compile/cache/utils.py
@@ -36,9 +36,15 @@ def _create_function_cache_key(function, args, kwargs):
             func_hash = hash(arg)
             signature_parts.append(f"externalfunction_{func_hash}")
         elif callable(arg):
-            # Function argument - use hash of function address for uniqueness
-            func_hash = hash(arg)
-            signature_parts.append(f"function_{func_hash}")
+            if hasattr(arg, "__code__"):
+                # Use bytecode and constants hash for Python functions/lambdas
+                code = arg.__code__
+                func_hash = hash((code.co_code, code.co_consts, code.co_names))
+                signature_parts.append(f"function_{func_hash}")
+            else:
+                # Function argument - use hash of function address for uniqueness
+                func_hash = hash(arg)
+                signature_parts.append(f"function_{func_hash}")
         else:
             # Unsupported type - use type name
             signature_parts.append(f"{type(arg).__name__}")
@@ -52,9 +58,15 @@ def _create_function_cache_key(function, args, kwargs):
             func_hash = hash(value)
             signature_parts.append(f"{key}_externalfunction_{func_hash}")
         elif callable(value):
-            # Function argument - use hash of function address for uniqueness
-            func_hash = hash(value)
-            signature_parts.append(f"{key}_function_{func_hash}")
+            if hasattr(value, "__code__"):
+                # Use bytecode and constants hash for Python functions/lambdas
+                code = value.__code__
+                func_hash = hash((code.co_code, code.co_consts, code.co_names))
+                signature_parts.append(f"{key}_function_{func_hash}")
+            else:
+                # Function argument - use hash of function address for uniqueness
+                func_hash = hash(value)
+                signature_parts.append(f"{key}_function_{func_hash}")
         else:
             # Unsupported type - use type name
             signature_parts.append(f"{key}_{type(value).__name__}")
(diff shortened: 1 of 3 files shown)


utils/mlir_wheels/setup.py
```diff
@@ -11,6 +11,27 @@
...
+    def get_tag(self):
+        python, abi, plat = _bdist_wheel.get_tag(self)
+        python, abi = "py3", "none"
+
+        CIBW_ARCHS = os.environ.get("CIBW_ARCHS")
+        if CIBW_ARCHS in {"arm64", "aarch64", "ARM64"}:
+            if "x86_64" in plat:
+                plat = plat.replace("x86_64", "aarch64")
+            elif "AMD64" in plat:
+                plat = plat.replace("AMD64", "aarch64")
```
andrej, line 31: What's the reasoning for this?

hunhoffe: Previously, we just renamed the wheels in the workflow to create a wheel for different arches. However, newer versions of pip check the file name against the internal platform tags. So I removed the code for the file rename and replaced it by setting the platform properly in the setup for building the wheel, which makes sure the metadata and the file naming convention remain in sync.

I believe the renaming logic itself is pretty much the same as it was before, just done in a more complete way. At least, that is my intention.

andrej: Sorry I might be a little slow or missing some context, but I still don't understand. To me it looks like we're replacing x86 with ARM which are different architectures?

.github/workflows/mlirDistro.yml
```diff
@@ -275,27 +275,6 @@ jobs:
...
-      if: ${{ matrix.OS == 'ubuntu-22.04' }}
-      working-directory: ${{ steps.workspace_root.outputs.WORKSPACE_ROOT }}
-      shell: bash
-      run: |
-        
-        rename 's/cp312-cp312/py3-none/' wheelhouse/mlir*whl
-        
-        if [ x"${{ matrix.OS }}" == x"ubuntu-22.04" ] && [ x"${{ matrix.ARCH }}" == x"aarch64" ]; then
-          rename 's/x86_64/aarch64/' wheelhouse/mlir*whl
-        fi
```
andrej, line 291: I see where you are talking about about previously doing it in the GitHub action...

utils/mlir_wheels/setup.py
andrej: Wait I think I get it now. The wheel actually contains no platform-dependent code, so any wheel will work on any platform. So if we are on ARM, we just replace the x86 string with aarch64 because I guess something down the chain looks for a wheel with that specific name. Is that correct?

andrej (approved):

utils/mlir_wheels/setup.py
hunhoffe: I've had to kind of reverse engineer the logic behind this, because I hadn't worked with this workflow before, but I think what is happening is:
* We use `CIBW_ARCH` to do crosscompilation for some arches because the github runners do not cover the whole matrix for the wheel building: https://github.com/Xilinx/mlir-aie/blob/b721e8257bd40cab973c331089f7e6716f43fa49/.github/workflows/mlirDistro.yml#L198C9-L198C40
* However, we don't actually care about cross compilation for the python portions (as you say) so we just renamed the wheels which was previously enough to allow pip to correctly choose the right file by file name
* However, newer versions of pip appear to check the file name to select the file for download the same way but then they also do some internal validation once the file is downloaded; I can't find a way to turn this off for `pip download` in particular. So I'm trying to fix it properly so the file name matches the metadata.

PR #2795, opened 2026-01-02
IRON test fixes for two intermittent failures

hunhoffe:
This PR addresses two intermittent test failures. 

# bfloat16 tensor.rand rounding error
The first, identified by @andrej , is:
```
=================================== FAILURES ===================================
______________________ test_rand[CPUOnlyTensor-bfloat16] _______________________

dtype = <class 'ml_dtypes.bfloat16'>
tensorclass = <class 'aie.iron.hostruntime.tensor.CPUOnlyTensor'>

    @pytest.mark.parametrize("dtype", TEST_DTYPES)
    @pytest.mark.parametrize("tensorclass", TENSOR_CLASSES)
    def test_rand(dtype, tensorclass):
        iron.set_iron_tensor_class(tensorclass)
        for d in tensorclass.DEVICES:
            t = iron.rand(2, 2, dtype=dtype, device=d)
            arr = t.numpy()
>           assert np.all((arr >= 0) & (arr < 1.0))
E           assert False
E            +  where False = <function all at 0x755a867e2670>((array([[0.683594, 1],\n       [0.451172, 0.746094]], dtype=bfloat16) >= 0 & array([[0.683594, 1],\n       [0.451172, 0.746094]], dtype=bfloat16) < 1.0))
E            +    where <function all at 0x755a867e2670> = np.all

../../../test/python/tensor.py:99: AssertionError
```

This error occurs -- I believe -- due to the low precision of bfloat16 such that when the numpy random array is concerted to bfloat16, sometimes values end up being 1. This can be fixed using numpy methods for clipping the random arrary after it is generated and converted to bfloat16. 

Does this keep the uniform distribution completely correct? I'm not sure, I haven't proven it either way, but I think this fix is "good enough." I also added a test for this specific rounding issue.

# Lambda function caching error
This is an error that occurs sometimes when test/python/test_cache_functionality.py is run:
```
 =================================== FAILURES ===================================
_________________________ test_cache_lambda_functions __________________________

    def test_cache_lambda_functions():
        """Test that caching works correctly with different lambda functions."""
        # Create input tensor
        input_tensor = iron.tensor((32,), dtype=np.int32)
        input_tensor[:] = np.arange(1, 33, dtype=np.int32)  # [1, 2, 3, ..., 32]
    
        # Test 1: First execution with lambda function
        transform(input_tensor, input_tensor, lambda x: x + 1)
        result1 = input_tensor.numpy().copy()
    
        # Reset tensor
        input_tensor[:] = np.arange(1, 33, dtype=np.int32)
    
        # Test 2: Second execution with same lambda function (should use cache)
        transform(input_tensor, input_tensor, lambda x: x + 1)
        result2 = input_tensor.numpy()
    
        # Results should be identical
        np.testing.assert_array_equal(result1, result2)
    
        # Test 3: Different lambda function (should generate new cache entry)
        input_tensor[:] = np.arange(1, 33, dtype=np.int32)
        transform(input_tensor, input_tensor, lambda x: x * 2)
        result3 = input_tensor.numpy()
    
        # Results should be different
>       np.testing.assert_raises(
            AssertionError, np.testing.assert_array_equal, result1, result3
        )

../../../test/python/test_cache_functionality.py:111: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python3.12/unittest/case.py:778: in assertRaises
    return context.handle('assertRaises', args, kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/usr/lib/python3.12/unittest/case.py:237: in handle
    with self:
/usr/lib/python3.12/unittest/case.py:260: in __exit__
    self._raiseFailure("{} not raised by {}".format(exc_name,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <unittest.case._AssertRaisesContext object at 0x77d62aa47050>
standardMsg = 'AssertionError not raised by assert_array_equal'

    def _raiseFailure(self, standardMsg):
        msg = self.test_case._formatMessage(self.msg, standardMsg)
>       raise self.test_case.failureException(msg)
E       AssertionError: AssertionError not raised by assert_array_equal
```

I believe this is because the cache of lambda functions is solely based on the hash of the code. However, some lambdas are defined not only by their bytecode operations but also by their constants. I added these constants to the hashing algorithm. I'm not sure if there are other things that need to be checked for lambda caching, but this is a start in the right direction.

### DIFF
--- a/python/iron/hostruntime/tensor.py
+++ b/python/iron/hostruntime/tensor.py
@@ -373,6 +373,20 @@ class Tensor(ABC):
 
         t = cls.__check_or_create(*size, out=out, dtype=dtype, device=device, **kwargs)
         random_val = np.random.uniform(0.0, 1.0, size=t.shape).astype(dtype)
+        # Ensure values are < 1.0 for low-precision types
+        is_bfloat16 = False
+        try:
+            from ml_dtypes import bfloat16
+
+            if dtype == bfloat16:
+                is_bfloat16 = True
+        except ImportError:
+            pass
+
+        if np.issubdtype(dtype, np.floating) or is_bfloat16:
+            max_val = np.nextafter(dtype(1.0), dtype(0.0))
+            random_val = np.clip(random_val, 0.0, max_val)
+
         if t.shape == ():
             t.data.fill(random_val)
         else:
--- a/test/python/tensor.py
+++ b/test/python/tensor.py
@@ -305,3 +305,15 @@ def test_mixed_device_operations(dtype):
 
     npu_tensor = npu_tensor.to("cpu")
     assert npu_tensor.device == "cpu"
+
+
+@pytest.mark.parametrize("dtype", [np.float32, bfloat16])
+@pytest.mark.parametrize("tensorclass", TENSOR_CLASSES)
+def test_rand_bfloat16_boundary(dtype, tensorclass):
+    """Test that bfloat16 rand never produces 1.0 due to rounding."""
+    iron.set_iron_tensor_class(tensorclass)
+    # Generate many values to increase chance of hitting boundary
+    t = iron.rand(10000, dtype=dtype)
+    arr = t.numpy()
+    assert np.all(arr < 1.0)
+    assert np.all(arr >= 0.0)


python/iron/hostruntime/tensor.py
```diff
@@ -373,6 +373,20 @@ def rand(cls, *size, out=None, dtype=None, device=None, **kwargs):
...
         random_val = np.random.uniform(0.0, 1.0, size=t.shape).astype(dtype)
+        # Ensure values are < 1.0 for low-precision types
+        is_bfloat16 = False
+        try:
+            from ml_dtypes import bfloat16
+
+            if dtype == bfloat16:
+                is_bfloat16 = True
+        except ImportError:
+            pass
```
andrej, line 384: Could be a useful utility function elsewhere.

hunhoffe:
@mawad-amd do you have any opinion on the lambda code?

PR #2801, opened 2026-01-06
WSL Compatibility: Allow common.cmake to find repo root in WSL

thomthehound:
Fixes WSL reversion from #2642 (Windows PowerShell was not seeing the correct directories). Now allows examples to be built in WSL checkouts. Mostly. There may still be issues with Python tests that will be addressed later.

Still requires the user to `source /opt/xilinx/xrt/setup.sh` before attempting compile. I am out of the loop, so I'm not sure why that was removed from environmental setup. This could also be a WSL-specific issue.

### DIFF
--- a/programming_examples/common.cmake
+++ b/programming_examples/common.cmake
@@ -7,30 +7,72 @@
 # Common CMake configuration for programming examples
 # This file provides common setup for test_utils library linking
 
-# Get MLIR_AIE_DIR from Python
-execute_process(
-    COMMAND python3 -c "from aie.utils.config import root_path; print(root_path())"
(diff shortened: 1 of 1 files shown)


hunhoffe:
We removed ```source /opt/xilinx/xrt/setup.sh``` because some newer install methods for xrt may not utilize this script.

thomthehound:
Oh, of course! I just don't want to be presumptuous. I'm trying to stay in my lane getting WSL back in working order. But sometimes there are things I just don't understand about the project ergonomics. I think I should have read the readme more literally instead of assuming *any* functional flavor of `xrt-smi`, including `xrt-smi.exe`, would be sufficient.

PR #2809, opened 2026-01-12
Fix tensor size mismatch

hunhoffe:
I believe I introduced an error in the ```test.py``` for the programming guide section 3 and section 4a in https://github.com/Xilinx/mlir-aie/pull/2737 that occasionally causes a segfault in the CI. This PR fixes the error (mismatch in tensor size).


andrej (approved):
gotta love these intermittent memory corruption errors

PR #2810, opened 2026-01-12
Add checks on DMA region ops

AndraBisca:
### Summary
DMA region operations for all tile types (`MemOp`,  `MemTileDMAOp`) are missing checks that verify:
- there can only be one `DMABdOp` in each block,
- there can only be one lock acquire operation in each block,
- there can only be one lock release operation in each block.

These checks are currently only present in the high-level DMA block primitive `DMAOp`.

### Changes
- Applied these checks to the existing `HasValidBDs` trait for DMA-like Ops.
- Added unit tests for all DMA-like tile types.
- `DMAOp` verification is kept separate and unchanged due to potential refactoring as a higher-level primitive for DMA-like Ops.


AndraBisca:
Thank you @yenjames for the changes, everything looks great!

Issue #2805, opened 2026-01-09
Dependencies, Versioning, and Wheels


hunhoffe:
The build/distribution system is pretty complex at the moment. To build reliable wheels that provide a consistent development experience, I think what needs to happen is:
- [x] The README(s) are updated to encourage folks to use the tag of the repo that matches the version they are using; this can help keep programming examples in sync with the wheels (I believe this may help with issues such as https://github.com/Xilinx/mlir-aie/issues/2804) -- PR is here: https://github.com/Xilinx/mlir-aie/pull/2808
- [x] The process of updating LLVM should be done automatically, unless there is an issue, and well documented in order to keep the dependencies in sync. When we bump MLIR, we also want to bump the eudsl extras to keep them in-sync. I've just started work towards this here: https://github.com/Xilinx/mlir-aie/pull/2845
- [x] Since the eudsl extras are kind of a pain to install (the environment variable precludes normal pip install; the flag that is supported to install via a requirements file is very specific to particular pip versions). https://github.com/Xilinx/mlir-aie/pull/2811 There are two options: 
  - 1. Encoding them as a dependency that gets installed as the wheels are installed. This is probably the best option but I couldn't figure out how to get it to work.
  - 2. Packaging them directly into the wheels (which is more resilient to if the extras wheels are also purged after a specific time limit) -- I've spent a small amount of time exploring the feasibility of this option: https://github.com/Xilinx/mlir-aie/pull/2786 but then made a follow-on PR that succeeded.

I was thinking that instead of building our own MLIR wheels, we should use distributions already publicly available, e.g.: https://github.com/llvm/eudsl/tree/main/projects/mlir-wheel However, these wheels expire after 30 days, so we would need to find a way to have them hosted or built on our own on the fly to ensure more consistent long-term dev environment (or also phase out our wheels every 30 days).

@jgmelber and @erwei-xilinx and @andrej please to feel free to weigh in!

PR #2811, opened 2026-01-12
Vendor eudsl in wheels/build (aie.extras)

hunhoffe:
# TL;DR
This PR vendors eudsl into the mlir-aie wheels and fixes some minor flaws in the CI.

# Full Explanation
This PR is associated with the desire for a cleaner install of eudsl (Part of this issue: https://github.com/Xilinx/mlir-aie/issues/2805). This should benefit those that use the wheels to install mlir-aie.

This was more difficult than expected. To include eudsl in the normal requirement.txt, I had to ensure pip was updated. This was fairly straightforward.

But what was not straightforward was:
* Instrumenting the setup.py that installed mlir-aie wheels dependencies upon install to automatically install the extras
* Instrumenting the cmake files to include the vendored files for tests from local builds e.g., ninja

The difficulties were a combination of:
* setup.py does not support parsing the install options required for eudsl
* because eudsl is installed into the same namespace (aie) installing it directly into aie.extras didn't work correctly; so I have to use pip to download it and then copy it into the correct place
* Updating pip changed the version of LLVM used in the MLIR AIE Distro workflow, which resulting in a different version of llvm. I change this workflow to use the same version of LLVM used in the clone_llvm utility, which should ensure future consistency.


python/requirements_dev.txt
```diff
@@ -10,3 +10,4 @@ matplotlib
 psutil
 pytest
 pyyaml
+cloudpickle # required by eudsl when it is vendored instead of installed
```
hunhoffe, line 13: This is a super hacky addition that should be able to be deleted in a follow-on PR. It only exists to satisfy the CI. Here's some explanation.

cloudpickle is a dependency of eudsl. Since (in this PR) we start to vendor eudsl, we need to include cloudpickle as a dependency. We do this in requirements.txt. This is not hacky and is correct.

However, the "Build and Test on Ryzen AI" job uses latest wheels to install dependencies. So *new* dependencies are not reflected in the CI job. So I added cloudpickle here until wheels can be built with the new dependency, at which point it should not be necessary.

I didn't try to fix the workflow dependencies because there is value in testing that dependencies are correctly reflected in the wheels, and I didn't know how to retain that property while fixing this issue.

PR #2812, opened 2026-01-13
fix buffer size in programming guide example

hunhoffe:
I noticed that one of the programming examples failed in the CI with two mismatches. I'm not sure if this will help or not, but this PR fixes an inconsistency in argument size between the IRON design and the test.cpp file and the python design. I didn't stumble upon anything else weird that may be causing intermittent CI errors with this example, but I'm open to suggestions.


hunhoffe:
Another PR could try to clean up dummy arguments throughout the programming examples. In this PR, I just want to figure out why the CI is failing with bad data in only two of the values only some of the time. Seems to me it could be a buffer alignment issue if the driver places the supposed inputs contiguously or something. Of course, this could also be due to the runtime PR with caching/concurrency issue messing up the XRT driver -- but I want to double check other possibilities before I turn the caching off.

If you have other suggestions to try, I'm open!

PR #2813, opened 2026-01-13
Try to update GitHub docs

hunhoffe:
I noticed that my changes to the GitHub documentation from https://github.com/Xilinx/mlir-aie/pull/2737 weren't being percolated to the webpage. I went to the gh-pages branch and found this error in the CI task:
```
Liquid Exception: Liquid syntax error (line 1266): Tag '{%' was not properly terminated with regexp: /\%\}/ in AIEDialect.md
```

I think this PR should fix that error and allow the gh-pages to be updated.


fifield (approved):
It's been a while since we hit this Liquid syntax error, but it has happened multiple times. Ideally we would have some way to prevent it or recover automatically since it fails on valid MLIR embedded in the docs.

PR #2831, opened 2026-01-26
Also inline locks in `aiex.run` lowering + codegen performance improvements

andrej:
- Make it possible to use locks referenced from runtime sequences called (inlined) via `aiex.run` (`WorkerRuntimeBarrier` generates locks in the runtime sequence)
- Clean up the `aiex.run` inlining
- Make new symbol names for inlined runtime sequences less likely to clash, drastically improving codegen time for large designs (less time spent in the loop trying to find a unique symbol name)

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
@@ -295,8 +295,7 @@ copyReferencedSSAValues(PatternRewriter &rewriter,
 
     } else if (auto lockOp = llvm::dyn_cast<AIE::LockOp>(definingOp)) {
       rewriter.restoreInsertionPoint(clonedSSAInsertPoint);
-      Operation *clonedLock = rewriter.clone(*lockOp, argMap);
-      argMap.map(definingOp->getResult(0), clonedLock->getResult(0));
+      rewriter.clone(*lockOp, argMap);
       clonedSSAInsertPoint = rewriter.saveInsertionPoint();
     } else {
       return errorReportOp->emitError()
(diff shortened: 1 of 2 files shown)


lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -250,24 +279,25 @@ copyReferencedSSAValues(PatternRewriter &rewriter,
...
+      rewriter.replaceOpUsesWithIf(
+          definingOp, clonedTile->getResult(0), [&](OpOperand &operand) {
+            return operand.getOwner()->getParentOfType<AIE::DeviceOp>() ==
+                   callerDevice;
+          });
+
+    } else if (auto lockOp = llvm::dyn_cast<AIE::LockOp>(definingOp)) {
+      rewriter.restoreInsertionPoint(clonedSSAInsertPoint);
+      Operation *clonedLock = rewriter.clone(*lockOp, argMap);
+      argMap.map(definingOp->getResult(0), clonedLock->getResult(0));
```
fifield: `clone(Operation &, IRMapping &)` should add the new op to the map while also re-mapping any operands. I don't think the `argMap.map` on line 299 is needed.

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -477,19 +499,39 @@ struct AIEMaterializeRuntimeSequencesPass
...
+        PatternRewriter rewriter(runOp->getContext());
+        rewriter.setInsertionPoint(runOp);
+        llvm::DenseMap<SymbolRefAttr, SymbolRefAttr> previouslyInlinedSymbolMap;
+        if (failed(inlineRunOp(runOp, rewriter, previouslyInlinedSymbolMap,
+                               ssaDefInsertPoint, symbolDefInsertPoint,
+                               allSymbolNames))) {
+          runOp.emitError() << "Failed to inline aiex.run operation";
+          signalPassFailure();
+        }
+      });
```
fifield: Why did you need to change away from a simpler rewrite pattern here? Do you need to control the order?

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
andrej: Oh cool, will give that a try.

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
andrej: That was the initial issue, I wanted to keep `previouslyInlinedSymbolMap` persistent across multiple run operations, and I think you're not supposed to keep state across calls to `matchAndRewrite`. However it turned out that didn't work and I actually needed separate symbol definitions for each run, so this is a leftover. I will see if it still works after reverting that change, thanks for pointing it out.

andrej: Rolled back to a pattern rather than a loop. This changed the order in which run ops are visited and so had to update a test, but everything is still functional.

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
andrej: Yup, works without explicitly adding to the argMap, thanks for the pointer!

PR #2837, opened 2026-01-30
Add integration tests + fixes for aiex.dma_configure_task on mem tiles and core tiles

fifield:
* Add NPU XRT integration tests for dma_configure_task targeting:
 -- Mem tiles (row 1): basic and with issue_token
 -- Core tiles (row 2): basic and with issue_token

* Add fixes to make the tests work


lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -224,8 +225,19 @@ struct AIEDMATasksToNPUPass : AIEDMATasksToNPUBase<AIEDMATasksToNPUPass> {
...
+                                 (buf_addr / 4) << 14, 0x0fffc000, nullptr,
+                                 nullptr, nullptr);
+      } else if (target_model.isMemTile(col, row)) {
+        NpuMaskWrite32Op::create(builder, bd_op.getLoc(), register_addr,
+                                 buf_addr / 4, 0x0007FFFF, nullptr, nullptr,
+                                 nullptr);
+      } else {
+        NpuWrite32Op::create(builder, bd_op.getLoc(), register_addr, buf_addr,
+                             nullptr, nullptr, nullptr);
+      }
```
fifield, line 240: note: can probably delete this

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -224,8 +225,19 @@ struct AIEDMATasksToNPUPass : AIEDMATasksToNPUBase<AIEDMATasksToNPUPass> {
...
-                           nullptr, nullptr, nullptr);
+      buf_addr += bd_op.getOffsetInBytes();
+      if (target_model.isCoreTile(col, row)) {
+        NpuMaskWrite32Op::create(builder, bd_op.getLoc(), register_addr,
+                                 (buf_addr / 4) << 14, 0x0fffc000, nullptr,
+                                 nullptr, nullptr);
+      } else if (target_model.isMemTile(col, row)) {
+        NpuMaskWrite32Op::create(builder, bd_op.getLoc(), register_addr,
+                                 buf_addr / 4, 0x0007FFFF, nullptr, nullptr,
+                                 nullptr);
```
andrej, line 236: Does this masking/shifting translate the buffer address from a global address to the local address space of the tiles? Is any of this (masks, amount of shifting) architecture-specific and could go into the target model?

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
andrej: Not needed for the shim tiles? I guess not because the address patch overwrites it?

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
fifield: > Does this masking/shifting translate the buffer address from a global address to the local address space of the tiles? Is any of this (masks, amount of shifting) architecture-specific and could go into the target model?

The BD holds the "32-bit word address" which is the reason for the divide by 4. There should probably be a comment there. 

The amount of masking and shifting is specific to the architecture revision. This could already be looked up in the register database (via target model), and is probably not a bad idea.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
fifield: > Not needed for the shim tiles? I guess not because the address patch overwrites it?

Yes, but I take back my earlier comment because there are simulation use cases where a shim tile might use a BufferOp, but it is not common.

erwei-xilinx (approved):

PR #2838, opened 2026-01-31
Fixup issue related to mapping cloned lock ops

erwei-xilinx:


lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -295,13 +295,22 @@ copyReferencedSSAValues(PatternRewriter &rewriter,
...
+      Operation *clonedLock = rewriter.clone(*lockOp, argMap);
       clonedSSAInsertPoint = rewriter.saveInsertionPoint();
+
+      // Map old lock result to new lock result
+      argMap.map(definingOp->getResult(0), clonedLock->getResult(0));
+      rewriter.replaceOpUsesWithIf(
+          definingOp, clonedLock->getResult(0), [&](OpOperand &operand) {
+            return operand.getOwner()->getParentOfType<AIE::DeviceOp>() ==
+                   callerDevice;
+          });
```
fifield, line 304: Doesn't `clone` already add the new op to the map? I guess the issue with this code is that the some uses of the locks are cloned before the lock itself, so the extra replace fixups are needed. A future refactoring could simply clone the referenced ops as they are encountered, adding them to the map and moving on

erwei-xilinx: Agreed. The `argMap.map` is redundant as `rewriter.clone` should already handle that well. Code updated.

PR #2839, opened 2026-02-02
Enable `dma.configure.task` for configuring core/mem tile DMAs

erwei-xilinx:
Add e2e unit tests for lowering `dma.configure.task` to core/memtile DMA configuration.

### DIFF
--- a/test/npu-xrt/memtile_dmas/dma_configure_task/aie.mlir
+++ b/test/npu-xrt/memtile_dmas/dma_configure_task/aie.mlir
@@ -1,0 +1,71 @@
+//===- aie.mlir ------------------------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// This test demonstrates using aiex.dma_configure_task to configure memtile
+// DMAs at runtime. Functionally equivalent to the writebd test.
+//
+// Data flow:
+//   Host -> Shim MM2S -> Memtile S2MM -> Buffer -> Memtile MM2S -> Shim S2MM -> Host
+
+module {
+  aie.device(npu1_1col) {
+    %shim_tile = aie.tile(0, 0) {controller_id = #aie.packet_info<pkt_type = 0, pkt_id = 1>}
+    %mem_tile = aie.tile(0, 1)
+    
+    // Local buffer in the memtile - address 0
+    %buffer = aie.buffer(%mem_tile) {address = 0 : i32, mem_bank = 0 : i32, sym_name = "buffer"} : memref<4096xi32>
+    
+    // Locks for synchronization between S2MM and MM2S on memtile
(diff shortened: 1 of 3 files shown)


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2293,9 +2293,20 @@ LogicalResult UseLockOp::verify() {
   if (HasSomeParent<CoreOp, func::FuncOp>::verifyTrait(*this).succeeded()) {
     return success();
   }
+  // Or it can be in a DMAConfigureTaskOp (for runtime DMA configuration)
+  // Check by operation name to avoid circular dependency with AIEX dialect
+  {
+    Operation *operation = (*this)->getParentOp();
+    while (operation) {
+      if (operation->getName().getStringRef() == "aiex.dma_configure_task")
```
fifield, line 2301: Use `llvm::isa<>` here

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -216,16 +237,21 @@ struct AIEDMATasksToNPUPass : AIEDMATasksToNPUBase<AIEDMATasksToNPUPass> {
...
-      NpuWrite32Op::create(builder, bd_op.getLoc(), register_addr, buf_addr,
-                           nullptr, nullptr, nullptr);
+      uint64_t buf_addr = *buffer.getAddress();
+      // For memtile buffers, we need to write the buffer address to the BD's
+      // address register. For core tile buffers, the address is encoded
+      // differently and NpuWriteBdOp handles it - no separate write32 needed.
+      if (target_model.isMemTile(tile.getCol(), tile.getRow())) {
+        NpuWrite32Op::create(builder, bd_op.getLoc(), register_addr, buf_addr,
+                             nullptr, nullptr, nullptr);
+      }
```
fifield, line 253: Can we move the memtile write32 into the WriteBdOp too? I don't see any reason why it's different than core tile address. Or use the fix in https://github.com/Xilinx/mlir-aie/pull/2837 which does a maskwrite32 for both.

erwei-xilinx:
> This PR adds lock support to dma tasks while #2837 does more extensive testing of task-compete-tokens and BD offsets. Let's combine them.

Sounds great to me. This PR also adds support for having locks inside the `dma.configure.task` so that should also be included in the https://github.com/Xilinx/mlir-aie/pull/2837. Happy to close this PR and cherry pick useful features from here to that PR.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
erwei-xilinx: Agreed. The fix in #2837 is more elegant.

lib/Dialect/AIE/IR/AIEDialect.cpp
erwei-xilinx: I cannot do that unfortunately, because this file must not include AIEXDialect, otherwise there is going to be a circular dependency (AIEXDialect depends on AIEDialect).

PR #2841, opened 2026-02-02
Change command to update pip

hunhoffe:
I found the following error in the CI only for some windows commands:
```
Run pip install --upgrade pip
Requirement already satisfied: pip in c:\hostedtoolcache\windows\python\3.10.11\x64\lib\site-packages (25.3)
Collecting pip
  Downloading pip-26.0-py3-none-any.whl.metadata (4.7 kB)
Downloading pip-26.0-py3-none-any.whl (1.8 MB)
   ---------------------------------------- 1.8/1.8 MB 49.2 MB/s  0:00:00

Notice:  A new release of pip is available: 25.3 -> 26.0
Notice:  To update, run: python.exe -m pip install --upgrade pip
ERROR: To modify pip, please run the following command:
C:\hostedtoolcache\windows\Python\3.10.11\x64\python.exe -m pip install --upgrade pip
```

I suspect this is because, due to file locks in Windows, pip can't necessarily update itself without the more full command. So I updated instances of:
```
pip install --upgrade pip
```
to
```
python -m pip install --upgrade pip
```
in the workflow actions.


erwei-xilinx:
> Ahha! So it may be a transient issue. Thanks for the heads up @erwei-xilinx !

I think the original CI issue on windows already exists before this incident, so your fix is going to be useful. I think it's just that we won't be able to run any CI tests unless the GitHub incident is resolved.

PR #2848, opened 2026-02-04
Try to debug "device not found" error

hunhoffe:
Over the last few days, "device not found" errors have been more prevalent in the CI. I've tried on equivalent architecture  of both CI runners to run stress tests simulating the concurrency, etc. that could be happening in the CI runners using the `CachedXRTRuntime` class, but I've been unable to trigger this error locally.

This PR does a few things:
* It adds an exponential backoff timeout for fetching the device up to a set number of retries
* It adds some debug logging that will attempt to be run should this error occurs. I'm hopeful this may shed some light on the root cause of this problem, since I am still unsure.

### DIFF
--- a/python/utils/hostruntime/xrtruntime/hostruntime.py
+++ b/python/utils/hostruntime/xrtruntime/hostruntime.py
@@ -12,6 +12,8 @@ import os
 import time
 import weakref
 import gc
+import sys
+import subprocess
 from pathlib import Path
 from typing import TYPE_CHECKING
 import numpy as np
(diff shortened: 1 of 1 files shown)


andrej (approved):
For even more info, maybe there's something useful from the XDNA driver. You could ask Jeff to enable verbose debugging longs from the XDNA driver on the runners:
```
modprobe -r amdxdna
modprobe amdxdna dyndbg=+pf
```
And add
```
dmesg | grep "xdna" | tail -n100
```
to the logging part of your script.

PR #2849, opened 2026-02-05
Fixup device symbol ref with runtime sequence materialization

erwei-xilinx:
Fixup a failure when runtime sequence fails to inline when the child op makes reference to device symbol.

For example, if a `load_pdi` op is present inside the device op's runtime_sequence and needs to be inlined to another device op. The load_pdi's symbol reference cannot be lost.


lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -129,6 +129,19 @@ struct InsertLoadPdiForConfigurePattern : RewritePattern {
...
+    // If so, don't add another one.
+    if (!configureBlock->empty()) {
+      if (auto existingLoadPdi =
+              llvm::dyn_cast<AIEX::NpuLoadPdiOp>(&configureBlock->front())) {
+        if (existingLoadPdi.getDeviceRef() ==
+            referencedDevice.getSymNameAttr()) {
+          return failure();
+        }
+      }
+    }
```
andrej: Can you elaborate on why this is needed?

erwei-xilinx: I think after calling this pass, the "main" device's runtime_sequence will always start with a `load_pdi` as the initial configuration. If the "callee" device op's runtime_sequence also starts with a runtime_sequence (and with the same sym name), then it would end up having two consecutive identical `load_pdi` ops at the start.

A longer term solution could be a canonicalizer on `load_pdi`, that removes any immediately adjacent clones.

andrej: I see, makes sense. Since load_pdi resets the entire partition, any two back-to-back load_pdis really don't make sense and only the second load_pdi will take effect (except for the effect that the first load_pdi will slow things down). Maybe a more general solution than just deleting the first load_pdi in a sequence would be to delete all but  the last load_pdi in any sequence of load_pdis without any other ops in between.

erwei-xilinx: Ok I removed the above logic, and replaced with a call to a canonicalizer of load_pdi, which now greedily checks and removes back-to-back reconfigurations (of the same config).

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -1010,3 +1010,32 @@ LogicalResult AIEX::RunOp::verify() {
...
+
+  // Check if they are identical (all attributes match)
+  if (op.getDeviceRefAttr() == nextLoadPdi.getDeviceRefAttr() &&
+      op.getId() == nextLoadPdi.getId() &&
+      op.getSize() == nextLoadPdi.getSize() &&
+      op.getAddress() == nextLoadPdi.getAddress()) {
+    // Erase the first one, keeping the second
+    rewriter.eraseOp(op);
+    return success();
+  }
```
andrej, line 1038: I'm not sure they have to be identical? Maybe I'm missing some case. But  if you do 
```
load_pdi(foo)
load_pdi(bar)
```
Only `bar` will be loaded, `foo` will be overwritten. Since you're not doing anything with `foo` before it is overwritten, might was well never configure `foo`. No?

erwei-xilinx: Yeah I agree that's what will happen eventually. Keeping both load_pdi's there might be helpful for debugging the design though, because if load-foo-load-bar happens, that could indicate a failure elsewhere.

PR #2858, opened 2026-02-10
Tensor torch bfloat16 conversion

hunhoffe:
The previous implementation was buggy; this PR adds tests that previously would have fail, and fixes the bug.

Thanks for the correct implementation @andrej !


andrej (approved):
LGTM

PR #2873, opened 2026-02-17
Implement aiecc compiler driver in C++ 

jgmelber:
This pull request introduces a new C++ implementation of the AIE compiler driver, `aiecc`, and integrates it into the build and test infrastructure. The changes add the driver, its documentation, CMake build instructions, and several comprehensive tests (including multi-device and NPU instruction generation scenarios). The Makefile for an example is updated to use the new C++ tool instead of the Python script.

**Summary of the most important changes:**

### New C++ AIE Compiler Driver

* Added the new `aiecc` C++ compiler driver, including its CMake build instructions (`tools/aiecc/CMakeLists.txt`), integration into the main build (`tools/CMakeLists.txt`), and a detailed user and developer documentation (`tools/aiecc/README.md`). [[1]](diffhunk://#diff-28d880e96094ee96843625d4206ef9b61730c9d5fe596021fb4d1fab95352747R1-R34) [[2]](diffhunk://#diff-70523fd3fe726ff8a7cba389b1deb5a5b761e02ebefe2fad6a1531d529e440dcR7) [[3]](diffhunk://#diff-70523fd3fe726ff8a7cba389b1deb5a5b761e02ebefe2fad6a1531d529e440dcR16) [[4]](diffhunk://#diff-691eff19303b15aa40c6ce0dcd54338dc3dab577d6387b0ff3daefde12598d3fR1-R217)

### Test Infrastructure and Coverage

* Introduced new tests for `aiecc` covering basic compilation, multi-device and sequence filtering, and NPU instruction generation, including MLIR input files and corresponding test scripts (`test/aiecc/cpp_basic.mlir`, `test/aiecc/cpp_multi_device_sequence.mlir`). [[1]](diffhunk://#diff-4b02954d2164060d781fec9d81004262a03baea2b39b2c3e52a1790aff75c8baR1-R67) [[2]](diffhunk://#diff-965b8f222a2a25d6a56476536eecbadbc80094c27ed06083b9df7513f910927eR1-R90)
* Added a full end-to-end test for generating and running xclbin/NPU instructions on hardware, including the MLIR design, a test harness in C++, and a LIT run script (`test/npu-xrt/add_one_cpp_aiecc/aie.mlir`, `test/npu-xrt/add_one_cpp_aiecc/test.cpp`, `test/npu-xrt/add_one_cpp_aiecc/run.lit`). [[1]](diffhunk://#diff-0edae0384a5f32c5bf7f49ef6bb58094c760f97beb29b8d6c639f3dae4476c45R1-R58) [[2]](diffhunk://#diff-a6aac8dfb099e9565a80de3699db0b96024f6868d7c90430489411d385923a05R1-R143) [[3]](diffhunk://#diff-89b9e7dd0d4c7adca1abcb410cb2f42476da764c89b1327dc6d5663e22b080e3R1-R12)

### Build System and Example Updates

* Updated the passthrough DMA example Makefile to use the new `aiecc` C++ tool instead of the Python `aiecc.py` script, reflecting the migration to the new driver (`programming_examples/basic/passthrough_dmas/Makefile`).

---

These changes collectively provide a performant C++ alternative to the Python AIE compiler driver.

### DIFF
--- a/tools/aiecc/aiecc.cpp
+++ b/tools/aiecc/aiecc.cpp
@@ -566,19 +566,28 @@ buildInputWithAddressesPipeline(StringRef aieTarget = "aie2") {
   std::ostringstream oss;
   oss << "builtin.module("
       << "convert-vector-to-aievec{aie-target=" << aieTarget.lower()
-      << " target-backend=llvmir}," << "lower-affine,"
-      << "aie-canonicalize-device," << "aie.device(" << "aie-assign-lock-ids,"
-      << "aie-register-objectFifos," << "aie-objectFifo-stateful-transform{"
+      << " target-backend=llvmir},"
+      << "lower-affine,"
+      << "aie-canonicalize-device,"
+      << "aie.device("
+      << "aie-assign-lock-ids,"
+      << "aie-register-objectFifos,"
+      << "aie-objectFifo-stateful-transform{"
       << "dynamic-objFifos=" << (dynamicObjFifos ? "true" : "false")
       << " packet-sw-objFifos=" << (packetSwObjFifos ? "true" : "false") << "},"
-      << "aie-assign-bd-ids," << "aie-lower-cascade-flows,"
-      << "aie-lower-broadcast-packet," << "aie-lower-multicast,"
+      << "aie-assign-bd-ids,"
+      << "aie-lower-cascade-flows,"
+      << "aie-lower-broadcast-packet,"
+      << "aie-lower-multicast,"
       << "aie-assign-tile-controller-ids,"
       << "aie-generate-column-control-overlay{route-shim-to-tile-ctrl="
       << (ctrlPktOverlay ? "true" : "false") << "},"
       << "aie-assign-buffer-addresses{alloc-scheme=" << allocScheme.getValue()
-      << "}," << "aie-vector-transfer-lowering{max-transfer-rank=1}" << "),"
-      << "convert-scf-to-cf" << ")";
+      << "},"
+      << "aie-vector-transfer-lowering{max-transfer-rank=1}"
+      << "),"
+      << "convert-scf-to-cf"
+      << ")";
   return oss.str();
 }
 
@@ -591,10 +600,17 @@ static std::string buildLLVMLoweringPipeline(StringRef deviceName,
       << "aie-standard-lowering{device=" << deviceName.str() << "},"
       << "aiex-standard-lowering,"
       << "convert-aievec-to-llvm{aie-target=" << aieTarget.lower() << "},"
-      << "canonicalize," << "cse," << "expand-strided-metadata,"
-      << "lower-affine," << "arith-expand," << "finalize-memref-to-llvm,"
+      << "canonicalize,"
+      << "cse,"
+      << "expand-strided-metadata,"
+      << "lower-affine,"
+      << "arith-expand,"
+      << "finalize-memref-to-llvm,"
       << "convert-func-to-llvm{use-bare-ptr-memref-call-conv=true},"
-      << "convert-to-llvm{dynamic=true}," << "canonicalize," << "cse" << ")";
+      << "convert-to-llvm{dynamic=true},"
+      << "canonicalize,"
+      << "cse"
+      << ")";
   return oss.str();
 }
 


tools/aiecc/README.md
```diff
@@ -0,0 +1,217 @@
...
+## Dependencies
+
+- MLIR/LLVM libraries
+- AIE dialect libraries (AIE, AIEX)
+- External tools (found in PATH or relative to aiecc):
+  - `aie-opt` - For running MLIR passes
+  - `aie-translate` - For MLIR to binary translation
+  - `bootgen` - For PDI generation (optional)
+  - `xclbinutil` - For xclbin generation (optional)
+  - `xchesscc` - For core compilation (optional)
```
fifield, line 194: `aiecc.py` has been evolving to do mlir passes and translations (`aie-opt` and `aie-translate`) in memory, via the python bindings to the pass manager and various translations (e.g. `translate_npu_to_binary`). There are still calls out to `aie-translate` in aiecc.py, but `aie-opt` has been eliminated, with output to disk optional. Is the any obstacle to a C++ version of `aiecc` doing the same? It seem to me a goal of a C++ `aiecc` should be to not write to disk if possible.

jgmelber: Yes, agreed. This started with vibe coding, then I wanted to actually make it work. I'll iterate on the implementation.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+    llvm::errs() << "Error parsing lowered MLIR file\n";
+    return failure();
+  }
+
+  // Find device
+  SmallVector<Operation *, 1> devices;
+  findAIEDevices(module.get(), devices);
+
+  for (auto *deviceOp : devices) {
+    if (auto devOp = dyn_cast<xilinx::AIE::DeviceOp>(deviceOp)) {
```
andrej: Simplification: In C++ I think we can more tightly integrate with LLVM so we could just `module.walk()` or `module.getOps<DeviceOp>()` here (eliminate `findAIEDevices`, less code better). Also applies to `findRuntimeSequences` etc.

ypapadop-amd:
> Cool!
> 
> This currently doesn't support full ELF. If we're already doing a rewrite, we could consider dropping the old xclbin+insts flow and only supporting full ELF? There might also be other flows/passes in `aiecc.py` that are maybe obsolete by now?
> 
> Regarding enabling full ELF, there is some logic in `aiecc.py` that does some minimal IR modification (assign PDI IDs for the load_pdi instructions). This would have to be replicated here as well if we wanted to enable that flow.

Would that drop also the PDI + insts flow? ROCR still depends on it until we switch to ELF (planned, but no ETA as of now).

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+  }
+  return result;
+}
+
+//===----------------------------------------------------------------------===//
+// AIE Device and Core Discovery
+//===----------------------------------------------------------------------===//
+
+// Walk the module to find AIE device operations
+static void findAIEDevices(ModuleOp module,
```
ypapadop-amd: `module` is a keyword in C++20 and above. It's probably best to avoid it.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+      if (sequenceName.empty() ||
+          (seqOp.getSymNameAttr() && seqOp.getSymName() == sequenceName)) {
+        sequences.push_back(op);
+      }
+    }
+  });
+}
+
+// Count cores in a device for progress reporting
+static unsigned countCoresInDevice(Operation *deviceOp) {
```
ypapadop-amd: C++ has been moving towards fixed width integrals, e.g., `std::uint32_t`

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+  });
+  return count;
+}
+
+//===----------------------------------------------------------------------===//
+// Pass Pipeline Construction
+//===----------------------------------------------------------------------===//
+
+static std::string buildInputWithAddressesPipeline() {
+  std::string pipeline = "builtin.module(";
```
ypapadop-amd: This causes multiple allocations as the sting is built. It's probably better to change it to a single multi-line string with a single concat at the end or use `std::stringstream`.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+  pipeline +=
+      "aie-assign-buffer-addresses{alloc-scheme=" + allocScheme.getValue() +
+      "},";
+  pipeline += "aie-vector-transfer-lowering{max-transfer-rank=1}";
+  pipeline += "))";
+  return pipeline;
+}
+
+static std::string buildNpuLoweringPipeline() {
+  std::string pipeline = "builtin.module(aie.device(";
```
ypapadop-amd: This can be a string_view.

tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1040 @@
...
+
+  SmallVector<StringRef, 8> cmd;
+  for (const auto &str : cmdStrs) {
+    cmd.push_back(str);
+  }
+
+  if (!executeCommand(cmd)) {
+    llvm::errs() << "Error running NPU lowering passes\n";
+    return failure();
+  }
```
ypapadop-amd: This looks like an odd pattern: string -> vector of strings -> vector of stringrefs. Maybe `executeCommand` can be revised with a different input argument type?

PR #2874, opened 2026-02-17
Interface and trait refactoring

yenjames:
### Summary

This PR refactors AIE interface and traits to reduce redundancy.

### Changes
- Removed manual code verification for tile type checks which are already covered by tile type traits.
- `FlowEndpoint` changed to a trait because it functions more like a "marker" trait rather than an interface.
- Removed duplicate `colIndex()`, `rowIndex()` from individual ops, now provided as interface methods.
- Changed `Interconnect` interface to inherit `TileElement` for less redundancy.
- `getTileOp()` for TileElement-like Ops is tricky to promote into InterfaceMethod, appears that it is due to TileElement inheriting DeclareOpInterfaceMethods for Asm; we have a circular dependency where it needs `TileOp` to declare the function.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -212,7 +212,7 @@ def AIE_SwitchboxOp: AIE_Op<"switchbox", [
     size_t getNumDestConnections(WireBundle bundle);
     void getAsmResultNames(
         llvm::function_ref<void(mlir::Value, llvm::StringRef)> setNameFn) {
-      return ::xilinx::AIE::TileElement::Trait<SwitchboxOp>::getAsmResultNames(setNameFn);
+      ::xilinx::AIE::TileElement::Trait<SwitchboxOp>::getAsmResultNames(setNameFn);
     }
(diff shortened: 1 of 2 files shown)


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -207,14 +207,12 @@ def AIE_SwitchboxOp: AIE_Op<"switchbox", [
...
+    TileOp getTileOp();
     size_t getNumSourceConnections(WireBundle bundle);
     size_t getNumDestConnections(WireBundle bundle);
-    int colIndex();
-    int rowIndex();
-    TileOp getTileOp();
     void getAsmResultNames(
         llvm::function_ref<void(mlir::Value, llvm::StringRef)> setNameFn) {
-      ::xilinx::AIE::TileElement::Trait<SwitchboxOp>::getAsmResultNames(setNameFn);
+      return ::xilinx::AIE::TileElement::Trait<SwitchboxOp>::getAsmResultNames(setNameFn);
```
yenjames: Remove return to keep it the same as before

PR #2875, opened 2026-02-17
Change XDNA driver commit

andrej:
The commit I put in there a bit ago already 404s for the firmware version when I just tried rebuilding the driver. This is the `1.7` tag which also seems to work and might be more stable in terms of firmware version. It seems to include full ELF support (why we needed a newer driver/XRT in the first place) but using this PR to run the CI and fully test.

### DIFF
--- a/utils/build_drivers.sh
+++ b/utils/build_drivers.sh
@@ -93,7 +93,8 @@ fi
 
 echo "Setting up XDNA driver repository..."
 # Clone or update the XDNA driver repository and initialize submodules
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Thanks for updating the driver tag, @andrej. Does this version include the `pyxrt` support on full-elf?

andrej:
Yes it should. They updated the XRT submodule and as far as I can tell it includes my commit to enable that. Let's see how CI does.

PR #2880, opened 2026-02-19
Add initial C++ aiecc compiler driver

jgmelber:
Implement a high-performance C++ version of the aiecc compiler driver that eliminates all subprocess calls to aie-opt and aie-translate by using direct MLIR C++ APIs.

Key features:
- Full in-memory MLIR compilation pipeline using PassManager API
- Direct C++ function calls for all translation operations
- External object file linking via link_with attribute
- Support for Peano toolchain compilation
- NPU instruction and xclbin generation

In-memory pipeline functions:
- runResourceAllocationPipeline(): Lock assignment, ObjectFIFO transform, buffer address allocation, BD assignment
- runRoutingPipeline(): Pathfinder flow routing
- runLLVMLoweringPipeline(): Per-core LLVM lowering with options support for aie-standard-lowering, convert-aievec-to-llvm, etc.
- runNpuLoweringPipeline(): BD chain materialization, DMA to NPU conversion

Direct translation APIs (no subprocess):
- AIETranslateToTargetArch(): Device target detection
- AIETranslateToLdScript(): Linker script generation
- AIETranslateNpuToBinary(): NPU instruction generation
- AIETranslateToCDODirect(): CDO file generation
- translateModuleToLLVMIR(): MLIR to LLVM IR translation

Performance improvements:
- Zero subprocess calls to aie-opt or aie-translate
- Minimal disk I/O (only LLVM IR files for Peano toolchain)
- Single MLIR parse, module stays in memory throughout compilation

Also adds:
- Options-based pass creation for AIEAssignBufferAddressesPass
- Options-based pass creation for AIECoreToStandardPass
- Options-based pass creation for ConvertAIEVecToLLVMPass
- LIT tests for C++ aiecc functionality
- NPU-XRT integration test

TODO:
- Extend support for calling xchesscc 
- Migrate all tests to use this rather than Python driver
- Deprecate aiecc.py or bind that interface to this implementation

### DIFF
--- a/aie_kernels/aie2p/softmax.cc
+++ b/aie_kernels/aie2p/softmax.cc
@@ -5,7 +5,7 @@
 // See https://llvm.org/LICENSE.txt for license information.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// Copyright (C) 2025, Advanced Micro Devices, Inc.
+// Copyright (C) 2026, Advanced Micro Devices, Inc.
 //
 //===-------------------------------------------------- --------===//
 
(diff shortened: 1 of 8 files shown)


tools/aiecc/aiecc.cpp
```diff
@@ -0,0 +1,1961 @@
...
+}
+
+//===----------------------------------------------------------------------===//
+// AIE Device and Core Discovery
+//===----------------------------------------------------------------------===//
+
+struct CoreInfo {
+  std::int32_t col;
+  std::int32_t row;
+  std::string linkWith; // External object files to link
```
ypapadop-amd, line 554: Can / will we support multiple files to link against? Or would that be handled by a single string?

ypapadop-amd (approved):
LGTM; one comment about if we should support multiple linked files.

tools/aiecc/aiecc.cpp
jgmelber: This PR preserves existing functionality but there's definitely opportunity for improving the linking process

PR #2898, opened 2026-02-26
Add C ObjectFIFO API for kernel-managed lock synchronization

jgmelber:
Enable precompiled C kernels to directly manage ObjectFIFO synchronization by receiving lock IDs and buffer references from MLIR, rather than relying on the compiler to insert acquire/release ops.

New MLIR ops:
- aie.objectfifo.lock: returns (acq_lock, rel_lock) for a port
- aie.objectfifo.buffer: returns a buffer memref at a given index

Stateful transform lowering resolves these ops to concrete lock/buffer SSA values, handling AIE1 single-lock and AIE2/AIE2P dual-lock semantics.

C API header (aie_runtime_lib/AIE2{,P}/aie_objectfifo.h): objectfifo_port_t struct with inline acquire/release functions using compiler-provided lock intrinsics. Use acq_value=-1 for AcquireGreaterEqual semantics matching the DMA protocol.

Python API: get_lock()/get_buffer() on object_fifo and ObjectFifoHandle.

Bug fixes discovered during hardware bring-up:
- python/dialects/aie.py: get_lock() passed incorrect extra args to ObjectFifoGetLockOp auto-generated constructor
- AIELocalizeLocks: used getParentOp() == coreOp which failed for lock uses nested inside scf.for; fixed to use isProperAncestor()

Includes:
- CMake install rules for aie_objectfifo.h into install tree
- LIT tests for lock localization with external function calls
- E2E NPU tests (lock_pass_to_c, objectfifo_lock_c_api)
- Programming example: passthrough_kernel_c_objfifo

### DIFF
--- a/aie_runtime_lib/AIE2P/aie_objectfifo.h
+++ b/aie_runtime_lib/AIE2P/aie_objectfifo.h
@@ -10,7 +10,8 @@
 //
 // C API for ObjectFIFO operations in AIE2P kernels.
 // Provides a self-contained struct that bundles locks, buffers, and depth,
-// hiding the dual-lock (producer + consumer) semantics of AIE2P semaphore locks.
+// hiding the dual-lock (producer + consumer) semantics of AIE2P semaphore
+// locks.
 //
 // On AIE2P, each ObjectFIFO element has two locks:
 //   - Producer lock (acq_lock for producer, rel_lock for consumer)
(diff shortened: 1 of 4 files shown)


jgmelber:
> Did you consider instead of `objectfifo_port_t` an `objectfifo_t` which also includes the buffers? This could be a defacto C compatible ABI for passing around object fifos analogous to the numpy array struct format memrefs lower to.
> 
> ```python
>             # C kernel owns the compute loop and buffer rotation
>             passThroughLine(of_out.get_value())
> ```
> 
>  ```c++
> typedef struct {
>   // some well defined layout
>   // ports ...
>   // buffers ...
> } objectfifo_t;
> 
> void passThroughLine(objectfifo_t *of) {
>     // ...
> 
>     objectfifo_acquire(Consume, of);
>     objectfifo_acquire(Produce, of);
> ```

Yes, I was thinking of something along those lines. It gets messy handling those buffers from the user perspective. The struct should also know how deep the objFIFO is so that it can do the `%` or better a `+/-` to access the objects in the correct order.

PR #2900, opened 2026-02-26
Introduce placement pass

yenjames:
### Summary

This PR implements the `--aie-place-tiles` pass with sequential placer algorithm, ~~completing logical-to-physical tile mapping~~.

### Changes
- New pass definition that can be extended with other placer algorithm.
- Follows structure similar to AIEPathfinder pass.
- Not registered to `aiecc` for now.
- Removed "Pure" class from aie.logical_tile, noticed canonicalization pass was removing logical_tiles because they appear identical in mlir.

### Future PR
- Enhance placer to handle ShimPL tiles.
- Consolidate with ObjectFifo refactor plans.
---

**Update (2026-03-18):** Change plans to merge this PR first as it's self isolated. Will introduce this pass without registering to `aiecc.cpp` and part of lowering pipeline. This pass has been verified with local tests.

---
**Update (2026-03-24):** 
Review Changes:
- Placer as a typed enum option.
- Support partially constrained tiles with sequential placer. These are provided a "filtered" list of candidate tiles for them to map to.
- Additional tests: mem/shim/compute tile exhaustion, `cores-per-col` exceeding device capacity, exact capacity fitting.
- Added checks to tests to ensure tiles are actually merged (by checking next tile in coordinate system wasn't created).


lib/Dialect/AIE/Transforms/AIEPlaceTiles.cpp
```diff
@@ -0,0 +1,107 @@
...
+    : xilinx::AIE::impl::AIEPlaceTilesBase<AIEPlaceTilesPass> {
+  void runOnOperation() override {
+    DeviceOp device = getOperation();
+
+    // Create placer
+    std::shared_ptr<Placer> placer;
+    if (clPlacerName == "sequential_placer") {
+      std::optional<int> coresPerCol = std::nullopt;
+      if (clCoresPerCol >= 0)
+        coresPerCol = clCoresPerCol;
```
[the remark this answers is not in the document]
yenjames: I think unnecessary

test/place-tiles/unknown_placer.mlir
```diff
@@ -0,0 +1,20 @@
+//===- unknown_placer.mlir --------------------------------------*- MLIR -*-===//
```
yenjames: Rename following test_* convention

lib/Dialect/AIE/Transforms/AIEPlacer.cpp
```diff
@@ -0,0 +1,577 @@
...
+      // Only remove fully constrained compute tiles from availability.
+      // Mem/Shim may still host additional logical tiles as long as
+      // channel/DMA capacity permits.
+      if (logicalTile.getTileType() == AIETileType::CoreTile)
+        availability.removeTile(tile, logicalTile.getTileType());
+      continue;
+    }
+
+    // Place unconstrained compute tiles sequentially
+    // NOTE: Partial constraints (col-only or row-only) are currently ignored
```
hunhoffe: I think this limitation needs to be better documented. Is this fix something that belongs in this PR? Or is there a reason to defer it to another PR? I'm ok with either option, as long as it's a clearly documented limitation and there is a plan in place.

lib/Dialect/AIE/Transforms/AIEPlacer.cpp
```diff
@@ -0,0 +1,577 @@
...
+              }
+            }
+          }
+        }
+      }
+    }
+
+    // Rounded average of core columns
+    if (totalCoreEndpoints > 0) {
+      groupCommonCol = (sumCols + totalCoreEndpoints / 2) / totalCoreEndpoints;
```
hunhoffe, line 257: With how we average, just to sanity check, we could use the last column, yes?

lib/Dialect/AIE/Transforms/AIEPlaceTiles.cpp
```diff
@@ -0,0 +1,99 @@
...
+};
+
+struct AIEPlaceTilesPass
+    : xilinx::AIE::impl::AIEPlaceTilesBase<AIEPlaceTilesPass> {
+  void runOnOperation() override {
+    DeviceOp device = getOperation();
+
+    // Create placer
+    std::shared_ptr<Placer> placer;
+    if (clPlacerName == "sequential_placer") {
```
hunhoffe: Should this be a type and not a string?

test/place-tiles/sequential_placer/test_place_basic.mlir
```diff
@@ -0,0 +1,155 @@
...
+    aie.objectfifo @of1 (%shim, {%mem}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+    // CHECK: aie.objectfifo @of2(%[[MEM]], {%[[CORE]]}, 2 : i32)
+    aie.objectfifo @of2 (%mem, {%core}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+    // CHECK-NOT: aie.logical_tile
+  }
+}
+
+// -----
+
+// Partial constraints are currently ignored by sequential placer
```
hunhoffe: Shouldn't this be an XFAIL? Because this behavior is known but not correct? Same with below.

test/place-tiles/sequential_placer/test_place_errors.mlir
```diff
@@ -0,0 +1,87 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: not aie-opt --split-input-file --aie-place-tiles %s 2>&1 | FileCheck %s
```
hunhoffe, line 11: Are there any other error cases you need to check?

test/place-tiles/sequential_placer/test_place_objectfifo.mlir
```diff
@@ -0,0 +1,96 @@
...
+    aie.objectfifo @broadcast (%shim, {%core1, %core2}, 2 : i32) : !aie.objectfifo<memref<16xi32>>
+
+    aie.core(%core1) { aie.end }
+    aie.core(%core2) { aie.end }
+    // CHECK-NOT: aie.logical_tile
+  }
+}
+
+// -----
+
```
hunhoffe, line 77: I think there need to be a few more tests about column grouping, exceeding channels (so place near column group)

test/place-tiles/sequential_placer/test_place_basic.mlir
```diff
@@ -0,0 +1,155 @@
+//===- test_place_basic.mlir -----------------------------------*- MLIR -*-===//
+//
```
hunhoffe, line 2: There might need to be a test for "exactly fills capacity" in various ways. It's good to double check the edge cases.

test/place-tiles/sequential_placer/test_place_options.mlir
```diff
@@ -0,0 +1,38 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: aie-opt -split-input-file --aie-place-tiles='cores-per-col=2' %s | FileCheck %s
+// RUN: aie-opt -split-input-file --aie-place-tiles %s | FileCheck %s --check-prefix=NO-LIMIT
+
```
hunhoffe, line 14: Should test error too -- if limit but request beyond limit?

test/place-tiles/test_placer_unknown.mlir
```diff
@@ -0,0 +1,20 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2026 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// RUN: not aie-opt --aie-place-tiles='placer=unknown_placer' %s 2>&1 | FileCheck %s
```
hunhoffe, line 11: If you don't give a placer argument at all, what happens?

lib/Dialect/AIE/Transforms/AIEPlacer.cpp
yenjames: Yes,  we search rightwards **starting** from the common (average) column until we find available tile.

test/place-tiles/test_placer_unknown.mlir
yenjames: Default pass option goes to sequential placer, and this will still be the case with enum placer.

lib/Dialect/AIE/Transforms/AIEPlacer.cpp
yenjames: Support now added for partial constraints as well

test/place-tiles/sequential_placer/test_place_options.mlir
yenjames: Missed validation for this, now added.

lib/Dialect/AIE/Transforms/AIEPlaceTiles.cpp
yenjames: Updated to enum type.

PR #2911, opened 2026-02-27
[aiecc] Skip core compilation for devices without ELFs

### LINKED ISSUE #2913: aiecc.py should support devices without core ELFs for lightweight PDI reload
### Use case

Between `air.launch` iterations, `load_pdi` reloads the full tile configuration including core ELF binaries (~1684 words per tile). Since the ELFs don't change between iterations, this reload is redundant and accounts for ~85% of the PDI data. Skipping it gives a **1.47x speedup** on flash attention (26.4ms → 18.0ms).

The ideal way to express this is a second device identical to the original but without core bodies — a "reset-only" device that only reconfigures DMA BDs, locks, switches, and core enables without reloading ELFs. The segment's runtime_sequence would have `load_pdi {device_ref = @my_segment_reset}` between iterations, generating a lightweight PDI without ELF data.

### Reproduction

`aiecc.py` tries to compile ALL CoreOps in ALL devices, including reset-only devices with empty cores (no `link_with`, no `elf_file`).

**Test input** (`test_empty_core.mlir`):
```mlir
module {
  aie.device(npu2) @reset_device {
    %tile_0_2 = aie.tile(0, 2)
    %lock_0_2 = aie.lock(%tile_0_2, 0) {init = 1 : i32}
    %buf = aie.buffer(%tile_0_2) {address = 1024 : i32, sym_name = "buf"} : memref<256xi32>

    %mem = aie.mem(%tile_0_2) {
      %0 = aie.dma_start(S2MM, 0, ^bb1, ^bb2, repeat_count = 3)
    ^bb1:
      aie.use_lock(%lock_0_2, AcquireGreaterEqual, 1)
      aie.dma_bd(%buf : memref<256xi32>, 0, 256) {bd_id = 0 : i32}
      aie.use_lock(%lock_0_2, Release, 1)
      aie.end
    ^bb2:
      aie.end
    }

    // CoreOp with empty body — present for initLocks (core reset +
    // lock zero-initialization) but has no program to compile.
    %core = aie.core(%tile_0_2) {
      aie.end
    }

    aie.runtime_sequence @seq() {
      aiex.npu.load_pdi {device_ref = @reset_device}
    }
  }
}
```

**Command:**
```bash
aiecc.py test_empty_core.mlir --no-aiesim --no-xchesscc --no-xbridge \
    --no-compile-host --generate-full-elf --expand-load-pdis \
    --peano $PEANO_INSTALL_DIR --tmpdir=/tmp/test_out -O 3
```

**Result:** `aiecc.py` calls `process_cores` (main.py:906) which tries to compile the empty core:
```
aiecc/main.py:1876  elf_paths[device_name] = await self.process_cores(
aiecc/main.py:906   in process_cores
aiecc/main.py:842   in chesshack
FileNotFoundError: ... chess-llvm-link
```

Even on a system with chess/Peano tools, `process_cores` compiles the empty core (generating an empty ELF) and `set_elf_file_for_core` (main.py:1088) re-adds the `elf_file` attribute, defeating the purpose of the reset-only device.

### What currently goes wrong

1. **`aiecc.py` recompiles all CoreOps**: The core compilation step at `main.py:1088` (`set_elf_file_for_core`) processes ALL devices and ALL CoreOps, including the reset-only device. It compiles the empty CoreOps (generating empty ELFs) and re-adds `elf_file` attributes, defeating the purpose.

2. **With `elf_file` restored, `addAieElfs` loads the ELFs**: Even though the ELFs are empty/trivial, the XAie API's tile iteration in `addAieElfs` sets up per-tile context that changes how `addInitConfig` serializes transactions — it generates blockwrite-style tile configs with core disable/channel pause wrappers instead of individual lock reset writes. The core disable interrupts the while-true loop mid-execution, causing a hardware timeout.

3. **Removing CoreOps entirely doesn't work either**: `initLocks` in `AIERT.cpp:536` gates lock zero-initialization on `tileOp.getCoreOp()`. Without CoreOps, locks aren't reset between iterations, causing a deadlock.

### Suggested changes

1. **`aiecc.py`**: Skip core compilation for CoreOps that have no `link_with` and no `elf_file`. This allows the reset-only device's empty CoreOps to pass through without getting `elf_file` assigned.

2. **`AIERT.cpp` `initLocks`**: Decouple lock zero-initialization from CoreOp presence. Initialize locks for all non-shim tiles, not just tiles with CoreOps. This allows reset-only devices to work without CoreOps if needed.

### Current workaround

PRs https://github.com/Xilinx/mlir-aie/pull/2911 and https://github.com/Xilinx/mlir-air/pull/1334 add a `skip_elfs` attribute to `npu.load_pdi` that skips `addAieElfs()` at the `generateTransactions` level, after `aiecc.py` has already compiled cores.

### Performance data

Flash attention (LQ=512, NUM_HEADS=12, LK=12288, 24 launch iterations):

| Metric | Full load_pdi | Without ELF reload | Improvement |
|--------|--------------|-------------------|-------------|
| Avg latency | 26,436 us | 18,038 us | 1.47x faster |
| Avg GFLOPS | 731 | 1,071 | 1.47x |

erwei-xilinx:
## Summary
- `aiecc.py`: Skip core compilation for CoreOps with empty bodies (no `elf_file`, no `link_with`, body is just `aie.end`)
- `AIERT.cpp`: Gracefully skip ELF loading for cores without `elf_file` in `addAieElfs`

This enables lightweight "reset-only" devices — clones of the segment device with empty CoreOps that only reconfigure DMA/lock/switch state without reloading core ELFs. Between `air.launch` iterations, `load_pdi` on the reset device generates a PDI ~85% smaller than the full version.

Resolves #2913.

## Performance impact

Flash attention (LQ=512, NUM_HEADS=12, LK=12288, 24 launch iterations):

| Metric | Before | After | Improvement |
|--------|--------|-------|-------------|
| Avg latency | 26,436 us | 18,038 us | **1.47x faster** |
| Avg GFLOPS | 731 | 1,071 | **1.47x** |

## Test plan
- [x] vector_vector_add_BDs_init_values example passes on NPU2 hardware
- [x] Flash attention LQ=128/256/512 passes on NPU2 hardware
- [x] error_require_lowered_core.mlir updated for graceful skip behavior
- [x] All CI build/test/lint/format jobs pass (59/60)

### DIFF
--- a/lib/Targets/AIERT.cpp
+++ b/lib/Targets/AIERT.cpp
@@ -990,10 +990,9 @@ LogicalResult xilinx::AIE::AIERTControl::addAieElfs(DeviceOp &targetOp,
         if (auto fileAttr = coreOp.getElfFile()) {
           fileName = fileAttr->str();
         } else {
-          coreOp.emitOpError()
-              << "Expected lowered ELF file to be given as attribute "
-                 "`elf_file` for this core. Compile cores first.";
-          return failure();
+          // Skip cores without elf_file (e.g., lightweight reset
+          // devices that only need DMA/lock reconfiguration).
+          continue;
         }
         // Check if fileName is already an absolute path.
         // If so, use it directly. Otherwise, concatenate with elfPath.
--- a/python/compiler/aiecc/main.py
+++ b/python/compiler/aiecc/main.py
@@ -336,6 +336,14 @@ def generate_devices_list(module):
     ]
 
 
+def _core_has_nonempty_body(core_op):
+    """Check if a CoreOp has a non-empty body (more than just aie.end)."""
+    for block in core_op.body:
+        if len(list(block)) > 1:
+            return True
+    return False
+
+
 def generate_cores_list(device_op):
     return [
         (
@@ -347,6 +355,9 @@ def generate_cores_list(device_op):
             device_op.operation,
             lambda o: isinstance(o.operation.opview, aiedialect.CoreOp),
         )
+        if c.elf_file is not None
+        or c.link_with is not None
+        or _core_has_nonempty_body(c)
     ]
 
 
--- a/test/Passes/expand-load-pdi/error_require_lowered_core.mlir
+++ b/test/Passes/expand-load-pdi/error_require_lowered_core.mlir
@@ -8,28 +8,31 @@
 //
 //===----------------------------------------------------------------------===//
 
-// RUN: aie-opt --verify-diagnostics --aie-expand-load-pdi %s
+// Cores without elf_file are gracefully skipped during PDI expansion.
+// This supports lightweight "reset-only" devices that have CoreOps for
+// lock/core initialization but no ELF to load.
 
-// This test should error as load_pdi operations can only be inlined after the
-// code in each core has been lowered to an ELF file and linked to the core
-// with the `elf_file` attribute.
+// RUN: aie-opt --aie-expand-load-pdi %s | FileCheck %s
 
 module {
 
     aie.device(npu2_1col) @my_core_device {
         %tile = aie.tile(0, 2)
-        // expected-error @+1 {{Expected lowered ELF file}}
+        %lock = aie.lock(%tile, 0) {init = 1 : i32}
+        // CoreOp without elf_file — should be skipped by addAieElfs
+        // but still trigger core reset/enable in initLocks/addCoreEnable
         aie.core(%tile) {
             aie.end
         }
     }
 
     aie.device(npu2_1col) @main {
+        // CHECK: aie.runtime_sequence
         aie.runtime_sequence (%arg0: memref<1xi32>) {
-            // expected-error @+1 {{Failed to generate configuration operations}}
+            // CHECK: aiex.npu.load_pdi
+            // CHECK: aiex.npu.write32
             aiex.npu.load_pdi { device_ref = @my_core_device }
         }
     }
-
 
 }


fifield:
Could you instead emit a second device without the core bodies (i.e. no elfs) and load_pdi on that?  It seems like `aiex.configure` and `aiex.run` make this easy(er) and I wonder how many reconfiguration special cases is the right number to encode in the compiler with boolean flags.

cc: @andrej

erwei-xilinx:
I tried the clone-device approach: clone the segment device, strip out the core bodies and `elf_file` attributes, and point the between-iteration `load_pdi` at the clone. This way the existing `load_pdi` expansion naturally skips ELF loading since there are no ELFs to load.

It works for single iteration but times out on multi-iteration. I haven't figured out why yet — the clone produces a much larger expanded output than expected (21k lines vs 6k with `skip_elfs`), so something in the expansion is duplicating work. I need to dig into this more.

The `skip_elfs` flag approach works for all configs. Should I keep investigating the clone-device approach, or is the flag acceptable?

fifield:
> I tried the clone-device approach. It doesn't work, and the reason is `aiecc.py`'s core compilation step.
> 
> **What I did:**
> 
> * Clone the segment device, remove core bodies and `elf_file` attributes, redirect between-iteration `load_pdi` to the clone
> 
> **Why it fails:**
> 
> * `aiecc.py` compiles ALL CoreOps in ALL devices as part of its flow (`set_elf_file_for_core` at main.py:1088). This re-adds `elf_file` attributes to the clone's CoreOps before `aie-expand-load-pdi` runs.
> * With `elf_file` restored, `addAieElfs` loads the (empty) ELFs, which sets up XAie per-tile context that changes how `addInitConfig` serializes transactions — it generates blockwrite-style tile configs with core disable/channel pause wrappers instead of individual lock reset writes. The core disable stops the while-true loop mid-execution, causing a timeout.

So this is the issue then: the CoreOp handling is broken for empty CoreOps.

> **Why not just remove CoreOps entirely from the clone?**
> 
> * `initLocks` (AIERT.cpp:536) gates lock zero-initialization on `tileOp.getCoreOp()`. Without CoreOps, locks aren't reset between iterations → deadlock.

So why not (re)set the locks in the MLIR instead of relying on the implicit behavior that could change? 

> **Why `skip_elfs` works:**
> 
> * It operates at the `generateTransactions` level, AFTER `aiecc.py` has already compiled cores and added `elf_file`. It simply skips the `addAieElfs()` call while keeping `addInitConfig` and `addCoreEnable` unchanged.
> 
> The clone-device approach would require modifying `aiecc.py` to skip core compilation for reset-only devices, which is a larger change. Would that be acceptable, or should we proceed with `skip_elfs`?

It sounds like aiecc needs some work to support this important use case cleanly. The input MLIR should be able express this explicitly without hacks in the compiler. Indeed this is one of the reasons for mlir-aie, to be able to express your program exactly. This PR is therefor a temporary workaround. Can you capture the use case in an issue?

PR #2922, opened 2026-03-04
func level `link_with`

hunhoffe:
## Summary

Removes the one-object-file-per-core restriction by moving `link_with` from `aie.core` to individual `func.func` declarations. The new `aie-assign-core-link-files` pass traces `func.call` edges from each core, aggregates referenced object files into a `link_files` array on the `CoreOp`, and both compiler backends (xbridge/Chess and lld/Peano) consume it.

## Changes

**Dialect:** `CoreOp` gains `link_files` (`StrArrayAttr`); `func.func` declarations carry `link_with`. Verifier rejects designs that set both on the same core.

**Pass `aie-assign-core-link-files`:** Traces direct `func.call` edges per core, deduplicates `link_with` values into `link_files`. Warns on indirect calls, unused `link_with` functions, and deprecated core-level `link_with`. Wired into the resource-allocation pipeline in `aiecc.cpp`.

**Backends:** BCF and ldscript emitters consume `link_files` with fallback to deprecated `link_with`. `atomicCopyFile` added for safe parallel-core `.o` copying.

**Python/IRON (breaking changes):**
- `Kernel(name, bin_name)` → `Kernel(name, object_file_name)`; `.bin_name` → `.object_file_name`
- `Core(link_with=...)` raises `TypeError`; set `link_with` on `external_func()` instead
- `tile_size()`, `arg_types()`, and arg-count validation lifted to `BaseKernel`
- `ExternalFunction`: removed `debug` parameter and no-op context manager
- `merge_object_files` / `link.py` removed (superseded by per-function `link_with`)
- Migrated `vector_exp`, `softmax` (×2), `color_detect`, `edge_detect` placed examples

## Tests

- `test/aiecc/`: 11 LIT pass tests covering canonical, multi-`.o`, dedup, deprecation, warnings, emitter fallback, no-op
- `test/npu-xrt/`: single-`.o` and two-`.o` end-to-end on NPU1/NPU2 (Chess + Peano)
- `test/python/`: `external_func(link_with=...)` MLIR output; JIT with two `ExternalFunction`s on one core

## Migration

```python
# Before
my_fn = external_func("my_fn", inputs=[...])
@core(tile, "my_kernel.o")
def core_body(): ...

# After
my_fn = external_func("my_fn", inputs=[...], link_with="my_kernel.o")
@core(tile)
def core_body(): ...
```

## Known limitations

* Only direct `func.call` edges are traced; transitive calls through helper functions are not followed automatically.
* Does not handle symbol collisions; that can be a separate PR.


programming_examples/basic/vector_exp/vector_exp_placed.py
```diff
@@ -60,7 +60,9 @@ def device_body():
 
         # AIE Core Function declarations
 
-        exp_bf16_1024 = external_func("exp_bf16_1024", inputs=[tile_ty, tile_ty])
+        exp_bf16_1024 = external_func(
+            "exp_bf16_1024", inputs=[tile_ty, tile_ty], link_with="kernels.a"
```
jgmelber, line 64: I admit I didn't dig into this example before writing this, but doesn't func-level `link_with` eliminate the need for archive files?

hunhoffe: My understanding of this example is that the archive format is a reflection of needing to bundle the function definition with a file of common helpers (e.g., kernel source requirement) and not a reflection of a need to archive multiple .o files due to mlir-aie.

PR #2924, opened 2026-03-05
[AIEX] Canonicalize contiguous NpuDmaMemcpyNdOp accesses to linear form

### LINKED ISSUE #2825: Linear transfers (without TensorAccessPattern) shouldn't use data layout transformation dimensions
Documenting this because I just ran into this:

If we have a large linear transfer of a type that is described as a multi-dimensional tensor, e.g. `(M, K)`, when using `iron.fill`, the lowered code ends up using the data layout transform dimensions:

```python
    ty = np.ndarray[(M, K,), dtype_in]
    with rt.sequence(ty) as (inp):
            rt.fill(fifo.prod(), B)
```
produces:
```mlir
        aie.dma_bd(%arg1 : memref<32x2048xbf16>, 0, 65536, [<size = 1, stride = 0>, <size = 1, stride = 0>, <size = 32, stride = 2048>, <size = 2048, stride = 1>]) {burst_length = 0 : i32}
        // (note that the two lowest data layout transformation dimensions are used.)
```

When one of the dimensions is too large (e.g., `K > 1024`), this leads to this error:
```
error: "-":29:9: 'aie.dma_bd' op Size 0 exceeds the [0:1023] range.
```

The hardware can be used in a "layout transformation" and "linear" mode. (For relevant special handling of linear transfers, grep for `isLinearTransfer` in the codebase.) In linear mode (I'm just calling it that, not sure it has an official name), the data layout transform dimensions are ignored and we can thus do larger, longer transfers without running into above error.

Currently, there are two workarounds:

1. Change the type to a single linear buffer, in the example above `ty = np.ndarray[(M*K,)], dtype_in]` rather than `ty = np.ndarray[(M, K,), dtype_in]`.
2. Use a simple linear TensorAccessPattern with the `fill`:  `B_tap = TensorAccessPattern(tensor_dims=(M,K,), offset=0, sizes=[1, 1, 1, num_batches * K], strides=[0, 0, 0, 1])`

I think it would be a usability improvement if for simple linear transfers (without any TAPs given), the second workaround was implemented as the default lowering.

cc @hunhoffe

hunhoffe:
Add a canonicalization pattern to `NpuDmaMemcpyNdOp` that rewrites a contiguous row-major access — `sizes=[s3,s2,s1,s0] strides=[st3,st2,st1,1]` where `st1==s0` and `st2==s0*s1` (with the usual "don't-care" exception when a size is 1) — to the canonical linear form `sizes=[s3,1,1,N] strides=[st3,0,0,1]` with `N=s0*s1*s2`. In linear form `isLinearTransferWithoutTransformation()` returns true, which bypasses the 10-bit `d0` wrap-size check and uses the hardware's wider buffer-length register instead. This fixes issue #2825, where fill/drain of a 2D buffer (e.g., `memref<M x K x bf16>`) generates `sizes=[1,1,M,K]` with `K>1023` and previously failed verification.

A correctness bug in the initial implementation was caught and fixed: the pattern was zeroing the d1/d2 strides without folding the corresponding offsets, causing `getOffsetInBytes()`— which computes `sum(offsets[i] * strides[i])` — to silently drop any non-zero byte offset contributed by a folded dimension. The fix collapses the innermost three offset/stride pairs into a single linear element index at `d0` before zeroing the intermediate strides. This bug produced wrong output in the `npu-xrt/packet_flow_fanout` and `npu-xrt/ctrl_packet_reconfig_4x1_cores` hardware tests.

The test suite covers: basic 2D and 3D folds; preservation of the repeat dimension, packet, issue_token, burst_length, metadata, and id attributes; negative cases for non-contiguous strides; idempotency; and regression cases for non-zero offsets in each folded dimension including the specific patterns from the two failing CI tests.

Closes #2825

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -915,9 +915,9 @@ LogicalResult AIEX::DMAStartBdChainOp::verify() {
   }
   for (unsigned i = 0, n = expectedArgTypes.size(); i < n; i++) {
     if (actualArgTypes[i] != expectedArgTypes[i]) {
-      return emitOpError("Argument ")
-             << (i + 1) << " types mismatch: " << "expected "
-             << expectedArgTypes[i] << " but got " << actualArgTypes[i];
+      return emitOpError("Argument ") << (i + 1) << " types mismatch: "
+                                      << "expected " << expectedArgTypes[i]
+                                      << " but got " << actualArgTypes[i];
     }
   }
   return success();


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -360,6 +360,94 @@ bool AIEX::NpuDmaMemcpyNdOp::isLinearTransferWithoutTransformation() {
...
+//   s1 == 1  ||  st1 == s0          (stride irrelevant when size is 1)
+//   s2 == 1  ||  st2 == s0 * s1
+// yielding a total of N = s0 * s1 * s2 contiguous elements.  The repeat
+// dimension s3 / stride st3 is unchanged by the fold.
+//
+// This fold is always semantically valid and never introduces new hardware
+// limit violations: in the resulting linear form, isLinearTransferWithout-
+// Transformation() returns true, so verifyStridesWraps() skips the 10-bit
+// d0 wrap-size check.  The hardware uses a wider transfer-length register in
+// linear mode, so arbitrarily large N is supported.
```
andrej: I think there are still limits but they are very large.

andrej (approved):
Nice clear PR description! I think this canonicalization is useful.

> A row-major contiguous access pattern such as sizes=[s3, s2, s1, s0] strides=[st3, st2, st1, 1]
where st1==s0 and st2==s0*s1 **(or the corresponding size is 1)**

For all tests, if size == 1 the corresponding stride == 0. Does this pass also canonicalize if the stride != 0? Do we want to canonicalize in that case, or perhaps error (since with a size of 1, the stride will never be applied, so might indicate user confusion)? Either way, can we add a test for some cases where size == 1 and stride != 0?

Tests for the DMA task syntax would be good too.

Is this same canonicalization also applied to the dimensions of an ObjectFifo (dimensionsToStream, dimensionsFromStream)?

aie_kernels/aie2/rgba2hue.cc
```diff
@@ -170,8 +170,14 @@ void rgba2hue_aie_scalar(uint8_t *rgba_in, uint8_t *hue_out,
...
 void rgba2hueLine(uint8_t *in, uint8_t *out, int32_t lineWidth) {
-  // rgba2hue_aie_scalar(in, out, 1, lineWidth);
+#ifdef __AIE2__
+  // Vectorized path: correct and fast on AIE2 (npu1).
   rgba2hue_aie(in, out, 1, lineWidth);
+#else
+  // AIE2P (npu2): the vectorized path produces incorrect results due to
+  // differences in acc32 SRS behaviour on AIE2P; use the scalar fallback.
+  rgba2hue_aie_scalar(in, out, 1, lineWidth);
+#endif
```
andrej, line 180: Change seems unrelated to compiler changes in the PR, maybe a Claude hallucination?

hunhoffe: Nope! This was 30 min of me scratching my head for why color detect failed on my dev machine but not the CI. I was doing local testing since I'd changed that design, and then it failed for both main and this branch. I could put this in a separate PR or remove it from the PR entirely since it's not a problem shown by the CI, but it is indeed real.

hunhoffe:
Moving back to draft because there's a bit more work I want to do before merge.

aie_kernels/aie2/rgba2hue.cc
hunhoffe: A few things:
1. the lit tests were missing RUN so this was previously not tested in the CI
2. I'm not sure if this is a good fix, but I'm glad it's at least running in the CI now.

include/aie/Dialect/AIEX/IR/AIEXDialect.h
```diff
@@ -56,10 +56,20 @@ bool isLinearTransfer(llvm::ArrayRef<int64_t> sizes,
...
+// is never applied during the transfer (the loop runs only once).
 // This is the vector-form counterpart of AIE::isContiguousBDTransfer.
 bool isContiguousTransfer(llvm::ArrayRef<int64_t> sizes,
                           llvm::ArrayRef<int64_t> strides);
 
+// Like isContiguousTransfer, but additionally requires every size-1 outer
+// dimension to have stride 0.  Use this when the BD descriptor physically
+// encodes those stride fields (e.g. aie.dma_bd in dma_configure_task), where
+// a non-zero stride in a size-1 dimension would occupy ND registers and
+// prevent the transfer from being lowered to flat (linear) mode.
```
andrej: Just trying to understand, why do we need this? What are the cases where we don't want to apply canonicalization because stride>1 but size=1, i.e. where would we use this function rather than `isContiguousTransfer`?

nit: duplicate "descriptor" on line 66

jgmelber (approved):

include/aie/Dialect/AIEX/IR/AIEXDialect.h
hunhoffe: I thought this function was needed because -- if canonicalize was not run -- you want to be extra careful to still catch linear transfers. But the more elegant way is to assume canonicalization (and add it to default flow). Thanks for bringing this up and for the second review!

PR #2925, opened 2026-03-05
Add remaining features to C++ aiecc and deprecate aiecc.py

jgmelber:
## Summary

Complete the C++ aiecc compiler driver to achieve full feature parity with the Python aiecc.py, then replace aiecc.py with a thin passthrough wrapper that delegates to the C++ binary.

### C++ aiecc features added

- **Parallel core compilation** (-j N / --nthreads N)
  - -j 0 auto-detects CPU count, -j 1 default (sequential)
  - Uses std::async for concurrent core compilation
- **Unified compilation** (--unified / --no-unified)
  - Compiles all cores into a single object file, links each separately
- **Host compilation** (--compile-host, -I, -L, -l, -o)
  - Invokes clang++ with auto-discovered runtime library paths (xaiengine, memory allocator)
  - Supports --host-target, --sysroot, --link_against_hsa
  - Accepts host source files (.cpp) as positional arguments alongside .mlir
- **AIE simulation** (--aiesim) with work folder generation
- **ELF generation** via aiebu-asm (--aie-generate-elf, --generate-full-elf)
- **AIEBU library integration** - direct ELF generation via C API when available, subprocess fallback
- **Bootgen library integration** - direct PDI generation via C API when available, subprocess fallback
- **Transaction binary** (--aie-generate-txn) and **control packet** (--aie-generate-ctrlpkt) generation
- **xclbin extension** (--xclbin-input) - merge new kernel/PDI into existing xclbin
- **PDI isolation fix** - PDI written to .prj tmpdir by default (CWD only with explicit --aie-generate-pdi), eliminating parallel test collisions
- **Repeater script generation** (--enable-repeater-scripts, --repeater-output-dir)
  - On pass pipeline failure, saves pre-transformation MLIR and generates a shell script to reproduce the failure with aie-opt
  - Captures MLIR diagnostics and includes them in the script for context
  - Enabled by default; all pipeline functions go through runPipelineWithRepeater()

### Python wrapper changes

- Replace 2,164-line aiecc.py with ~130-line thin passthrough wrapper
- All arguments passed directly to C++ aiecc binary unchanged (no filtering)
- Preserves deprecated run() API with warning for backward compatibility

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIETargetModel.h
+++ b/include/aie/Dialect/AIE/IR/AIETargetModel.h
@@ -292,6 +292,17 @@ public:
     return getNumBDs(getTileType(col, row));
   }
 
+  /// Return the number of buffer descriptors accessible on channel `channel`
+  /// for the tile at (`col`, `row`). For tiles with no per-channel BD
+  /// partitioning this equals getNumBDs(col, row).
+  uint32_t getNumBDsForChannel(int col, int row, int channel) const {
+    uint32_t count = 0;
+    for (uint32_t bd = 0; bd < getNumBDs(col, row); ++bd)
+      if (isBdChannelAccessible(col, row, bd, channel))
+        ++count;
+    return count;
+  }
+
   /// Return true iff buffer descriptor `bd_id` on tile (`col`, `row`) can be
   /// submitted on channel `channel`.
   virtual bool isBdChannelAccessible(int col, int row, uint32_t bd_id,
--- a/lib/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEAssignBufferDescriptorIDs.cpp
@@ -91,9 +91,11 @@ struct AIEAssignBufferDescriptorIDsPass
               std::optional<int32_t> next_id =
                   gen.nextBdId(dmaOp.getChannelIndex());
               if (!next_id) {
+                int channelIndex = dmaOp.getChannelIndex();
                 bd.emitOpError()
                     << "Allocator exhausted available BD IDs (maximum "
-                    << targetModel.getNumBDs(col, row) << " available).";
+                    << targetModel.getNumBDsForChannel(col, row, channelIndex)
+                    << " available for channel " << channelIndex << ").";
                 return signalPassFailure();
               }
               bd.setBdId(*next_id);
@@ -131,9 +133,11 @@ struct AIEAssignBufferDescriptorIDsPass
             std::optional<int32_t> next_id =
                 gen.nextBdId(blockChannelMap[&block]);
             if (!next_id) {
+              int channelIndex = blockChannelMap[&block];
               bd.emitOpError()
                   << "Allocator exhausted available BD IDs (maximum "
-                  << targetModel.getNumBDs(col, row) << " available).";
+                  << targetModel.getNumBDsForChannel(col, row, channelIndex)
+                  << " available for channel " << channelIndex << ").";
               return signalPassFailure();
             }
             bd.setBdId(*next_id);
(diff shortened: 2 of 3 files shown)


tools/aiecc/aiecc.cpp
```diff
@@ -378,6 +475,44 @@ static void printVersion(raw_ostream &os) {
...
+  if (!positionalArgs.empty())
+    return positionalArgs[0];
+  return "";
+}
+
+// Get host source files from positional arguments.
+static std::vector<std::string> getHostSourceFiles() {
+  std::string mlirFile = getInputFilename();
+  std::vector<std::string> hostFiles;
+  for (const auto &arg : positionalArgs) {
```
ypapadop-amd, line 510: If you know how many elements `positionalArgs` has you can `reserve()` to avoid reallocations.

tools/aiecc/aiecc.cpp
```diff
@@ -1704,107 +2266,803 @@ static LogicalResult compileCore(MLIRContext &context, ModuleOp moduleOp,
...
+          mlir::registerLLVMDialectTranslation(threadContext);
+          xilinx::xllvm::registerXLLVMDialectTranslation(threadContext);
+
+          // Parse the module in this thread's context
+          ParserConfig parseConfig(&threadContext);
+          OwningOpRef<ModuleOp> threadModule =
+              parseSourceString<ModuleOp>(moduleStr, parseConfig);
+
+          if (!threadModule) {
+            llvm::errs() << "Error: Failed to parse module for core ("
```
ypapadop-amd: You probably need a mutex here for output.

tools/aiecc/aiecc.cpp
```diff
@@ -1704,107 +2266,803 @@ static LogicalResult compileCore(MLIRContext &context, ModuleOp moduleOp,
...
+        }));
+  }
+
+  // Wait for all tasks to complete
+  for (auto &future : futures) {
+    future.wait();
+  }
+
+  if (hasFailure.load()) {
+    return failure();
```
ypapadop-amd, line 2568: I can't follow all the logic, but what happens if there are multiple failures? Is there a race condition at setting the failure or do we just get one of them?

jgmelber: Left a comment: hasFailure is atomic — multiple threads may set it concurrently without a race. We report failure if any core compilation failed.

python/compiler/aiecc/main.py
```diff
@@ -3,2189 +3,129 @@
...
-echo "Try running this script multiple times."
-"""
-
-    with open(output_file, "w") as f:
-        f.write(script_content)
-
-    # Make script executable
-    os.chmod(output_file, os.stat(output_file).st_mode | stat.S_IEXEC)
+    # Check relative to this file (installed location)
+    # python/compiler/aiecc/main.py -> bin/aiecc
```
hunhoffe: Just out of curiosity, does this ever work? I think the python normally gets installed into python venv dependencies (if installed from wheels) or build directory or install directory.

tools/aiecc/README.md
```diff
@@ -238,13 +249,27 @@ The C++ implementation provides **near-full feature parity**:
...
 - ✅ AIE simulation (`--aiesim`)
 - ✅ Host compilation (`--compile-host`, `--host-target`)
+- ✅ Repeater scripts (`--enable-repeater-scripts`, `--repeater-output-dir`)
+
+### Features Not Yet Implemented
+
+| Flag | Status | Notes |
+|------|--------|-------|
+| `--profile` | TODO | Command execution timing |
+| `--progress` | No-op | Rich progress bar (not planned for C++) |
```
hunhoffe, line 259: If not planned, should we remove it? Should this documentation be cleaned up at all?

python/compiler/aiecc/main.py
jgmelber: Good catch, you're right

tools/aiecc/README.md
jgmelber: I think I will leave it for now for those familiar with the original `aiecc.py`

PR #2928, opened 2026-03-05
Remove XRT_HACK_UNSECURE_LOADING_XCLBIN workaround

hunhoffe:
Remove all uses of the XRT_HACK_UNSECURE_LOADING_XCLBIN=1 env var, which bypassed xclbin firmware signature verification. I do not believe it is required any longer.


jgmelber:
Great! This definitely isn't needed

PR #2940, opened 2026-03-08
Allow high-rank vector.transfer_read with leading unit dims

erwei-xilinx:
## Summary
- Make `vector.transfer_read` dynamically legal in the CPP backend vectorizer
- Only mark illegal when rank <= 4 (which AIEVec patterns can handle)
- High-rank vectors with leading unit dims (e.g., `vector<1x1x4x32x4x8xbf16>`) are left for downstream passes

## Context
LLVM 23's `fold_unit_extent_dims_via_reshapes` pattern doesn't converge on memref IR with `IsolatedFromAbove` ops (like `air.herd`) containing unit-extent memref dimensions. This produces vectors like `vector<1x1x4x32x4x8xbf16>` that the `test-lower-vector-to-aievec` pass can't lower, causing `failed to legalize operation 'vector.transfer_read'`.

On LLVM 22, the fold pattern converged and eliminated unit dims before vectorization. On LLVM 23, the patterns oscillate (documented as "should not be used with a greedy driver" in `Transforms.h:2095`).

## Test plan
- [ ] Verify matmul programming examples compile with `--direct-codegen` on both aie2 and aie2p
- [ ] Verify existing mlir-aie tests still pass

### DIFF
--- a/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
+++ b/lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
@@ -25,11 +25,13 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
+#include "mlir/Dialect/Vector/Transforms/VectorRewritePatterns.h"
 #include "mlir/IR/PatternMatch.h"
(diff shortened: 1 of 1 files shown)


erwei-xilinx:
Closing - the fix is no longer needed. The root cause was addressed in mlir-air by overriding the collapseFn in the FoldUnitExtentDimsOp to handle strided memrefs with rank-reducing subviews, so the fold properly eliminates unit dims before vectorization. The aiecc vectorizer never sees vector<1x1x...> types.

PR #2946, opened 2026-03-10
Fix convert-to-llvm missing dynamic=true in C++ aiecc

### LINKED ISSUE #2945: C++ aiecc missing dynamic=true on convert-to-llvm pass
## Bug

The C++ `aiecc` binary introduced in #2925 (`dde51cb`) calls `createConvertToLLVMPass()` without setting `dynamic=true`, causing `builtin.unrealized_conversion_cast` failures during LLVM translation.

## Symptoms

`torch_mlir_e2e` tests in mlir-air fail with:
```
error: LLVM Translation failed for operation: builtin.unrealized_conversion_cast
note: see current operation: %15 = "builtin.unrealized_conversion_cast"(%14) : (i64) -> index
Error translating to LLVM IR for core (0, 3)
```

Affected tests: `torch_mlir_e2e/relu.py`, `torch_mlir_e2e/mul.py` (mlir-air).

## Root Cause

In `tools/aiecc/aiecc.cpp`, lines 1689 and 1764:
```cpp
pm.addPass(createConvertToLLVMPass());  // no options — dynamic defaults to false
```

The old Python `aiecc.py` had:
```python
.convert_to_llvm(dynamic=True)
```

Without `dynamic=true`, the pass doesn't use `DataLayoutAnalysis` to configure the type converter, so `index` types are not properly converted to `i64`, leaving behind `unrealized_conversion_cast` ops.

## Fix

```cpp
ConvertToLLVMPassOptions llvmOpts;
llvmOpts.useDynamic = true;
pm.addPass(createConvertToLLVMPass(llvmOpts));
```

This needs to be applied at both call sites (lines 1689 and 1764).

## Reproducer

Bump mlir-aie to any commit after `dde51cb` (e.g., `d4f709d`) in mlir-air and run the `torch_mlir_e2e` tests:
```bash
ninja check-air-python  # relu.py and mul.py will fail
```

erwei-xilinx:
## Summary

Fix a regression introduced in #2925 where the C++ `aiecc` binary calls `createConvertToLLVMPass()` without `dynamic=true`, causing `builtin.unrealized_conversion_cast` failures during LLVM translation.

Fixes #2945

## Problem

The old Python `aiecc.py` had:
```python
.convert_to_llvm(dynamic=True)
```

The new C++ `aiecc` has:
```cpp
pm.addPass(createConvertToLLVMPass());  // dynamic defaults to false
```

Without `dynamic=true`, the pass doesn't use `DataLayoutAnalysis` to configure the type converter, so `index` types are not properly converted to `i64`, leaving behind `builtin.unrealized_conversion_cast` ops:
```
error: LLVM Translation failed for operation: builtin.unrealized_conversion_cast
note: see current operation: %15 = "builtin.unrealized_conversion_cast"(%14) : (i64) -> index
```

This breaks `torch_mlir_e2e` tests in mlir-air (relu.py, mul.py) when bumping to any mlir-aie version after #2925.

## Fix

Set `useDynamic = true` on `ConvertToLLVMPassOptions` at both call sites in `tools/aiecc/aiecc.cpp` (per-core lowering at line 1689 and unified lowering at line 1769).

## Test plan
- [x] Fix compiles cleanly
- [x] Existing aiecc tests should pass (the option restores old behavior)
- [x] mlir-air torch_mlir_e2e tests pass after bumping to this fix

### DIFF
--- a/programming_examples/basic/event_trace/Makefile
+++ b/programming_examples/basic/event_trace/Makefile
@@ -16,7 +16,7 @@ srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 include ${srcdir}/../../makefile-common
 
-devicename ?= npu
+devicename ?= npu1
 targetname = mlir_trace_example
 
 in1_size = 16384 # in bytes
@@ -31,7 +31,7 @@ build/insts.bin: build/final.xclbin
 # Compile the AIE kernel
 build/scale.o: ${srcdir}/vector_scalar_mul.cc
 	mkdir -p ${@D}
-ifeq ($(devicename),npu)
+ifeq ($(devicename),npu1)
 	cd ${@D} && ${PEANO_INSTALL_DIR}/bin/clang++ ${PEANOWRAP2_FLAGS} -c $< -o ${@F}
 else ifeq ($(devicename),npu2)
 	cd ${@D} && ${PEANO_INSTALL_DIR}/bin/clang++ ${PEANOWRAP2P_FLAGS} -c $< -o ${@F}
--- a/programming_examples/basic/event_trace/aie_trace.mlir
+++ b/programming_examples/basic/event_trace/aie_trace.mlir
@@ -18,7 +18,7 @@
 //===----------------------------------------------------------------------===//
 
 module {
-  aie.device(npu1_1col) {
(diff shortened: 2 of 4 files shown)


jgmelber:
Add a unit test for this?

PR #2953, opened 2026-03-11
Fix ObjectFifoLink MemTile buffer allocation with padDimensions

RobinMelhuish:
## Summary

When ObjectFifoLink connects two ObjectFifos with different element sizes and the output has padDimensions, the stateful transform had two bugs:

1. MemTile buffer allocation always used the larger (output) type. When the output has padDimensions, the MemTile buffer should use the input (smaller) size because padding is applied on-the-fly by the DMA BD during MM2S transfer, not pre-allocated in the buffer.

2. The lenOut override in createMemTileDMA() forced both S2MM and MM2S to use the target FIFO's element count. Since createMemTileDMA is called separately for each FIFO in a link, lenOut is already correctly initialized from op.getElemType() for each call. The override was unnecessary and caused S2MM to use the wrong (larger) length.

Without this fix, the S2MM BD length exceeds the actual data sent per transfer, causing the MemTile DMA to stall waiting for data that never arrives.

Also exposes the existing padDimensions attribute through the IRON Python ObjectFifo API via a new pad_dimensions parameter.

## Test plan

  - Added lit test: `link_via_shared_mem_padDimensions.mlir` — verifies MemTile buffers use input
   size, S2MM BD length matches input element count, MM2S BD length matches output element count
  with dims+padding
  - Verified on NPU hardware (Strix Point) with 4ch→8ch DMA padding design


test/objectFifo-stateful-transform/data_movement_patterns/link/link_via_shared_mem_padDimensions.mlir
```diff
@@ -0,0 +1,95 @@
+//===- link_via_shared_mem_padDimensions.mlir --------------------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2025, Advanced Micro Devices, Inc.
```
jgmelber, line 7:
```suggestion
// Copyright (C) 2026, Advanced Micro Devices, Inc.
```

PR #2982, opened 2026-03-17
Fix non-deterministic crash in RunOp::verify() during parallel pass execution

hunhoffe:
**Warning: I am not fully confident this is the right cause/fix for this issue**

## Root Cause

**Confidence: high on mechanism, unconfirmed by thread instrumentation (not locally reproducible).**

After `AIEMaterializeRuntimeSequencesPass` (module-level, sequential) completes, MLIR dispatches the nested device-level passes to its thread pool — one thread per sibling `DeviceOp`. The crashing test has three sibling `DeviceOp`s (`@main`, `@add_two`,`@add_three`). Thread A verifying ops in `@main` called `SymbolTable::lookupSymbolIn(@add_two, seqName)` from `RunOp::verify()` while thread B was concurrently transforming `@add_two` — a read/write race on MLIR's symbol table.

## Fix

Remove `hasVerifier` from `RunOp` and delete `RunOp::verify()`. Move cross-device validation (sequence existence, argument count/types) into `AIEMaterializeRuntimeSequencesPass::runOnOperation()`, which iterates `DeviceOp`s sequentially on the main thread before any parallel execution begins.

Two crashes found during implementation:
- `Region::getOps<T>()` has an ilist sentinel issue on single-op blocks in this MLIR version; fixed by using `getBody().front().getOps<T>()`.
- `aie.runtime_sequence @name () {}` creates a **zero-block** region; calling `.front()` crashes. Fixed by guarding: `calleeRegion.empty() ? 0 : calleeRegion.getNumArguments()`.

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -786,9 +786,9 @@ LogicalResult AIEX::DMAStartBdChainOp::verify() {
   }
   for (unsigned i = 0, n = expectedArgTypes.size(); i < n; i++) {
     if (actualArgTypes[i] != expectedArgTypes[i]) {
-      return emitOpError("Argument ")
-             << (i + 1) << " types mismatch: " << "expected "
-             << expectedArgTypes[i] << " but got " << actualArgTypes[i];
+      return emitOpError("Argument ") << (i + 1) << " types mismatch: "
+                                      << "expected " << expectedArgTypes[i]
+                                      << " but got " << actualArgTypes[i];
     }
   }
   return success();


andrej:
> Thread A verifying ops in @main called SymbolTable::lookupSymbolIn(@add_two, seqName) from RunOp::verify() while thread B was concurrently transforming @add_two — a read/write race on MLIR's symbol table.

That makes a lot of sense and is likely the cause of this.

I fear we might get a lot more of these parallelism errors as more multi-device designs come up. I guess the rule for verifiers in particular should be to only verify w.r.t.  the operation itself, no other ops, since any of those could be modified in parallel if a pass is defined at that level.

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -506,6 +506,54 @@ struct InlineRuntimeCallsPattern : RewritePattern {
...
+
+    // Validate argument count and types against the callee signature.
+    // An empty body region (no blocks) means the sequence takes no arguments.
+    Region &calleeRegion = runtimeSeq.getBody();
+    unsigned numCalleeArgs =
+        calleeRegion.empty() ? 0 : calleeRegion.getNumArguments();
+    ValueRange args = runOp.getArgs();
+    if (numCalleeArgs != args.size())
+      return runOp.emitOpError() << "argument count mismatch: expected "
+                                 << numCalleeArgs << " but got " << args.size();
```
andrej, line 544: That's a nice improvement over the existing verifier

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -521,6 +569,24 @@ struct AIEMaterializeRuntimeSequencesPass
...
+        // DeviceOp from a verifier causes a data race on its symbol table.
+        for (ConfigureOp configureOp :
+             runtimeSequenceOp.getOps<ConfigureOp>()) {
+          AIE::DeviceOp referencedDev = configureOp.getReferencedDeviceOp();
+          if (!referencedDev) {
+            // ConfigureOp::verify() already reported the error (no such
+            // device, not a device, or device type mismatch) at parse time;
+            // this is a safety net if verification was disabled.
+            return signalPassFailure();
+          }
```
andrej, line 586: Might want to check if the same race condition that this PR tries to fix doesn't also exist for the verifier of configureOps.

hunhoffe: Good point -- I'll take a look and if I find anything else, I'll put it in an additional PR.

PR #2986, opened 2026-03-18
Remove redundant depth-1 exception from objectFifo lowering

abisca:
The `findObjectFifoSize()` function in AIEObjectFifoStatefulTransform had a special case that returned 1 buffer when both the declared scalar depth and the max acquire count were 1, suppressing the usual `maxAcquire + 1` ping-pong increment. This exception was redundant because the array-depth form (e.g. `[1, 1]`) already provides a way to explicitly enforce per-endpoint depths. Remove the exception so that a scalar depth-1 objectFifo consistently lowers to two buffers like any other scalar depth, and document that `[1, 1]` is the correct way to enforce a single buffer.

Changes:
- Remove the `maxAcquire == 1 && objFifo.size() == 1` guard from `findObjectFifoSize()` in AIEObjectFifoStatefulTransform.cpp
- Update `depth_one_objectfifo_test.mlir` to use `[1, 1]` array depth so the test continues to verify single-buffer lowering via the explicit array-depth mechanism
- Add `depth_one_scalar_test.mlir`: scalar depth 1 now lowers to 2 buffers (ping-pong)
- Add `depth_one_array_test.mlir`: array depth `[1, 1]` still lowers to 1 buffer, confirming the enforcement mechanism works
- Add `depth_two_scalar_test.mlir`: scalar depth 2 with max acquire 1 lowers to 2 buffers (not 3), confirming buffer count is driven by maxAcquire rather than the declared size
- Update `01_single_double_buffer/` example designs to use depth 2 (ping-pong) and add a note that `[1, 1]` enforces a single buffer

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -642,7 +642,14 @@ static void printObjectFifoInitValues(OpAsmPrinter &p, ObjectFifoCreateOp op,
                                       Attribute initValues) {
   if (op.getInitValues()) {
     p << "= [";
-    int depth = llvm::dyn_cast<mlir::IntegerAttr>(numElem).getInt();
+    int depth;
+    if (isa<ArrayAttr>(numElem)) {
+      depth = llvm::dyn_cast<mlir::IntegerAttr>(
+                  llvm::dyn_cast<mlir::ArrayAttr>(numElem)[0])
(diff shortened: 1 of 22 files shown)


abisca:
This started out as an experiment and it showed that changing this feature results in several exceptions in the IRON ObjectFifo API when multiple depths of 1 need to be expressed and passed on to the compiler. The current implementation is much cleaner.

PR #2998, opened 2026-03-22
Windows: improve wheel packaging parity

thomthehound:
This patch brings the Windows distribution wheels closer to parity with the existing Linux wheels.

To summarize, the Windows wheel now additionally packages:

- a bundled, statically built `xaiengine` (`LibXAIE`)
- the `aie_runtime_lib` helper payload
- the MLIR runtime support binaries under `python/aie/_mlir_libs`
- a Windows-appropriate subset of `test_utils` interface headers

The following are still excluded by design because they are not supported on Windows, are Vitis-only, or both:

- the HSA runtime payload
- Linux dma-buf/ION allocator bits
- simulator-only helper libs
- Chess wrapper artifacts

With this, the Windows wheel packaging better resembles a beta-level feature scope than an early alpha.

The only remaining, potentially platform-appropriate, file still missing is `mlir_apfloat_wrappers`. This omission is upstream in LLVM (["# Conditionally link apfloat wrappers only on Linux.", etc.](https://llvm.googlesource.com/llvm-project/mlir/+/refs/heads/main/lib/ExecutionEngine/CMakeLists.txt)). It is not breaking, as the remaining MLIR runtime libraries on Windows do not link against it.

---

When the Windows wheel path was first introduced, packaging the bundled `xaiengine` runtime payload was impractical because the CI runner did not have a Windows XRT SDK available to link against. Furthermore, the codebase already supported using an externally built `LibXAIE`, which could itself be produced from the `aie-rt` submodule, so the deliberate omission was a tolerable stop-gap. Now that CI can stage an SDK for Windows, the bundled `xaiengine` payload can be packaged as well, as it is for Linux wheels.

While working on that, a few additional CMake oversights also surfaced:

- some helper payload used by examples/tests was being gated unnecessarily broadly on Vitis availability in `aie_runtime_lib/CMakeLists.txt`
- some Python runtime libraries were being skipped in `python/CMakeLists.txt` because `mlir_apfloat_wrappers` does not exist on Windows, and CMake was using `break()` instead of `continue()`.
- runtime-helper packaging logic in `runtime_lib/CMakeLists.txt` and `runtime_lib/test_lib/CMakeLists.txt` assumed Linux-oriented helper libraries always exist; this needed to be split carefully so Windows still gets the subset of utilities that make sense there without pulling in HSA/simulator payloads that are not intended for that platform

In total, the following paths are now packaged in the Windows wheels that were not packaged before:

- `aie_runtime_lib/AIE2/...`
- `aie_runtime_lib/AIE2P/...`
- `runtime_lib/x86_64/xaiengine/include/...`
- `runtime_lib/x86_64/xaiengine/lib/xaiengine.lib`
- `runtime_lib/x86_64/test_lib/include/cxxopts.hpp`
- `runtime_lib/x86_64/test_lib/include/test_utils.h`
- `runtime_lib/x86_64/test_lib/include/xrt_test_wrapper.h`
- `python/aie/_mlir_libs/...`

### DIFF
--- a/aie_runtime_lib/CMakeLists.txt
+++ b/aie_runtime_lib/CMakeLists.txt
@@ -10,12 +10,12 @@ add_custom_target(aie-runtime-libs ALL)
 
 function(add_aie_runtime_libs arch)
   # Precompile the intrinsic wrapper when chess-clang and AIETools are available.
-  if(TARGET xchesscc_wrapper AND DEFINED VITIS_AIETOOLS_DIR)
+  if(AIE_BUILD_CHESS_CLANG AND DEFINED VITIS_AIETOOLS_DIR)
     string(TOLOWER ${arch} aiearch)
     add_custom_target(${arch}_chess_intrinsic_wrapper ALL DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/chess_intrinsic_wrapper.ll)
     add_custom_command(OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/chess_intrinsic_wrapper.ll
             COMMAND ${PROJECT_BINARY_DIR}/bin/xchesscc_wrapper ${arch}
-            -aietools ${AIETOOLS_DIR}
+            -aietools ${VITIS_AIETOOLS_DIR}
             -c ${CMAKE_CURRENT_SOURCE_DIR}/chess_intrinsic_wrapper.cpp
             -d # Disassemble output
             -f # Use LLVM frontend


thomthehound:
I think the problem was that the Chess wrapper was implicitly getting built before, and the lit tests were depending on it without explicitly declaring it. My changes altered that because, obviously, Windows doesn't have Chess, so the wrapper is no longer *always* getting built. Now it should when it is needed.

Oy. And to think I still have another month planned of mucking around with the CMakes...

Issue #3000, opened 2026-03-26
AIEInsertTraceFlows: redundant DMA register writes in single-shim routing mode


FIM43-Redeye:
## Problem

In `-aie-insert-trace-flows` with `routing = single` (the default and currently only strategy), the `shimInfos` map is keyed by **source column**, with every entry pointing to the same shim tile. Phase 4c-4f then iterates all `shimInfos` entries and emits BD/DMA/task-queue register writes for each one, producing duplicate writes to the same shim.

For example, with traces on tiles (0,2) and (1,2) both routing to shim (0,0), Phase 4 emits the BD setup twice for shim column 0 -- identical values to identical registers.

## Impact

Low. The writes are idempotent (same values to same registers), so behavior is correct. The redundancy is in the generated instruction sequence only -- a few extra `npu.write32` / `npu.writebd` ops that program already-configured registers.

## Location

`lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp`, the `// 4c-4f. Insert per-shim configurations` loop iterates `shimInfos` which has one entry per source column, not per unique target shim.

## Suggested fix

Either deduplicate before the loop (iterate unique shim target columns instead of all `shimInfos` entries), or track already-configured shim columns in a `std::set<int>` and skip duplicates.

PR #2962, opened 2026-03-12
Insert trace flows pass 

yenjames:
### Summary
This PR implements a new MLIR pass to automatically insert trace packet flows and runtime sequence configuration for AIE trace operations. Following #2705, this should eliminate the need for manual trace `npu.write32` configurations. 

### Changes
- The pass inserts packet flows from trace sources to shim DMA and generates all required runtime sequence operations (timer control, BD setup, DMA config, broadcast triggers). Operations are inserted at the beginning of runtime sequence but after any `aie.trace.start_config` ops.
- Updated test in `test/npu-xrt/vec_mul_event_trace` to be the full end to end trace test. Checks trace output after hardware run.
- When `aie.trace.packet` type is not specified, the pass auto-detects it from the tile type. For core tiles, it defaults to core instead of mem.
- Fixed `AIETraceToConfig.cpp` which incorrectly allows MemTile and ShimTiles to be configured with a Mode field in their trace. Based on my understanding, only `core` module has the Mode field in its trace control register.
- If a traced shim is a trace destination, it will use USER_EVENT_1/0 as start stop events; for non-destination shims, it will use BROADCAST_A_15/14.

### Example
What user writes:
``` mlir

aie.device(npu1_1col) {
  %tile02 = aie.tile(0, 2)
  
    // Trace configuration for compute tile (0,2) - core events
    aie.trace @core_trace(%tile_0_2) {
      // Set trace mode (Event-Time captures timestamps)
      aie.trace.mode "Event-Time"

      // Configure packet routing (ID and type for packet-switched routing)
      aie.trace.packet id=1 type=core

      // Specify which events to capture (up to 8 events)
      aie.trace.event<"INSTR_EVENT_0">        // User event 0 (start marker)
      aie.trace.event<"INSTR_EVENT_1">        // User event 1 (end marker)
      aie.trace.event<"INSTR_VECTOR">         // Vector instructions
      aie.trace.event<"MEMORY_STALL">         // Memory access stalls
      aie.trace.event<"STREAM_STALL">         // Stream buffer stalls
      aie.trace.event<"LOCK_STALL">           // Lock acquisition stalls
      aie.trace.event<"PORT_RUNNING_0">       // DMA:0 S2MM running
      aie.trace.event<"PORT_IDLE_1">          // DMA:1 MM2S idle
      aie.trace.port<0> port=DMA channel=0 direction=S2MM
      aie.trace.port<1> port=DMA channel=0 direction=MM2S

      // Specify start/stop control (broadcast events)
[...]

PR #3001, opened 2026-03-26
feat(trace): lateral routing and S2MM channel distribution for InsertTraceFlows

FIM43-Redeye:
## Summary

Adds two new opt-in features to `-aie-insert-trace-flows`, building on #2962:

- **Lateral routing** (`--lateral-routing`): redirects trace packet flows to spare columns (no active `aie.core` ops) so trace traffic avoids competing with data flows on vertical stream switch ports. Auto-detects the nearest spare NOC shim column, or accepts a forced target via `--lateral-target-col=N`. Falls back to the original target when no spare exists.

- **S2MM channel distribution** (`--distribute-channels`): round-robins traces across both S2MM DMA channels per shim tile, doubling available trace bandwidth when multiple traces target the same shim. Each channel gets its own BD and runtime sequence argument index.

Both features default to off and compose with each other and with the existing `routing = single` strategy. Designed as orthogonal post-processing steps so they will automatically compose with future routing strategies (e.g., when per-column routing returns).

## Design notes

- **Lateral routing is a post-processing step**, not a routing strategy. It runs after Phase 2b shim selection and redirects targets. When `TraceShimRouting::PerColumn` is reintroduced, lateral routing will compose with it automatically (each column's shim redirects to its nearest spare).
- **CLI-only for now.** These are pass options, not `host_config` attributes. This keeps the door open for @yenjames to wire them into IRON/Python (#2988) however he sees fit -- either as `TraceShimRouting` enum values or as separate `host_config` fields.
- **Known pre-existing issue:** `shimInfos` keyed by source column causes redundant (but idempotent) DMA register writes in single-shim mode. Filed as #3000, not introduced by this PR.

## Test plan

- [ ] 4 FileCheck RUN lines for lateral routing (auto-detect, no-spare fallback, forced target, default-off)
- [ ] 2 FileCheck RUN lines for channel distribution (two-trace distribute, single-trace no-op, default-off)
- [ ] 1 FileCheck RUN line for combined lateral + distribute composition
- [ ] Existing 9 trace tests unaffected (both features default to off)
- [ ] Compilation pipeline validated end-to-end on NPU1 (Phoenix) hardware: event_trace example compiles, runs, produces correct output and valid trace data with the updated toolchain

Generated using Claude Code.


FIM43-Redeye:
Hey Jack, appreciate the thorough feedback! Went ahead and addressed items 1-4:

**1. >2 traces with distribute** - added a 4-trace-across-3-columns test on `npu1_3col`. Round-robin assignment handles N traces across the 2 S2MM channels, traces alternate between channel 0 and 1.

**2. Shim already used** - added S2MM conflict detection. The pass now scans existing `aie.flow`/`aie.packet_flow`/`ShimDMAAllocationOp` before claiming channels. If 1 channel's free, it uses that one (and skips distribute). If both are taken and `lateral-routing` is on, it redirects to a spare column. Otherwise it errors with a clear diagnostic.

**3. Auto column selection** - already implemented: `lateral-routing=true` without `lateral-target-col` auto-selects the nearest spare shim NOC column. No changes needed here.

**4. Forced column + partial use** - falls out of the conflict detection in #2. If the forced target has 1 channel free, it uses it single-channel.

**5. End-to-end** - this is where I'd love your input. I can write MLIR FileCheck tests all day, but for the hardware-path validation you're describing (stacked offsets with distribute, routing over disabled tiles with lateral, `host_config`/`start_config` interaction with BD offsets and `ddr_id`), I'm not sure what the right test infrastructure looks like on your end. Is there an existing e2e pattern I should follow, or would it be better for us to pair on that?

FIM43-Redeye:
Pushed a few commits addressing your feedback:

**arg_idx offset fix** -- both S2MM channels now share the same arg_idx and split by buffer offset (channel 0 at offset 0, channel 1 at offset buffer_size). Host allocates 2x buffer_size. This replaces the old arg_idx/arg_idx+1 approach that bumped into the XRT buffer id limit. Also updated the pass description in AIEPasses.td.

**E2e test** -- added `test/npu-xrt/vec_mul_trace_distribute_lateral/` modeled on `vec_mul_event_trace`. 2-column NPU1 design, core on column 0, column 1 spare, 2 traces with distribute + lateral. Uses a pre-lower workaround (aie-opt with pass options first, then aiecc.py) until pass-through options are available.

**Hardware results on NPU1:**
- Data computation passed -- lateral routing through column 1's spare tiles works correctly, no data path perturbation
- Core trace captured (4x INSTR_EVENT_0 + 4x INSTR_EVENT_1, matching the 4 kernel iterations)
- Trace buffer readback is truncated because `TraceConfig.write_trace()` strips zero words and drops the channel boundary. The raw extraction buffer is correctly sized - seems to be a host infrastructure thing, not a pass issue

Let me know how you'd like to handle the aiecc.py integration for the pass options, happy to iterate on the e2e test once that's in place.

jackl-xilinx:
@FIM43-Redeye Great progress. I was wondering if the parser is able to handle the 0's gap and it sounds like it cannot. This is one of the things that needs to be updated in order to handle split trace data. You're welcome to dig into the trace parser to see if it's an easy fix or else you can create a check for two sets of non-zero trace data in the single trace.txt file. As for the aiecc.py changes, I need to get back to you on that, hopefully later today.

FIM43-Redeye:
Peeked in the trace parser, you were right - choked on the zero gap. Three mixed issues:

1. `write_trace()` stripped ALL zero words, which collapsed the gap between channels and merged ch1's data onto ch0
2. `trim_trace_pkts()` scanned forward and truncated at the first `fefefefe` + two zeros, so it cut off ch1 entirely at ch0's end marker
3. Even if zeros survived into the stream, `trace_pkts_de_interleave()` processed in 8-word packets and the gap wasn't necessarily packet-aligned, so zero words landed at arbitrary positions within the packet structure

Fix is in commit 66ef67e on the PR branch:

- `write_trace()` now strips only trailing zeros, preserving the internal gap
- New `split_trace_segments()` in utils.py detects contiguous non-zero data blocks separated by 8+ consecutive zeros (one packet width), splitting the buffer into per-channel segments before parsing
- `parse_trace()` and `main()` process each segment independently through trim/de-interleave, then merge by (type, location)

No-op for single-channel traces, `split_trace_segments` returns one segment and the pipeline is identical. Also added 17 pytest cases to cover segment splitting edge cases and the write_trace fix. Standing by on the aiecc pass integration, let me know how to proceed.

FIM43-Redeye:
Quick note on the latest push (c3a9510): the CI format check was a stray blank line (trivial), but the Ryzen AI test failure was more interesting.

The `write_trace()` fix from 66ef67e preserved internal zeros correctly, but trailing zeros after channel 1's last data word were still getting stripped. When `read_trace()` returned the truncated buffer, channel 1's offset was past the end. Fix: `read_trace()` now pads back to `trace_size`. No-op for single-channel traces since they already fill the buffer.

This touches `config.py` which is shared trace infra -- the change is backward-compatible (padding is a no-op when the buffer is already full), but happy to split it out or discuss if you'd prefer a different approach.

Also relaxed the e2e assertion from `DMA_S2MM_0_START_TASK >= 1` to `DMA_* >= 1` -- the mem trace captures whichever events fire, and starvation events are more common than start-task events depending on timing.

jackl-xilinx:
@FIM43-Redeye Thanks and good changes. I looked into the aiecc.py changes and it led to some good internal discussions on possibly making it more robust for custom passes. For now, can you see if you can run your trace pass first on the source mlir file and then run the standard aiecc.py on that? I believe that may just work out of the box. Barring that, we may need to look into expanding the underlying commands executed by aiecc.py manually and insert your pass in that list. We do not want to directly edit aiecc.py as it's a central tool and not one we want to bloat with too many custom options. For the second option, we don't have a ready-to-go expansion of the aiecc.py commands, particularly the python-binded commands so we would need to figure out how to translate those into direct calls.

FIM43-Redeye:
Rebased on main and verified two-step works on my hardware stepwise:

```
# Step 1: pre-lower with custom trace options
aie-opt aie.mlir \
  -aie-insert-trace-flows="distribute-channels=true lateral-routing=true" \
  -aie-trace-to-config \
  -aie-trace-pack-reg-writes \
  -aie-inline-trace-config \
  -o trace_lowered.mlir

# Step 2: standard aiecc.py, trace pipeline is a no-op
aiecc.py --no-xchesscc --no-xbridge --no-aiesim \
  --aie-generate-xclbin --aie-generate-npu-insts \
  --no-compile-host --xclbin-name=final.xclbin \
  --npu-insts-name=insts.bin trace_lowered.mlir
```

Works cleanly because `hasTraceOps()` in the aiecc pipeline checks for `aie.trace` ops, which are fully consumed by the pre-lowering chain. aiecc skips the entire trace pipeline and proceeds with resource allocation/routing/codegen normally.

NPU1 hardware results (distribute + lateral, 2-column design):
- Data: PASS
- Channel 0 (core trace): 48 non-zero words, 4x INSTR_EVENT_0 + 4x INSTR_EVENT_1
- Channel 1 (mem trace): 48 non-zero words, 7 DMA events
- Trace parser handles the split buffer correctly

The e2e test in the PR already uses this exact flow. Happy to adjust the test structure if you'd prefer a different pattern. Standing by on aiecc.py extensibility thoughts, but the two-step seems good to me.

jackl-xilinx:
@FIM43-Redeye Nice work! This is exactly what is needed and I appreciate the extra tests for the split trace data code you added. I think we're almost there. The last bit is actually something we had missed but with end-2-end trace tests, we need to check the generated trace.txt file rather than just check for a successful run. Can you test and add the following 2 lines from https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/basic/vector_scalar_mul/Makefile#L115-L116 to your test.py so it runs the trace parse script which checks for valid trace generation? This is actually missing from https://github.com/Xilinx/mlir-aie/blob/main/test/npu-xrt/vec_mul_event_trace/test.py so if you don't mind adding it there as well, that'd be very helpful. After that, I think we're good to go.

jackl-xilinx (approved):

FIM43-Redeye:
Added CLI trace parse validation to both tests (c60818d):

Both `vec_mul_trace_distribute_lateral/test.py` and `vec_mul_event_trace/test.py` now run `parse.py --input trace.txt --mlir <mlir> --output trace.json` followed by `get_trace_summary.py --input trace.json` as subprocess calls after the existing in-memory event validation. Scripts are located via the installed `aie.utils.trace` package path, so it works in any test environment without hardcoded relative paths.

Test fails if either script returns non-zero, and prints the cycle summary on success. No-op if `trace.txt` or the MLIR file isn't available (guards already in place for non-hardware CI).

PR #3002, opened 2026-03-26
Runtime TXN generation

jgmelber:
## Summary

Add **dynamic runtime TXN generation**: compile a single XCLBIN once, then generate NPU instruction streams at runtime for arbitrary problem sizes. The compiler emits a standalone C++ function that builds TXN binaries parameterized by runtime values (e.g., matrix dimensions M, K, N).

### Core infrastructure

- **`TxnEncoding.h`** — Header-only runtime library for encoding NPU TXN instructions. Zero MLIR/LLVM dependencies; shared between compiler and generated host code.

- **`ConvertAIEXToEmitC` pass** — Lowers AIEX runtime sequence ops (`npu.write32`, `npu.sync`, `npu.address_patch`, `npu.blockwrite`) to EmitC dialect, producing compilable C++ via `translateToCpp`. Includes blockwrite fusion: dynamic BD word overrides (write32s tagged with `bd_group`) are folded into a single `txn_append_blockwrite` call, producing identical TXN binary to the static path.

- **`--aie-generate-txn-cpp` flag** — New `aiecc` option that generates a C++ header alongside the XCLBIN containing `generate_txn_sequence(...)` parameterized by runtime values.

- **`BdLowering.{h,cpp}`** — Shared utility for emitting hardware BD register encodings as SSA arith chains. Used by both `AIEDmaToNpu` and `AIEDMATasksToNPU` dynamic paths. Includes d0_stride underflow guard and compile-time warning for 10-bit d0_size overflow.

### Dynamic DMA support

- **`npu.dma_memcpy_nd`** — Extended to support SSA-parameterized sizes, strides, and offsets via `emitDynamicHwBdEncoding()`.

- **`aie.dma_bd`** — New `dyn_sizes`, `dyn_strides`, `dyn_offset`, `dyn_len` operands for runtime-parameterized buffer descriptors within `aiex.dma_configure_task`.

- **`npu.write_rtp`** — Supports dynamic SSA values via `dyn_value` operand. Single implementation in `aie.py`, delegated from `aiex.py`.

- **`RuntimeSequenceOp`** — Removed `IsolatedFromAbove` to allow referencing parent `DeviceOp` values. SCF-to-CF conversion scoped to `aie.core` ops only.

### IRON Python support

`RuntimeScalar` type for scalar runtime parameters, `write_rtp()` for runtime-tunable parameters, BD ID allocator for dynamic DMA tasks, direct `npu_dma_memcpy_nd` emission path.

### Test coverage

- FileCheck tests for EmitC conversion (basic, dynamic values, unsupported ops)
- Dynamic BD encoding tests with numeric arith chain verification
- MemTile rejection negative test
- Python round-trip and lowering tests (`dma_tasks_dynamic.py`)
- End-to-end `aiecc` pipeline test (`cpp_dynamic_txn.mlir`)
- Static-vs-dynamic TXN binary equivalence test

## Test plan

- [x] `test/python/` — 74 pass, 3 unsupported (expected)
- [x] `test/dialect/AIEX/` — 21 pass
- [x] `test/dialect/AIE/` — 110 pass, 1 xfail (expected)
- [x] `test/Conversion/DmaToNpu/` — 16 pass
- [x] `test/Conversion/AIEXToEmitC/` — 3 pass
- [x] Hardware: 5 shapes x 2 variants = 10/10 pass on Ryzen AI NPU
- [ ] CI: Full build matrix + Ryzen AI hardware tests


include/aie/Conversion/AIEXToEmitC/AIEXToEmitC.h
```diff
@@ -0,0 +1,25 @@
+//===- AIEXToEmitC.h - AIEX to EmitC conversion -----------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
```
andrej: 2026

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -748,13 +748,19 @@ def AIE_NpuWriteRTPOp: AIEX_Op<"npu.rtp_write", []> {
...
+  let hasVerifier = 1;
   let description = [{
-    rtp write operator
+    rtp write operator.
+    When `dyn_value` is provided, it supplies the RTP value at runtime
+    instead of the static `value` attribute.
+  }];
+  let extraClassDeclaration = [{
+    bool hasDynamicValue() { return getDynValue() != nullptr; }
   }];
```
andrej, line 786: I'm a little worried about code bloat with having every parameter for these ops duplicated, once as an attribute and once as an SSA value, along with the added custom verifier and assembly format for each op. 

Could we consider removing the attributes altogether and instead use SSA values, with `arith.constant` for the static case? All existing lowerings can get the value from `arith.constant` and throw an error if it's not a constant, this emitC pass can use the actual SSA values. This approach would of course touch a lot of code (all examples etc. that use these ops with attributes would have to be rewritten to use `arith.constant`), but I think AI could handle it. I think it would be cleaner and might remove the need for customAssemblyFormat and hasVerifier for every op (haven't gotten to those yet but assume they're there because of this).

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -831,10 +868,38 @@ def AIE_NpuMaskWrite32Op: AIEX_Op<"npu.maskwrite32", []> {
     into the memory space of aie.tile(column, row).
     If 'buffer' is not present and 'column' and 'row' are not present then
     'address' is interpreted as a full 32-bit address in the AIE array.
+
+    Optionally, SSA values can be provided for 'dyn_address', 'dyn_value', and
+    'dyn_mask' to enable runtime-parameterized sequences.
+
+    Static syntax (unchanged): `aiex.npu.maskwrite32 {address = 123 : ui32, ...}`
```
andrej: Suggest removing "(unchanged)" from these comments

include/aie/Runtime/TxnEncoding.h
```diff
@@ -0,0 +1,173 @@
+//===- TxnEncoding.h - Standalone TXN instruction encoding -------*- C++
+//-*-===//
```
andrej: suggest formatting onto a single line

lib/Targets/AIETargetNPU.cpp
```diff
@@ -80,123 +42,86 @@ reserveAndGetTail(std::vector<uint32_t> &instructions, uint64_t tailSize) {
...
-  words[2] = op.getAddress();
+  // The encoding library leaves word[1] as 0. If col/row are present, set it.
   auto col = op.getColumn();
   auto row = op.getRow();
   if (col && row) {
-    words[1] = (*col & 0xff) | ((*row & 0xff) << 8);
+    // word[1] is at position (current_size - headerSize - count + 1)
+    size_t headerPos = instructions.size() - 4 - payload.size();
+    instructions[headerPos + 1] = (*col & 0xff) | ((*row & 0xff) << 8);
   }
```
andrej, line 123: Why doesn't the encoding library aie_runtime::txn_append_blockwrite take row/col?

lib/Targets/AIETargetNPU.cpp
andrej, line 1: I like the refactor of this to use the "encoding library". The wrappers around the "append*" functions (appendWrite32, appendPreempt, ...) seem very thin and the same for each function. Could they be eliminated?

include/aie/Runtime/TxnEncoding.h
```diff
@@ -0,0 +1,173 @@
...
+#define AIE_RUNTIME_TXNENCODING_H
+
+#include <cstdint>
+#include <cstring>
+#include <vector>
+
+namespace aie_runtime {
+
+// Transaction opcodes - mirroring xaie_txn.h from aie-rt.
+// See aie-rt commit a6196eb, xaiengine/xaie_txn.h.
```
andrej, line 31: I understand this is just moving code but maybe let's earmark this not-ideal code duplication from aie-rt. I know we want to be light on dependencies, but maybe a sed script could generate a header here by pulling from aie-rt. That way we'd be on top of any firmware changes to these opcodes.

include/aie/Runtime/TxnEncoding.h
andrej, line 1: Like having this encoding logic factored out into a reusable file for both runtime and compilation a lot!

include/aie/Targets/AIENpuLowering.h
```diff
@@ -0,0 +1,43 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2025 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+//
+// Shared NPU lowering pipeline used by both aiecc and aie-translate.
```
andrej, line 11: Might become clear as I look through the code more, but what is meant by this exactly? Doesn't aiecc just call into aie-translate or aie-opt when needed?

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+
+    for (auto &[seqOp, deviceOp] : sequences) {
+      Location loc = seqOp.getLoc();
+      std::string seqName = seqOp.getSymName().str();
+
+      // Extract device info for TXN header before we erase the device.
+      uint8_t devGen = 3;
+      uint8_t numRows = 6;
+      uint8_t numCols = 5;
+      uint8_t numMemTileRows = 1;
```
andrej: Can we make these values uninitialized/garbage to cause early failure if the below if doesn't execute? I think the runtime sequence should always be inside of a device.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+      Value txnVec = emitc::LiteralOp::create(builder, loc, txnVecType, "txn");
+
+      // Emit: uint32_t op_count = 0;
+      // Tracked purely via VerbatimOp - no SSA value needed since we only
+      // increment it (via VerbatimOp) and use it in the footer (via
+      // VerbatimOp).
+      emitc::VerbatimOp::create(builder, loc, "uint32_t op_count = 0;");
+      // Dummy value passed to helpers (unused since emitIncrementOpCount
+      // emits VerbatimOp directly).
+      Value opCountLval = txnVec; // placeholder, not used
```
andrej: I don't like this. I'd say either commit to the "proper" way (create an associated SSA value, then pass that and increment it, my preference, seems more resilient to future modifications) or commit to the hard-coded variable name `op_count` and remove the dummy `Value` both there and from the helper functions that require it.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+          if (!defOp)
+            continue;
+          // Check if defOp is outside the runtime_sequence.
+          if (!seqOp->isAncestor(defOp)) {
+            // Create EmitC equivalent for this external value.
+            if (auto constOp = dyn_cast<arith::ConstantOp>(defOp)) {
+              auto emitResult = emitOp(builder, constOp, loc, txnVec,
+                                       opCountLval, argMapping);
+              (void)emitResult;
+            }
```
andrej: Suggest erroring here if the defining op is not an `arith.constant`

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+    // leftover aie.device/runtime_sequence ops are harmless. However,
+    // leaving them causes verifier issues, so we need to erase them.
+    //
+    // To safely erase, we must ensure no cross-references from EmitC ops
+    // back to old Values. Walk all EmitC funcs and clone any operands
+    // that point to ops defined outside the func (i.e., in the old IR).
+    // Fix cross-references: EmitC ops may reference Values defined in the
+    // old IR (e.g., arith.constant values that weren't mapped through
+    // argMapping). For each such reference, create an EmitC constant
+    // with the same value directly in the EmitC function.
```
andrej: Comment seems to re-state the same thing twice unless I'm reading it wrong. Also, isn't that what we're already doing above on line 308?

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+                  std::to_string(static_cast<uint64_t>(intVal));
+              if (emitcType == emitc::OpaqueType::get(ctx, "uint32_t"))
+                valStr += "u";
+              replacement = emitc::ConstantOp::create(
+                  localBuilder, defOp->getLoc(), emitcType,
+                  emitc::OpaqueAttr::get(ctx, valStr));
+            }
+          }
+          if (!replacement) {
+            // Fallback: just create a 0u constant.
```
andrej: I'd rather have this error, seems like a debugging nightmare otherwise. If it works with `0`, the user can always just use `arith.constant 0`

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+    }
+    return success();
+  }
+
+  /// Emit a single operation.
+  LogicalResult emitOp(OpBuilder &builder, Operation *op, Location loc,
+                       Value txnVec, Value opCountLval, IRMapping &argMapping) {
+    Location opLoc = op->getLoc();
+
+    // AIEX write32 - handles both static and dynamic forms.
```
andrej, line 566: Is there a way to make these blocks, along with the `emit*` functions less repetitive? No worries if not just would like us to consider it. Maybe a helper function that takes a vector of operands, a vector of types, and a function pointer to the encoding library function.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+        } else {
+          valStr = std::to_string(static_cast<uint64_t>(intVal));
+          if (emitcType == getU32Type(ctx))
+            valStr += "u";
+        }
+        result = emitc::ConstantOp::create(builder, opLoc, emitcType,
+                                           emitc::OpaqueAttr::get(ctx, valStr));
+      } else {
+        // Unsupported constant type - just emit 0.
+        result = createU32Constant(builder, opLoc, 0);
```
andrej: Again I think an error would be better

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+          valStr = "static_cast<" +
+                   cast<emitc::OpaqueType>(emitcType).getValue().str() + ">(" +
+                   std::to_string(intVal) + ")";
+        } else {
+          valStr = std::to_string(static_cast<uint64_t>(intVal));
+          if (emitcType == getU32Type(ctx))
+            valStr += "u";
+        }
+        result = emitc::ConstantOp::create(builder, opLoc, emitcType,
+                                           emitc::OpaqueAttr::get(ctx, valStr));
```
andrej: Is there no existing MLIR infrastructure to turn an `arith.constant` into a emitc-equivalent? If not, can we factor this out into a function with a descriptive name, like `arithConstantToEmitcConstant`?

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,991 @@
...
+      return success();
+    }
+
+    if (auto extOp = dyn_cast<arith::ExtSIOp>(op)) {
+      Value input = argMapping.lookupOrDefault(extOp.getIn());
+      Type resultType = extOp.getType();
+      Value result = emitc::CastOp::create(builder, opLoc, resultType, input);
+      argMapping.map(extOp.getResult(), result);
+      return success();
+    }
```
andrej: There's got to be an existing pass that converts `arith` -> `emitc`, no? If not, this could at least be factored out into its own pass.

Edit: AI says there is: `--convert-arith-to-emitc`

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
andrej, line 1: I think this diff could be made a lot smaller (less code = less maintenance burden and opportunities for bugs) by reusing existing MLIR infrastructure such as `--convert-arith-to-emitc`. I'm also worried the more complicated aspects of this, such as handling `yield`s in `scf.for`s and `scf.if`s might be hard to get right in all edge-cases, so relying on the MLIR existing pass might be more resilient (although to be fair I didn't look spot any errors). I think Claude might have worked around some other limitations here that caused it to think reinventing this wheel was the easiest way forward, but I'd like to push for trying again with the existing infrastructure.

lib/Dialect/AIE/Transforms/AIEVectorTransferLowering.cpp
```diff
@@ -63,7 +63,17 @@ struct AIEVectorTransferLoweringPass
...
-    if (failed(applyPatternsGreedily(deviceOp, std::move(patterns))))
+    // Disable cross-region constant CSE to prevent the greedy rewriter from
+    // hoisting arith.constant ops from inside aie.runtime_sequence (which has
+    // IsolatedFromAbove) up to the aie.device scope. Without this, the default
+    // cseConstants=true causes constants to be moved from the runtime_sequence
+    // body into the device scope, making aie.core bodies reference device-scope
+    // values. This breaks AIECoreToStandardPass which cannot clone the core
+    // body into a func.func when the core references values defined outside it.
+    GreedyRewriteConfig config;
+    config.enableConstantCSE(false);
```
andrej, line 74: I don't see how the other changes in this PR are related to this change.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
andrej, line 1: If we can deprecate attributes in favor of SSA values for these ops, we could get rid of a lot of the parse/print code that was added in here.

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -191,6 +215,50 @@ struct PushQueuetoWrite32Pattern : OpConversionPattern<NpuPushQueueOp> {
...
+  if (val.getType() != intType) {
+    unsigned valBits = val.getType().getIntOrFloatBitWidth();
+    unsigned tgtBits = intType.getIntOrFloatBitWidth();
+    if (valBits > tgtBits)
+      val = arith::TruncIOp::create(builder, loc, intType, val);
+    else
+      val = arith::ExtUIOp::create(builder, loc, intType, val);
+  }
+  return val;
+}
```
andrej: This might also fit in a more general-purpose "utils" class sicne it's not directly related to the AieDmaToNpu pass afaict.

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -229,69 +297,281 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
...
+          rewriter, op->getLoc(), column, row, infoOp.getChannelDirAttr(),
+          infoOp.getChannelIndexAttr(), issue_token, repeat_count, bd_id);
+
+      rewriter.eraseOp(op);
+      return success();
+    }
+
+    // =====================================================================
+    // DYNAMIC CODE PATH -- some sizes/strides are SSA values
+    // =====================================================================
```
andrej, line 522: The `// ==========` here seems to me a good indication that it might make sense to split this up into smaller functions: One for the static path (everything included in the closing brace above) and one for the dynamic path.

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -229,69 +297,281 @@ struct DmaToNpuPattern : OpConversionPattern<NpuDmaMemcpyNdOp> {
...
+          return getConstantIntValue(s).has_value();
+        });
+    bool allStridesConstant =
+        llvm::all_of(op.getMixedStrides(), [](OpFoldResult s) {
+          return getConstantIntValue(s).has_value();
+        });
+
+    if (allSizesConstant && allStridesConstant) {
+      // =====================================================================
+      // STATIC CODE PATH (unchanged) -- all sizes/strides are constants
```
andrej: remove (unchanged)

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
andrej, line 1: I think this file could benefit from some code deduplication and cleanup. If NpuWriteBd op were changed to also accept SSA values, maybe there would be less of a need for separate code paths here. I'd like to avoid having to make every change in two places (dynamic and static path) for future changes to these ops.

lib/Dialect/AIEX/Transforms/AIEMaterializeBDChains.cpp
```diff
@@ -128,6 +128,7 @@ struct AIEMaterializeBDChainsPass
     GreedyRewriteConfig rewriter_config = GreedyRewriteConfig();
     rewriter_config.setRegionSimplificationLevel(
         GreedySimplifyRegionLevel::Disabled);
+    rewriter_config.enableConstantCSE(false);
```
andrej, line 131: Why is this needed?

lib/Targets/AIETargetCppTxn.cpp
```diff
@@ -0,0 +1,72 @@
+//===- AIETargetCppTxn.cpp - EmitC-based C++ TXN translation ------*- C++
+//-*-===//
```
andrej: format

programming_examples/basic/matrix_multiplication/single_core_dynamic/dynamic_gemm_txn.h
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  if (!found_rtp)
+    throw std::runtime_error("Could not find RTP write in static instructions");
+  if (!found_s2mm)
+    throw std::runtime_error(
+        "Could not find S2MM maskwrite in static instructions");
+
+  return c;
+}
```
andrej: Might it make sense to add a compiler option to export some of these "magic values" at compile time, into say a JSON that could be ingested at runtime? Some of them, like the controller ID, probably should also just get baked into the code generated by emitC.

programming_examples/basic/matrix_multiplication/single_core_dynamic/dynamic_gemm_txn.h
andrej, line 1: This doesn't seem to use the code generated from the emitC, but instead reconstructs the instruction sequence manually using the encoding library.

andrej (changes requested):
Nice to see this starting to take shape. The biggest question is if we want to deprecate the attributes; I'd be in favor of it, although it would mean touching potentially a lot of tests (but the actual code would be smaller).

The GEMM test unfortunately doesn't seem to use the added infrastructure. I think we are already on the same page, but just to make sure, what I'm envisioning looks more like this:

User writes a single runtime sequence in MLIR/Python (pseudocode):
```
aiex.runtime_sequence @my_sequence(%A: memref, %B: memref, %C: memref, %param_M: int, %param_K: int, %param_N: int) {\
  ...
  aie.dma_memcpy_nd(...)
  ...
}
```

User calls compiler roughly like so:
```
aie-opt --aie-to-cpp aie.mlir -o my_runtime_sequence.h
```

which produces something like this using emitC (important -- this is compiler generated from the above MLIR, not manually written like in the GEMM test):
```
#include <txn_encoding.h>
std::vector<uint32_t> my_sequence(void *A, void *B, void *C, int param_M, int param_K, int param_N) {
    std::vector<uint32_t> txn;
    aie_runtime::txn_append_write32(txn, param_M, ...)
    ...
}
```

and then can use that generated file in their `test.cpp` like so:
```
#include "my_runtime_sequence.h"

int main(){
   // setup XRT
   xrt::kernel my_kernel = // get out of xclbin
   std::vector<uint32_t> insts = my_sequence( my params ... )
   my_kernel(insts, a, b, c);
}
```

So ideally the GEMM test's `test.cpp` at the end of this would not look significantly more complicated than the existing ones do.

Again, cool to see this taking shape!

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
yenjames: These two upstream passes might be relevant: https://mlir.llvm.org/doxygen/ArithToEmitC_8cpp_source.html https://mlir.llvm.org/doxygen/SCFToEmitC_8cpp_source.html.

The issue I think is **all** values are mapped to `uint32_t` so there is the IR walking + hand-rolled patterns. Do we need to convert MLIR signless integers to `uint32_t` before doing arithmetic or can we let translateToCpp treat them as int32_t and only casting to `uint32_t` at txn calls?
for example C++ output like:
```cpp
int32_t addr = base + offset;
aie_runtime::txn_append_write32(txn, static_cast<uint32_t>(addr)...) // casting at the txn call site
```
This way we can directly apply conversion patterns `populateArithToEmitCPatterns()` `populateSCFToEmitCConversionPatterns()` and we cast `uint32_t` for AIEX patterns.

include/aie/Dialect/AIEX/IR/AIEX.td
jgmelber: This is fair, and I agree. Though I want to get this working without impacting the current flow. For now, it is entirely optional. Later I'd like to debate and decide if it should remain a separate path or transition to SSA values with constants in the static case.

hunhoffe:
Superseded by https://github.com/Xilinx/mlir-aie/pull/3168

PR #3003, opened 2026-03-27
Issue with assign address buffer at bank-aware strategy

joeldushouyu:
I’ve been encountering this issue for a while but only recently had the chance to investigate it more thoroughly.

In **bank-aware mode**, I noticed that when assigning an address to a buffer using the Python placement API, the allocation does not always honor the specified address. This becomes problematic when working with RTP addresses in my designs and is likely the reason why the `address_aware_alloc.mlir` test case fails.

- Python placement API:  
  https://github.com/Xilinx/mlir-aie/blob/3a1aba217920bc4475b5d705c165e7b678a2646a/python/dialects/aie.py#L325

After examining the implementation, I found that this behavior is expected: the current logic only guarantees correct placement when **both the address and the memory bank ID are explicitly specified**.

- Relevant implementation:  
  https://github.com/Xilinx/mlir-aie/blob/3a1aba217920bc4475b5d705c165e7b678a2646a/lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp#L413

However, the Python API does not expose an interface for setting the memory bank ID, which limits usability.

In this PR, I changed the code to use the following order. 

1. Address-driven bank inference
- If an `address` attribute is provided, the system infers the corresponding `bank_id`.
- If a `bank_id` is also specified, a consistency check ensures it matches the inferred value.

2. Bank-driven address inference
- If only the `bank_id` is provided, the system infers a valid address within that memory bank.

The changes is pushed in [this commit](1b35b228ec88ac9b693e18a045389be79eae8afc) and passed the address_aware_alloc.mlir test case


lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -404,16 +423,35 @@ bool simpleBankAwareAllocation(TileOp tile) {
...
         return false;
-      if (!has_addr.value() && !has_bank.value())
-        buffersToAlloc.push_back(buffer);
-      else
+      if (has_bank.value()) {
         preAllocatedBuffers.push_back(buffer);
+        continue;
+      }
+      // Shouldn't reach here since isBufferPreAllocated was true.
+      return false;
```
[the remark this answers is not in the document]
joeldushouyu: Perhaps we should just let this fail at compile time. That way, it's up to the developer to figure out if they should change the address or use another buffer—especially since this is restricted to the 'placed' API.

joeldushouyu:
I am not sure why an error is expected for the first test case in [`bank_aware_prealloc_error.mlir`](https://github.com/Xilinx/mlir-aie/blob/main/test/assign-buffer-addresses/bank_aware_prealloc_error.mlir#L13-25).

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -404,16 +426,35 @@ bool simpleBankAwareAllocation(TileOp tile) {
   // the above.
   for (auto buffer : allBuffers) {
     if (buffer.getTileOp() == tile) {
+
+
+
+      if (!isBufferPreAllocated(buffer)) {
+        buffersToAlloc.push_back(buffer);
+        continue;
+      }
```
abisca, line 430: With this in mind, changing `isBufferPreAllocated()` to check that both attributes are set (returning true only in that case) and using this result to avoid the other checks will yield a cleaner implementation. A sanity check could also be added to ensure that the address and mem bank are coherent.

abisca (commented):
It's my current understanding that the limitation you are trying to solve is: buffers that already have both an address and a mem_bank should not be checked (`checkAndAddBufferWithAddress()` and `checkAndAddBufferWithMemBank()`) and should instead be immediately added to `preAllocatedBuffers`.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
joeldushouyu: If I understand correctly, you want to enable customize buffer allocation only if caller gives both the address and bank_id?  Currently, the function returns true if either the bank_id or address is set

fifield: The bank aware allocator has never worked for me for manual buffer placement, I always use the sequential allocator if doing manual placement. One of my complaints is that the bank aware does things like "A sanity check ... to ensure that the address and mem bank are coherent", when I almost always want the allocator to ignore any buffers with address already set, and simply allocate around the hole(s) created by them.

abisca: > If I understand correctly, you want to enable customize buffer allocation only if caller gives both the address and bank_id?

That wasn't quite what I was trying to say. I am a little out of touch with this part of the codebase so I likely misunderstood the intent of the PR. Let me try to reason it out.

During bank-aware allocation, buffers are checked for their address and mem_bank. If one is missing, the other is completed. My point above was that if a buffer already has both of these attributes set, then there is no need to do both of the checks, even at all. However, custom buffer allocation should still enable the user to provide either only:
- the address, and then the compiler finds the corresponding bank,
- the mem_bank, and the compiler assigns an address to the buffer within that bank.
 
Currently, the allocator seems to overwrite the custom field if there's no room for it. I think it'd be better to enforce the custom choice and let the error happen.

joeldushouyu: I see. I should probably put more comments in the code.

The if (!isBufferPreAllocated(buffer)) {} part of the code is for normal bank-aware allocation, aka the caller did not set either the address or bank ID.

As far as I am aware of, the code should return an error if the allocator failed to allocate the required address/bank id.

joeldushouyu:
> Thank you @joeldushouyu for adding the support. With the pre-allocated buffers, , we might end up creating memory fragments that go unused with the simple allocation techniques we currently have in place. In particular, with bank-aware allocation, this could lead to more conflicts than the user might expect. Because of this, it seems more reasonable to allow only the bank ID to be specified by the user for this feature, rather than the address, since specifying the address could introduce the issues mentioned above. Please let us know if this approach would be sufficient for your application.
> 
> We can also look into smarter allocation strategies in the future if needed.

For my usecase, it is actually crucial for the user to specify the address due to the rtp bufer and the runtimesequence is generated at **runtime**.

Regarding the internal bank-aware allocation, I have write some about it in this [issue](https://github.com/Xilinx/mlir-aie/issues/3019?reload=1). Not sure if you guys like it or not.

jgmelber:
> I am not sure if I should add a warning message

I don't think we need a warning

joeldushouyu:
Good to merge if the new fix passes the lint test?

jgmelber:
> Good to merge if the new fix passes the lint test?

@pvasireddy-amd @abisca @fifield

pvasireddy-amd:
It might be worth adding a test to show that a preallocated buffer occupies space in a bank, which can prevent allocating other buffers and cause a fallback to sequential placement. This would help future contributors better understand the allocator's behavior. Other than that LGTM :)

test/assign-buffer-addresses/fallback_alloc_error.mlir
```diff
@@ -0,0 +1,24 @@
+//===- bank_aware_prealloc_error.mlir --------------------------*- MLIR -*-===//
```
pvasireddy-amd: Change name

test/assign-buffer-addresses/fallback_alloc_error.mlir
```diff
@@ -0,0 +1,24 @@
...
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2024, Advanced Micro Devices, Inc.
```
pvasireddy-amd: Update copyright

pvasireddy-amd (approved):
LGTM, thank you @joeldushouyu for all the changes!

test/assign-buffer-addresses/fallback_alloc_error.mlir
joeldushouyu: 0eed5ce2dd7554bb0cf797de00c58aba002c59b5

test/assign-buffer-addresses/fallback_alloc_error.mlir
joeldushouyu: 0eed5ce2dd7554bb0cf797de00c58aba002c59b5

jgmelber (approved):

Issue #2738, opened 2025-11-25
SequentialPlacer Channel Counting


hunhoffe:
Right now the ```SequentialPlacer``` uses heuristics to explore the placement of ObjectFifo endpoints to attempt to avoid overallocating channels. This is done without querying whether the device supports shared memory, so this heuristic can diverge a fair amount from the number of channels actually required by ObjectFifos in a design.

Ideally, the ```SequentialPlacer``` will use logic in the compiler to power it's heuristics, e.g., query if shared memory can be used instead of a channel to avoid over-counting channels. This is a bit of a complex task, and I'm not sure how well it will work, so I'm breaking it up into smaller pieces.

These pieces include:
- [x] Exposing the ```is_legal_mem_affinity``` function in the device class (https://github.com/Xilinx/mlir-aie/pull/2691)
- [x] Creating and testing the usefulness of this function for this purpose as an incremental step (https://github.com/Xilinx/mlir-aie/pull/2733)
- [ ] Modifying the ```SequentialPlacer```to use this function in Channel counting
- [ ] Reduce duplicated logic between the ```SequentialPlacer``` and the compiler for estimating channels

In addition to the overall goal of having the ```SequentialPlacer``` decision making align with decision making within the compiler, this should also --- hopefully --- make the ```SequentialPlacer``` more robust (e.g., https://github.com/Xilinx/mlir-aie/issues/2666)

hunhoffe:
Deprecating this issue since the python sequential placer has been deprecated.

PR #3016, opened 2026-03-31
[New] Deprecate IRON sequential placer

yenjames:
### Summary
Migrate IRON tile placement from the Python SequentialPlacer to the C++ --aie-place-tiles compiler pass. IRON Tile objects now emit aie.logical_tile ops with a tile type constraint and optional coordinates instead of absolute aie.tile coordinates, and the compiler resolves physical placement. 

### Changes
Compiler
  - Register --aie-place-tiles into the aiecc pipeline when aie.logical_tile ops are detected in the IR
  - Expose getTileType() through bindings for direct tile type queries from Python

  IRON Python API
  - Delete placeable.py and placers.py — remove SequentialPlacer, Placeable, and PlacementTile type alias
  - Rename placement= parameter to tile= across Worker, RuntimeEndpoint, ObjectFifoLink, join/split/forward, fill/drain
  - Tile sentinels (AnyComputeTile, AnyMemTile, AnyShimTile) are now Tile instances with tile.copy() instead of separate sentinel classes
  - Add Tile.copy(), coordinate validation in Device.get_tile_type(), and coordinate merging in Device.resolve_tile()
  - Maintain device query functions (get_tile_type, is_mem_accessible, connection counts) — these error if tiles don't have coordinates

### Regression issue in `test_compile_cache_functionality.py` unrelated to changes in this PR
- Temporarily disabled float32 jit test in #3023


programming_examples/ml/bottleneck/bottleneck.py
```diff
@@ -138,7 +137,7 @@ def bottleneck4AIEs():
     # AIE-array data movement with object fifos
     of_inOF_act_L3L2 = ObjectFifo(tensorLayer1In_ty, name="inOF_act_L3L2")
     of_skip_buf = of_inOF_act_L3L2.cons(4).forward(
-        depth=2, placement=AnyMemTile, name="skip_buf"
+        depth=2, tile=Tile(), name="skip_buf"
```
andrej: Is this the same as `AnyMemTile` or more like `AnyTile`?

python/iron/device/tile.py
```diff
@@ -6,41 +6,73 @@
...
+    def op(self, op: LogicalTileOp | TileOp):
         if self._op and self._op != op:
             raise ValueError("Cannot change operation once it is set.")
         self._op = op
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, Tile):
             return NotImplemented
-        return self.col == other.col and self.row == other.row
+        return id(self) == id(other)
```
andrej, line 71: What happens if `self` and `other` are distinct objects, but both are placed with the same row and col?

test/python/npu-xrt/test_objectfifo.py
```diff
@@ -146,3 +76,74 @@ def test_split():
...
+    assert ep.tile.tile_type == AIETileType.ShimNOCTile
+    # None tile_type is accepted (RuntimeEndpoint sets it to ShimNOCTile)
+    ep2 = RuntimeEndpoint()
+    assert ep2.tile.tile_type == AIETileType.ShimNOCTile
+
+
+def test_workers_cannot_share_tile():
+    """Two workers cannot be placed on the same tile object."""
+    n_ty = np.ndarray[(1024,), np.dtype[np.int32]]
+    shared_tile = Tile(0, 2)
```
andrej, line 185: I suggest adding a test `tile_1 = Tile(0, 2)`, `tile_2 = Tile(0, 2)` and checking that this still errors.

andrej (approved):

programming_examples/ml/bottleneck/bottleneck.py
yenjames: This was an incorrect pattern match rewrite by Claude, thank you for the catch!

python/iron/device/tile.py
yenjames: Currently we need to error on this one because worker tiles are not allowed to refer to the same physical tile. In the future when we extend support to logically place **any** tile component the error handling can be removed.

python/iron/dataflow/objectfifo.py
```diff
@@ -228,8 +229,8 @@ def tiles(self, cons_only: bool = False) -> list[PlacementTile]:
         tiles += [cons.endpoint.tile for cons in self._cons]
         return tiles
 
-    def can_used_shared_mem(self, device: Device, cons_only: bool = False) -> bool:
-        """Checks if all endpoints of the object fifo have a legal memory affinity."""
+    def can_used_shared_mem(self, device, cons_only=False):
```
hunhoffe: Nit: Please keep type hints as long as they are accurate

python/iron/dataflow/objectfifo.py
```diff
@@ -11,6 +11,7 @@
 
 from ... import ir  # type: ignore
 from ...dialects._aie_enum_gen import ObjectFifoPort  # type: ignore
+from ..device.tile import AIETileType
```
hunhoffe: Nit: I don't love prefixing things with AIE.... it's already implied by the python module hierarchy

python/iron/dataflow/objectfifo.py
```diff
@@ -467,11 +468,16 @@ def __str__(self) -> str:
 
     @endpoint.setter
     def endpoint(self, endpoint: ObjectFifoEndpoint) -> None:
-        if self._endpoint and self._endpoint != endpoint:
-            raise ValueError(
-                f"Endpoint already set for ObjectFifoHandle {self.name}.{self.handle_type}: "
-                f"Set to {self._endpoint}, trying to set to {endpoint}"
-            )
+        if self._endpoint is not None:
```
hunhoffe: It's not only the type that needs protection, right? You can't specify more than one shim per objectfifo, which was ensured in the previous check but not now?

python/iron/dataflow/objectfifo.py
```diff
@@ -496,7 +502,7 @@ def join(
 
         Args:
             offsets (list[int]): Offsets into the current producer, each corresponding to a new consumer.
-            placement (PlacementTile, optional): The placement where the Join operation occurs. Defaults to AnyMemTile.
+            tile (Tile | None, optional): The tile where the Join operation occurs. Defaults to a new unplaced mem tile.
```
hunhoffe: Nit: It does not default to an unplaced mem tile any more

python/iron/dataflow/objectfifo.py
```diff
@@ -578,7 +584,7 @@ def split(
 
         Args:
             offsets (list[int]): The offset into the current consumer for each new ObjectFifo producer.
-            placement (PlacementTile, optional): The placement tile where the Split operation takes place. Defaults to AnyMemTile.
+            tile (Tile | None, optional): The tile where the Split operation takes place. Defaults to a new unplaced mem tile.
```
hunhoffe: Default is non, not unplaced mem tile

python/iron/dataflow/objectfifo.py
```diff
@@ -657,7 +663,7 @@ def forward(
         is forwarded to the producer of a newly-constructed ObjectFifo.
 
         Args:
-            placement (PlacementTile, optional): The placement of the Forward operation. Defaults to AnyMemTile.
+            tile (Tile | None, optional): The tile for the Forward operation. Defaults to a new unplaced mem tile.
```
hunhoffe: Default is None not unplaced mem tile

python/iron/dataflow/objectfifo.py
```diff
@@ -722,7 +728,8 @@ def __init__(
         Args:
             srcs (list[ObjectFifoHandle] | ObjectFifoHandle): A list of consumer ObjectFifoHandles to link.
             dsts (list[ObjectFifoHandle] | ObjectFifoHandle): A list of producer ObjectFifoHandles to link.
-            placement (PlacementTile, optional): The place the link occurs. Defaults to AnyMemTile.
+            tile (Tile | None, optional): The tile where the link occurs. Defaults to a new unplaced mem tile.
+                This is the only component where tile_type can be overridden (e.g., to CoreTile).
```
hunhoffe: Default is not unplaced mem tile. And I don't understand the overridden piece? You can specify types elsewhere?

python/iron/dataflow/objectfifo.py
```diff
@@ -756,7 +763,10 @@ def __init__(
             s.endpoint = self
         for d in self._dsts:
             d.endpoint = self
-        ObjectFifoEndpoint.__init__(self, placement)
+        tile = Tile.resolve(tile)
+        if tile.tile_type is None:
+            tile.tile_type = AIETileType.MemTile
```
hunhoffe, line 761: What if the user set it to None? Since None is a valid placement in this context, I think?

python/iron/device/device.py
```diff
@@ -5,107 +5,77 @@
...
+        return self._tm.columns()
+
+    @property
+    def rows(self) -> int:
+        """Number of rows in the device tile array."""
+        return self._tm.rows()
+
+    def _infer_tile_type(self, col, row):
+        """Infer the AIETileType from device coordinates."""
+        if self._tm.is_core_tile(col, row):
```
hunhoffe: Can you add a convenient wrapper to _tm that's something like get_tile_type() instead of putting the if/else ehre?

python/iron/device/device.py
```diff
@@ -5,107 +5,77 @@
...
+            return AIETileType.CoreTile
+        elif self._tm.is_mem_tile(col, row):
+            return AIETileType.MemTile
+        elif self._tm.is_shim_noc_tile(col, row):
+            return AIETileType.ShimNOCTile
+        elif self._tm.is_shim_pl_tile(col, row):
+            return AIETileType.ShimPLTile
+        else:
+            raise ValueError(f"Cannot determine tile type for ({col}, {row})")
```
hunhoffe: If you give an invalid col/row I htink that should be an explicit error that happens before even checking the type

python/iron/device/device.py
```diff
@@ -252,8 +241,36 @@ def resolve_tile(
...
+
+        if tile.tile_type is None:
+            if tile.col is not None and tile.row is not None:
+                tile.tile_type = self._infer_tile_type(tile.col, tile.row)
+            else:
+                raise ValueError(
+                    f"Cannot resolve {tile}: tile_type must be set or inferred from coordinates."
+                )
+        elif tile.col is not None and tile.row is not None:
+            inferred = self._infer_tile_type(tile.col, tile.row)
```
hunhoffe: For both conditions, can't you infer the type with just the row?

python/iron/device/tile.py
```diff
@@ -6,41 +6,73 @@
...
+        self,
+        col: int | None = None,
+        row: int | None = None,
+        *,
+        tile_type: AIETileType | None = None,
+        allocation_scheme: str | None = None,
+    ) -> None:
+        self.col: int | None = col
+        self.row: int | None = row
+        self.tile_type: AIETileType | None = tile_type
```
hunhoffe, line 43: Should AnyType be an option in the enum? e.g., is it actually valid for this to be None throughout? I am genuinely unsure of the answer to this question.

python/iron/device/tile.py
```diff
@@ -61,4 +93,8 @@ class AnyComputeTile:
     pass
 
 
-PlacementTile = Tile | AnyShimTile | AnyMemTile | AnyComputeTile
+_ANY_TILE_MAP = {
```
hunhoffe: If I remember correctly, sometimes AIETileType is exposed at the top level user... I think we need to pick just one. And I don't think there's any point in AnyCompute, etc. if the AIETileTypes represent the same thing?

python/iron/runtime/endpoint.py
```diff
@@ -9,16 +9,23 @@
...
 class RuntimeEndpoint(ObjectFifoEndpoint):
     """The Endpoint of an ObjectFifo that is produced or consumed by the Runtime.
     The placement of this Endpoint should be a Shim Tile.
     """
 
-    def __init__(self, placement: PlacementTile) -> None:
-        super().__init__(placement)
+    def __init__(self, tile: Tile | None = None) -> None:
+        tile = Tile.resolve(tile)
+        if tile.tile_type is not None and tile.tile_type != AIETileType.ShimNOCTile:
```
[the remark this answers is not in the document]
hunhoffe: Good note, copilot

python/iron/runtime/runtime.py
```diff
@@ -145,7 +145,7 @@ def fill(
         tap: TensorAccessPattern | None = None,
         task_group: RuntimeTaskGroup | None = None,
         wait: bool = False,
-        placement: PlacementTile = AnyShimTile,
+        tile: Tile | None = None,
```
hunhoffe: I'm confused about when things can be a LogicalTile vs a Tile vs a TileType. I think we need to clarify how they interact and then ensure at most 2 (but ideally just 1) is exposed to the user in the top-level API.

python/iron/runtime/runtime.py
```diff
@@ -157,7 +157,7 @@ def fill(
                 If None is given, this will default to a linear transfer containing all data in the source buffer. Defaults to None.
             task_group (RuntimeTaskGroup | None, optional): A TaskGroup to associate this task with. Defaults to None.
             wait (bool, optional): Whether this Task should be awaited on or not. If not, it will be freed when the task group is finished. Defaults to False.
-            placement (PlacementTile, optional): The Shim tile to associate the data transfer with. Defaults to AnyShimTile.
+            tile (Tile | None, optional): The shim tile for this data transfer. Defaults to a new unplaced shim tile.
```
hunhoffe: Does not default to unplaced shim.

python/iron/runtime/runtime.py
```diff
@@ -194,7 +194,7 @@ def drain(
                 If None is given, this will default to a linear transfer containing all data in the destination buffer. Defaults to None.
             task_group (RuntimeTaskGroup | None, optional): A TaskGroup to associate this task with. Defaults to None.
             wait (bool, optional): Whether this Task should be awaited on or not. If not, it will be freed when the task group is finished. Defaults to False.
-            placement (PlacementTile, optional): The Shim tile to associate the data transfer with. Defaults to AnyShimTile.
+            tile (Tile | None, optional): The shim tile for this data transfer. Defaults to a new unplaced shim tile.
```
hunhoffe: Does not default to unplaced shim.

python/iron/buffer.py
```diff
@@ -44,7 +43,7 @@ def __init__(
             type (type[np.ndarray] | None, optional): The type of the buffer. Defaults to None.
             initial_value (np.ndarray | None, optional): An initial value to set the buffer to. Should be of same datatype and shape as the buffer. Defaults to None.
             name (str | None, optional): The name of the buffer. If none is given, a unique name will be generated. Defaults to None.
-            placement (PlacementTile | None, optional): A placement location for the buffer. Defaults to None.
+            tile (Tile | None, optional): The tile for the buffer. Usually set automatically by the Worker. Defaults to None.
```
hunhoffe: Maybe a little more descriptive: "Usually set to match the tile of the Worker that accesses the Buffer."

python/iron/worker.py
```diff
@@ -45,18 +46,24 @@ def __init__(
...
-                Will override any allocation scheme set on the tile given as placement.
+                Will override any allocation scheme set on the tile.
             trace (int, optional): If >0, enable tracing for this worker.
             trace_events (list | None, optional): Custom list of trace events for this worker. Defaults to None.
 
         Raises:
             ValueError: Parameters are validated.
         """
-        self._tile = placement
+        tile = Tile.resolve(tile)
```
hunhoffe: You should not resolve things in Init I don't think (I could be misremembering the code I wrote.... so please feel free to tell me if I'm wrong). In my mind, resolve() implies that you're changing the tile from a purely python object into an MLIR operation, which can/should only be done during the call to resolve_program().

python/iron/worker.py
```diff
@@ -45,18 +46,24 @@ def __init__(
         Args:
             core_fn (Callable | None): The task to run on a core. If None, a busy-loop (`while(true): pass`) core will be generated.
             fn_args (list, optional): Pointers to arguments, which should include all context the core_fn needs to run. Defaults to [].
-            placement (PlacementTile | None, optional): The placement for the Worker. Defaults to AnyComputeTile.
+            tile (Tile | None, optional): The compute tile for the Worker. Defaults to a new unplaced compute tile.
```
hunhoffe: Does not match default value of None

python/iron/program.py
```diff
@@ -67,15 +65,23 @@ def device_body():
...
-                        self._device, self._rt, self._rt.workers, all_fifos
-                    )
-
-                # Collect all tiles
+                # Collect all tiles, validating no two workers share the same tile
                 all_tiles = []
+                worker_tile_ids = set()
+                worker_tile_coords = set()
                 for w in self._rt.workers:
+                    if id(w.tile) in worker_tile_ids:
```
hunhoffe: Can't you just always check for no duplicate coords? What does it matter if it's the same object?

python/utils/trace/setup.py
```diff
@@ -407,14 +410,15 @@ def configure_trace(
     packet_id = 1
     seen_core_tiles = set()
 
-    for tile_op in tiles_to_trace:
+    for tile_val in tiles_to_trace:
+        tile_op = tile_val if isinstance(tile_val, OpView) else tile_val.owner
```
hunhoffe: Why is this needed?

test/python/npu-xrt/test_objectfifo.py
```diff
@@ -11,87 +11,17 @@
...
     Tile,
     AnyMemTile,
     AnyComputeTile,
     AnyShimTile,
 )
-from aie.iron.dataflow.objectfifo import ObjectFifo
-from aie.iron.dataflow.endpoint import ObjectFifoEndpoint
-
-
-@pytest.fixture(params=[NPU1Col1, NPU1, NPU2])
```
hunhoffe, line 27: Are you not checking for multiple device types any more?

hunhoffe (commented):
Overall, most of my comments are about small things. However, there is a general theme where I'm not sure how this interface is exposed to the user? Do we need to also explain logicaltiles/types/etc in the programming guide? Is it consistent throughout the API?

python/iron/dataflow/objectfifo.py
yenjames: Sorry this is an incorrect indirect import, it should be imported from dialects._aie_enum_gen like ObjectFifoPort. My idea was to keep AIETileType internal as it's an enum generated from mlir. Users should be expected to use the original Any*Tile to directly declare a type of tile. The AIE prefix is due to the mlir attribute naming.

python/iron/device/device.py
yenjames: Yes we could based on current architectures. This is using the function binded from mlir target model which requires both coordinates. My idea is this ground truth lookup allows easier extensions for future architectures. The trade-off is we cannot infer tile type off of partial coordinates unless we hardcode.

python/iron/device/tile.py
yenjames: We could definitely add that if there is a use case, it would require some minor change to the placer. This would be added on the mlir side.

python/iron/device/tile.py
yenjames: Yes to clarify, my plan is to actually not expose AIETileType but only keep it internal to IRON classes, and keep API the same as before. As you mentioned AIETileType is wordy but it's generated from bindings, so if we would like to modify we would have to write some IRON class to wrap it, which the Any*Tile classes accomplish. I recommend not changing AIETileType in MLIR as that's the prefix convention we use.

python/utils/trace/setup.py
yenjames: Claude did a roundabout solution due to aie.logical_tile op not having the same overwrite to return op views as aie.tile.

python/iron/program.py
yenjames: This is a check for the limitation on worker tiles being unmergable. For example:
```python
shared = Tile()
w1 = Worker(None, tile=shared)
w2 = Worker(None, tile=shared)
```
However, the latest refactor removes this check anyways but it is ensured by creating a new instance of tile each time.

test/python/npu-xrt/test_objectfifo.py
yenjames: This was removed by mistake, thanks for the catch!

python/iron/dataflow/objectfifo.py
yenjames: With some refactoring this is now restored.

python/iron/dataflow/objectfifo.py
yenjames: Yes this was an unwanted change, restored all Tile defaults to be the type described in docstring.

python/iron/dataflow/objectfifo.py
yenjames: Incorrect description. You can always specify a different type but most cases enforce a type check as tile type is implied by caller. Forward is allowed to be called on core tiles type as well as memory tile.

python/iron/dataflow/objectfifo.py
yenjames: Removed None from possibility. Must be a Tile class, Any*Tile are instances of Tile class.

python/iron/device/device.py
yenjames: Binded `get_tile_type()` from mlir target model, and added an additional helper to device to check invalid coordinates outside device range.

hunhoffe:
When this is merged, we can close this issue: https://github.com/Xilinx/mlir-aie/issues/2738

jgmelber (approved):
Looks great! This was a lot for me quality work

PR #3021, opened 2026-04-07
fix(cmake): remove redundant add_definitions(${LLVM_DEFINITIONS})

### LINKED ISSUE #2965: CppTests fail to build with GCC due to duplicate compile definitions
## Problem

Building mlir-aie with GCC fails on the CppTests targets (`target_model`, `target_model_rtti`, `register_database`) with errors like:

```
<command-line>: error: token '=' is not valid in preprocessor expressions
```

The generated compile command has garbled `-D` flags where multiple definitions are concatenated into a single quoted string:

```
-D_GLIBCXX_USE_CXX11_ABI=1
-D_GLIBCXX_USE_CXX11_ABI="1 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS"
-D_GLIBCXX_USE_CXX11_ABI="1 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS -D_GLIBCXX_USE_CXX11_ABI=1 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS"
```

GCC rejects these; Clang silently accepts the malformed flags (which is why CI likely does not catch this).

## Root Cause

The definitions are added twice through different mechanisms:

1. `test/CMakeLists.txt:91` includes `HandleLLVMOptions`, which calls `add_compile_definitions()` for `_GLIBCXX_USE_CXX11_ABI`, `__STDC_CONSTANT_MACROS`, `__STDC_FORMAT_MACROS`, and `__STDC_LIMIT_MACROS` individually.

2. `test/CMakeLists.txt:97` calls `add_definitions(${LLVM_DEFINITIONS})`, which re-adds the same flags as a single space-separated string (`"-D_GNU_SOURCE -D_GLIBCXX_USE_CXX11_ABI=1 -D__STDC_CONSTANT_MACROS -D__STDC_FORMAT_MACROS -D__STDC_LIMIT_MACROS"`).

3. The top-level `CMakeLists.txt` has the same pattern (lines 157 + 165), adding a third layer via directory property inheritance.

When CMake generates the compile command, the space-separated string from `add_definitions()` gets quoted as a single token, producing the garbled output.

The same pattern exists in the top-level `CMakeLists.txt` (lines 157 and 165), though it only manifests as a build failure in CppTests because those are the targets where the duplication reaches the threshold that GCC rejects.

## Verified Fix

Removing the redundant `add_definitions(${LLVM_DEFINITIONS})` from `test/CMakeLists.txt:97` resolves the build failure. All three CppTests (`TargetModel`, `TargetModelRtti`, `RegisterDatabase`) then build and pass with GCC 15.

A harmless redefinition warning remains from the top-level `add_definitions()`, but the build succeeds.

## Environment

- GCC 15 (Ubuntu 25.10)
- LLVM 23 (built from mlir-aie's pinned commit `979132a`)
- CMake 3.31
- Ubuntu 25.10 (kernel 6.19)

FIM43-Redeye:
## Summary

- Remove redundant `add_definitions(${LLVM_DEFINITIONS})` from all six CMakeLists.txt files that also `include(HandleLLVMOptions)`
- `HandleLLVMOptions` already adds every flag individually via `add_compile_definitions()`, making these calls purely duplicative
- The redundant calls caused garbled `-D` flags with GCC because `LLVM_DEFINITIONS` is a space-separated string, not a CMake list

Fixes #2965.

## Test plan

- [x] Verified `target_model`, `target_model_rtti`, and `register_database` CppTests build and pass with Clang after removal
- [x] CI: GCC compile matrix should now pass cleanly

Generated using Claude Code.


jgmelber (approved):

FIM43-Redeye:
CI checks out,but the queue bot kicked it out on Apr 14. Still seems okay, though. Anything else to take care of before merge?

PR #3024, opened 2026-04-08
Build test_utils on Windows to fix missing lib/ in wheel

erwei-xilinx:
## Summary

- The `WIN32` guard in `runtime_lib/test_lib/CMakeLists.txt` unconditionally disabled both `BUILD_TEST_LIBRARY` and `BUILD_TEST_UTILS_LIBRARY`, causing the Windows wheel to ship headers but no compiled libraries under `runtime_lib/x86_64/test_lib/lib/`
- `test_library.cpp` requires POSIX APIs (`<sys/mman.h>`) and legitimately cannot build on Windows, but `test_utils.cpp` is portable standard C++ that compiles fine with MSVC
- Remove the `BUILD_TEST_UTILS_LIBRARY OFF` line so that `test_utils.lib` is compiled and included in the Windows wheel

## Test plan

- [ ] Verify Windows CI wheel build succeeds and `test_utils.lib` appears in the wheel under `runtime_lib/x86_64/test_lib/lib/`
- [ ] Verify Linux build is unaffected (`libtest_utils.a` still built as before)

### DIFF
--- a/runtime_lib/test_lib/CMakeLists.txt
+++ b/runtime_lib/test_lib/CMakeLists.txt
@@ -7,6 +7,9 @@
 cmake_minimum_required(VERSION 3.30)
 project("test lib for ${AIE_RUNTIME_TARGET}")
 
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
+
 function(add_runtime_pic_flag target_name)
   if (NOT WIN32)
     target_compile_options(${target_name} PRIVATE -fPIC)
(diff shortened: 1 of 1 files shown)


jackl-xilinx (approved):

thomthehound:
@erwei-xilinx Thank you for handling this. I had intended to address it in a follow-up patch, but personal issues intervened and I had to step away from the project for a while.

I understand that mlir-air is causing some problems with Windows Python 3.13/3.14 support in your Triton project. I will set aside some time next week to look into that in exchange. I've been trying to get that modernized across the board, anyway (the Windows XRT SDK ships with the assumption of Python 3.13 in the `pyxrt.pyd` extension module, so it is needed).

Issue #2273, opened 2025-05-06
IRON API bindings for packet switching


jgmelber:
Add support for IRON (unplaced) API using packet switching. 

Follow up to #2242 #2235 #2266

hunhoffe:
Addressed in https://github.com/Xilinx/mlir-aie/pull/3025

Issue #2206, opened 2025-04-15
Improvements to JIT workflow


mawad-amd:
PR https://github.com/Xilinx/mlir-aie/pull/2201 added JIT compilation workflows to IRON and some future improvements and suggestions are below. This issue is the umbrella issue for other future sub-tasks.

- [ ] Better syntax for launching deferred placement IRON kernels (See: https://github.com/Xilinx/mlir-aie/pull/2201#issuecomment-2806863573)
- [ ] Improvements to `aiecc.py` to facilitate using the Python module and redirecting outputs to specific locations (See: https://github.com/Xilinx/mlir-aie/pull/2201#discussion_r2045021145)
- [ ] Automatically detecting the device and not requiring users to specify them (see: https://github.com/Xilinx/mlir-aie/pull/2201#discussion_r2042762849) 
- [ ] Better/consistent naming for implicit/deferred placement (See: https://github.com/Xilinx/mlir-aie/pull/2201#discussion_r2045004090)
- [ ] Control over where the JIT-compiled binaries are stored (See: https://github.com/Xilinx/mlir-aie/pull/2201#discussion_r2045014041)

ypapadop-amd:
One thing that https://github.com/Xilinx/mlir-aie/pull/2201 has removed is an easy way to do offline compilation of kernels. E.g., before I would pick up the vector_vector_add.py and dig through the makefiles to figure out how to generate an xclbin and then use it directly w/ XRT in my application. Now this is not possible.

jgmelber:
> One thing that [#2201](https://github.com/Xilinx/mlir-aie/pull/2201) has removed is an easy way to do offline compilation of kernels. E.g., before I would pick up the vector_vector_add.py and dig through the makefiles to figure out how to generate an xclbin and then use it directly w/ XRT in my application. Now this is not possible.

Yes, that is a good point. This is why the plan will be to keep some examples with the Makefiles and bare XRT C++ around.

ypapadop-amd:
I think that having the same example working with two drivers (Python and C++) in the same directory would make it clear how to transition from one to the other.

mawad-amd:
I will add an example that showcases both approaches. Any preference on which example? I plan to _iron out_ the deferred placement comments first before touching other code.

jgmelber:
GEMM for sure should have all approaches since that is a design that might be integrated into a variety of frameworks. Up for suggestions on another

ypapadop-amd:
[...]

PR #3025, opened 2026-04-09
Unify `@iron.jit` compilation workflow across all programming examples

hunhoffe:
Implements the RFC at #3152 (please read first for motivation and design rationale). This PR is the migration.

Every example under `programming_examples/` and `programming_guide/` is now a single `@iron.jit`-decorated Python file; hand-rolled `pyxrt` harnesses collapse to `iron.tensor` + `DefaultNPURuntime.run_test`; per-device Makefile rule trees collapse to one `jit_xclbin` macro call.

## Issues
Closes:
* #2206 — Improvements to JIT workflow (umbrella; final unchecked item [Add tracing support] landed via `programming_examples/basic/event_trace`)
* #2273 — IRON API bindings for packet switching (`packet=` DMA kwarg + `--packet-sw-objFifos` via `@iron.jit(aiecc_flags=…)`; end-to-end test in `basic/transposes -s dma_packet`)
* #2509 — Code object cache for external functions (content-based dedupe via `ExternalFunction._content_digest`; `_EXTERN_CACHE` memoization in `kernels/_common.py`; on-disk skip-if-exists in `compile_external_kernel` under `NPU_CACHE_HOME`)
* #2541 — Add trace with JIT example (this PR supersedes; `basic/event_trace` is the production version)

## CI savings

```
                                       main    branch   delta
─────────────────────────────────────────────────────────────
programming_examples/  lit files        284      155    -129  (-45%)
```

Biggest contributors: `basic/matrix_multiplication` -34, `basic/vector_reduce_max` -10, `ml/block_datatypes` -6, and the `transposes/` consolidation (combined + dma + dma_packet + shuffle → one example) -14.

## Reviewer's reading order

957 files is a lot. Three pointers that save 30 minutes:

1. **The new core** — `python/iron/` (decorator, runtime, dataflow) + `python/utils/compile/jit/` (`compilabledesign.py`, `_hash.py`, `_serialization.py`, `_introspect.py`). This is where the abstraction lives.
2. **One migrated example end-to-end** — `programming_examples/basic/vector_scalar_mul/`. Small, exercises `Compile[T]`, the kernels library, `transform_typed`, and the dual standalone/Makefile entry path. Compare against `git show origin/main:programming_examples/basic/vector_scalar_mul/` to see the collapse.
3. **The docs** — `programming_guide/` diff. 12 backfill sections covering the new surface; what new users will read first.


hunhoffe:
note to self: kernel lib wheel build?

thomthehound:
I'm still looking at this, but it is going to take me more time. One (very minor) concern I have so far is that I think some of the JIT/external-kernel paths may be relying on include/header staging more than I would personally prefer.

In particular, I think `ninja` is doing a lot of heavy lifting to make sure the staged include tree has the right headers before the lit tests run. For centralization work like this, I think it would be cleaner if the needed kernel source/header roots were made explicit, perhaps somewhere like the new `python/iron/kernels/_common.py`, just to remove the ambiguity.

But this is by no means a critical issue. It just seems to me that a centralized JIT compilation path should try to be as self-contained as possible.

thomthehound:
OK, I've taken a longer look now. It does not appear to me that this introduces any new **classes** of cross-platform compatibility issues that I haven't dealt with already, but there are still many things that will obviously need to be reconciled with the CMake-conversion patch I've been working on. I don't think that can be helped at this point. It's not a fault of this patch, and I don't think it would be constructive for me to derail you by detailing them now. I can fold in the needed changes once this patch lands.

The only thing I see that is actionable right now (aside from my previous comment about explicit header ownership) is an issue with `NPU_CACHE_HOME` versus `IRON_CACHE_HOME`, but I'm almost certain that is just branch drift and not intent.

hunhoffe:
@thomthehound if you want to proceed and merge your patch in first, that's fine with me. I think there will be some discussion around this.... I'm not 100% sure it'll get merged at all.

thomthehound:
> @thomthehound if you want to proceed and merge your patch in first, that's fine with me. I think there will be some discussion around this.... I'm not 100% sure it'll get merged at all.

Thanks for the heads-up. I'm still a week or so out from being ready with the cmake example patch. I think I'll need the lit test patches merged first for it to fully make sense, anyway. If those don't get merged, I'll be in the same boat.

aie_kernels/aie2/bottleneck/bn_conv2dk1_skip.cc
```diff
@@ -910,290 +910,6 @@ void conv2dk1_skip_ui8_i8_i8_get(uint8_t *input0, int8_t *kernels,
...
-                                 skip_scale);
-}
-
-#endif // UNSIGNED_SKIP
-
-#else // Vector
-
-#endif // Vector
-#endif // BN9
-
```
hunhoffe, line 1196: Can you verify for me that no kernel functionality is lost with these modifications?

aie_kernels/aie2p/mm_bfp.cc
```diff
@@ -168,6 +170,18 @@ extern "C" {
 #define DIM_N 64
 #endif
 
+// MATMUL_ONLY / ZERO_ONLY / SHUFFLE_ONLY let callers (e.g. @iron.jit
+// ExternalFunction) compile a .o containing exactly one of the three
+// entry points, avoiding duplicate-symbol errors when the same .cc is
+// compiled multiple times for distinct ExternalFunctions in one design.
+// Without any macro, all three symbols are emitted (legacy behaviour).
+#if !defined(MATMUL_ONLY) && !defined(ZERO_ONLY) && !defined(SHUFFLE_ONLY)
```
hunhoffe, line 178: Is this the best way to handle these dependencies? Is it elegant, correct, and maintainable?

docs/conferenceDescriptions/asplos26TutorialDescription.md
```diff
@@ -32,7 +32,7 @@ Location: Pittsburgh, PA, USA
 | 08:15am | "Hello World" from Ryzen™ AI | Joe | [AI Engine Basic Building Blocks](../../programming_guide/section-1/) |
 | 08:35am | Exercise 1: Build and run your first program | All | [Passthrough](../../programming_examples/basic/passthrough_kernel/) |
 | 08:50am | Data movement on Ryzen™ AI with objectFIFOs | Erika | [Data Movement](../../programming_guide/section-2/) |
-| 09:10am | Exercise 2: Explore AIE DMA capabilities | All | [DMA Transpose](../../programming_examples/basic/dma_transpose/) |
```
hunhoffe, line 35: These files are legacy, let's leave them alone.

docs/conferenceDescriptions/ipdps25TutorialDescription.md
```diff
@@ -32,7 +32,7 @@ Location: Building 3, Room 3.2.2
 | 14:15 | 08:45 | "Hello World" from Ryzen™ AI | Andra | [AI Engine Basic Building Blocks](../../programming_guide/section-1/) |
 | 14:35 | 09:05 | Exercise 1: Build and run your first program | All | [Passthrough](../../programming_examples/basic/passthrough_kernel/) |
 | 14:50 | 09:20 | Data movement on Ryzen™ AI with objectFIFOs | Andra | [Data Movement](../../programming_guide/section-2/) |
```
hunhoffe, line 34: This file is legacy, let's make no changes here.

docs/conferenceDescriptions/isca25TutorialDescription.md
```diff
@@ -32,7 +32,7 @@ Location: Tokyo, Japan
 | 08:15am | "Hello World" from Ryzen™ AI | Joe | [AI Engine Basic Building Blocks](../../programming_guide/section-1/) |
 | 08:35am | Exercise 1: Build and run your first program | All | [Passthrough](../../programming_examples/basic/passthrough_kernel/) |
 | 08:50am | Data movement on Ryzen™ AI with objectFIFOs | Joe | [Data Movement](../../programming_guide/section-2/) |
```
hunhoffe, line 34: This file is historical, do not modify it.

docs/conferenceDescriptions/micro24TutorialDescription.md
```diff
@@ -25,7 +25,7 @@ Prerequisite: Please bring your laptop so that you can SSH into our Ryzen™ AI-
 | 08:15am | "Hello World" from Ryzen™ AI | Joe | [AI Engine Basic Building Blocks](../../programming_guide/section-1/) |
 | 08:35am | Exercise 1: Build and run your first program | All | [Passthrough](../../programming_examples/basic/passthrough_kernel/) |
 | 08:50am | Data movement on Ryzen™ AI with objectFIFOs | Joe | [Data Movement](../../programming_guide/section-2/) |
```
hunhoffe, line 27: This file is historical, do not modify it.

.github/workflows/buildRyzenWheels.yml
```diff
@@ -221,14 +221,14 @@ jobs:
 
   build-windows:
     name: Build and upload mlir_aie wheels (Windows)
-
```
hunhoffe: If the only changes to this file are whitespace, can we revert those changes?

.github/workflows/mlirDistro.yml
```diff
@@ -157,7 +157,7 @@ jobs:
     - name: set ENV
       shell: bash
       run: |
-
```
hunhoffe: If the only changes to this file are whitespace, can we revert those changes?

docs/Building.md
```diff
@@ -1,6 +1,6 @@
 # Building the MLIR-AIE Codebase on Linux
 
-These instructions will guide you through everything required for building and executing a program on the Ryzen™ AI NPU, starting from a fresh bare-bones **Ubuntu 24.04** or **Ubuntu 24.10** install. It is possible to use **Ubuntu 22.04** however you must follow the documentation on the [xdna-driver](https://github.com/amd/xdna-driver) repository to configure the Linux kernel, driver and runtime for deployment. 
+These instructions will guide you through everything required for building and executing a program on the Ryzen™ AI NPU, starting from a fresh bare-bones **Ubuntu 24.04** or **Ubuntu 24.10** install. It is possible to use **Ubuntu 22.04** however you must follow the documentation on the [xdna-driver](https://github.com/amd/xdna-driver) repository to configure the Linux kernel, driver and runtime for deployment.
```
hunhoffe, line 3: Do not make whitespace only changes to files like this, where the format is immaterial.

programming_examples/algorithms/run_jit.lit
```diff
@@ -1,7 +1,7 @@
 // (c) Copyright 2026 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// REQUIRES: ryzen_ai, peano
```
hunhoffe, line 4: remove whitespace only changes

programming_examples/basic/chaining_channels/Makefile
```diff
@@ -1,56 +1,54 @@
...
+# The script's --xclbin-path / --insts-path mode drives
+# compile_mlir_module directly, so the Makefile shape matches the
+# @iron.jit ports.
 
 srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 include ${srcdir}/../../makefile-common
 
-all: build/final.xclbin
+SHELL := /bin/bash
```
hunhoffe, line 20: Can you verify all makefile changes are platform independent? e.g., that they will run on both windows and linux?

programming_examples/basic/combined_transpose/diagrams/transpose_L1.svg
hunhoffe, line 1: do we need to replace the pictures somehow?

programming_examples/basic/combined_transpose/combined_transpose.py
hunhoffe, line 1: Can you verify we didn't lose test coverage when all these lit files were removed?

programming_examples/basic/dma_compression/dma_compression.py
```diff
@@ -290,31 +302,33 @@ def enable_processor_bus():
...
     if config == "regdump":
-        return _build_regdump(dev)
+        return _build_regdump()
 
-    # Reset the @func cache so a previous build's FuncOp doesn't leak.
+    # @func caches its emitted FuncOp on the decorator instance, but each
+    # @iron.jit invocation builds in its own MLIR context. Reset the cache
+    # so passthrough_line emits a fresh FuncOp into THIS context (otherwise
+    # the second config in a multi-config sweep tries to reference a
+    # FuncOp owned by a now-dead context — KeyError at func call site).
```
hunhoffe: This is pretty ugly. can we find a way around this???

programming_examples/basic/dma_compression/README.md
```diff
@@ -236,13 +236,13 @@ python3 test_dma_compression.py both -v
...
-import numpy as np
 from aie.iron.device import NPU1Col1
-from aie.extras.context import mlir_mod_ctx
+from aie.utils.hostruntime import set_current_device
+from aie.utils.compile.jit.compilabledesign import CompilableDesign
 from dma_compression import dma_compression
-a = np.arange(4096, dtype=np.int32); c = np.zeros(4096, dtype=np.int32)
-with mlir_mod_ctx() as ctx:
-    print(dma_compression(a, c, config='both', dev=NPU1Col1()))
+set_current_device(NPU1Col1())
```
hunhoffe, line 242: Why do we do this? Shouldn't we get is dynamically?

programming_examples/basic/dma_transpose_packet/test.cpp
hunhoffe, line 1: Do we keep at least a few test.cpp files/examples?

programming_examples/basic/event_trace/aie_trace.mlir
hunhoffe, line 1: Why was this file checked in originally?

programming_examples/basic/event_trace/aie_trace.py
```diff
@@ -5,101 +5,96 @@
...
-#
-# Python example using the declarative trace API:
-# - configure_trace() outside runtime_sequence to set up trace ops
-# - start_trace() inside runtime_sequence to activate tracing
-#
-# Usage:
-#   python3 aie_trace.py > aie_trace_from_py.mlir
-#
-# ===-----------------------------------------------------------------------===#
+"""Vector × scalar with custom hardware event tracing — Iron + @iron.jit.
```
hunhoffe: I think it's IRON not Iron, here and everywhere.

programming_examples/basic/event_trace/test.py
```diff
@@ -255,10 +255,14 @@ def main():
...
             continue
 
-        # Read output buffer
-        # pyxrt read() returns numpy array directly
-        bufOut_bytes = bo_out.read(OUT_SIZE, 0)
-        bufOut = np.frombuffer(bufOut_bytes, dtype=DATATYPE_OUT)
+        # Read output buffer.  pyxrt's bo.read() returns a numpy buffer that
+        # may not be C-contiguous on some XRT/pyxrt versions; coerce to bytes
+        # before np.frombuffer to be robust.
+        # pyxrt's bo.read() has produced stale/empty data on some XRT
```
hunhoffe: Uh this is a weird comment.

programming_examples/basic/matrix_multiplication/cascade/cascade.py
```diff
@@ -3,397 +3,458 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024-2025 AMD Inc.
+# (c) Copyright 2024-2026 Advanced Micro Devices, Inc.
+"""Cascade matrix multiply — Iron API design with ``@iron.jit`` compilation.
```
hunhoffe: Can you refrain from labelling all the designs as being iron.jit?? We ported all of them, so it's kind of taken for granted now....

programming_examples/basic/matrix_multiplication/matrix_vector/tests/run_makefile.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test
 // RUN: cd test
 // RUN: make -f %S/../Makefile clean
-// RUN: make -f %S/../Makefile
```
hunhoffe, line 9: whitespace only change.

programming_examples/basic/matrix_multiplication/single_core/README.md
```diff
@@ -1,63 +1,52 @@
-<!---//===- README.md --------------------------*- Markdown -*-===//
-//
-// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
-// See https://llvm.org/LICENSE.txt for license information.
-// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-//
-// Copyright (C) 2022, Advanced Micro Devices, Inc.
-// 
-//===----------------------------------------------------------------------===//-->
-
```
hunhoffe, line 10: Is all the important information from this readme still retained?

programming_examples/basic/matrix_multiplication/whole_array/tests/run_makefile_2_col.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test_2_col
 // RUN: cd test_2_col
 // RUN: make -f %S/../Makefile clean
-// RUN: env n_aie_cols=2 make -f %S/../Makefile
```
hunhoffe, line 9: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_makefile_4_col.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test_4_col
 // RUN: cd test_4_col
 // RUN: make -f %S/../Makefile clean
-// RUN: env n_aie_cols=4 make -f %S/../Makefile
```
hunhoffe, line 9: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_makefile_4_col_i8.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test_4_col_i8
 // RUN: cd test_4_col_i8
 // RUN: make -f %S/../Makefile clean
-// RUN: env n_aie_cols=4 dtype_in=i8 dtype_out=i8 M=512 K=512 N=512 m=64 k=128 n=64 make -f %S/../Makefile 
+// RUN: env n_aie_cols=4 dtype_in=i8 dtype_out=i8 M=512 K=512 N=512 m=64 k=128 n=64 make -f %S/../Makefile
```
hunhoffe: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_makefile_i16_i32_b_c_col_maj.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test_i16_i32_b_c_col_maj
 // RUN: cd test_i16_i32_b_c_col_maj
 // RUN: make -f %S/../Makefile clean
-// RUN: env dtype_in=i16 dtype_out=i32 b_col_maj=1 c_col_maj=1 make -f %S/../Makefile
```
hunhoffe, line 9: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_makefile_i16_i32_c_col_maj.lit
```diff
@@ -6,5 +6,5 @@
 // RUN: mkdir -p test_i16_i32_c_col_maj
 // RUN: cd test_i16_i32_c_col_maj
 // RUN: make -f %S/../Makefile clean
-// RUN: env dtype_in=i16 dtype_out=i32 c_col_maj=1 make -f %S/../Makefile
```
hunhoffe, line 9: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_strix_makefile_bf16_c_col_maj.lit
```diff
@@ -1,10 +1,10 @@
 // (c) Copyright 2025 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// REQUIRES: ryzen_ai_npu1, peano
```
hunhoffe, line 4: Did we keep both npu1 and npu2 tests here?

programming_examples/basic/matrix_multiplication/whole_array/tests/run_strix_makefile_bf16_emulated_chess.lit
```diff
@@ -1,7 +1,7 @@
 // (c) Copyright 2025 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// REQUIRES: ryzen_ai_npu2, chess 
+// REQUIRES: ryzen_ai_npu2, chess
```
hunhoffe: whitespace only

programming_examples/basic/matrix_multiplication/whole_array/tests/run_strix_makefile_i16_i32_c_col_maj.lit
```diff
@@ -1,10 +1,10 @@
 // (c) Copyright 2024 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// REQUIRES: ryzen_ai_npu1, peano
```
hunhoffe, line 4: did we retain npu1 tests too?

programming_examples/basic/matrix_multiplication/whole_array/Makefile
```diff
@@ -1,23 +1,20 @@
...
-# 
+# Copyright (C) 2025-2026, Advanced Micro Devices, Inc.
+#
 ##===----------------------------------------------------------------------===##
 srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 subdir=whole_array
 targetname=whole_array
 
-
-# M, K, N: final GEMM dimensions
```
hunhoffe, line 15: Can we keep these comments?

programming_examples/basic/matrix_multiplication/whole_array/README.md
```diff
@@ -18,53 +18,44 @@ At a high level, the code does the following (in order):
...
 
-1. [**Defining Data Movement Inside the NPU:**](#3-defining-data-movement-inside-the-npu) ObjectFIFOs are a data movement abstraction for buffering data and synchronizing between AIE components. We configure ObjectFIFOs for `A`, `B` and `C` to transfer and buffer data between AIE components in chunks of the previously defined sizes (`m`&times;`k`, `k`&times;`n` and `m`&times;`n`, respecively).
+1. [**Defining Data Movement Inside the NPU:**](#3-defining-data-movement-inside-the-npu) ObjectFifos are a data movement abstraction for buffering data and synchronizing between AIE components. We configure ObjectFifos for `A`, `B` and `C` to transfer and buffer data between AIE components in chunks of the previously defined sizes (`m`&times;`k`, `k`&times;`n` and `m`&times;`n`, respectively).
```
hunhoffe, line 21: Do we use ObjectFIFO vs ObjectFifo consistently throughout the programming examples, programming guide, and Python API?

programming_examples/basic/matrix_multiplication/whole_array/README.md
```diff
@@ -18,53 +18,44 @@ At a high level, the code does the following (in order):
...
 
-1. [**Defining Core Computations:**](#4-defining-core-computations) The `core_body()` function contains the code that will be loaded onto each AIE core. This code describes the matrix multiplication using the input submatrices `a` and `b` acquired through the ObjectFIFOs. The results are accumulated in the output submatrix `c`.
+1. [**Defining Core Computations:**](#4-defining-core-computations) The `core_fn()` function — wrapped in a `Worker` — contains the code that will be loaded onto each AIE core. This code calls the matrix-multiply microkernel from the library (`kernels.mm`) on the input sub-matrix elements acquired through the ObjectFifos, accumulating into the output sub-matrix.
 
-1. [**Defining External Data Transfer Sequences:**](#5-defining-external-data-transfer-sequences) The `aie.runtime_sequence()` op sets up matrix data movement from the host into the AIE compute cores, and back to the host after computation. It initializes Direct Memory Access (DMA) transfers, sets memory access patterns, and performs synchronization.
```
hunhoffe, line 25: I liked the definition of DMA here, including the point about synchronization.

programming_examples/basic/matrix_multiplication/whole_array/README.md
```diff
@@ -96,68 +87,59 @@ We describe it further below, in section *"5. Defining External Data Transfer Se
...
+1. **Compute tiles** (rows 2–5): the AIE cores that run the matmul microkernel.  Across `n_aie_cols` columns × 4 rows we get a 4 × `n_aie_cols` grid of cores (16 by default with `n_aie_cols=4`).
 
-1. **Compute tiles:** In each of the four columns, there are 4 rows of computation tiles above the memory tiles. This makes for a total of 16 computation cores, which in this design are configured to perform the matrix multiplication. In our code, they are represented by a list of lists, `cores`, showing their two-dimensional arrangement.
-
-### 3. Defining Data Movement Inside the NPU: 
+In iron we don't usually enumerate tiles by name.  Instead, the design picks the device family (and column count) via `from_name(opts.dev, n_cols=…)`, builds explicit `Tile(col, row)` handles where needed (for the per-worker placement and for the shim/memtile endpoints of `ObjectFifo.split()` / `forward()` / `join()` calls), and lets the rest of the placement fall out of the FIFO topology:
```
hunhoffe: Can you please elaborate a little bit on how placement is handled by the compiler if left unspecificed; specification only useful when the internal placer makes suboptimal decisions.

programming_examples/basic/matrix_multiplication/whole_array/whole_array.py
```diff
@@ -3,103 +3,60 @@
...
-            args.emulate_bf16_mmul_with_bfp16,
-            args.trace_size,
-            args.generate_taps,
-        )
-        # print(ctx.module.operation.verify())
-        print(ctx.module)
 
-    if args.generate_taps:
-        return maybe_taps
+def _device_for(dev_str, n_aie_cols):
```
hunhoffe, line 48: why do we need this? Can't we do this type of thing in a more object-oriented/library way?

programming_examples/basic/matrix_multiplication/whole_array/whole_array.py
```diff
@@ -3,103 +3,60 @@
...
+    # On NPU1 pick the matching ColN variant (or NPU1 itself when
+    # n_aie_cols == max = 4).  On NPU2 use the unrestricted device
+    # regardless of n_aie_cols so the placer has the full 8-column array.
+    return from_name(dev_str, n_cols=n_aie_cols if dev_str == "npu" else None)
 
 
 def ceildiv(a, b):
```
hunhoffe: Is this replicated anywhere?

programming_examples/basic/matrix_scalar_add/run_makefile.lit
```diff
@@ -4,5 +4,5 @@
 // REQUIRES: ryzen_ai_npu1, peano
 //
 // RUN: make -f %S/Makefile clean
-// RUN: make -f %S/Makefile
```
hunhoffe, line 7: whitespace only.

hunhoffe (commented):
Initial review of first 40 files.

programming_examples/basic/dma_compression/README.md
```diff
@@ -236,13 +236,13 @@ python3 test_dma_compression.py both -v
...
+spec = CompilableDesign(dma_compression, compile_kwargs={'config': 'both'})
+print(spec.as_mlir(spec.generator))
```
hunhoffe: Are you sure there's not a more elegant/canonical way to do this? Seems kinda ugly.

programming_examples/basic/dma_transpose/transpose_data.png
hunhoffe, line 1: Is this something that should have been kept or regenerated?

programming_examples/basic/event_trace/aie_trace.py
```diff
@@ -141,35 +136,56 @@ def core_body():
...
+    )
+
+
+def _run_and_verify(opts):
+    rng = np.random.default_rng(seed=42)
+    a_np = rng.integers(1, 100, size=opts.tensor_size, dtype=np.int32)
+    f_np = np.array([3], dtype=np.int32)
+    c_np = np.zeros(opts.tensor_size, dtype=np.int32)
+
+    a_t = iron.tensor(a_np, dtype=np.int32, device="npu")
```
hunhoffe, line 167: Can we use native tensor.zeros, etc. instead of creating numpy array THEN wrapping in iron.tensor? Here any anywhere else it applies to?

programming_examples/basic/event_trace/Makefile
```diff
@@ -1,118 +1,71 @@
...
+# the kernels.scale library kernel is auto-built into the JIT work dir
+# (no separate Makefile rule needed).  Custom trace event lists live
+# in aie_trace.py (rt.enable_trace kwargs).
 
 srcdir := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
 
 include ${srcdir}/../../makefile-common
 
-devicename ?= npu1
+SHELL := /bin/bash
```
hunhoffe: Why are we adding shell? Is this more appropriate for linux AND windows builds?

programming_examples/basic/event_trace/Makefile
```diff
@@ -1,118 +1,71 @@
...
+
 targetname = mlir_trace_example
 
-in1_size = 16384 # in bytes
-in2_size = 4 # in bytes, should always be 4 (1x int32)
-out_size = 16384 # in bytes, should always be equal to in1_size
+in1_size   = 16384
+in2_size   = 4
+out_size   = 16384
```
hunhoffe: Should we keep comments?

programming_examples/basic/event_trace/test.py
```diff
@@ -255,10 +255,11 @@ def main():
...
             continue
 
-        # Read output buffer
-        # pyxrt read() returns numpy array directly
-        bufOut_bytes = bo_out.read(OUT_SIZE, 0)
-        bufOut = np.frombuffer(bufOut_bytes, dtype=DATATYPE_OUT)
+        # bo.read() has been unreliable across XRT versions (non-contiguous
+        # buffers, stale/empty data); bo.map() returns the mapped host
+        # memoryview directly, matching the C++ bo.map<T*>() path.
+        bufOut_view = bo_out.map()
```
hunhoffe: Do we REALLY have to keep using bo's direction here? Can't we use the python helper runtime objects????

programming_examples/basic/matrix_multiplication/matrix_vector/tests/run_makefile_scalar.lit
```diff
@@ -0,0 +1,10 @@
+// (c) Copyright 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// REQUIRES: ryzen_ai_npu1, peano
```
hunhoffe, line 4: I noticed both exisitng lit tests here appear to be for scalar. Do we need to test vectorized too?

programming_examples/basic/matrix_multiplication/single_core/single_core.py
```diff
@@ -3,373 +3,320 @@
...
+        output_dtype=dtype_out,
+        b_col_maj=bool(b_col_maj),
+        use_chess=use_chess,
+        emulate_bf16_mmul_with_bfp16=emulate_bf16_mmul_with_bfp16,
+    )
+    zero_kernel = matmul_kernel.zero
+    r, s, t = matmul_kernel.mac_dims
+    assert m % r == 0
+    assert k % s == 0
+    assert n % t == 0
```
hunhoffe, line 95: Since we fetch them form the kernel, do we actually need these asserts here?

programming_examples/basic/matrix_scalar_add/matrix_scalar_add.py
```diff
@@ -4,72 +4,138 @@
...
+"""Matrix scalar add — IRON API design with ``@iron.jit`` compilation.
 
-from aie.iron import ObjectFifo, Program, Runtime, Worker
-from aie.iron.device import NPU1Col1, NPU2, XCVC1902
-from aie.iron.controlflow import range_
-from aie.helpers.taplib import TensorTiler2D
+A single AIE compute core reads one ``TILE_HEIGHT x TILE_WIDTH`` tile from
+the top-left corner of a ``MATRIX_HEIGHT x MATRIX_WIDTH`` matrix (via
+``TensorTiler2D.simple_tiler``), adds 1 to each element of that tile, and
+writes it back.  The rest of the matrix is untouched — this demonstrates
```
hunhoffe: untouched compared to what? This comment seems too historical rather than discussing what it.

programming_examples/basic/matrix_scalar_add/matrix_scalar_add.py
```diff
@@ -4,72 +4,138 @@
...
+input matrix); the algorithms library's ``transform_typed`` would flatten
+to 1D and produce a different DMA stride pattern, which is why this
+primitive is unified in the matvec / vector_scalar_mul way rather than the
+vector_vector_add way.
 
-# Size of the entire matrix
-MATRIX_HEIGHT = 16
-MATRIX_WIDTH = 128
-MATRIX_SHAPE = (MATRIX_HEIGHT, MATRIX_WIDTH)
+Three invocation modes (mirrors vector_vector_modulo):
```
hunhoffe: We don't need to compare against other designs in comments, just describe this one.

programming_examples/basic/memcpy/memcpy.py
```diff
@@ -94,111 +90,100 @@ def core_fn(of_in, of_out, passThroughLine):
...
+        bypass=_bypass_bool(opts.bypass),
+    )
+
+
+def _run_and_verify(opts):
+    in_np = np.arange(opts.length, dtype=np.int32)
+    out_np = np.zeros_like(in_np)
+
+    a_t = iron.tensor(in_np, dtype=np.int32, device="npu")
+    b_t = iron.tensor(out_np, dtype=np.int32, device="npu")
```
hunhoffe: Let's use iron.arange and iron.zeros_like instead of doing numpy then iron.tensor....

programming_examples/basic/packet_switch/packet_switch.py
```diff
@@ -0,0 +1,484 @@
...
+    )
+    p.add_argument("-n", "--length", type=int, default=256, help="vector length")
+    return p
+
+
+def _compile_kwargs(opts):
+    return dict(in_out_size=opts.length, input_packet_id=_OP_PACKET_ID[opts.op])
+
+
+def _emit_mlir(opts):
```
hunhoffe: I feel like this is a common pattern. We gotta find a better way to do this rather than replicating _emit_mlir() functions everywhere...

programming_examples/basic/passthrough_dmas/passthrough_dmas.py
```diff
@@ -4,49 +4,160 @@
...
+Three target/topology modes:
+
+  * ``--plio none`` (default):  shim → memtile → shim, NPU or VCK5000.
+  * ``--plio input``:           PLIO ObjectFifo on the input side, forwarded
+                                through a compute tile to a non-PLIO shim.
+                                VCK5000 only.
+  * ``--plio output``:          mirror of above — non-PLIO input, PLIO output.
+                                VCK5000 only.
+
+Three invocation modes (mirrors matrix_scalar_add):
```
hunhoffe: Let's not compare designs against each other in comments; just describe what is.

programming_examples/basic/transposes/README.md
```diff
@@ -0,0 +1,154 @@
...
+Switch strategy / sizes:
+
+```shell
+make STRATEGY=shuffle run                       # uint8 16x16
+make STRATEGY=combined M=128 K=128 run          # int32 128x128 with built-in m=n=32, s=8
+```
+
+## What replaces this directory
+
+This directory consolidates four earlier examples:
```
hunhoffe: Do not include hsitorical information in comments; describe what currently IS

programming_examples/basic/vector_reduce_min/README.md
```diff
@@ -5,97 +5,41 @@
...
 
-1. **`reduce_min.cc`**: A C++ implementation of a vectorized `min` reduction operation for AIE cores. The code uses the AIE API, which is a C++ header-only library providing types and operations that get translated into efficient low-level intrinsics, and whose documentation can be found [here](https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_api/aie_api/doc/index.html).  The source can be found [here](../../../aie_kernels/aie2/reduce_min.cc).
+   * standalone — `python3 vector_reduce_min.py`
+   * compile-only — `... --xclbin-path=PATH --insts-path=PATH` (used by the `Makefile`)
 
-## Usage
+1. `reduce_min.cc`: A C++ implementation of a vectorized `min` reduction for AIE cores. The kernel uses the AIE API ([docs](https://www.xilinx.com/htmldocs/xilinx2023_2/aiengine_api/aie_api/doc/index.html)). Source: [here](../../../aie_kernels/aie2/reduce_min.cc).
 
-### JIT Approach (Just-In-Time Compilation)
+1. `test.cpp`: C++ testbench. Loads the compiled XCLBIN, supplies input, runs on the NPU, and verifies the result.
```
hunhoffe: Eventually we might move to full elf workflow. In comments can we make it clear it's xclbin+insts OR elf?

For variable names, unless it's literally for the xclbin or insts file, let's keep such things generic.

programming_examples/basic/vector_scalar_mul/run_strix_makefile.lit
```diff
@@ -6,12 +6,12 @@
 // RUN: mkdir -p test_stx
 // RUN: cd test_stx
 // RUN: make -f %S/Makefile clean
-// RUN: make -f %S/Makefile devicename=npu2 
+// RUN: make -f %S/Makefile devicename=npu2
 // RUN: %run_on_npu2% make -f %S/Makefile run devicename=npu2
 // RUN: make -f %S/Makefile clean
 // RUN: %run_on_npu2% make -f %S/Makefile trace devicename=npu2
-// CHECK:  cycles is 46
```
hunhoffe, line 13: It seems odd that this would have changed....

programming_examples/basic/vector_scalar_mul/test.py
```diff
@@ -5,11 +5,15 @@
...
+
+import numpy as np
+
 import aie.iron as iron
 from aie.utils import DefaultNPURuntime
+from aie.utils.hostruntime.argparse import add_runtime_args
+from aie.utils.test import create_npu_kernel
 
 
 def main(opts):
```
hunhoffe, line 19: Why do we have a separate python test file here instead of just using JIT?

programming_examples/basic/matrix_multiplication/cascade/cascade.py
```diff
@@ -3,397 +3,458 @@
...
+    matmul_get_only = cascade_kernel.get_only
+    matmul_put_only = cascade_kernel.put_only
+    matmul_put_get = cascade_kernel.put_get
+    zero_kernel = cascade_kernel.zero
 
     if dtype_in_str == "bf16":
-        r = 4
-        s = 8
-        t = 4
+        r, s, t = 4, 8, 4
```
hunhoffe: Tehre's a pattern in matmul to include this information with the kernel. Can we use that same technique here?

programming_examples/getting_started/03_matrix_multiplication_single_core/diagrams/matmul_l2l1.svg
```diff
@@ -348,7 +348,7 @@
          id="tspan3-1-1-9-3"
          style="font-size:3.88056px;fill:#666666;stroke-width:0.264583"
          x="7.6688719"
-         y="58.162884">s = 2</tspan></text>
```
hunhoffe, line 351: why is this file edited????

programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config2/n32_core.py
```diff
@@ -0,0 +1,259 @@
...
+from aie.utils.hostruntime.argparse import add_compile_args
+from aie.utils.hostruntime.cli import run_design_cli
+
+_KERNEL_SRC = Path(__file__).resolve().parent / "mm_bfp_mixed.cc"
+_AIE_KERNELS_INC = Path(__file__).resolve().parents[5] / "aie_kernels"
+
+DIV = 6
+
+
+def ceildiv(a, b):
```
hunhoffe: I thought we caught all instances of this and used the helper function?

programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/README.md
```diff
@@ -58,12 +58,14 @@ combined tradeoff) see Sections 3 – 5 of the paper.
 
 ## Examples
 
-These designs use the placed (`*_placed.py`) IRON flow and are currently only
-supported with the **chess** compiler. Build any of them with:
+These designs use IRON and currently require the **chess** compiler (each
+`config*/n32_core.py` declares its bfp16 mac kernel with `use_chess=True`
+on the `ExternalFunction`, since Peano's AIE2P backend can't yet legalize
+those kernels). Build any of them with:
```
hunhoffe: Is this actually accurate? Can we verify with a newer version of llvm-aie/peano?

programming_examples/ml/block_datatypes/matrix_multiplication/single_core/single_core.py
```diff
@@ -1,67 +1,101 @@
...
+    *,
+    M: Compile[int] = 128,
+    K: Compile[int] = 128,
+    N: Compile[int] = 128,
+    m: Compile[int] = 64,
+    k: Compile[int] = 64,
+    n: Compile[int] = 64,
+):
+    # bfp16ebs8 matmul mac unit is 8x8x8; m/k/n must be multiples of these.
+    r = s = t = 8
```
hunhoffe, line 61: Can r/s/t be inferred from the kernel in the kernel library?

programming_examples/ml/conv2d/README.md
```diff
@@ -1,95 +1,33 @@
...
-```
-
-To run the design:
-```shell
-make run_py
-```
-
-## Configure design
-To configure the parameters of the convolution such as data width, height and the number of input and output channels, you can edit the top of the `Makefile`. Choosing the scalar or vectorized version of the kernel can likewise be selected in the `Makefile` by modifying the `vectorized` variable. 
+<!---//===- README.md --------------------------*- Markdown -*-===//
```
hunhoffe, line 1: It seems this README lost a LOT of informaiton. Where did it go? Why is it gone?

programming_examples/ml/conv2d_14x14/conv2dk14.py
```diff
@@ -1,228 +1,467 @@
...
 # This file is licensed under the Apache License v2.0 with LLVM Exceptions.
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
-# (c) Copyright 2024 AMD Inc.
-import numpy as np
+# (c) Copyright 2024-2026 Advanced Micro Devices, Inc. or its affiliates
+"""14x14 Conv2D — IRON API designs (single-core + 32-core) with @iron.jit.
+
+KNOWN: ``make run_py`` fails on Strix Halo (npu2). The aie2p variant of
```
hunhoffe: This sounds like a problem...

programming_examples/ml/eltwise_unary/eltwise_unary.py
```diff
@@ -0,0 +1,127 @@
...
+from aie.iron.algorithms import transform_parallel_typed
+from aie.iron.device import device_from_args
+from aie.utils.hostruntime.argparse import add_compile_args
+from aie.utils.hostruntime.cli import run_design_cli
+from aie.utils.verify import assert_pass
+
+_KERNEL_FACTORIES = {"relu": kernels.relu, "silu": kernels.silu, "gelu": kernels.gelu}
+
+
+def _relu_ref_f32(x):
```
hunhoffe: shouldn't ref's live with the kernels themselves?

programming_examples/ml/mobilenet/bottleneck_A/gen_golden_bn7.py
```diff
@@ -12,6 +12,7 @@
 from aie.utils.ml import DataShaper
 import time
 import os
+from pathlib import Path
 import numpy as np
 import aie.iron as iron
```
hunhoffe: I see a LOT of duplication in these "gen_golden" -- is there a way to simplify them? Or use a single switched file?

programming_guide/mini_tutorial/exercise_5/exercise_5a/answer_2.py
```diff
@@ -9,23 +9,26 @@
...
+from aie.iron import Out, In, Compile, Program, Runtime, Worker, ObjectFifo
 from aie.iron.controlflow import range_
 from aie.helpers.taplib import TensorTiler2D
 
 import aie.iron as iron
 
 
 @iron.jit
-def exercise_5a(input0, output):
+def exercise_5a(
```
hunhoffe, line 20: I see a lot of repetition in question/answer files. Can we make it a bit cleaner?

programming_guide/mini_tutorial/README.md
```diff
@@ -4,7 +4,7 @@
 
 ## <ins>Key Components: Workers, ObjectFifos, Runtime</ins>
 
-IRON provides an unplaced (deferred placement) [API](../../python/iron/) for NPU programming. Below are examples describing AIE compute code and the Object FIFO data movement primitive:
+IRON provides a higher-level (deferred-placement) [API](../../python/iron/) for NPU programming. Below are examples describing AIE compute code and the Object FIFO data movement primitive:
```
hunhoffe: Make sure to remove ALL references to the idea that higher-levle and lower-level IRON have different placement methods.

programming_guide/section-3/README.md
```diff
@@ -24,30 +24,99 @@ The host code can be written in either C++ (as shown in the figure) or in Python
...
+        rt.fill(of_factor.prod(), f)
+        rt.drain(of_out.cons(), c, wait=True)
+
+    return Program(iron.get_current_device(), rt).resolve_program()
+```
+
+The host side is just three tensors and one call:
+
+```python
+a_in = iron.tensor(np.arange(1, tensor_size + 1, dtype=np.int32), device="npu")
```
hunhoffe: we don't need to wrap tensor around numpy creations....

programming_guide/section-3/vector_scalar_mul.py
```diff
@@ -0,0 +1,133 @@
+# section-3/vector_scalar_mul.py -*- Python -*-
```
hunhoffe, line 1: Do we need a seaparte copy of this? Or should we point to the programming example?

programming_guide/section-4/section-4b/vector_scalar_mul.py
```diff
@@ -0,0 +1,144 @@
+# section-4/section-4b/vector_scalar_mul.py -*- Python -*-
```
hunhoffe, line 1: Again, do we need a whole seaparte copy here or should we lean on the existing programming example?

programming_guide/whats_new_in_unify_compilation_workflow.ipynb
```diff
@@ -0,0 +1,1474 @@
+{
```
hunhoffe: Can we make sure programming guide + examples cover all of this, and then delete this notebook in favor of the official documentation?

python/iron/algorithms/for_each.py
```diff
@@ -9,12 +9,96 @@
...
+        data: In-place tensor placeholder (``InOut``).
+        func: Function or :class:`~aie.iron.kernel.ExternalFunction` to apply.
+        N: Number of elements in ``data``.
+        dtype: Element dtype.
+        tile_size: Number of elements per tile. Defaults to 16.
+
+    Returns:
+        mlir.ir.Module: The compiled MLIR module.
+    """
+    tensor_ty = np.ndarray[(N,), np.dtype[dtype]]
```
hunhoffe: Do you think we could use the compile config parameter context thing instead of having typed/untyped?

python/iron/algorithms/transform.py
```diff
@@ -9,13 +9,21 @@
...
 
-from aie.iron import ObjectFifo, Program, Runtime, Worker
+from aie.iron import Compile, In, Out, ObjectFifo, Program, Runtime, Worker
 from aie.helpers.taplib.tap import TensorAccessPattern
 from aie.iron.controlflow import range_
 import aie.iron as iron
 
 
-def _transform_gen(func, inputs: list, output, *params, tile_size=16):
+def _check_num_channels(num_channels: int) -> None:
```
hunhoffe, line 18: Can we fetch this kind of thing from the target model/device?

python/iron/algorithms/transform.py
```diff
@@ -360,89 +459,435 @@ def core_body(*of_args):
...
+        func,
+        tensor_ty,
+        tile_size=tile_size,
+        trace_size=trace_size,
+        num_channels=num_channels,
+        pass_size_to_kernel=pass_size_to_kernel,
+    )
+
+
+def transform_parallel_binary(
```
hunhoffe: Are there any other examples we can port to use the algorithms library?

python/iron/dataflow/flow.py
```diff
@@ -0,0 +1,252 @@
...
+    when there are multiple sinks (uncommon, but the underlying op supports
+    it).
+    """
+
+    tile: object  # Tile (avoid forward-ref import)
+    port: WireBundle = WireBundle.DMA
+    channel: int = 0
+
+
+class PacketFlow(Resolvable):
```
hunhoffe, line 159: should flow and packetflow be in two different files?

python/iron/dataflow/objectfifo.py
```diff
@@ -6,8 +6,17 @@
...
-from typing import Sequence
+from typing import Sequence, TypeAlias
+
+# Named aliases for the (size, stride) pair-lists used by DMA stream
+# layout transforms and pad-then-stream descriptors. Both are list[(size,
+# stride)] from highest to lowest dimension; the names exist purely to
+# make signatures and docs self-documenting — they are not validated
+# types at runtime.
+StreamDims: TypeAlias = list[Sequence[int]]
+PadDims: TypeAlias = list[Sequence[int]]
```
hunhoffe, line 19: Are these used consistently across the code base?

python/iron/dataflow/tile_dma.py
```diff
@@ -0,0 +1,273 @@
...
+        channel: hardware channel index.
+        bds: ordered list of :class:`Bd` entries that form the chain.
+    """
+
+    direction: DMAChannelDir
+    channel: int
+    bds: list[Bd]
+
+
+class TileDma(Resolvable):
```
hunhoffe, line 118: Is this utilized when needed in the programming examples?

python/iron/device/__init__.py
```diff
@@ -24,3 +25,112 @@
     XCVC1902,
 )
 from .tile import AnyShimTile, AnyMemTile, AnyComputeTile, Tile
+
+# Map from CLI device name → base class name (max-column variant).
+_NAME_TO_BASE: dict[str, str] = {
```
hunhoffe, line 30: Is this used exhaustively across the code base?

python/iron/device/__init__.py
```diff
@@ -24,3 +25,112 @@
...
+        cls_name = f"{base}Col{n_cols}"
+    else:
+        raise ValueError(
+            f"n_cols must be in 1..{max_cols} (or None for max) for "
+            f"{base}, got {n_cols}"
+        )
+    return getattr(_device_module, cls_name)()
+
+
+def device_from_args(
```
hunhoffe: I think args lives in utils --- so maybe this should too?

python/iron/device/device.pyi
```diff
@@ -0,0 +1,138 @@
...
+
+The stubs let IDEs, mypy, and ``from aie.iron.device.device import NPU2``-style
+imports resolve cleanly.  They do NOT affect runtime behaviour; the real
+classes are still synthesised at import time.
+"""
+
+from .. import ir
+from .device import Device
+
+class XCVC1902(Device):
```
hunhoffe: What!?! I thought we already had device types parsed magiclaly vai on-the-fly class creation. Do we REALLY need these? I really think not.

python/iron/kernels/__init__.py
```diff
@@ -0,0 +1,98 @@
+# kernels/__init__.py -*- Python -*-
```
hunhoffe, line 1: keeping full functionality, are we able to consolidate any of the kernels?

python/iron/kernel.py
```diff
@@ -14,13 +14,71 @@
...
 from .. import ir  # type: ignore
+from ..dialects import memref  # type: ignore
 from ..extras.dialects.func import FuncOp  # type: ignore
 from ..helpers.dialects.func import call
 from ..dialects.aie import external_func
 from .resolvable import Resolvable
 from .buffer import Buffer
 
 
+def _is_contiguous_row_major(mr):
```
hunhoffe, line 25: is this helper useful anywhere else in the code base or examples?

python/utils/compile/jit/_dma_size_parser.py
```diff
@@ -0,0 +1,165 @@
...
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def parse_dma_sizes(kernel_dir: Path) -> list[int] | None:
```
hunhoffe, line 47: does this logic have comprehensive tests?

python/utils/compile/jit/compilabledesign.py
```diff
@@ -0,0 +1,1015 @@
+# compilabledesign.py -*- Python -*-
```
hunhoffe, line 1: This file is huge. Can we decompose? Use common helpers? Simplify without losing functionality?

python/utils/callabledesign.py
```diff
@@ -0,0 +1,432 @@
...
+        ``elf_path`` is optional: when set, aiecc also wraps the NPU
+        instructions into an ELF (via ``aiebu-asm``) at that path.  Needed by
+        C++ testbenches that load instructions through ``xrt::elf`` +
+        ``xrt::module``; requires explicit ``xclbin_path`` + ``inst_path``.
+        """
+        return self.compilable.compile(
+            xclbin_path=xclbin_path, inst_path=inst_path, elf_path=elf_path
+        )
+
+    def as_mlir(self, *runtime_args, **runtime_kwargs) -> str:
```
hunhoffe, line 413: Is this used comprehensively throughout the programming_guide and programming_examples where appropriate?

.gitignore
```diff
@@ -29,3 +29,10 @@ utils/vitisVariables.config
 runtime_lib/xaiengine/src
 /ironenv
 /.cache
+
+# Local run/test artifacts under programming_examples/ and programming_guide/
+**/log/
+**/test_stx/
+**/trace.txt
+**/trace_*.json
+**/_whats_new_executed.*
```
hunhoffe: we're going to remove the notebook, so please remove this line too!

python/requirements_dev.txt
```diff
@@ -5,6 +5,10 @@ wheel
 ninja!=1.13.0
 cibuildwheel
 pre-commit>=4.6.0,<5.0
+# black[jupyter] so the format check covers .ipynb code cells too
```
hunhoffe: I don't think the comments here are necessary.

python/utils/verify.py
```diff
@@ -0,0 +1,201 @@
+# verify.py -*- Python -*-
```
hunhoffe, line 1: Do the programming_examples and programming_guide use this consistently?

hunhoffe (commented):
Yet another review for Claude

aie_kernels/aie2p/mm_bfp.cc
```diff
@@ -185,13 +199,18 @@ void matmul_vectorized_bfp16(bfp16ebs8 *__restrict pA, bfp16ebs8 *__restrict pB,
 
   matmul_vectorized_2x2_bfp16<m / r, k / s, n / t, r, s, t>(pA, pB, pC);
 }
+#endif
 
+#ifdef ZERO_ONLY
 void zero_kernel(bfp16ebs8 *__restrict cOut) {
```
ypapadop-amd, line 205: I thought that we can now have multi file linking. Maybe a follow-up PR?

programming_examples/algorithms/transform.py
```diff
@@ -5,63 +5,66 @@
...
+from aie.utils.verify import assert_pass
 
-from aie.iron.algorithms import transform
 
+@iron.jit
+def transform(
+    input: In,
+    output: Out,
+    *,
+    num_elements: Compile[int],
```
ypapadop-amd: Can we also use `Compile` or does it always need to be typed?

How do I read `Compile`? Compile the value or is it a compile-time constant?

programming_examples/basic/chaining_channels/Makefile
```diff
@@ -1,56 +1,54 @@
...
 
-devicename ?= npu2
 targetname = chaining_channels
+devicename ?= npu2
```
ypapadop-amd, line 23: I know it's not from this PR, but is this code only for aie2p?

programming_examples/basic/matrix_multiplication/matrix_vector/matrix_vector.py
```diff
@@ -3,181 +3,201 @@
...
-    C_sz_div_n_cores = C_sz // n_cores
-
-    M_div_m = M // m
+    M_div_n_cores = M // n_cores
     M_div_m_div_n_cores = M // (m * n_cores)
     K_div_k = K // k
 
-    m_x_k = m * k
-    m_x_K = m * K
+    matvec_kernel = kernels.mv(
```
ypapadop-amd, line 54: I'm a little confused how do we end up with the kernels here.

ypapadop-amd (commented):
I think I got the gist of it. Overall it looks great.

The only red flag I see is with the external functions that somehow appear in a dict. It abstracts things a bit too much for a lower-level framework and it's not immediately obvious how it works.

aie_kernels/aie2p/mm_bfp.cc
hunhoffe: Multi-file linking does work now — but we don't have a full solution for symbol renaming/collisions yet. There are multiple definitions of zero, and so that's why the `#ifdef MATMUL_ONLY` / `ZERO_ONLY` / `SHUFFLE_ONLY` split exists (mm_bfp.cc:173-210). One .cc is compiled three times with different -D flags into three .os, each emitting only one of the symbols, then linked together. The guards are there precisely so the separate builds of the same .cc file don't both emit a symbol and clash at link (see the static comment at line 30).

We could abstract each function into their own file, but I'd rather just use the guards until the time comes if/when we address symbol collisions more fully.

programming_examples/algorithms/transform.py
hunhoffe: Yes, bare Compile works. `_is_compile_param` (python/utils/compile/jit/_introspect.py:35) checks annotation is `Compile` first, then falls through to the subscripted `Compile[T]` case. So `num_elements: Compile` is accepted exactly like `num_elements: Compile[int]`.

The `[T]` is purely for static type-checkers (mypy/pyright)  -- none of which are, at present, integrated into this repo. But that's an addition I would really love to make.

For the second question: Compile = compile-time constant analogous to Triton's `tl.constexpr` — the value is baked into the MLIR and changing it triggers a recompile. I'm open to renaming is `CompileConst`or similar if there's something you think is more descriptive?

programming_examples/basic/chaining_channels/Makefile
hunhoffe: Let me look into this -- if I can make it more generic, I'll submit a separate PR to loosen the constraint

programming_examples/basic/matrix_multiplication/matrix_vector/matrix_vector.py
hunhoffe: `kernels.mv(...)` is a factory (python/iron/kernels/linalg.py:202) that returns a fully-configured `ExternalFunction`. You can see exactly what comes back by introspecting it live — no compile or device needed beyond pinning an arch so the .cc path / mac_dims tables resolve:

```python
import aie.iron as iron
from aie.iron import kernels
from aie.utils import set_current_device

set_current_device(iron.device.NPU2Col1())   # arch only; no HW touched
mv = kernels.mv(dim_m=32, dim_k=32, vectorized=True)

mv._name            # '7f47d893_matvec_vectorized_i16_i32'  (digest-prefixed symbol)
mv.object_file_name # 'matvec_vectorized_i16_i32_7f47d893.o'
mv.arg_types()      # [ndarray[(1024,), int16], ndarray[(32,), int16], ndarray[(32,), int32]]
mv._source_file     # .../install/include/aie_kernels/aie2/mv.cc
mv._compile_flags   # ['-DDIM_M=32', '-DDIM_K=32']
mv.zero._name       # 'zero_vectorized_i32'   (companion symbol, same .o)
```

The kernel wrappers just provide an easy way to introspect on `ExternalKernel` metadata for "library" kernels. A user can still define an `ExternalKernel` directly -- they are NOT constrained to only using "library" kernels.

To illustrate/maintain this flexibility, I purposely did not port all kernels to the "library". As an example:
```python
# programming_examples/getting_started/01_SAXPY/saxpy.py:46 — the canonical minimal form, pointing at a local  saxpy.cc:
saxpy_kernel = ExternalFunction(
     "saxpy",
    source_file=str(Path(__file__).parent / "saxpy.cc"),
     arg_types=[in_ty, in_ty, out_ty],
     include_dirs=[cxx_header_path()],
)
```

Both "library kernels" and traditional `ExternalKernels` end up in the exact same registry → discovery → compile flow as the factory output (it's the same `ExternalFunction` class either way); they just skip the library-lookup convenience. 

I think it's an open question whether the "kernel library" lives here long-term or in `amd/IRON` but we need to do a great deduplication before we address that question.

hunhoffe:
@ypapadop-amd If I am understanding your dict/function comment correctly, you are refering to the `ExternalFunction._instances` — a class-level registry (actually a set), not a hidden global lookup. 

I think this is the only clean/reliable way to provide a solutions for https://github.com/Xilinx/mlir-aie/pull/2509. The mechanism is deliberately thin: every `ExternalFunction(...)` constructor call adds self to                  `ExternalFunction._instances` (python/iron/kernel.py:361). At `@iron.jit` time the design body runs once to build MLIR;
`compile()` then snapshots whatever got registered (compilabledesign.py:681, _instances.clear() before/after so a   failed run can't leak stale entries) and compiles exactly those source files.                                      
                                                                                                                
So there's no magic resolution by name — the registry is just "which C++ sources did this design body instantiate." he user never touches it; they call `kernels.mm(...)` (or construct an `ExternalFunction` directly) and the object   self-registers. Happy to rename `_instances` or document it more visibly if the set-as-registry pattern is the part that reads as over-abstracted.

Also -- thank you for the review!! Please do let me know if I've addressed your comments to your satisfaction.

aie_kernels/aie2p/mm_bfp.cc
ypapadop-amd: OK, that makes sense. You can also template it and avoid the ifdefs, but it's not changing the fundamental issue.

If we have two compilation units that use the same symbol internally, can we run into issues? I don't think we restrict visibility.

programming_examples/algorithms/transform.py
ypapadop-amd: I was only wondering if it actually compiles something at the argument definition which would be confusing. `CompileConst` or `ConstExpr` may indeed be more descriptive.

programming_examples/basic/matrix_multiplication/matrix_vector/matrix_vector.py
ypapadop-amd: Got it. I think that moving the registry and the `aie_kernels` closer would benefit this.

ypapadop-amd:
> @ypapadop-amd If I am understanding your dict/function comment correctly, you are refering to the `ExternalFunction._instances` — a class-level registry (actually a set), not a hidden global lookup.
> 
> I think this is the only clean/reliable way to provide a solutions for #2509. The mechanism is deliberately thin: every `ExternalFunction(...)` constructor call adds self to `ExternalFunction._instances` (python/iron/kernel.py:361). At `@iron.jit` time the design body runs once to build MLIR; `compile()` then snapshots whatever got registered (compilabledesign.py:681, _instances.clear() before/after so a failed run can't leak stale entries) and compiles exactly those source files.
> 
> So there's no magic resolution by name — the registry is just "which C++ sources did this design body instantiate." he user never touches it; they call `kernels.mm(...)` (or construct an `ExternalFunction` directly) and the object self-registers. Happy to rename `_instances` or document it more visibly if the set-as-registry pattern is the part that reads as over-abstracted.
> 
> Also -- thank you for the review!! Please do let me know if I've addressed your comments to your satisfaction.

It's kind of sore subject because I had to roll out my own registry due to various technical reasons (kernels needed a specific name, .o in specific directories, etc.) I think that the `kernels` factory has potential to replace my solution but only if the artifact naming, storage and kernel names can be configured.

hunhoffe:
@ypapadop-amd I'm sorry that you had to roll your own artifact registry. I do believe this is flexible enough to fit your use case -- now that we have a basic infra, I think we can also adjust it if there are bugs. The only note I have is that storage of source files is currently flexible/override-able, but storage of the objects themselves is somewhat tied to the cache keys used here. Is that a problem?

Edit: You can prefix/suffix .o files even in the library I think, to support multiple versions being cached at once.

aie_kernels/aie2p/mm_bfp.cc
hunhoffe: I think that's still a problem but it's not a new problem -- I'd like to leave that for a separate PR.

As a side note, I have looked into handling symbol collisions better. I was able to find what I think is a somewhat robust solution for peano but for some reason I haven't been able to debug it doesn't work with chess. If this continues to be an issue, let's revisit the issue of symbols.

programming_examples/algorithms/transform.py
hunhoffe: I did `CompileTime` because I thought being explicit for WHEN it's evaluated was good.... maybe let's try this for a while, and if there is confusion, we can try `ConstExpr`?

fifield (approved):

ypapadop-amd:
> @ypapadop-amd I'm sorry that you had to roll your own artifact registry. I do believe this is flexible enough to fit your use case -- now that we have a basic infra, I think we can also adjust it if there are bugs. The only note I have is that storage of source files is currently flexible/override-able, but storage of the objects themselves is somewhat tied to the cache keys used here. Is that a problem?
> 
> Edit: You can prefix/suffix .o files even in the library I think, to support multiple versions being cached at once.

With objects you mean compilation artifacts? Yes, I manage where they live and their name.

PR #3036, opened 2026-04-22
Fix LinearizeContiguousBDTransfer for 4D BDs with iteration dimension 

hunhoffe:
Edit: description rewritten after feedback, thanks @andrej !

## Bug

`LinearizeContiguousBDTransfer` collapsed any contiguous shim BD into a linear transfer. For a 4D shim BD this is wrong: the outermost dim is the hardware **iteration** dim (programmed into `iteration_size` / `iteration_stride` by
`AIEDMATasksToNPU`), and `len` covers only the inner three dims. Linearizing dropped the iteration stride, so the NPU re-read the same `len` bytes from offset 0 every repeat.

Surfaced by IRON `transpose[M_2048-N_64-m_64-n_64-s_8]`: BD with `dims = [<32, 4096>, <1, 64>, <64, 64>, <64, 1>]`, `len = 4096`, canonicalized into a 4096-byte transfer repeated 32 times → 89.5% wrong.

## Fix

Bail when `dims.size() == 4 && dims.front().size() > 1`. Outer-size 1 is a trivial iteration and remains foldable. Structural, not length-based: a 4D BD with `len == product` still bails because outer-size > 1 implies iteration semantics.

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -2424,14 +2424,12 @@ struct LinearizeContiguousBDTransfer : public mlir::OpRewritePattern<DMABDOp> {
     if (dims->size() == 1 && dims->front().getStride() == 1)
       return mlir::failure();
 
-    // Compute total element count from all dimension sizes.
     int64_t product = 1;
     for (BDDimLayoutAttr dim : *dims)
       product *= dim.getSize();
 
-    // If len < product, the outermost dim is a BD iteration dim (per-
-    // iteration transfer = len, repeat count/stride from the outer dim).
-    // Linearizing would drop the iteration stride.
+    // len < product means the outermost dim is a BD iteration dim;
+    // linearizing would drop the iteration stride.
     if (auto lenVal = op.getLen())
       if (static_cast<int64_t>(*lenVal) != product)
         return mlir::failure();
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -2429,18 +2429,12 @@ struct LinearizeContiguousBDTransfer : public mlir::OpRewritePattern<DMABDOp> {
     for (BDDimLayoutAttr dim : *dims)
       product *= dim.getSize();
 
-    // If the op has an explicit len that disagrees with the total dims product,
-    // the outermost dimension is a BD iteration (d3) dimension — len covers
-    // only the inner dims (per-iteration transfer), while the outermost dim
-    // controls how many times the BD is re-issued with address advancement.
-    // Linearizing would collapse the iteration into len, but the iteration
-    // stride (address advancement) would be lost, causing the hardware to
-    // repeatedly transfer only the first len elements instead of sweeping
-    // across the full buffer.  Bail out in this case.
-    if (auto lenVal = op.getLen()) {
+    // If len < product, the outermost dim is a BD iteration dim (per-
+    // iteration transfer = len, repeat count/stride from the outer dim).
+    // Linearizing would drop the iteration stride.
+    if (auto lenVal = op.getLen())
       if (static_cast<int64_t>(*lenVal) != product)
         return mlir::failure();
-    }
     int32_t len = static_cast<int32_t>(product);
 
     // Drop the dimensions attribute in-place; all other attributes (offset,
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -2428,8 +2428,8 @@ struct LinearizeContiguousBDTransfer : public mlir::OpRewritePattern<DMABDOp> {
     for (BDDimLayoutAttr dim : *dims)
       product *= dim.getSize();
 
-    // len < product means the outermost dim is a BD iteration dim;
-    // linearizing would drop the iteration stride.
+    // len < product: the outermost dim describes a hardware BD iteration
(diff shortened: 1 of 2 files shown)


andrej (approved):
Thanks for fixing this. 

Happy to approve since it fixes the issue. If you have the time, I think I would need a little help understanding this in detail for future reference. I've stared at it for a while and am not sure I understand all the nuance.

<details>
<summary>If you have the time ...</summary>



> When a 4D shim DMABDOp has an explicit len representing the per-iteration transfer size (not the total) [...]

Isn't `len` _always_ supposed to encode a single BD iteration's length, i.e. should be the product of the lowest three dimensions? What happens if it's something else, or if it's not set?

> Concrete failure: transpose M_2048-N_64-m_64-n_64:

Is this shuffle_transpose or combined_transpose? Glancing at it I don't see an "explicit length" set on either of those.

> BD has len=4096, dims [<32, 4096>, <1, 64>, <64, 64>, <64, 1>], total=131072. After canonicalization → linear transfer of 4096 bytes repeated 32 times.

Initially you'd think this is supposed to be a linear transfer of 32*4,096=131,072 elements, but if you know that the highest dimension is iteration stepsize/wrap, you see that it's actually 32 separate transfers of contiguous blocks of 4,096 elements. The only places I can think of where this small difference would matter would be (a) lock behavior (if locks are associated with this BD, they will be triggered 32 times, not once) and (b) BD chaining (if there are other BDs in this DMA task, they will be interspersed after every iteration). In those cases the canonicalization would change the semantics by turning it into a single 131,072 element transfer -- if you're not reliant on lock behavior or BD chaining, those semantic changes would not show up. Is that understanding correct?

If so, it feels to me like that difference has nothing to do with whether an explicit length is set or not. It's just whether the highest dimension is used or not.

I think the fundamental difficulty is that iteration stepsize and iteration wrap are encoded as the highest dimensions, just like stride/wrap, although there semantics are different.

Please correct me, just trying to understand.

</details>

hunhoffe:
@andrej Thank you for the thoughtful feedback! You are quite right on all counts. It was my turn to stare at this PR for a while. If you have a few minutes, would you mind reviewing one more time?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2424,18 +2424,22 @@ struct LinearizeContiguousBDTransfer : public mlir::OpRewritePattern<DMABDOp> {
...
-        product *= dim.getSize();
-      len = static_cast<int32_t>(product);
-    }
+    // A 4-dimensional shim BD reserves the outermost dim as the hardware BD
+    // iteration dim, programmed into iteration_size / iteration_stride by
+    // AIEDMATasksToNPU; `len` covers only the inner three dims (the per-
+    // iteration transfer).  Linearizing would collapse the iteration into
+    // len and silently drop the iteration stride, so the hardware would
+    // repeatedly transfer the first `len` elements from offset 0 instead
+    // of sweeping the buffer.  An outer size of 1 is a trivial iteration
```
andrej: Not sure I follow this. If linearizing properly works, it should still send the whole buffer, not do a repeat. If linearization turns a contiguous transfer into a repeat, that's a bug. 

I suspect what happened is: Linearization dropped the highest dim altogether (due to the shorter lowest-three-dim `len`). The BDs in the example were configured as an infinite ping-pong loop, so the same BD was repeated infinitely. Claude observed this and it appeared like a "repeat". 

But I think the explanation "linearizing would collapse the iteration in to len" is nonsense/doesn't explain the repeat behavior.

I think this "Linearizing would..." hypothetical is misleading/confusing at this point.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2424,18 +2424,22 @@ struct LinearizeContiguousBDTransfer : public mlir::OpRewritePattern<DMABDOp> {
...
+    // (no address advancement, no semantic loss) and remains foldable.
+    if (dims->size() == 4 && dims->front().getSize() > 1)
+      return mlir::failure();
```
andrej: Might be worth double checking which "dimension" index the iteration stepsize/wrap is? Might not always be at index 4. Maybe it is but I know some other tiles have fewer dimensions but I don't know if those also have iteration stepsize/wrap.

test/dialect/AIE/canonicalize_no_linearize_iter_bd.mlir
```diff
@@ -0,0 +1,120 @@
...
+//
+//===----------------------------------------------------------------------===//
+//
+// Regression test for LinearizeContiguousBDTransfer: a 4-dimensional shim BD
+// reserves the outermost dim as a hardware BD iteration dim (programmed into
+// iteration_size / iteration_stride by AIEDMATasksToNPU); `len` covers only
+// the inner three dims (the per-iteration transfer).  Linearizing such a BD
+// would collapse the iteration into len and silently drop the iteration
+// stride, causing the NPU to repeatedly transfer only the first `len`
+// elements from offset 0 instead of sweeping the whole buffer.
```
andrej: Again I find this "Linearizing would..." hypothetical hard to follow

andrej (approved):

lib/Dialect/AIE/IR/AIEDialect.cpp
andrej: To put this differently, if we make `len` the product of all four dimension lengths (including `iteration_wrap`), set `iteration_count=1`, I think hardware would _not_ do a repeat. It would give a single correct linear transfer. It would only differ in the lock/BD chaining (if present) semantics. That's why I find the "Linearizing would ..." argument unconvincing.

PR #3038, opened 2026-04-24
Use tablegen Python bindings for trace events instead of separate generation

### LINKED ISSUE #3012: Use MLIR Python binding enums for trace events
Remove `generate_events_enum.py` and import the generated event enums from `aie.dialects.aie`

yenjames:
### Summary
Trace event enums were duplicated across two generation pipelines that both parsed the same aie-rt C headers. We remove the redundant Python enum generation.

### Changes
- **`python/utils/trace/events/__init__.py`**: Import event enums from aie.dialects._aie_enum_gen instead of generated per-arch modules. Use SimpleNamespace for get_events_for_device(). Fix GenericEvent.__init__ to check Enum before int since IntEnum is a subclass of both.
- **`python/CMakeLists.txt`**: Remove Python enum generation, GenerateTraceEventsEnum target, and associated copy/install steps.
- **`include/aie/Dialect/AIE/IR/CMakeLists.txt`**: Fold `events_database.json` generation into the existing `GenerateAIEEventsTD` target alongside the tablegen files, sharing the same aie-rt header inputs.
- **`utils/generate_events_json.py`**: Renamed from `generate_events_enum.py`, stripped of Python enum generation. Now JSON-only, invoked at build time by `GenerateAIEEventsTD`.
- **`test/python/trace_events.py`**: Updated to assert `MemTileEvent is None` for AIE1 instead of iterating an empty enum.


hunhoffe (approved):
Looks like a clean improvement! I had some questions, but overall, I like it!

hunhoffe:
Does this PR address this issue? https://github.com/Xilinx/mlir-aie/issues/3012

yenjames:
> Does this PR address this issue? #3012

Yes, just linked it to that issue.

python/utils/trace/events/__init__.py
```diff
@@ -6,27 +6,42 @@
...
+    MemTileEventAIE2,
+    CoreEventAIE2P,
+    MemEventAIE2P,
+    ShimTileEventAIE2P,
+    MemTileEventAIE2P,
 )
 
 from aie.dialects.aie import WireBundle, DMAChannelDir
 
+# Default to AIE2 for backwards compatibility
```
hunhoffe, line 38: I'm a little confused abut why we set defaults when there's a function below that does more correct conditional sets? Can you clarify a bit?

python/utils/trace/events/__init__.py
```diff
@@ -68,8 +98,12 @@ class GenericEvent:
     def __init__(
         self, code: typing.Union[CoreEvent, MemEvent, ShimTileEvent, MemTileEvent]
     ):
-        # For backwards compatibility, allow integer as event
-        if isinstance(code, int):
+        # For backwards compatibility, allow plain integer as event.
```
hunhoffe: Honestly curious: do we need backwards compatibility here?

utils/generate_events_json.py
```diff
@@ -214,7 +146,7 @@ def find_default_input_files():
 
 def main():
     argparser = argparse.ArgumentParser(
-        description="Generate Python enums and JSON database from AIE event headers"
+        description="Generate JSON event database from AIE event headers"
```
hunhoffe, line 149: Just to double check -- the JSON file is checked in in GitHub on this PR. In the past, I thought this was generated at compile time, and not checked in. I don't feel strongly about which one is best, but just to check -- was that purposeful?

fifield (changes requested):
This PR goes farther than removing the duplicated event enums by also removing the generated events_database.json. Generating the events_database.json is intended and not duplication. Can we restore that part, or fold the generation of the json into the cmake rules for generating the tablegen files (if reducing lines of cmake is also a goal of the PR)?

python/utils/trace/events/__init__.py
yenjames: Yes this is a part where I don't have a clear solution. The issue is the tablegen events suffixes the event names with the architecture (CoreEventAIE2.), but I still wanted to keep the aliases CoreEvent etc to paper over this name. The backwards compatibility comment is not very correct.

The get_events_for_device is more correct but it will require changing all examples to first acquire events with it.

utils/generate_events_json.py
yenjames: As pointed out by Jeff, this was a mistake on my end.

yenjames:
> This PR goes farther than removing the duplicated event enums by also removing the generated events_database.json. Generating the events_database.json is intended and not duplication. Can we restore that part, or fold the generation of the json into the cmake rules for generating the tablegen files (if reducing lines of cmake is also a goal of the PR)?

Reverted, this was overdone and a mistake on my end. Thank you for catching this!

python/utils/trace/events/__init__.py
yenjames: I think not really needed, we should use events by the enum type and error for other types.

fifield (approved):

Issue #2864, opened 2026-02-11
Port placement to MLIR and placer enhancement


yenjames:
This issue will track a series of effort to port IRON `SequentialPlacer` into MLIR infrastructure and enhancements to placement in MLIR-AIE.

erwei-xilinx:
mlir-air RFC https://github.com/Xilinx/mlir-air/issues/1567 is migrating AIR to emit `aie.logical_tile` across all tile-creation sites; PR https://github.com/Xilinx/mlir-air/pull/1570 has Stage A complete (memtiles + shim tiles + compute tiles routed through the placer with tight constraints, byte-identical placements vs. before).

Stage B (constraint relaxation, where AIR genuinely gives up placement decisions) needs additional placer capabilities. Concrete asks based on a survey of the placer at `origin/main`:

**G1. Row-range constraint on `LogicalTileOp`** — `(col, [row_min, row_max])` so AIR can express "this herd starts at row N" without hardcoding individual rows. Today only exact rows or full freedom is expressible. This is the **primary blocker for AIR compute-tile relaxation**: when AIR emits `(col, ?)`, the placer fills the column from the bottom (row 1 on VC1902), ignoring `row_offset=3`.

Suggested syntax:
```mlir
%t = aie.logical_tile<CoreTile>(2, 3..5)
%t = aie.logical_tile<CoreTile>(2, 3..)     // open upper bound
```

**G2. Other connectivity sources beyond objectfifos** — read `aie.flow`, `aie.packet_flow`, `aie.cascade_flow`, and shared-buffer cross-references for channel-requirement and grouping. AIR's DMA path (when `clUseObjFifo=false`) never creates objectfifos; today the placer is blind to it. Cascade likewise — `AIELowerCascadeFlows` validates adjacency at lowering time but the placer plays no role.

I see this is partly already in progress (the `aie.flow` / `aie.packet_flow` parts in your `placer-read-flow-ops` branch). The remaining piece is `aie.cascade_flow` so the placer can natively enforce cascade adjacency rather than relying on lowering-time errors.

**G3. Adjacency / mem-affinity constraint primitives between `LogicalTileOp`s** — express "these N tiles must be physically adjacent (cardinal neighbours, or `isLegalMemAffinity`)." Cascade requires cardinal adjacency; shared L1 requires mem-affinity. With this primitive AIR could emit `(?, ?)` and let the placer find any valid placement for cascade kernels.

[...]

PR #3041, opened 2026-05-01
[AIEPlacer] Read aie.flow / aie.packet_flow as channel-requirement and grouping sources

erwei-xilinx:
## Summary

Extends `SequentialPlacer` to read `aie.flow` and `aie.packet_flow` ops as additional connectivity sources, alongside the existing `aie.objectfifo` reader. Frontends that lower channels directly to `aie.flow` (without going through objectfifo) now get the same channel-aware placement that the objectfifo path enjoys today.

This is the first of a planned series (companion issue forthcoming, related to #2864) to make `--aie-place-tiles` work from any combination of high-level (`aie.objectfifo`) and lowered (`aie.flow`, `aie.packet_flow`, `aie.cascade_flow`, `aie.buffer`) connectivity ops.

### Motivation

The placer's current input contract (single `device.walk` over `LogicalTileOp` + `ObjectFifoCreateOp` + `ObjectFifoLinkOp`) means any IR that reaches the placer without objectfifos is treated as having zero connectivity:

- channel capacity is never validated → no error if a tile is over-subscribed
- mem/shim tiles fall through to the column-0 fallback → no proximity to connected cores

This affects callers that lower channels straight to `aie.flow` / `aie.dma_bd` / `aie.lock` / `aie.buffer` (e.g. mlir-air's DMA path).

## What this PR does

Additive. The objectfifo path is unchanged; pure-objectfifo IR produces bit-identical placement.

- **Phase 1**: also collect `FlowOp` and `PacketFlowOp`.
- **`addChannelRequirementsFromFlows`**: for each flow, if the source bundle is `DMA` and the source's defining op is a `LogicalTileOp`, increment its MM2S counter; symmetrically for `dst` and S2MM. `PacketFlowOp` is walked into for `PacketSourceOp` / `PacketDestOp`. Results are summed into the same `channelRequirements` map produced by the objectfifo path.
- **`buildFlowGroups`**: BFS over the flow graph treating each `LogicalTileOp` as a node and each flow as an undirected edge. Each connected component is one group; non-core tiles in the group need placement, core tiles in the group define the group's common column.
- **Phase 4b**: mirrors phase 4 but iterates flow groups instead of objectfifo groups. Tiles already placed in phase 4 are skipped, so objectfifo + flow IR coexists cleanly.

## Test plan

- [x] New lit test `test/place-tiles/sequential_placer/test_place_flows.mlir` (6 cases): mem-near-core via flow, shim→mem→core chain, multi-core grouping, channel capacity from flows, packet flows, non-DMA bundle (no demand).
- [x] All 7 tests in `test/place-tiles/` pass (6 existing + 1 new).
- [x] All 108 tests in `test/dialect/AIE/` pass (+ 1 pre-existing XFAIL).
- [x] Downstream verification: rebuilt mlir-air against this change; all 368 tests in `check-air-mlir` pass (+ 7 pre-existing XFAIL).

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEPlacer.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEPlacer.cpp
@@ -120,6 +120,8 @@ LogicalResult SequentialPlacer::place(DeviceOp device) {
   SmallVector<LogicalTileOp> logicalTiles;
   SmallVector<ObjectFifoCreateOp> objectFifos;
   SmallVector<ObjectFifoLinkOp> objectFifoLinks;
+  SmallVector<FlowOp> flows;
+  SmallVector<PacketFlowOp> pktFlows;
 
   device.walk([&](Operation *op) {
     if (auto lt = dyn_cast<LogicalTileOp>(op))
@@ -128,11 +130,18 @@ LogicalResult SequentialPlacer::place(DeviceOp device) {
       objectFifos.push_back(of);
     if (auto link = dyn_cast<ObjectFifoLinkOp>(op))
       objectFifoLinks.push_back(link);
+    if (auto f = dyn_cast<FlowOp>(op))
+      flows.push_back(f);
+    if (auto pf = dyn_cast<PacketFlowOp>(op))
+      pktFlows.push_back(pf);
   });
 
-  // Phase 2: Build channel requirements from ObjectFifo connectivity
+  // Phase 2: Build channel requirements. ObjectFifo connectivity is the
+  // primary source; `aie.flow` / `aie.packet_flow` ops contribute additively
+  // so the placer can be channel-aware on lowered IR (e.g. AIR's DMA path).
   auto channelRequirements =
       buildChannelRequirements(objectFifos, objectFifoLinks);
+  addChannelRequirementsFromFlows(flows, pktFlows, channelRequirements);
 
   // Phase 3: Place constrained tiles then compute tiles
   size_t nextCompIdx = 0;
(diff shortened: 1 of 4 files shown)


erwei-xilinx:
Thanks for the review. Addressed in commits e11311ab and a4dedf9a:

**1. `flow_memtile_two_cores` "different columns" comment** — Reworded to "Memtile shared by two cores: common column is the rounded average." (e11311ab)

**2. `flow_shim_channel_capacity` intro saying it errors out** — Reworded to "Two MM2S flows fit within a npu1 shim's 2 source DMA channels." which matches the actual assertion. (e11311ab)

**3. `groupToCoreCols` vs `groupToCoreTiles` mismatch in header docstring** — The stale docstring was removed entirely in e11311ab during a comment-trim pass.

**4 & 5. Nondeterministic iteration over `DenseMap`/`DenseSet` in `buildFlowGroups` and phase 4b** — Valid concern, fixed in a4dedf9a. Switched the adjacency graph to `MapVector<LogicalTileOp, SetVector<LogicalTileOp>>` and the group maps to `MapVector<int, SmallVector<LogicalTileOp>>`. Iteration now follows insertion order, which itself follows flow IR order, so group IDs and per-group tile order are stable across runs.

Note: the existing objectfifo path (`buildObjectFifoGroups` + phase 4) uses the same `DenseMap` pattern and has the same theoretical exposure. Left unchanged here to keep this PR scoped to the flow-reading addition; happy to file a follow-up if you'd like that path tightened too.

erwei-xilinx:
Cc: @yenjames

lib/Dialect/AIE/Transforms/AIEPlacer.cpp
```diff
@@ -472,72 +398,196 @@ SequentialPlacer::buildChannelRequirements(
...
+    for (Value consumerTile : ofOp.getConsumerTiles())
+      recordEndpoint(consumerTile, group);
+  }
+}
 
-    groupToFifos[groupId].push_back(ofOp);
+// Already-resolved TileOps have no placer-visible budget, so only flows whose
+// endpoints are LogicalTileOps contribute to channelRequirements.
+void SequentialPlacer::addChannelRequirementsFromFlows(
+    ArrayRef<FlowOp> flows, ArrayRef<PacketFlowOp> pktFlows,
```
yenjames, line 457: I think this function will overcount channel requirement for one-to-many or many-to-one flows. We need to deduplicate based on the DMA channel:
```MLIR
// Broadcast: same source channel 0, should only 
// consume one output channel for producer tile
aie.flow(%shim, DMA : 0, %c1, DMA : 0)
aie.flow(%shim, DMA : 0, %c2, DMA : 0)
```

erwei-xilinx: Thanks for the comment. Good catch. Fixed in 7d335727.

hunhoffe (approved):
Thank you for this work Erwei!

PR #3042, opened 2026-05-01
[AIEPlacer] Read aie.cascade_flow as adjacency constraint

erwei-xilinx:
## Summary

Teaches `SequentialPlacer` to honor `aie.cascade_flow` adjacency at placement time. The constraint is enforced both for fully-constrained tiles (validated as a precondition) and for partial / unconstrained `CoreTile`s (filtered in the candidate loop). Cardinal orientations that pass the cascade_flow verifier but the cascade lowering rejects are caught here, so users get a placement-time error instead of a late lowering error or invalid placement.

This is the second of the planned series following PR #3041 (flow / packet_flow channel-requirements). Companion to issue #2864 G2.

### What changes

- Phase 1 collects `CascadeFlowOp`.
- New `buildCascadeAdjacency()` constructs a per-`LogicalTileOp` peer adjacency map.
- New `satisfiesCascadeAdjacency()` checks whether a candidate position satisfies every adjacency relative to already-placed peers.
- Phase 3 fully-constrained branch validates adjacency before placing.
- Phase 3 partial / unconstrained branch filters candidates by adjacency.

The constraint matches `AIELowerCascadeFlows` exactly: for `cascade_flow(src, dst)`, `src` must be one row above (north) or one column to the left (west) of `dst`. The cascade_flow verifier accepts any cardinal direction; the lowering only accepts south-or-west; this PR aligns the placer with the lowering.

### Behavior preserved

- Pure-objectfifo IR is unchanged.
- IR with no `aie.cascade_flow` is unchanged.
- For `aie.cascade_flow` IR with all endpoints fully-constrained at compatible positions, the placer is a pass-through (no behavioral change vs. today, just an added consistency check).

## Test plan

- [x] New `test/place-tiles/sequential_placer/test_place_cascade.mlir` (3 cases): unconstrained chain (greedy fills east along row 2), vertical chain with pinned head (fills south down column 0), pre-pinned compatible cascade.
- [x] New error cases in `test_place_errors.mlir`: 5-tile chain on `npu1_1col` (4 core rows) is unsatisfiable; cascade with src north of dst (verifier-legal, lowering-illegal) caught.
- [x] All 7 lit-files in `test/place-tiles/` pass.
- [x] All 109 tests in `test/dialect/AIE/` pass (+ 1 pre-existing XFAIL).
- [x] Downstream: rebuilt mlir-air against this; all 368 `check-air-mlir` tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3044, opened 2026-05-02
[AIEPlacer] Read aie.buffer as tile-local memory capacity constraint

erwei-xilinx:
## Summary

Teaches `SequentialPlacer` to honor tile-local memory capacity at placement time by summing `BufferOp::getAllocationSize()` per `LogicalTileOp` and rejecting candidate physical CoreTiles whose `getLocalMemorySize()` cannot host the demand.

This is the third entry in the planned reader series following PR #3041 (flow / packet_flow channel-requirements) and PR #3042 (cascade_flow adjacency). Companion to issue #2864.

## Why on top of `AIEAssignBuffers`

`AIEAssignBuffers` already validates buffer-byte totals per physical `TileOp`, but it runs **after** placement. Two concrete improvements:

1. **Diagnostic provenance.** Errors point at the user's `aie.logical_tile` op (with source location and explicit `bytes vs capacity` numbers) rather than a downstream physical tile the user never wrote.
2. **Avoid wasted greedy choices.** The greedy core branch now skips physical candidates that cannot host the demand, so an unconstrained logical tile won't pick a doomed-to-fail physical tile when a fitting one exists later in the sweep.

Symmetric with the existing channel-capacity validation: both are finite per-tile resources, both gated at the same point.

## What changes

- Phase 1 also collects `BufferOp`.
- `buildBufferRequirements()` aggregates bytes per `LogicalTileOp` op pointer; `TileOp` peers are skipped (they're physical and remain `AIEAssignBuffers`' responsibility).
- `tileMemoryCapacity(TileID)` dispatches on tile type to the matching target-model accessor (`getLocalMemorySize` / `getMemTileSize`).
- `fitsBufferCapacity()` returns true for non-CoreTile candidates: the existing placer shares MemTiles across multiple `LogicalTileOp`s (cf. `removeTile` skipping non-CoreTile), and validating accumulated bytes there requires per-physical-tile tracking like `inputChannelsUsed`. Left to a follow-up; `AIEAssignBuffers` still catches MemTile overflow.
- Phase 3 fully-constrained branch: validates and emits an error naming bytes vs capacity.
- Phase 3 partial / unconstrained branch: filters candidates and, on exhaustion, suffixes the error with `(buffer capacity exceeded)` / `and buffer capacity`.

## Behavior preserved

- IR with no `aie.buffer` is unchanged.
- IR whose buffers all reference physical `aie.tile` (not `aie.logical_tile`) is unchanged.
- AIE1 / AIE2 / AIE2P and the npu1/npu1_*/npu2 variants pick up the right capacity automatically via the target model.

## Test plan

- [x] New `test/place-tiles/sequential_placer/test_place_buffers.mlir` (2 positive cases): pinned tile + small buffer, unconstrained tile + small buffer.
- [x] New error cases in `test_place_errors.mlir`: `buffer_overflow_pinned` (128KB on a 64KB CoreTile) and `buffer_overflow_unconstrained` (every candidate fails the filter).
- [x] All 7 lit-files in `test/place-tiles/` pass.
- [x] All 109 tests in `test/dialect/AIE/` pass (+ 1 pre-existing XFAIL).
- [x] Downstream: rebuilt mlir-air against this change; no API breakage.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -131,6 +131,14 @@ def AIE_LogicalTileOp: AIE_Op<"logical_tile", [
 
   let extraClassDeclaration = [{
     TileID getCanonicalTileID();
+
+    CoreOp getCoreOp() {
+      auto users = getResult().getUsers();
+      for(auto user : users)
+        if(auto coreOp = llvm::dyn_cast<CoreOp>(*user))
+          return coreOp;
+      return nullptr;
+    }
   }];
 
   let extraClassDefinition = [{
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -131,14 +131,6 @@ def AIE_LogicalTileOp: AIE_Op<"logical_tile", [
 
   let extraClassDeclaration = [{
     TileID getCanonicalTileID();
-
-    CoreOp getCoreOp() {
-      auto users = getResult().getUsers();
-      for(auto user : users)
-        if(auto coreOp = llvm::dyn_cast<CoreOp>(*user))
-          return coreOp;
-      return nullptr;
-    }
   }];
 
   let extraClassDefinition = [{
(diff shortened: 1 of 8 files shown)


erwei-xilinx:
Closing — this PR's scope (per-LogicalTileOp L1 capacity check at placement time) duplicates the overflow detection that AIEAssignBuffers already does, with a strictly less informative diagnostic (flat byte total vs AssignBuffers' per-buffer memory map). The intended third entry in the #3041/#3042 series — buffer-as-adjacency reader, mirroring the cascade pattern with isLegalMemAffinity — will be opened as a separate PR.

PR #3046, opened 2026-05-04
[AIEPlacer] Read shared-L1 aie.buffer references as memory-affinity adjacency

erwei-xilinx:
## Summary

Teaches `SequentialPlacer` to honor memory-affinity adjacency when an `aie.core(%lt_consumer)` reads or writes an `aie.buffer(%owner)` attached to a different tile. The consumer's physical position must satisfy `targetModel->isLegalMemAffinity(consumerPos, ownerPos)` so the L1 access actually resolves to a reachable neighbor-memory window. Today the placer doesn't model this; cross-tile buffer accesses survive placement and only break later in lowering.

This is the third of the planned reader series after #3041 (`flow` / `packet_flow` channel-requirements) and #3042 (`cascade_flow` adjacency), completing the path for IR that reaches the placer with raw buffers instead of objectfifos. Companion to #2864.

## What changes

- New `BufferAdjacency` struct (parallel to `CascadeAdjacency` from #3042): per-LTO peer edges indexed by either endpoint.
- New `buildBufferAdjacency()` walks each `aie.core` body, traces each operand back to its underlying `BufferOp` through view-like memref ops via `ViewLikeOpInterface::getViewSource`, records edges `(consumer LTO, owner TileLike)` — de-duplicated per consumer/owner pair. Self-references and same-LTO accesses are skipped.
- New `satisfiesBufferAdjacency()` checks each edge using `isLegalMemAffinity`; defers when a peer is unplaced and unpinned, so one symmetric predicate at either endpoint catches any violation.
- New `attachBufferPeerNotes()` decorates the diagnostic with peer positions.
- Phase 3 fully-constrained branch validates adjacency before placing.
- Phase 3 partial / unconstrained branch filters candidates by adjacency, with a `(shared-L1 buffer adjacency unsatisfiable)` diagnostic suffix when the filter was the cause of failure.

## Behavior preserved

- IR with no `aie.buffer` is unchanged.
- IR whose buffers are all attached to the same tile as the consuming core is unchanged (no cross-tile edges built).
- Pure objectfifo IR is unchanged (objectfifo-derived buffers are materialized after the placer runs).
- AIE1 / AIE2 / AIE2P pick up the right affinity rules automatically via the target model.

## Test plan

- [x] New `test/place-tiles/sequential_placer/test_place_buffer_adjacency.mlir` (6 positive cases): unconstrained consumer steered to N of pinned owner; 3-consumer star around pinned owner filling W/N/S slots; both endpoints pinned at compatible coords; access through `memref.subview`; self-reference (no constraint); npu2 target-model dispatch.
- [x] New error cases in `test_place_errors.mlir`: both endpoints pinned far apart (peer-note diagnostic), unconstrained consumer with column constraint that has no affinity slot, 4-consumer star oversubscribing the owner's affinity neighborhood.
- [x] All 7 lit-files in `test/place-tiles/` pass.
- [x] All 110 tests in `test/dialect/AIE/` pass (1 pre-existing XFAIL).
- [x] Downstream: rebuilt mlir-air against this change; no API breakage.

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
@@ -137,6 +137,10 @@ private:
   // Adjacency<Predicate> once both have landed.
   struct BufferAdjacency {
     llvm::SmallVector<std::pair<LogicalTileOp, TileLike>, 4> edges;
+    llvm::DenseMap<mlir::Operation *, llvm::SmallVector<unsigned, 2>>
+        tileToEdges;
+  };
+
   // `tileToEdges` indexes into `edges` only for `LogicalTileOp` endpoints
   // (the ones the placer visits); `TileOp` peers contribute coords via
   // `TileLike::tryGetCol`/`tryGetRow`.
--- a/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
@@ -113,6 +113,36 @@ private:
   buildChannelRequirements(
       llvm::SmallVector<ObjectFifoCreateOp> &objectFifos,
       llvm::SmallVector<ObjectFifoLinkOp> &objectFifoLinks);
+
+  // Adjacency constraints inferred from cross-tile L1 buffer accesses.
+  //
+  // When `aie.core(%lt_consumer)` reads or writes an `aie.buffer(%owner)`
+  // attached to a different tile, the consumer's physical placement must
+  // satisfy `targetModel->isLegalMemAffinity(consumerPos, ownerPos)` so the
+  // L1 access actually resolves to a reachable neighbor-memory window.
+  //
+  // `edges` records each cross-tile reference once; `tileToEdges` indexes by
+  // any LogicalTileOp endpoint so a check at either side catches the
+  // violation regardless of placement order. Owners that are physical TileOps
+  // contribute their fixed coords via `TileLike::tryGetCol/tryGetRow` and are
+  // not indexed (the placer never visits them).
+  struct BufferAdjacency {
+    llvm::SmallVector<std::pair<LogicalTileOp, TileLike>, 4> edges;
+    llvm::DenseMap<mlir::Operation *, llvm::SmallVector<unsigned, 2>>
+        tileToEdges;
+  };
+
+  BufferAdjacency
+  buildBufferAdjacency(llvm::ArrayRef<LogicalTileOp> logicalTiles);
+
+  // Defers when a peer is unplaced and unpinned; symmetric so a check at one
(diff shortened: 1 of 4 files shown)


include/aie/Dialect/AIE/Transforms/AIEPlacer.h
```diff
@@ -113,6 +113,25 @@ class SequentialPlacer : public Placer {
   buildChannelRequirements(
       llvm::SmallVector<ObjectFifoCreateOp> &objectFifos,
       llvm::SmallVector<ObjectFifoLinkOp> &objectFifoLinks);
+
+  // Cross-tile L1 buffer accesses: consumer must satisfy isLegalMemAffinity
+  // of owner. Same shape as CascadeAdjacency in #3042; TODO: factor a generic
+  // Adjacency<Predicate> once both have landed.
```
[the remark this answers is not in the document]
erwei-xilinx: Addressed in d48445d340 — comment now spells out the parameter mapping (coreCol=consumerCol, coreRow=consumerRow, memCol=ownerCol, memRow=ownerRow).

erwei-xilinx:
Cc: @yenjames

include/aie/Dialect/AIE/Transforms/AIEPlacer.h
```diff
@@ -113,6 +113,27 @@ class SequentialPlacer : public Placer {
   buildChannelRequirements(
       llvm::SmallVector<ObjectFifoCreateOp> &objectFifos,
       llvm::SmallVector<ObjectFifoLinkOp> &objectFifoLinks);
+
+  // Cross-tile L1 buffer access: consumer LTO's core must pass
+  // targetModel->isLegalMemAffinity(coreCol=consumerCol, coreRow=consumerRow,
+  //                                 memCol=ownerCol,     memRow=ownerRow).
+  // Same shape as CascadeAdjacency (#3042); TODO: factor a generic
+  // Adjacency<Predicate> once both have landed.
```
hunhoffe: This TODO would be nice, but can wait for future work.

hunhoffe (approved):
Looks good!

include/aie/Dialect/AIE/Transforms/AIEPlacer.h
erwei-xilinx: Makes sense. Let me work on that today.

erwei-xilinx: I'll just edit this PR in place.

erwei-xilinx: Done.

hunhoffe: Thank you!

PR #3053, opened 2026-05-07
[AIEPlacer] Read aie.lock and aie.objectfifo.allocate as memory-affinity adjacency

erwei-xilinx:
## Summary

Follows on from #3046 (buffer adjacency). Closes two more memory-affinity gaps that today either fail at lowering or silently degrade. Both reuse the `Adjacency` machinery from #3046 with the same `isLegalMemAffinity` predicate; the only new code is the walkers and labels.

| Constraint | Today | After this PR |
|---|---|---|
| `aie.use_lock(%lock_at_other_LTO, ...)` in core body | `LockOp::verify` skips LTO users -- violation surfaces later as a verifier error on the resolved `TileOp` | Caught at placement; unconstrained consumer steered into a memory-affinity neighbor |
| `aie.objectfifo.allocate @of (%delegate)` with non-adjacent endpoints | `AIEObjectFifoStatefulTransform` silently degrades to a DMA fifo, ignoring the user's `allocate` intent | Caught at placement; unconstrained endpoints steered to satisfy delegate affinity |

Reference: `LockOp::verify` skips LTO users at [AIEDialect.cpp:251-254](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/IR/AIEDialect.cpp#L251-L254). The objectfifo.allocate contract is documented in `ObjectFifoAllocateOp`'s docstring ("Only works if all the objectfifo's tiles, producer and consumers, have shared memory access to the specified tile's memory module"); the silent-DMA fallback is in `requiresDMAs` at [AIEObjectFifoStatefulTransform.cpp:339-348](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp#L339-L348).

### Lock adjacency (commit 1)

- New `buildCoreBodyAdjacency(logicalTiles, extractOwner)` private helper that owns the "walk every core body, dedup cross-tile owners, emit edge" scaffolding. `buildBufferAdjacency` (refactored from #3046) and the new `buildLockAdjacency` are 5-line wrappers that just supply the `Value -> TileLike` extraction. Pre-empts duplication that would otherwise grow with each new "core body op references X attached to another tile" constraint.
- New `buildLockAdjacency` walker: traces each `aie.use_lock`'s lock operand to its defining `LockOp`, then to the owner `TileLike`. Same cross-tile / dedup rules as buffer.

### ObjectFifo.allocate adjacency (commit 2)

- Phase 1 walk also collects `ObjectFifoAllocateOp`.
- New `buildObjectFifoAllocateAdjacency` walker: for each `allocate`, emits edges `(producer, delegate)` and `(consumer_i, delegate)` for every cross-tile endpoint.

### Phase 3 wiring (both commits)

- Lock and objectfifo.allocate share the `memAffinityPred` lambda with buffer (same `isLegalMemAffinity` rule), each with its own edges and labels.
- Diagnostic suffix and peer-note attachment extended to cover both new edge kinds. The unconstrained-placement "memory-affinity was cause" flag triggers when buffer, lock, or objectfifo.allocate fails.

## Behavior preserved

- IR with no `aie.lock` and no `aie.objectfifo.allocate` is unchanged.
- IR whose locks/allocates are already memory-affinity-compatible is unchanged.
- Pure `aie.objectfifo` (no `allocate`) is intentionally **not** constrained -- without an `allocate`, lowering is free to pick DMA or shared-L1 based on post-placement positions, so adding edges would over-constrain valid designs.
- AIE1 / AIE2 / AIE2P pick up the right affinity rules via the target model.

## Test plan

- [x] New `test/place-tiles/sequential_placer/test_place_lock_adjacency.mlir` (5 positive cases): unconstrained consumer steered to N of pinned owner; both endpoints pinned at compatible coords; self-reference (no constraint); buffer + lock from same owner agree; npu2 target-model dispatch.
- [x] New `test/place-tiles/sequential_placer/test_place_objectfifo_allocate_adjacency.mlir` (4 positive cases): two unconstrained endpoints steered to memory-affinity slots around pinned delegate; all three endpoints pinned-compatible; producer-as-delegate self-reference; npu2 target-model dispatch.
- [x] New error cases in `test_place_errors.mlir`: lock pinned-far-apart violation; lock unconstrained-consumer-with-column-constraint unsatisfiable; objectfifo.allocate three-pinned with one far away (peer notes for all endpoints); objectfifo.allocate pinned-delegate + unsatisfiable column.
- [x] All 11 lit-files in `test/place-tiles/` pass.

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
@@ -154,12 +154,29 @@ private:
     }
   };
 
+  // Generic walker: for each `CoreTile` LTO, walk its core body and use
+  // `extractOwner` to map each operand value to the owning `TileLike` (or
+  // null if the operand is irrelevant). For every distinct cross-tile owner,
+  // emit edge `(consumer, owner)`. Used by buffer and lock adjacency, both
+  // of which derive from "core body op references something attached to
+  // another tile".
+  Adjacency buildCoreBodyAdjacency(
+      llvm::ArrayRef<LogicalTileOp> logicalTiles,
+      llvm::function_ref<TileLike(mlir::Value)> extractOwner);
+
   // Cross-tile L1 buffer access: consumer LTO's core must pass
   // targetModel->isLegalMemAffinity(coreCol=consumerCol, coreRow=consumerRow,
   //                                 memCol=ownerCol,     memRow=ownerRow).
   // The first endpoint of each edge is the consumer LTO; the second is the
   // owner tile.
   Adjacency buildBufferAdjacency(llvm::ArrayRef<LogicalTileOp> logicalTiles);
+
+  // Cross-tile L1 lock access: same `isLegalMemAffinity` rule as buffers,
+  // since AIE2 locks live in the L1 region of their owning tile and are only
+  // visible to memory-affinity neighbors. The `UsesAreAccessible` trait on
+  // `LockOp::verify` enforces this for placed `TileOp` users but skips
+  // `LogicalTileOp` users, which is exactly the gap the placer fills.
+  Adjacency buildLockAdjacency(llvm::ArrayRef<LogicalTileOp> logicalTiles);
 
   // The first endpoint of each edge is the cascade source; the second is the
   // destination.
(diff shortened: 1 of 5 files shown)


erwei-xilinx:
Closing in favor of a unified objectfifo → Adjacency refactor. The work in this PR was too narrow — it added rules on `aie.lock` and `aie.objectfifo.allocate` but didn't actually redirect the existing objectfifo placement work through the Adjacency machinery, which was the original intent. A new PR will replace `ConnectivityGroup` / `buildObjectFifoGroups` / `buildFlowGroups` with `Adjacency`-backed equivalents and rewire Phase 4 centroid placement accordingly.

PR #3054, opened 2026-05-07
Fix llvm update and mlir distro

hunhoffe:
Supersedes #3052

Round 2, this PR should have better permissions for CI tasks since it's not from a fork.

**update-llvm orchestration.** `utils/update_llvm_version.py` and `.github/workflows/update-llvm.yml` now follow an identify-then-build-then-apply flow. A new `--identify-only` mode resolves the target LLVM commit and reports `target_commit` / `wheel_exists` / `bump_reason` via `$GITHUB_OUTPUT` without mutating files; if no `mlir-distro` wheel exists yet, the workflow dispatches `mlirDistro.yml` itself (passing a `DISPATCH_NONCE` input echoed into `mlirDistro`'s `run-name`), polls `gh run list` filtering by that nonce in `displayTitle` to deterministically locate its run, then `gh run watch`es it to completion before re-invoking the script to mutate `clone-llvm.sh` etc. The previous silent-fallback in `update_files()` (use commit date when wheel is missing) is gone; missing wheel is now a hard error since the orchestration step is supposed to have built it. `find_wheel_in_distro`'s regex is widened to tolerate the optional `-<run_id>(_<attempt>)?` build-tag segment introduced below, so it matches both canonical and PR-built wheels.

**mlirDistro hardening + race-free, bounded wheel uploads.** `mlirDistro.yml`'s previous unauthenticated `curl` against `api.github.com/repos/llvm/llvm-project/commits/main` was hitting the 60/hr per-IP limit and silently producing `LLVM_PROJECT_COMMIT=null`, which then built broken wheels — the fetch is now authenticated with `--fail-with-body` and an explicit empty/`null` check that hard-fails the run, and the `Release current commit` step is gated to skip on fork PRs (no token write access anyway, just a confusing 403). Three layered fixes then make uploads both race-free and bounded: (1) **PEP 427 build tags on PR builds** in both Distro workflows rename wheels to embed `<run_id>_<attempt>` so concurrent CI runs can never collide on a `dev-wheels` asset name — the build tag is invisible to `pip install mlir==<version>`, and canonical `workflow_dispatch` builds keep deterministic filenames so `find_wheel_in_distro` still works; (2) **`concurrency:` blocks** on both workflows serialize `workflow_dispatch` (and `schedule`) so two publishes can't overlap on the consumer-facing tags — structural fix for the race PR #2789 worked around with `replacesArtifacts: false`, which stays as defense-in-depth (and `mlirAIEDistro.yml` is brought in line, it was still `true`); (3) **`pruneReleaseAssets.yml`** janitor runs daily and deletes `dev-wheels` assets older than 30 days, since (1) means new uploads never overwrite old ones and `dev-wheels` would otherwise grow until it hit GitHub's 1000-asset-per-release cap (exactly what's been failing this PR's upload jobs).

### DIFF
--- a/.github/workflows/mlirAIEDistro.yml
+++ b/.github/workflows/mlirAIEDistro.yml
@@ -32,6 +32,14 @@ on:
         required: false
         default: ''
 
+concurrency:
+  # PR runs from the same PR cancel earlier in-flight runs (standard supersede).
+  # workflow_dispatch serializes per-event so the second publish waits for the
+  # first — preventing the upload race that PR #2789 worked around with
+  # `replacesArtifacts: false`.
+  group: ${{ github.workflow }}-${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.event_name }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
 
   get_aie_project_commit:
--- a/.github/workflows/mlirAIEDistro.yml
+++ b/.github/workflows/mlirAIEDistro.yml
@@ -492,16 +492,38 @@ jobs:
           name: build_artifact_${{ matrix.OS }}_${{ matrix.ARCH }}_rtti_${{ matrix.ENABLE_RTTI }}
           path: dist
 
+      # See mlirDistro.yml for the rationale: a PEP 427 build tag makes the
+      # uploaded asset name unique per CI run so concurrent uploads cannot
+      # collide. Skipped on workflow_dispatch so the canonical `latest-wheels`
+      # release keeps deterministic filenames.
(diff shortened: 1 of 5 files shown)


jgmelber (approved):
LGTM!

hunhoffe:
Workflow test: https://github.com/Xilinx/mlir-aie/actions/runs/25701807329

hunhoffe:
Note: from the test, the 422 self-fixes once this is merged due to a chicken-and-egg problem. The error is "input doesn't exist on the dispatched ref"; once mlirDistro.yml on main gains DISPATCH_NONCE, the next update-llvm run dispatches cleanly.

erwei-xilinx (approved):

PR #3056, opened 2026-05-07
Support firmware scratchpad feature

andrej:
Adds support for the XRT scratchpad memory feature, see example usage [test.cpp](https://github.com/andrej/mlir-aie/blob/171857916ff8c1b646fc5581779eb6ead8b68607/test/npu-xrt/scratchpad_regwrite/test.cpp) and [aie.mlir](https://github.com/andrej/mlir-aie/blob/171857916ff8c1b646fc5581779eb6ead8b68607/test/npu-xrt/scratchpad_regwrite/aie.mlir) as well as [docs in the code](https://github.com/andrej/mlir-aie/blob/171857916ff8c1b646fc5581779eb6ead8b68607/include/aie/Dialect/AIEX/IR/AIEX.td#L952).

I'm working on a more usable higher-level abstraction for run-time parameters using this mechanism.

image


lib/Targets/AIETargetNPU.cpp
```diff
@@ -199,6 +199,44 @@ void appendPreempt(std::vector<uint32_t> &instructions, NpuPreemptOp op) {
...
+  // Byte 3: padding
+  // Bytes 4-7: FuncArg
+  // Bytes 8-11: RegOff (absolute offset from AIE array base to register pair)
+  auto words = reserveAndGetTail(instructions, 3);
+
+  words[0] = XAIE_IO_UPDATE_REG;
+  words[0] |= (static_cast<uint32_t>(op.getStateTableIdx()) << 8);
+  words[0] |= (static_cast<uint32_t>(op.getFunc()) << 16);
+  words[1] = op.getFuncArg();
+  words[2] = *op.getAbsoluteAddress();
```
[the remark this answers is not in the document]
andrej: Other operations rely on getAbsoluteAddress() as well.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1025,6 +1154,12 @@ def AIE_NpuLoadPdiOp: AIEX_Op<"npu.load_pdi", []> {
     runtime by the driver or host program.
 
     If a symbol reference is provided, the compiler driver (aiecc.py) will match it to a device symbol name and assign the PDI ID field based on it.
+
+    ### Watch Out!
+
+    The firmware optimizes out repeated `load_pdi` operations if they refer to the same PDI.
+    To force a reload / device reset, intersperse a `load_pdi` to a different PDI between same-device reloads.
+    The easiest workaround is to create an empty `aie.device(npu2) @empty {}` and load that before the main device load.
```
jgmelber, line 1162: Interesting suggestion to force more configuration overhead 😄

include/aie/Dialect/AIEX/IR/AIEXEnums.h
```diff
@@ -0,0 +1,19 @@
+//===- AIEXEnums.h ----------------------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2024 Advanced Micro Devices, Inc.
```
hunhoffe: nit: year?

hunhoffe (approved):
Double check the modules, but otherwise looks good! I'm excited to see what you do with this feature.

include/aie/Dialect/AIEX/IR/AIEX.td
andrej: It's unfortunately what I've had to do in a lot of places. I guess maybe we should raise this as an issue with firmware people because it really is silly

jgmelber: What is it needed for?

ypapadop-amd:
Is this a new memory type, right?

include/aie/Dialect/AIEX/IR/AIEX.td
andrej: In [this test](https://github.com/andrej/mlir-aie/blob/f95f46b30f19447683e67275c69a3eb3e6cb86b0/test/npu-xrt/scratchpad_regwrite/aie.mlir#L77): (a) Cause core to re-execute from the start (alternative to `while(true)`which in this case did not work), basically a reset and (b) reset the scratchpad parameter buffer to zero (admittedly a little heavy-handed, if it was only for that a write32 would probably work too).

More generally: If I write `load_pdi`, I'd like a `load_pdi` to happen. The above two examples are two cases where just omitting causes unexpected different semantics (e.g. buffers not initialized to the values that are in the PDI that we supposedly loaded).

andrej:
> Is this a new memory type, right?

It's a new firmware feature. I've described how it works [here](https://github.com/andrej/mlir-aie/blob/f95f46b30f19447683e67275c69a3eb3e6cb86b0/include/aie/Dialect/AIEX/IR/AIEX.td#L964).

ypapadop-amd:
> > Is this a new memory type, right?
> 
> It's a new firmware feature. I've described how it works [here](https://github.com/andrej/mlir-aie/blob/f95f46b30f19447683e67275c69a3eb3e6cb86b0/include/aie/Dialect/AIEX/IR/AIEX.td#L964).

Is there more information on it? Does it persist among dispatches / hwctx reconfiguration for example?

andrej:
It persists across `xrt::runs`, for as long as the buffer object returned by `get_ctrl_scratchpad_bo` lives.

PR #3061, opened 2026-05-08
Add scratchpad run-time parameter functionality

andrej:
Builds on the XRT/firmware scratchpad feature (#3056) to enable more efficient run-time parameters. The XRT scratchpad supports up to 32 parameters in a "StateTable" that can be written to directly from DDR on the host, copied over by the firmware into a local memory, and from there used to write to registers in the NPU partition. 

The primary advantage/difference over the existing run-time parameter implementation is that these parameters are truly read from DDR. They are not hard-coded as write32s in the instruction sequence, and therefore do not require instruction sequence patching or different instruction sequences for different parameter values.

This PR abstracts the XRT scratchpad mechanism to make it more user-friendly through an `aiex.parameter` declaration (at the module level since parameters can be shared across devices), a `aiex.sync_parameters_from_host` operation that performs the copy from DDR -> command processor memory -> NPU registers/buffers, and a `aiex.read_parameter` operation that allows reading the parameters from the core. For the host side, a convenience class `ScratchpadParameter` provides an abstraction to write the parameters to the correct offset in DDR (no manual tracking of offsets, instead we can use named parameters as declared in the MLIR using `aiex.parameter`). 

The mapping of parameter name to DDR offset/StateTable index (where the host C++ writes to and runtime sequence reads from) is exported as additional metadata to be consumed by the runtime libs in a `params.txt` file. The mapping of parameter name to buffers on NPU cores is kept inline in the MLIR.

```
    test.cpp (host code) 
          |
          |    ScratchpadParameter abstraction in test_utils
          |    ScratchpadParameter.write("foo", value)
          |
 parameter in DDR 
          |
          |    aiex.sync_parameters_from_host in runtime sequence
          |
  parameter in scratchpad
      /  |  \
    /    |    \      parameters written to NPU cores, one buffer per parameter per core
  /      |      \     (this is also done by aiex.sync_parameters_from_host)
/        |        \
                 aiex.read_parameter("foo")

```

C++ usage example: [NPU-side (MLIR)](https://github.com/andrej/mlir-aie/blob/423a6c1aa80ab47bd13f15504e4c2243bc3b9817/test/npu-xrt/scratchpad_params/aie.mlir#L32), [host-side (C++)](https://github.com/andrej/mlir-aie/blob/423a6c1aa80ab47bd13f15504e4c2243bc3b9817/test/npu-xrt/scratchpad_params/test.cpp#L66)

Python usage example: [NPU-side (IRON Python)](https://github.com/andrej/mlir-aie/blob/423a6c1aa80ab47bd13f15504e4c2243bc3b9817/test/npu-xrt/scratchpad_params_python/aie_design.py#L46), [host-side (Python)](https://github.com/andrej/mlir-aie/blob/423a6c1aa80ab47bd13f15504e4c2243bc3b9817/test/npu-xrt/scratchpad_params_python/test.py#L58)

There is also another mode to use this where rather than use it as run-time parameters on NPU cores, you can use it to patch DMA BD addresses, [C++ example](https://github.com/andrej/mlir-aie/tree/params/test/npu-xrt/scratchpad_addr_offset), [Python example](https://github.com/andrej/mlir-aie/tree/params/test/npu-xrt/scratchpad_addr_offset_python).

Note: Depends on [XRT PR #9813](https://github.com/Xilinx/XRT/pull/9813) / [XRT PR #9814](https://github.com/Xilinx/XRT/pull/9814) for the Python bindings


include/aie/Dialect/AIE/IR/AIEDialect.h
```diff
@@ -59,6 +59,16 @@ uint32_t getShimBurstLengthBytes(const AIE::AIETargetModel &tm,
...
 
+// Generate a symbol name guaranteed to be unique within the symbol table of
+// `symbolTableOp`. Names are formed as "<prefix><n>" for increasing n; the
+// counter is advanced past the chosen value so repeated calls with the same
+// counter remain efficient and produce distinct names. The returned name is
+// not inserted into the symbol table; the caller is responsible for creating
+// a symbol with that name before the next call (otherwise the same name will
+// be returned again).
+std::string generateUniqueSymbolName(mlir::Operation *symbolTableOp,
+                                     llvm::StringRef prefix, unsigned &counter);
```
andrej, line 70: Verify what the blockwrite lowering uses for generating the memref global buffers' names. Either reuse this function there or what they are using there for the parameter buffer names.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1011,6 +1021,8 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
         OptionalAttr<AIEI32Attr>:$bd_id,
         OptionalAttr<PacketInfoAttr>:$packet,
         DefaultValuedOptionalAttr<AIEI32Attr, "0">:$burst_length,
+        // if set, the aiex.parameter that will override the BD's address
```
andrej: comment not accurate, the parameter will be _added_ to the BD's address (as an additional offset, hence the name)

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -2259,7 +2271,13 @@ def AIE_RuntimeSequenceOp : AIE_Op<"runtime_sequence", [
     The input arguments are arguments passed in from the host at kernel invocation time. This may include buffers on the host.
   }];
   let arguments = (
-    ins DefaultValuedAttr<SymbolNameAttr, "getDefaultRuntimeSequenceName()">:$sym_name
+    ins DefaultValuedAttr<SymbolNameAttr, "getDefaultRuntimeSequenceName()">:$sym_name,
+    // If set to false, suppresses automatic insertion of the parameter-sync
+    // preamble (create_scratchpad + set_lock) by `--aie-lower-parameters`.
```
andrej: preamble also contains update_from_scratchpad ops, not just create_scratchpad and set_lock

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -690,7 +690,9 @@ def AIE_NpuDmaMemcpyNdOp: AIEX_Op<"npu.dma_memcpy_nd", [
         DefaultValuedOptionalAttr<I64Attr, "0">:$d0_zero_after,
         DefaultValuedOptionalAttr<I64Attr, "0">:$d1_zero_after,
         DefaultValuedOptionalAttr<I64Attr, "0">:$d2_zero_after,
-        DefaultValuedOptionalAttr<I64Attr, "0">:$burst_length
+        DefaultValuedOptionalAttr<I64Attr, "0">:$burst_length,
+        // if set, the aiex.parameter that will override the BD's address
```
andrej: inaccurate, the aiex.parameter value will be added on top of the address as an offset, not override

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1390,6 +1392,77 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
 
+//===----------------------------------------------------------------------===//
+// Parameter ops
+//===----------------------------------------------------------------------===//
+
+def AIEX_ParameterOp: AIEX_Op<"parameter", [Symbol]> {
+  let summary = "Declare a scratchpad runtime parameter";
+  let description = [{
+    Declares a named runtime parameter that can be set from the host by writing to DDR and read by AIE cores using `aiex.read_parameter`. 
+    Parameters are communicated via the scratchpad memory mechanism (CREATE_SCRATCHPAD + UPDATE_FROM_SCRATCHPAD firmware opcodes).
```
andrej: Reword this to avoid passive voice

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1390,6 +1392,77 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
+
+    `aiex.parameter` ops are declared at **module scope** (outside any `aie.device`).  
+    The scratchpad is a single hardware resource shared by all PDIs loaded by a 
+    runtime sequence, so parameters are global to the whole module and may be 
+    referenced from any device.
+
+    Parameters can alternatively also be used to offset BD addresses by using them as the `offset_parameter` attribute in `aiex.dma_bd` and `aiex.dma_memcpy_nd`.
+    The two kinds of use are exclusive. If used this way, they cannot also be read from the cores.
```
andrej: "they" in this sentence is unclear

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1390,6 +1392,77 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
+    If used as an address offset on a BD, the parameter is a multiple of the BD's element size.
+
+    Each parameter occupies one StateTable entry (4 bytes in the scratchpad) in DDR.
+    The `--aie-lower-parameters` pass assigns the `state_table_idx` and the `kind` attribute (derived from the parameter's usage: 
+    `core` if read by a core via `aiex.read_parameter`, `addr` if used as a DMA offset via the `offset_parameter` attribute on a DMA op). 
+    Indices are unique across the entire module.
+
+    The `type` attribute specifies the data type of the parameter (bf16, f32, or an integer type up to i32). 
+    For `kind == addr`, the `type` must be `i32`.
+    The actual encoding uses a 30-bit value range due to the firmware's 2-bit masking.
```
andrej: "the firmware's 2-bit masking" assumes reader knows what this is; refer reader to appropriate lower level ops for discussion (this might be explained in the docstring for either read_parameter, npu.update_from_scratchpad or npu.create_scratchpad -- if not, add CONCISE discussion in the most fitting place)

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1390,6 +1392,77 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
+  );
+  let results = (outs);
+  let assemblyFormat = [{ $sym_name `:` $type attr-dict }];
+}
+
+def AIEX_ReadParameterOp: AIEX_Op<"read_parameter", []> {
+  let summary = "Read a scratchpad runtime parameter value on an AIE core";
+  let description = [{
+    Reads a runtime parameter previously declared with `aiex.parameter` in an `aie.core`.
+    You must first synchronize the scratchpad to the core buffers from the runtime sequence using `aiex.sync_parameters_from_host`.
```
andrej: No longer true with new automatic preamble lowering -- drop sentence

include/aie/Dialect/AIEX/IR/AIEXAttrs.td
```diff
@@ -23,4 +23,13 @@ def StateTableFunc : I32EnumAttr<"StateTableFunc",
   let cppNamespace = "::xilinx::AIEX";
 }
 
+def ParameterKindCore : I32EnumAttrCase<"Core", 0, "core">;
+def ParameterKindAddr : I32EnumAttrCase<"Addr", 1, "addr">;
+
+def ParameterKind : I32EnumAttr<"ParameterKind",
+    "Usage kind of a scratchpad runtime parameter",
```
andrej: Explain what the kinds are, as it stands this adds zero additional value over what is already clear from the enum name alone, this sort of bloat is unnecessary documentation

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -270,6 +270,42 @@ def AIELowerSetLock : Pass<"aie-lower-set-lock", "AIE::DeviceOp"> {
   ];
 }
 
+def AIELowerParameters : Pass<"aie-lower-parameters", "mlir::ModuleOp"> {
+  let summary = "Lower parameter ops and emit parameter-sync preambles";
+  let description = [{
```
andrej, line 275: Add a one-sentence high-level summary at the top before diving into the details.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1818,6 +1827,19 @@ TileOp CoreOp::getTileOp() {
...
+  if (auto attr = getEmitParameterSyncPreambleAttr())
+    return attr.getValue();
+  // Default: true iff any 'aiex.read_parameter' op is present in the core
+  // body. Type-erased check to avoid circular dependency with AIEX dialect.
+  bool found = false;
+  getBody().walk([&](Operation *op) {
+    if (op->getName().getStringRef() == "aiex.read_parameter")
+      found = true;
+  });
+  return found;
```
andrej: I do not like this forward-reference to AIEX from within AIE and the type-erase workaround making it possible. AIEX should be self-contained. Move this function as a helper into the pass that uses it, fine for it to be local there unless it's used elsewhere too. Don't forget to remove the header and any stale documentation.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -3085,12 +3107,14 @@ LogicalResult RuntimeSequenceOp::verifyBeforeMaterialization() {
               !llvm::isa<DeviceOp>(symbolDefOp) &&
               !llvm::isa<RuntimeSequenceOp>(symbolDefOp) &&
               !llvm::isa<BufferOp>(symbolDefOp) &&
-              !llvm::isa<memref::GlobalOp>(symbolDefOp)) {
+              !llvm::isa<memref::GlobalOp>(symbolDefOp) &&
+              symbolDefOp->getName().getStringRef() != "aiex.parameter") {
```
andrej: Double check if, by the time this runs, parameters aren't already lowered anyways and there shouldn't be any references to aiex.parameter any more. If so, remove this. If it's easy to make this happen, do so. The forward-reference to AIEX is ugly. Don't forget to update the error message below as well (line 3116-3117).

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -3119,6 +3143,27 @@ LogicalResult RuntimeSequenceOp::verifyBeforeMaterialization() {
   return success();
 }
 
+bool RuntimeSequenceOp::shouldEmitParameterSyncPreamble() {
+  if (auto attr = getEmitParameterSyncPreambleAttr())
+    return attr.getValue();
```
andrej: Throughout remainder of changeset: Always wrap one-line blocks in curly braces. Just because you can avoid them doesn't mean you should

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -3119,6 +3143,27 @@ LogicalResult RuntimeSequenceOp::verifyBeforeMaterialization() {
...
+    return false;
+  bool found = false;
+  device.walk([&](Operation *op) {
+    if (found)
+      return;
+    if (op->getName().getStringRef() == "aiex.read_parameter")
+      found = true;
+    else if (op->hasAttr("offset_parameter"))
+      found = true;
+  });
```
andrej: Again, forward reference to AIEX is ugly. Can we move this function as a helper into the pass that uses it and remove it here (don't forget the header)?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -811,6 +812,28 @@ LogicalResult AIEX::NpuCreateScratchpadOp::verify() {
     return emitOpError("size (")
            << getSize() << " bytes) exceeds maximum scratchpad size of "
            << kMaxScratchpadSizeBytes << " bytes.";
+
+  // At most one create_scratchpad may appear per runtime sequence. Walk the
+  // parent RuntimeSequenceOp to check; only report from the duplicate (i.e.
+  // the op that is NOT the first occurrence) to avoid emitting the same error
+  // twice.
+  if (auto runtimeSeq =
+          getOperation()->getParentOfType<AIE::RuntimeSequenceOp>()) {
```
andrej: odd indentation, also don't like wrapping the whole logic in an if block. Assign to a variable instead, exit early if variable is NULL/no runtime sequence op parent found.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -1117,6 +1140,27 @@ AIE::DeviceOp AIEX::ConfigureOp::getReferencedDeviceOp() {
...
+LogicalResult AIEX::ReadParameterOp::verify() {
+  auto device = (*this)->getParentOfType<AIE::DeviceOp>();
+  if (!device) {
+    return emitOpError("must be inside an aie.device");
+  }
+  if (!(*this)->getParentOfType<AIE::CoreOp>()) {
+    return emitOpError("must be inside an aie.core");
+  }
+  auto module = (*this)->getParentOfType<ModuleOp>();
+  if (!module || !module.lookupSymbol<AIEX::ParameterOp>(getParameter())) {
```
andrej: name variable moduleOp rather than module to avoid conflict with C++ keyword

lib/Dialect/AIEX/Transforms/AIECtrlPacketToDma.cpp
```diff
@@ -84,7 +84,7 @@ struct AIECtrlPacketToDmaPass
       IRMapping mapping;
 
       auto newSeq =
-          AIE::RuntimeSequenceOp::create(builder, loc, f.getSymNameAttr());
+          AIE::RuntimeSequenceOp::create(builder, loc, f.getSymName(), BoolAttr{});
```
andrej: Can we add a new default builder for RuntimeSequenceOp / make the emit_parameter_preamble a `DefaultValuedAttr` so we can avoid changing all the creation sites like this. i.e., revert this change and make it work by adding a properly overloaded `create` --this might be possible "for free" without adding code by just making the attribute a `DefaultValuedAttr`

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -242,6 +242,14 @@ struct AIEDMATasksToNPUPass
...
+      // If this BD has an offset_parameter, emit update_from_scratchpad to add
+      // the runtime offset to the BD address register.
+      if (bd_op.getOffsetParameterAttr()) {
+        auto bufType = llvm::cast<BaseMemRefType>(bd_op.getBuffer().getType());
+        if (failed(emitUpdateBdAddressFromOffsetParameter(
+                builder, bd_op, bufType, register_addr)))
+          return failure();
+      }
     } else if (AIE::BufferOp buffer =
                    llvm::dyn_cast<AIE::BufferOp>(buf.getDefiningOp())) {
```
andrej: Why is the `offsetParameter` handled here as mutually exclusive with the BufferOp branch? What if I want my BD to use a buffer's address as the base, but add an offset on top of that address as a parameter? Verify that this is sound and if not fix

andrej:
After it came up in the compiler/data movement meeting, I investigated whether it is allowed to call `create_scratchpad` multiple times. Turns out it is possible, and the firmware executes it each time, copying host main memory -> firmware memory. Even with multiple instructions, XRT creates only _one_ scratchpad, of the size specified in the _first_ `create_scratchpad` instruction. (Different-sized `create_scratchpad` operations therefore would be buggy and should be avoided.)

This means that, in theory, we could re-write different parameter values during a single XRT run and the firmware will copy the new value if a `create_scratchpad` instruction is present. However, there's no real way to make use of this because we lack a sync primitive: After some experimenting, I couldn't come up with a way to sync between CPU and NPU within a single XRT run. I think in some ways, the XRT run _is_ the synchronization barrier between CPU and NPU, so maybe this is not surprising. 

So, while you could change parameters while the XRT run runs, it would always be racy without a proper way of letting the NPU runtime sequence know that the CPU is done writing the new parameters. (There is a way to make the CPU wait for the NPU by spinning on a value in memory, but, again, I couldn't come up with a way to do the sync in the opposite direction.)

As such, I've decided to restrict runtime sequences to contain at most one `create_scratchpad` op, with an explicit check for this added to the verifier. It just doesn't make much sense to have multiple without being able to sync, and it is way more likely a user error that would be better to flag early than a legitimate use case.

For the case when a user inlines multiple runtime sequences into each other (`aiex.run`), the "materialize runtime sequences" pass now hoists a single `create_scratchpad` op to the top of the runtime sequence and drops all others (after making sure they match in size, erroring otherwise -- scratchpad size is a module-level property as there can only be one scratchpad per run).

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+// `emit_parameter_sync_preamble` attribute on aie.core and
+// aie.runtime_sequence.
+//
+// State table indices are assigned globally across the entire module: every
+// aiex.parameter in any aie.device gets a unique index in [0, 32).  This is
+// necessary because the scratchpad is a single hardware resource shared by all
+// PDIs loaded by a runtime sequence.
+//
+// A single params.txt file is emitted describing every parameter in the module
+// (name, global state_table_idx, type, and kind: "core" or "addr").
```
andrej: Merge this with the pass description in AIEXPasses.td. Remove the comment here, make sure the docs in AIEXPasses.td contain all the info here if they don't already. Some duplication here

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+
+    DenseMap<std::pair<StringRef, Operation *>, BufferOp> seen;
+
+    device.walk([&](ReadParameterOp readOp) {
+      auto coreOp = readOp->getParentOfType<CoreOp>();
+      TileOp tile = coreOp.getTileOp();
+      StringRef paramName = readOp.getParameter();
+      auto key = std::make_pair(paramName, tile.getOperation());
+
+      if (!seen.count(key)) {
```
andrej: Prefer less indentation: Check for the opposite condition, and if hit, return early. Then unident the whole block below.

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+      builder.setInsertionPoint(readOp);
+      Value c0 = builder.create<arith::ConstantIndexOp>(readOp.getLoc(), 0);
+      Value raw = builder.create<memref::LoadOp>(readOp.getLoc(), buf, c0);
+      Value c2 = builder.create<arith::ConstantOp>(
+          readOp.getLoc(), builder.getI32IntegerAttr(2));
+      Value decoded = builder.create<arith::ShRUIOp>(readOp.getLoc(), raw, c2);
+
+      Type resultType = readOp.getResult().getType();
+      Value result = decoded;
+      if (resultType != builder.getI32Type()) {
```
andrej: This check appears to be unnecessary as it is mutually exclusive with the contained if conditions. Remove and unindent contained block

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+  // create an aie.lock (no lockID; AIEAssignLockIDs will assign one later) and
+  // insert aie.use_lock(Acquire, 1) at the top of the core body.  Returns the
+  // list of lock values created (one per qualifying core, in walk order).
+  SmallVector<Value> emitCorePreambles(DeviceOp device, OpBuilder &builder) {
+    SmallVector<Value> syncLocks;
+    device.walk([&](CoreOp coreOp) {
+      if (!coreOp.shouldEmitParameterSyncPreamble())
+        return;
+
+      // Create aie.lock on the core's tile with no lockID (will be assigned).
```
andrej: stop obsessing over the fact that you're not giving it a lock ID

andrej: these comments restate the code 1:1 without adding value, remove

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+      TileOp tile = coreOp.getTileOp();
+      builder.setInsertionPointAfter(tile);
+      auto lockOp = builder.create<LockOp>(
+          coreOp.getLoc(), builder.getIndexType(), tile.getResult(),
+          /*lockID=*/IntegerAttr{}, builder.getI32IntegerAttr(0),
+          /*sym_name=*/StringAttr{});
+      syncLocks.push_back(lockOp.getResult());
+
+      // Insert use_lock(lock, Acquire, 1) at the very start of the core body.
```
andrej: remove; generally remove any comment that doesn't explain anything on top of what isn't already obvious from code

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+                        LockAction::Acquire, 1);
+
+      // Mark as done so the pass is idempotent.
+      coreOp.setEmitParameterSyncPreambleAttr(builder.getBoolAttr(false));
+    });
+    return syncLocks;
+  }
+
+  // For each runtime sequence in `device` with shouldEmitParameterSyncPreamble
+  // ()==true, insert at the beginning of the sequence body:
```
andrej: weird line break before ()

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+  LogicalResult emitParamsFile(ArrayRef<ParameterOp> allParams) {
+    if (outputParamsFile.empty())
+      return success();
+
+    std::error_code ec;
+    llvm::raw_fd_ostream out(outputParamsFile, ec);
+    if (ec)
+      return emitError(UnknownLoc::get(&getContext()),
+                       "failed to open params output file '")
+             << outputParamsFile << "': " << ec.message();
```
andrej: eww ... add braces

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+        NpuUpdateFromScratchpadOp::create(
+            builder, loc, stateIdx, StateTableFunc::Incr,
+            /*func_arg=*/static_cast<uint32_t>(0),
+            /*address=*/static_cast<uint32_t>(0), bufRef,
+            /*column=*/nullptr, /*row=*/nullptr);
+      }
+
+      for (Value lock : syncLocks)
+        SetLockOp::create(builder, loc, lock, builder.getI32IntegerAttr(1));
+
```
andrej: add braces everywhere

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+        [&](ReadParameterOp op) { usedAsCore[op.getParameter()] = true; });
+    auto markAddr = [&](Operation *op, FlatSymbolRefAttr ref) {
+      if (ref)
+        usedAsAddr[ref.getValue()] = true;
+    };
+    module.walk([&](NpuDmaMemcpyNdOp op) {
+      markAddr(op, op.getOffsetParameterAttr());
+    });
+    module.walk(
+        [&](AIE::DMABDOp op) { markAddr(op, op.getOffsetParameterAttr()); });
```
andrej: Use the same indentation as for the three lines above here

lib/Dialect/AIEX/Transforms/AIELowerParameters.cpp
```diff
@@ -0,0 +1,355 @@
...
+          p.getKind().value() == ParameterKind::Addr ? "addr" : "core";
+      out << p.getSymName() << " "
+          << static_cast<unsigned>(p.getStateTableIdx().value()) << " "
+          << typeStr << " " << kindStr << "\n";
+    }
+    return success();
+  }
+
+  void runOnOperation() override {
+    ModuleOp module = getOperation();
```
andrej: Rename to `moduleOp`  to avoid conflict with `module` keyword

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -480,13 +480,39 @@ struct InlineRuntimeCallsPattern : RewritePattern {
...
+    // execute once before anything else.
     rewriter.setInsertionPoint(runOp);
     mlir::OpBuilder::InsertPoint clonedOpInsertionPoint =
         rewriter.saveInsertionPoint();
     for (Operation &op : calleeBody.getOps()) {
+      bool shouldHoist =
+          llvm::isa<NpuCreateScratchpadOp>(&op);
+      if (shouldHoist) {
+        // Clone into the caller's RuntimeSequenceOp body before any
+        // non-hoisted ops. No symbol definitions need inlining for these ops.
```
andrej: Remove comment, restates code, confuses by referring to multiple "ops" when currently it's only the one that doesn't have a symbol ref, so the symbol inlining is further confusing. just remove

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -643,6 +669,39 @@ struct AIEMaterializeRuntimeSequencesPass
...
+                  << " != " << firstScratchpad.getSize();
+              return signalPassFailure();
+            }
+            duplicates.push_back(scratchpadOp);
+          }
+        }
+        for (NpuCreateScratchpadOp dup : duplicates)
+          dup.erase();
+      }
+
```
andrej: That's a whole lot of logic to just dump into the main function. Factor out into a helper. Consider if a GreedyPatternRewriter like the other components of this pass use would be more fitting (potentially parallelizable, more idiomatic). Don't put everything in a "duplicates" intermediate, just erase as you see them in the loop above. Unindent the else clause and "continue" early if you see if(!firstScratchpad). double-check if this logic could be added to the "InlineRuntimeCallsPattern"

lib/Dialect/AIEX/Utils/AIEUtils.cpp
```diff
@@ -155,3 +156,46 @@ memref::GlobalOp AIEX::getOrCreateDataMemref(OpBuilder &builder,
...
+    OpBuilder &builder, Operation *bdOp, BaseMemRefType bufType,
+    uint64_t registerAddr) {
+  auto paramRef = bdOp->getAttrOfType<FlatSymbolRefAttr>("offset_parameter");
+  if (!paramRef)
+    return bdOp->emitOpError(
+        "emitUpdateBdAddressFromOffsetParameter called without "
+        "offset_parameter attribute.");
+
+  auto module = bdOp->getParentOfType<ModuleOp>();
+  if (!module)
```
andrej: Rename to `moduleOp` to avoid conflict with C++ keyword

lib/Dialect/AIEX/Utils/AIEUtils.cpp
```diff
@@ -155,3 +156,46 @@ memref::GlobalOp AIEX::getOrCreateDataMemref(OpBuilder &builder,
...
 }
+
+LogicalResult AIEX::emitUpdateBdAddressFromOffsetParameter(
+    OpBuilder &builder, Operation *bdOp, BaseMemRefType bufType,
+    uint64_t registerAddr) {
+  auto paramRef = bdOp->getAttrOfType<FlatSymbolRefAttr>("offset_parameter");
+  if (!paramRef)
+    return bdOp->emitOpError(
+        "emitUpdateBdAddressFromOffsetParameter called without "
+        "offset_parameter attribute.");
```
andrej: Assume call sites know what they're doing. This should be at most an assert, not a user-facing error (bug in code, not in input MLIR)

lib/Dialect/AIEX/Utils/AIEUtils.cpp
```diff
@@ -155,3 +156,46 @@ memref::GlobalOp AIEX::getOrCreateDataMemref(OpBuilder &builder,
...
+    err.attachNote(paramOp.getLoc()) << "Parameter declared here.";
+    return err;
+  }
+
+  uint8_t stateIdx = static_cast<uint8_t>(paramOp.getStateTableIdx().value());
+  uint32_t elemBytes = bufType.getElementTypeBitWidth() / 8;
+  // Use func=mul with func_arg=elemBytes so the firmware computes
+  // StateTable[idx] * elemBytes = byte offset, added into the BD address
+  // register. UpdateReg only reads from the state table (it never writes
+  // back), so this is safe to repeat across runtime sequence invocations.
```
andrej: Remove second part of the comment, confusing

python/dialects/aiex.py
```diff
@@ -345,3 +351,25 @@ def dma_start_task(*args: DMAConfigureTaskForOp):
...
+
+_orig_read_parameter = read_parameter
+
+
+def read_parameter(name: str, result_type: Type) -> _orig_read_parameter:
+    """Read a runtime parameter inside an ``aie.core`` body.
+
+    Args:
+        name: The ``@sym_name`` of the ``aiex.parameter`` declaration.
+        result_type: The MLIR scalar type of the result (e.g. ``T.bf16()``, ``T.i32()``).
```
andrej: Prefer single backticks ` throughout documentation

python/iron/runtime/runtime.py
```diff
@@ -291,6 +319,15 @@ def set_barrier(self, barrier: WorkerRuntimeBarrier, value: int):
...
 
+    def sync_parameters(self):
+        """Deprecated: parameter-sync preamble is now emitted automatically.
+
+        This method is a no-op. The ``--aie-lower-parameters`` pass
+        automatically inserts the necessary lock and scratchpad-sync ops into
+        every core and runtime sequence that uses parameters.
+        """
+        pass  # preamble is now emitted by --aie-lower-parameters
+
```
andrej, line 330: remove the method

runtime_lib/test_lib/parameter_scratchpad.h
```diff
@@ -0,0 +1,169 @@
...
+  ParameterScratchpad(uint32_t *buffer, const std::string &paramsPath)
+      : boMap(buffer) {
+    parseParams(paramsPath);
+    clear();
+  }
+
+  /// Write raw bytes (up to 4) by name, interpreted as a little-endian
+  /// uint32.  The bits are left-shifted by 2 (firmware requirement) and
+  /// delta-encoded against the previous write.
+  /// For addr-kind parameters, the value is written raw (no shift, no delta).
```
andrej: Are we using this triple forward-slash convention elsewhere in the runtime lib or codebase in general? If not, switch to use the same style as elsewhere throughout

python/iron/runtime/runtime.py
```diff
@@ -63,6 +64,7 @@ def __init__(
         self._tasks: list[RuntimeTask] = []
         self._fifos = set()
         self._workers = []
+        self._parameters: list[Parameter] = []
```
hunhoffe: "parameter" feels pretty vague. Is there anything else we can call this that's a little more specific/description?

python/iron/runtime/runtime.py
```diff
@@ -391,3 +428,16 @@ def finish_task_group(tg, task_group_actions):
...
+
+class _SyncParametersTask(Resolvable):
+    """Emits ``aiex.sync_parameters_from_host`` during runtime sequence resolution."""
+
+    def resolve(
+        self,
+        loc: ir.Location | None = None,
+        ip: ir.InsertionPoint | None = None,
+    ) -> None:
+        from ...dialects.aiex import sync_parameters_from_host
```
hunhoffe: Does this import need to be here vs at the module level? If it is, no worries, I just try to keep imports all in one place per file unless there's a reason not to.

python/iron/parameter.py
```diff
@@ -0,0 +1,90 @@
+# parameter.py -*- Python -*-
+#
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
+"""Parameter: a named runtime value set from the host and read by Workers."""
```
hunhoffe: Still trying to think of a better name... H2WParam? e.g., Host-to-worker? I haven't come up with something good, but I think parameter is too vague.

python/utils/hostruntime/xrtruntime/parameter_scratchpad.py
```diff
@@ -0,0 +1,72 @@
...
+from pathlib import Path
+
+import pyxrt
+
+from aie._mlir_libs._parameter_scratchpad import (
+    ParameterScratchpad as _ParameterScratchpadImpl,
+)
+
+
+def _to_bytes(value) -> bytes:
```
hunhoffe, line 37: Are you sure there's not a library function we can use here?

python/utils/hostruntime/xrtruntime/parameter_scratchpad.py
```diff
@@ -0,0 +1,72 @@
...
+    params = ParameterScratchpad(run, "params.txt")
+    params.write("seq_len", 42)
+    params.sync()
+    run.start()
+"""
+
+import struct
+from pathlib import Path
+
+import pyxrt
```
hunhoffe, line 30: This file needs to go within the XRT runtime; we avoid XRT-specific imports in all other places so that you can still generate MLIR/import portions of IRON without XRT installed on the host.

test/npu-xrt/scratchpad_addr_offset_python/aie_design.py
```diff
@@ -0,0 +1,86 @@
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
```
hunhoffe: Would it make sense to include this new functionality in one of the programming examples/programming guide? That could easily be a follow-on PR too.

test/npu-xrt/scratchpad_addr_offset_python/test.py
```diff
@@ -0,0 +1,91 @@
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
```
hunhoffe: To fully test the python environment, I tend to put this type of test in test/python/npu-xrt -- tests in that directory are run with pytest through lit.

However, the important thing is that you have a test that's functional; I can see the argument for keeping the tests together so feel free to keep it where it is if that is your preference.

test/python/npu-xrt/scratchpad_addr_offset/test.py
```diff
@@ -0,0 +1,91 @@
...
+from aie.utils.parameter_scratchpad import ParameterScratchpad
+
+
+def main():
+    N_INPUT = 32
+    N_OUTPUT = 8
+
+    device = pyxrt.device(0)
+    elf = pyxrt.elf("aie.elf")
+    context = pyxrt.hw_context(device, elf)
```
hunhoffe, line 39: Is there a reason why you need to use pyxrt directly? Instead of the runtime wrappers? If the pyxrt interface changed in the future, I'm trying to reduce the number of places we make assumptions on the direct API.

test/npu-xrt/scratchpad_params_python/aie_design.py
```diff
@@ -0,0 +1,78 @@
+# (c) Copyright 2025 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+# IRON design: generates the MLIR for the scratchpad parameter test.
+#
+# REQUIRES: dont_run
```
hunhoffe: Don't run??

test/python/npu-xrt/scratchpad_params/test.py
```diff
@@ -0,0 +1,94 @@
...
+    f32_bytes = b"\x00\x00" + raw
+    return struct.unpack("<f", f32_bytes)[0]
+
+
+def main():
+    FOO_1, BAR_1 = bfloat16(3.0), bfloat16(4.0)
+    FOO_2, BAR_2 = bfloat16(2.0), bfloat16(5.0)
+
+    device = pyxrt.device(0)
+    elf = pyxrt.elf("aie.elf")
```
hunhoffe, line 36: Same comment about using pyxrt directly.

hunhoffe (commented):
Good work! I have a couple comments which I think are worth reviewing, but the main thing blocking my approval is that I'm trying to be very sparse/careful/disciplined with where we import pyxrt in the python code. I think you need to move where some of your code sits to avoid adding pyxrt dependencies where they don't need to be.

test/npu-xrt/scratchpad_params_python/aie_design.py
andrej: dont_run just for the aie_design.py file; the test that actually runs and builds it gets picked up in test.py. This is required because lit picks up all .py files as tests

python/iron/runtime/runtime.py
andrej: no, just something Claude loves to do but good catch!

python/utils/hostruntime/xrtruntime/parameter_scratchpad.py
andrej: Not really, since this makes a couple assumptions that are specific to this use case, specifically that integers are 32-bit little-endian, and that we want to be able to handle Python native integers as well as numpy values. The alternative would be to requiere users to always call with `np.uint32(...)` as the argument, but I do like the convenience of being able to just pass in an integer (though I don't feel strongly about it)

test/npu-xrt/scratchpad_addr_offset_python/aie_design.py
andrej: yes I already was thinking an example in the programming guide will probably be good, i'll make a note of it

test/npu-xrt/scratchpad_addr_offset_python/test.py
andrej: moved

test/python/npu-xrt/scratchpad_addr_offset/test.py
andrej: switched to XRTTensor and got device out of hostruntime; full ELF flow still requires some direct interaction with pyxrt

python/iron/runtime/runtime.py
andrej: renamed to scratchpad parameter

test/python/npu-xrt/scratchpad_addr_offset/test.py
```diff
@@ -0,0 +1,91 @@
+# (c) Copyright 2026 Advanced Micro Devices, Inc.
```
hunhoffe, line 1: All of the tests in this directory are run with pytest. I'd like to keep it that way. If you want to merge this, I'd be ok with you addressing this in a follow up PR but I do want to keep things tidy

hunhoffe (approved):
Sorry to have more test nits; I'm ok if you merge as-is and do a follow up later.

PR #3064, opened 2026-05-11
[AIEPlacer] Extend tile capacity check to search both directions

### LINKED ISSUE #3049: AIEPlacer: findTileWithCapacity only searches rightward from target column
## Description

In `lib/Dialect/AIE/Transforms/AIEPlacer.cpp`, `findTileWithCapacity` (line ~547) only searches columns to the right of `targetCol`:

```cpp
while (targetCol + offset < maxCol) {
    // try targetCol + offset
    offset++;
}
```

If the target column and all higher columns are exhausted, placement fails even if lower columns have capacity. A bidirectional search (expanding outward from `targetCol`) would provide better placement quality.

## Suggested Fix

Replace the unidirectional search with a bidirectional spiral:

```cpp
int offset = 0;
while (true) {
    if (targetCol + offset < maxCol && tryPlace(targetCol + offset))
        return success;
    if (offset > 0 && targetCol - offset >= minCol && tryPlace(targetCol - offset))
        return success;
    offset++;
    if (targetCol + offset >= maxCol && targetCol - offset < minCol)
        break;  // exhausted all columns
}
```

## Context

Found during PR review of `dynamic-runtime-sequences` branch. This is a design improvement, not a correctness bug — the current behavior is intentional for the sequential placer but could be surprising when columns to the left have available capacity.

## Files

- `lib/Dialect/AIE/Transforms/AIEPlacer.cpp:547`
- `include/aie/Dialect/AIE/Transforms/AIEPlacer.h`

yenjames:
## Summary
Design improvement to allow `findTileWithCapacity` to search both leftwards and rightwards, preferring the nearest column in either direction.

## Changes
`lib/Dialect/AIE/Transforms/AIEPlacer.cpp `                                                                                        
  - `findTileWithCapacity`: replaced the for (offset = 0; offset < maxCol; ++offset) rightward-only loop with a bidirectional loop
  that iterates over {+offset, -offset} at each step, skipping duplicates and out-of-bounds columns                                 
                                                                                                   
`test/place-tiles/sequential_placer/test_place_objectfifo.mlir `                                                               
  - Added `@channel_overflow_leftward`: 3 cores explicitly placed to column 3. 3 unplaced shims: first two merge to (3, 0) (with centroid logic), third overflows to (2, 0) since no column=4.

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEPlacer.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEPlacer.cpp
@@ -720,22 +720,26 @@ std::optional<TileID> SequentialPlacer::findTileWithCapacity(
     int requiredOutputChannels, AIETileType requestedType) {
   int maxCol = targetModel->columns();
 
-  // Search columns rightward
-  for (int offset = 0; offset < maxCol; ++offset) {
-    int searchCol = targetCol + offset;
-    if (searchCol >= maxCol)
-      continue;
-
-    for (auto &tile : tiles) {
-      AIETileType tileType = targetModel->getTileType(tile.col, tile.row);
-      if (tileType != requestedType)
+  // Search columns outward from targetCol bidirectionally
+  for (int offset = 0; targetCol + offset < maxCol || targetCol - offset >= 0;
+       ++offset) {
+    for (int dir : {1, -1}) {
+      int searchCol = targetCol + dir * offset;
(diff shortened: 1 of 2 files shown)


hunhoffe (approved):
Looks good once CI tests pass! Make sure to mark which issue this will close when it is merged.

erwei-xilinx (approved):

PR #3067, opened 2026-05-12
Fix incorrect matmul output for small m on AIE2

### LINKED ISSUE #2793: GEMM produces incorrect output for certain MNK tilings on AIE2 (XDNA1) with latest wheels
Hi, I'm encountering incorrect GEMM results when using the latest wheels build of mlir-aie targeting AIE2 (XDNA1). The same example and configurations worked correctly with release v1.0.

I'm following setup instructions in commit `e952e58d8c1ec7c2759696318731527de8dcc0b5`. I install `mlir-aie` with
```bash
python3 -m pip install mlir_aie==0.0.1.2025122304+e952e58 -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/latest-wheels-2/
```
and `llvm-aie` wheel with 
```bash
python3 -m pip install llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
```

When running `programming_examples/basic/matrix_multiplication/single_core` on `RyzenAI-npu1`, I observe incorrect computation results for certain MNK tiling configurations.

For example:
- when `m = 16, k = 64, n = 16`, the output matrix is all zeros

the same example and tiling configurations produced correct results when I was using release v1.0. I'm trying to understand whether this is caused by a misconfiguration on my side or any recent compiler change.

Thanks in advance!

hunhoffe:
Closes #2793

Bug: For certain tilings (e.g. m=16, k=64, n=16 i16/i32) the vectorized matmul on Phoenix/XDNA1 returns an all-zero output. Worked in                                                                                       
                                                                                                                                     
  Root cause                                                                                                                         
  - aie_kernels/aie2/mm.cc uses AIE_LOOP_MIN_ITERATION_COUNT(N) on the outer z-loop of matmul_vectorized_{2x2_mmul, 4x2_mmul, 4x4}.
  - The hint values (4, 4, 2) were sized for larger tilings. Smaller m (still satisfying the kernel's static_asserts) makes the loop run fewer iterations than promised.                              
  - The hint expands (under peano) to #pragma clang loop min_iteration_count(N), which requires a literal integer; a                 
  template-dependent constexpr like rowA / 2 is silently mishandled.
                                                                                                                                     
  Fix                                                       
  - Wrap each outer-loop body in a [[gnu::always_inline]] lambda.                                                                    
  - Dispatch on outer_iters (a constexpr unsigned derived from rowA) via if constexpr into one of three branches with literal hints (4), (2), (1). Each branch's hint matches its actual iteration count.                                                            
  - Lambda inlines fully — confirmed by llvm-objdump: the i16/i32 m=32 hot path keeps the same symbol table and basic-block layout as before.

### DIFF
--- a/aie_kernels/aie2/mm.cc
+++ b/aie_kernels/aie2/mm.cc
@@ -89,9 +89,11 @@ static inline void matmul_vectorized_2x2_mmul(const T_in *__restrict pA,
 
   event0();
 
-  AIE_PREPARE_FOR_PIPELINING
-  AIE_LOOP_MIN_ITERATION_COUNT(4)
-  for (unsigned z = 0; z < rowA; z += 2) {
+  // Outer-loop body factored into a lambda so the same code can be wrapped
+  // by three differently-parameterised loops below. Per-instantiation iter
+  // count must match the actual count: the clang loop pragma takes a literal
+  // and silently misbehaves on template-dependent expressions.
+  auto outer_body = [&](unsigned z) [[gnu::always_inline]] {
     T_out *__restrict pC1;
     T_out *__restrict pC2;
     if constexpr (c_row_maj) {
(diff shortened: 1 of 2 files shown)


andrej (approved):
Very nice find, thank you! Also good intel that these pragmas don't accept constexprs. I might have run into this before without noticing.

PR #3071, opened 2026-05-13
wheels: PEP 440-compliant versioning for mlir-aie (PyPI prep)

hunhoffe:
## Summary                                                                                                                                                  
Prep `mlir_aie_wheels` for PyPI uploads by fixing the version string format and exposing build metadata on the installed package.                           

## Changes                                                                                                                                                  
- **Versioning**: `get_version()` now derives the dev counter from `git describe`. Tagged builds: clean `1.2.3`. Between tags: `1.2.4.devN`. `+g<sha>` local segment is appended only when `AIE_WHEEL_KEEP_LOCAL=1` (set by GH Release upload paths; left unset for future PyPI uploads). The existing `AIE_WHEEL_VERSION` tag override is unchanged.
- **No-rtti as a distinct PyPI package**: `mlir-aie-no-rtti` instead of a non-PEP-440 `-no_rtti` version suffix. Matches the existing `mlir`/`mlir-no-rtti` convention.                                                                                                                                                 
- **Runtime introspection**: `aie.__version__`, `aie.__commit__`, `aie.__build_date__` (`_version.py` written + re-exported from `aie/__init__.py`).
                                                                                                                                                              
## Verified                                               
- `twine check` passes on the built sdist; `packaging.Version()` parses every output form.                                                                  
- Both `ENABLE_RTTI=0/1` produce the expected package names.                                                                                                
                                                                                                                                                              
## Out of scope                                                                                                                                             
Same fix for `utils/mlir_wheels/` and the actual PyPI publish workflow — separate PRs.      

## Update
Found via empirical wheel validation that CI's actions/checkout@v6 defaults to a depth-1 clone with no tags, so git describe returned nothing and get_version() fell into the 0.0.0.dev{count} fallback (with count == 1) instead of 1.3.2.dev<distance>. Latest commit:                                                                                                                                 
- Sets fetch-depth: 0 + fetch-tags: true on the wheel-build checkouts in buildRyzenWheels.yml.
- Excludes pre-release tags (v*-*) from the git describe matcher so e.g. a future v1.4.0-rc1 tag doesn't poison the version computation.

 Locally verified: HEAD-style builds now produce 1.3.2.dev49+g2030517 (PEP 440 valid; sorts correctly before the eventual 1.3.2 release).

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -96,6 +96,8 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: "true"
+          fetch-depth: 0
+          fetch-tags: true
 
       - uses: uraimo/run-on-arch-action@v3
         name: Build mlir-aie
@@ -248,7 +250,9 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: "true"
-  
+          fetch-depth: 0
+          fetch-tags: true
+
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -182,6 +182,7 @@ jobs:
             fi
             export AIE_VITIS_COMPONENTS='AIE2;AIE2P'
             export DATETIME=$(date +"%Y%m%d%H")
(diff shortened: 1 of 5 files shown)


jgmelber (approved):

thomthehound:
This is super exciting! Great work, Erika!

I am so jazzed for that future day (even if it is some time off) when users can just `pip install mlir-aie`!

PR #3072, opened 2026-05-13
wheels: harden supply chain, slim wheels, rehearse PyPI publish

hunhoffe:
Originally this PR just hash-pinned the mlir_aie wheel's build deps. It grew to cover three related threads that all needed to land before moving towards PyPI publication.

**Supply chain.** Hash-pin every pip dep we touch (mlir_aie wheel, mlir distro wheel, `python/requirements_dev.txt`). SHA-pin every GitHub Action with a `# vX.Y.Z` comment so dependabot keeps maintaining them. Default-deny `permissions: contents: read` on workflows; widen at job level only where needed. Add `actions/attest-build-provenance` on every wheel so `gh attestation verify` works end-to-end, even before any PyPI upload exists.

**Wheel hygiene.** Ship `LICENSE` (it was missing), set `project_urls` / `classifiers` so the PyPI page renders properly, fix author/maintainer attribution on the mlir distro wheel. Drop ~70 MB of dev-only content from the wheel: `aie-lsp-server` and `aie-visualize` (editor tools), static `.a` archives, C++ headers, cmake configs, plus a handful of CMake-leaked artifacts that shouldn't have been shipping at all (`src/` duplicates, `lib/objects-Release/` .o files, stray `__pycache__`).

**PyPI rehearsal.** Each build now strips a copy of the wheel, runs `twine check --strict`, and writes a size + cap-verdict table to the run summary. No upload happens. The wheel still goes to GitHub releases unstripped as before — the rehearsal just gives us measured data for the eventual size of wheels we'd like to publish to PyPI.

A few things worth flagging:
- `aie-lsp-server` / `aie-visualize` still build by default for local dev (new `AIE_BUILD_LSP_SERVER` / `AIE_BUILD_VISUALIZE` CMake options, both `ON`); only the wheel build passes `OFF`.
- After everything here the Linux wheel is ~205 MB unstripped / ~165 MB stripped; Windows is ~140 MB and already at its floor (MSVC builds don't embed debug info). Both are still above PyPI's 100 MB default cap.
- Deferred: actual PyPI publish workflow, still currently above PyPI size cap.


andrej (approved):

erwei-xilinx:
Air is deeply coupled to AIE C++ headers and links against libAIE.a. The new slim wheel dropped both headers and static libs. May I ask if there is any chance to bring them back?

What's gone (both needed by air's C++ build):

libAIE.a — the MLIR AIE dialect static archive air links against. The RECORD's only 4 .a files are runtime test libs (libtest_lib.a, etc.), not the dialect lib.
The MLIR dialect headers air #includes — aie/Dialect/AIE/IR/AIEDialect.h, AIETargetModel.h, AIEPlacer.h — all absent.

Cc: @hunhoffe @andrej

Issue #2347, opened 2025-05-29
Feature to detect insufficient stack size


joeldushouyu:
Hello all,
I am not sure if this is the best place to raise this issue, let me know if I should raise in other repo or not.

Recently, we encountered an issue where the result of the computation is wrong due to stack size overflow(I believe the stack size is defaulted to 1024 bytes).  Fortunately, there was a method for us to enlarge the stack size, as shown in this [pr](https://github.com/Xilinx/mlir-aie/pull/2280).

Although setting a larger stack fixes the issue for now, I was wondering if it will be worth discussing having some type of check or even automatically changing the stack size in the future? I find a similar issue for [iree-amd-aie](https://github.com/nod-ai/iree-amd-aie/issues/1281)

While using Peano compiler(probably also has a similar feature in xchess?), I was able to add the "-fstack-size-section" flag during the compile stage, and the stack usage of each function by doing 
```bash
llvm-readobj --stack-size xxx.o
```
to see the stack size usage for each function, for example, I got something below
```yaml
StackSizes [
  SectionEntry {
    Address: 0x0
    Size: 16
    FunctionName: main
  }
  SectionEntry {
    Address: 0x20
    Size: 32
    FunctionName: foo
  }
]
```
Admittedly, the stack size for main is not accurate, because the actual stack size should be (16+ 32), given that main() calls foo(). In other words, I think it only gave the stack size of each function, not the maximum stack depth. I believe the LLVM and MLIR can produce a Callgraph that shows the dependency graph of how the functions is being called.

After doing some research, I find that the maximum-stack-depth is almost impossible to implement in normal general compiler due to 
1. External Library and indirect function calls
2. Unbounded Recursion calls, for example,e calculating Fibonacci number recursively.

For point 1, I think it should not be a problem?
Because from my experience, most external library function calls I used are [user Guide API](https://www.xilinx.com/htmldocs/xilinx2024_2/aiengine_api/aie_api/doc/index.html), which should be known by the compiler?  Moreover, I haven't encountered any issues where I have to use indirect function calls instead of explicit function calls.

Regarding point 2, I haven't met any use case where I need **UnBounded Recusion** for doing computation in Core. Taking a step back, maybe there are cases where the use of **UnBounded Recusion** is justified, but I imagine that do not happen frequently?

### Ideas:
[...]

PR #3075, opened 2026-05-15
Peano lit enablement No1

thomthehound:
This updates the lit configuration files (and affected test files) so that AIE/NPU tests can run with the open-source Peano backend. These changes are compatible with (and were tested on) Windows, but are *not* exclusive to it. Running the tests with the Peano backend should now be possible on Linux as well.

#### Highlights:
- Detect Peano early enough for Ryzen AI/XRT feature gating, so NPU tests are not incorrectly disabled just because Vitis/AIETOOLS components are absent.
- Gate `npu-xrt` tests on lit features (`peano`, `chess`, `aietools_aie2`, `aietools_aie2p`) instead of raw Vitis state.
- Add `%backend_flags` substitution that keeps the existing Chess path when Chess is available, but automatically expands to `--no-xchesscc --no-xbridge` when only Peano is present.
- Add cross-platform host compiler/link substitutions, replacing Linux-only `clang ... -lrt -lstdc++` command lines with `%host_clang` and `%host_link_flags`.
- Harden Windows execution by avoiding POSIX-isms such as `|&`, stdin `echo` pipelines, Makefile-based tests, and extensionless executables.
- Fix CMake configuration needed by the lit flow, including defaulting Chess tests only if/after Vitis detection and correcting the `LibXAIE_${AIE_RUNTIME_TEST_TARGET}_DIR` variable.

This is intended as lit enablement only. Experimentation has demonstrated some separate Python host runtime ownership/lifetime quirks -- at least on Windows  -- that still require additional work.

---

#### Test command conventions

I am recommending several new cross-platform hygiene standards here for AIE/NPU test commands. Ideally, these can be extended to guides, examples, and other parts of the tree later. My intent is two-fold: 
- 1) keep tests from depending on `PATH`, POSIX shell behavior, or Linux-only linker flags and 
- 2) avoid the need to push commits to hundreds of individual files every time something like a backend flag changes (for instance, the recent hardening around `--no-xbridge` as a further example in addition to the present case).

For example, tests should prefer:

- `%aiecc %backend_flags ...` for compiler-driver invocations
- `%host_clang ... %host_link_flags ...` for host test binaries
- `%PYTHON ...` for Python scripts
- explicit `.exe` names for host executables used by lit (this is safe on Linux and required for Windows)
- lit features such as `peano`, `chess`, `aietools_aie2`, `aietools_aie2p`, and `xrt` for gating

And, in future tests, guides, examples, and scripts, etc., it would be best to **avoid**:

- bare `aiecc` or `clang`
- `aiecc.py` (in most cases)
- hard-coded `--no-xchesscc --no-xbridge` (`%backend_flags` supplies this when needed)
- hard-coded `-lstdc++` / `-lrt` (`%host_link_flags` supplies this)
- `|&`, stdin `echo` pipelines, Makefile-only test drivers, and extensionless executable assumptions

The shared substitutions used by the files touched in this PR are defined in `lit_config_helpers.py`.

---

#### Reproduction

On Windows, the tests may be exercised locally using the following PowerShell commands. The absolute path of an extracted copy of MLIR must be supplied for `$mlir`:

```powershell
# Setup env from fresh clone
python utils/iron_setup.py
python utils/iron_setup.py env --shell pwsh | iex 

# Configure lit tests
$mlir = "F:\dev\mlir"      # Replace with root containing lib\cmake\mlir and lib\cmake\llvm
$xrt = "C:\Xilinx\XRT"     # Adjust if your XRT SDK was extracted elsewhere
$prefix = "$PWD\build_lit\install"

cmake -S . -B build_lit -G Ninja `
  -DCMAKE_BUILD_TYPE=Release `
  -DCMAKE_INSTALL_PREFIX="$prefix" `
  -DMLIR_DIR="$mlir\lib\cmake\mlir" `
  -DLLVM_DIR="$mlir\lib\cmake\llvm" `
  -DPEANO_INSTALL_DIR="$env:PEANO_INSTALL_DIR" `
  -DXRT_ROOT="$xrt" `
  -DCMAKE_PREFIX_PATH="$xrt" `
  -DLLVM_EXTERNAL_LIT=lit `
  -DAIE_COMPILER=NONE `
  -DAIE_LINKER=NONE `
  -DHOST_COMPILER=NONE

# Build tests and tools
cmake --build .\build_lit

# Run test sweep (parallel)
lit -sv --show-unsupported --time-tests --timeout 600 .\build_lit\test 2>&1 |
  Tee-Object -FilePath .\lit_test_sweep.log
```

And the equivalent `cmd` commands from an `x64 Native Tools Command Prompt`:

```bat
python utils\iron_setup.py
python utils\iron_setup.py env --shell cmd > "%TEMP%\iron_env.bat" && call "%TEMP%\iron_env.bat"

set "mlir=F:\dev\mlir"
set "xrt=C:\Xilinx\XRT"
set "prefix=%CD%\build_lit\install"

cmake -S . -B build_lit -G Ninja ^
  -DCMAKE_BUILD_TYPE=Release ^
  -DCMAKE_INSTALL_PREFIX="%prefix%" ^
  -DMLIR_DIR="%mlir%\lib\cmake\mlir" ^
  -DLLVM_DIR="%mlir%\lib\cmake\llvm" ^
  -DPEANO_INSTALL_DIR="%PEANO_INSTALL_DIR%" ^
  -DXRT_ROOT="%xrt%" ^
  -DCMAKE_PREFIX_PATH="%xrt%" ^
  -DLLVM_EXTERNAL_LIT=lit ^
  -DAIE_COMPILER=NONE ^
  -DAIE_LINKER=NONE ^
  -DHOST_COMPILER=NONE

cmake --build .\build_lit

lit -sv --show-unsupported --time-tests --timeout 600 .\build_lit\test > .\lit_test_sweep.log 2>&1
type .\lit_test_sweep.log
```
> `cmd` does not have an equivalent to `tee`, so this will appear to hang for several minutes as the log is being written.

#### (!! IMPORTANT: You **must** be running Python 3.13 for this to work on Windows, because that is the only version supported by `pyxrt.pyd` in the XRT SDK !!)

---

~~At this point, the remaining failures may be triaged into the following four categories, all of which are distinct from lit test configuration issues:~~

~~1. Full-ELF/XRT coverage~~

   ~~Six (6) failures are in `loadpdi`, reconfiguration, and scratchpad-register-write tests that exercise full-ELF attachment through XRT. As tested on Windows, attaching most ELFs to a `hw_context` works fine, but these particular kinds of full-ELFs do not yet appear to be supported end-to-end, and I suspect that is because they are written in such a way that they touch the firmware. These tests are left ungated for now, but I think it is unlikely they will ever fully work in a signed-driver environment.~~

~~2. Python host-runtime ownership/lifetime behavior~~

   ~~Twelve (12) tests fail due to object lifetime ordering concerning `hw_context`-attached resources. Specifically, the Windows runtime is sensitive to destroying a `hw_context` before attached objects, such as BOs, which usually causes a seg-fault. These are correctable through hardening lifetime/ownership in `hostruntime.py`. I've prototyped that separately and will submit the fixes if/after this PR is merged. Frankly, the destruction order (attachments first, `hw_context` last) that Windows forces makes more intuitive sense to me, anyway.~~

~~3. Context-cache stress behavior~~

   ~~`test_cached_xrt_runtime.py` still has one (1) failure involving the creation of an additional `hw_context` when the configured cache limit is reached. This likely needs separate investigation of the cache eviction semantics.~~

~~4. Remaining generated-code/runtime correctness issue~~

   ~~`test_algorithms.py` still has one (1) scalar float32 add correctness failure, where-in one 16-element ObjectFifo/data-tile region (corresponding to output entries 96-111) is observed as zeroed. The root cause of this is not yet clear to me, but may involve either Peano-specific code-generation or our own runtime memory handling. In this *specific* case, the affected ObjectFifo buffer for data-tile 6 is allocated adjacent to the worker stack region, and my testing has indicated that increasing `stack_size` to `0x0800` (from the default `0x0400`) resolves the symptom but not the underlying cause. Altering the instruction to multiply, which uses libcalls, also prevents the corruption. This is *consistent with* (but potentially not *exclusive to*) a stack-pressure or spill-related mechanism: Peano’s AIE2P lowering custom-legalizes scalar float32 adds/subs through vector floating-point operations, which use zero-padded ACC2048 registers (and corresponding stack spill/reload machinery). However, other mechanisms cannot yet be ruled out. That is as far as I got on this one. It's a stumper.~~

~~`6 + 12 + 1 + 1 = 20`~~
All failures are accounted.


thomthehound:
Converting this to draft. The new scratchpad test introduced a few hours ago is giving me trouble parsing the new custom op.

thomthehound:
I think I have it nailed down now. The wheel wasn't as fresh as I expected it to be, so I needed to recompile `aiecc`. Interesting. I thought we distributed new wheels with each merge.

Regardless, looking at the new test gave me another idea to implement while I futzed around. Just a few more hours and I'll have this ready for review.

thomthehound:
- centralize tool substitutions
- exercise `aiecc` executable backend where appropriate (instead of deprecated Python wrapper)

hunhoffe:
Interesting to hear about the lifetime/ownership issues for pyxrt! Is that for input/output buffers, or just the buffer used internally for the instruction bo, etc. associated with the hardware context?

thomthehound:
> Interesting to hear about the lifetime/ownership issues for pyxrt! Is that for input/output buffers, or just the buffer used internally for the instruction bo, etc. associated with the hardware context?

The funny thing is, I don't see anything in upstream `pyxrt.cpp` that could even plausibly cause this behavior. In fact, I couldn't find anything *anywhere* in the publicly available XRT source that explains it. So I'd hesitate to call this a `pyxrt` issue specifically; I think it may be a coincidence that only the Python binding tests trigger it, because the C++ tests tend to be much more narrowly scoped and do not stress lifetime/ownership/cache behavior (at least not in the way that the bindings tests are designed to do).

My interpretation is that this is probably something specific to the Windows runtime/driver. But the source code for that isn't available to me, so I had to reason about it experimentally. To answer your question, though: pretty much every time, the trigger I observed was eviction/destruction of the cached instruction BO. I did not personally observe problems with the input/output buffers, but I also don't have enough visibility into the closed-source side of Windows to say that could *never* be an issue, and I didn't try to force it as a negative test, either.

That's a great question, though. I'll try to probe around more regarding this issue while I finalize my `hostruntim.py` patch, but I'm not sure how much I'll turn up.

thomthehound:
I don't think I've ever been so excited to see errors on the CI before (and, really, I am truly charmed when it comes to CI luck). Both of these are actually hugely helpful.

@hunhoffe A new wild transient has appeared!
```
terminate called after throwing an instance of 'xrt_core::system_error'
what(): Open /dev/accel/accel0 failed (err=-19): No such device
```
I will be modifying `run_on_npu.py` to be less specific about the wording (it was looking for "No such device with index" before). I'm not sure if there has been an update to the `xrt-smi` in the CI runner, or if we caught a rare legendary. And, as usual, pretty much no information as to why it happened:

```
-- Testing: 1276 of 1277 tests, 12 workers --
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60
FAIL: AIE_TEST :: npu-xrt/dynamic_object_fifo/sliding_window/aie2.py (820 of 1276)
******************** TEST 'AIE_TEST :: npu-xrt/dynamic_object_fifo/sliding_window/aie2.py' FAILED ********************
Exit Code: 250

Command Output (stdout):
--


****** Bootgen v2023.2
  **** Build date : May 18 2026-09:48:41
    ** Copyright 1986-2022 Xilinx, Inc. All Rights Reserved.
    ** Copyright 2022-2023 Advanced Micro Devices, Inc. All Rights Reserved.


[INFO]   : Bootimage generated successfully

XRT Build Version: 2.21.0 (HEAD)
       Build Date: 2026-02-18 10:35:42
          Hash ID: 4eb1f4392a012b4e6eca759762389c612537f7c7
Creating a default 'in-memory' xclbin image.

Section: 'MEM_TOPOLOGY'(6) was successfully added.
Size   : 88 bytes
Format : JSON
File   : 'aie2.mlir.prj/main_mem_topology.json'

Section: 'AIE_PARTITION'(32) was successfully added.
Size   : 3608 bytes
Format : JSON
File   : 'aie2.mlir.prj/main_aie_partition.json'
Info: Embedded Metadata section is missing project.platform.device.core element, adding it.
Successfully wrote (9641 bytes) to the output file: final.xclbin
Leaving xclbinutil.
Compilation completed successfully
terminate called after throwing an instance of 'xrt_core::system_error'
  what():  Open /dev/accel/accel0 failed (err=-19): No such device

--
Command Output (stderr):
--
RUN: at line 10: /home/github/actions-runner/_work/mlir-aie/mlir-aie/build/bin/xchesscc_wrapper aie2 -I /opt/ryzen_ai-1.3.0.1/vitis_aie_essentials/include -c /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/kernel.cc -o ./kernel.o
+ /home/github/actions-runner/_work/mlir-aie/mlir-aie/build/bin/xchesscc_wrapper aie2 -I /opt/ryzen_ai-1.3.0.1/vitis_aie_essentials/include -c /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/kernel.cc -o ./kernel.o
RUN: at line 11: "/home/github/actions-runner/_work/mlir-aie/mlir-aie/aie-venv/bin/python" /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/aie2.py > ./aie2.mlir
+ /home/github/actions-runner/_work/mlir-aie/mlir-aie/aie-venv/bin/python /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/aie2.py
RUN: at line 12: "/home/github/actions-runner/_work/mlir-aie/mlir-aie/build/bin/aiecc"  --no-aiesim --aie-generate-npu-insts --aie-generate-xclbin --no-compile-host --xclbin-name=final.xclbin --npu-insts-name=insts.bin ./aie2.mlir
+ /home/github/actions-runner/_work/mlir-aie/mlir-aie/build/bin/aiecc --no-aiesim --aie-generate-npu-insts --aie-generate-xclbin --no-compile-host --xclbin-name=final.xclbin --npu-insts-name=insts.bin ./aie2.mlir
warning: overriding the module target triple with pdarch-unknown-unknown-elf [-Woverride-module]
1 warning generated.
RUN: at line 13: "/usr/bin/clang" /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/test.cpp -o test.exe -std=c++17 -Wall -I/opt/xilinx/xrt/include -L/opt/xilinx/xrt/lib -luuid -lxrt_coreutil -lrt -lstdc++ -I/home/github/actions-runner/_work/mlir-aie/mlir-aie/build/runtime_lib/x86_64/test_lib/include -L/home/github/actions-runner/_work/mlir-aie/mlir-aie/build/runtime_lib/x86_64/test_lib/lib -ltest_utils
+ /usr/bin/clang /home/github/actions-runner/_work/mlir-aie/mlir-aie/test/npu-xrt/dynamic_object_fifo/sliding_window/test.cpp -o test.exe -std=c++17 -Wall -I/opt/xilinx/xrt/include -L/opt/xilinx/xrt/lib -luuid -lxrt_coreutil -lrt -lstdc++ -I/home/github/actions-runner/_work/mlir-aie/mlir-aie/build/runtime_lib/x86_64/test_lib/include -L/home/github/actions-runner/_work/mlir-aie/mlir-aie/build/runtime_lib/x86_64/test_lib/lib -ltest_utils
RUN: at line 14: "/home/github/actions-runner/_work/mlir-aie/mlir-aie/aie-venv/bin/python" "/home/github/actions-runner/_work/mlir-aie/mlir-aie/utils/run_on_npu.py" "npu1" ./test.exe
+ /home/github/actions-runner/_work/mlir-aie/mlir-aie/aie-venv/bin/python /home/github/actions-runner/_work/mlir-aie/mlir-aie/utils/run_on_npu.py npu1 ./test.exe

--
```


The other useful piece of information gained here is that my float32 error in `test_algorithms.py` was reproduced on the CI runner. Because of an error in my own code, the CI runner was accidentally forced to run using Peano (and NPU2) and not Chess, but that proves the failure is *not* specific to Windows. We will not be able to implement fully-Peano CI runners for any NPU2 platform until it is fixed (or gated).

thomthehound:
- `run_on_npu.py` now properly detects new transient error language
- Stop having Peano force execution on NPU2 when Chess wants NPU1

thomthehound:
@erwei-xilinx I'm sorry to bother you, but can you kick this thing back off again for me?

One of these days I need to talk to the management about becoming a collaborator around here...

thomthehound:
> Interesting to hear about the lifetime/ownership issues for pyxrt! Is that for input/output buffers, or just the buffer used internally for the instruction bo, etc. associated with the hardware context?

@hunhoffe I investigated this more thoroughly, and I can say with a *very* high degree of confidence now that input/output BOs are not affected by this issue; the instruction BOs are the problem (tested with code configured for arbitrary destruction ordering). I was also able to replicate the same errors on Windows using C++ code that forces destruction in the same order that `hostruntime.py` currently does. Previously, the C++ tests on Windows were shielded from this issue because RAII enforces destruction of the objects in the correct order unless explicitly made not to. At this point, I think the Python tests are passing on Linux because Linux temporarily preserves some handle, resource, or other state after `hw_context` destruction that Windows does not. I no longer think this is an XRT bug at any level aside from documentation. It looks to be a pure platform issue.

~~Later tonight~~ Soon I'll be pushing my `hostruntime.py` update, and I'll include some new lit tests with an XFAIL on Windows around this issue. It should be interesting to see if the C++ version fails on Linux as well, given that I have to force the bad destruction order there.

Edit: I might have to hold off on the lit tests, given I've slightly altered their preferred syntax here and I don't want to have to drift against my own recommendations.

hunhoffe:
@thomthehound Thanks for digging into this! And for keeping an eye out for legendary device not found errors! Re: float32 errors, we've had some issues with stack overflows when peano uses more stack space than expected, particularly for float32 JIT functions. Estimating/checking the stack size requirements is a bit of a sticky problem (see: https://github.com/Xilinx/mlir-aie/issues/2347) so we've disabled some of the float32 tests for now for that reason. Not sure if that's related to your errors or not.

thomthehound:
> @thomthehound Thanks for digging into this! And for keeping an eye out for legendary device not found errors! Re: float32 errors, we've had some issues with stack overflows when peano uses more stack space than expected, particularly for float32 JIT functions. Estimating/checking the stack size requirements is a bit of a sticky problem (see: #2347) so we've disabled some of the float32 tests for now for that reason. Not sure if that's related to your errors or not.

No problem! I thought it was a great question! And thank you for the reference regarding the float32 bug. From what you just said, if that isn't directly related, it's at least a next-door neighbor. I really appreciate that. I'll look into this more tomorrow. 

But, for a first Peano pass, I think this current work is done. I did not intend this to actually fix issues; just to identify them. I'd just like @jgmelber to take a look, because I think this is kind of thing that would interest him.

jgmelber (approved):

thomthehound:
@hunhoffe I understand you are super-busy right now, but would it be a problem to get this merged soon? #3100 will not pass CI without it.

hunhoffe:
@thomthehound Sorry for the delay! I think either @jackl-xilinx, @erwei-xilinx, or @jgmelber is going to do some windows environment setup trials but I don't think either of them will have time to address this until mid-next week at earliest. I do think this is a good contribution, and I've definitely not forgotten that this is waiting in the queue.

... if you happen to be looking for something to do.... I'm hoping to put out a RFC issue later today about some compile/build/jit/programming example consolidation I'm working on. Here's an exploratory notebook: https://github.com/Xilinx/mlir-aie/blob/ae4cb4e131de1cc9a4d21eca977b5f5c3b011f2b/programming_guide/whats_new_in_unify_compilation_workflow.ipynbf (WIP PR is here: https://github.com/Xilinx/mlir-aie/pull/3025. although I don't plan to merge it in one piece).

I'd be interested to see if you think my proposed changes will play well with Windows. I'll post the official issue with a better writeup soon!

thomthehound:
> @thomthehound Sorry for the delay! I think either @jackl-xilinx, @erwei-xilinx, or @jgmelber is going to do some windows environment setup trials but I don't think either of them will have time to address this until mid-next week at earliest. I do think this is a good contribution, and I've definitely not forgotten that this is waiting in the queue.
> 
> ... if you happen to be looking for something to do.... I'm hoping to put out a RFC issue later today about some compile/build/jit/programming example consolidation I'm working on. Here's an exploratory notebook: https://github.com/Xilinx/mlir-aie/blob/ae4cb4e131de1cc9a4d21eca977b5f5c3b011f2b/programming_guide/whats_new_in_unify_compilation_workflow.ipynbf (WIP PR is here: #3025. although I don't plan to merge it in one piece).
> 
> I'd be interested to see if you think my proposed changes will play well with Windows. I'll post the official issue with a better writeup soon!

No worries! It's just always hard as an outsider to know what's going on. I was just checking in. I appreciate the update!

That work you pointed out looks impressive, exciting and... substantial. I will try to come up with some tests for it against the tree as it stands as well as my backlog of Windows work. From my preliminary look, I don't see anything *obviously* breaking on the Windows-side, but I did run into some JIT issues while working on my upcoming programming_examples refactor. I think I can probably spin off the institutional fixes for those into a tiny pre-patch.

thomthehound:
Is there any way that I could help with whatever concerns there may be about this patch? I was relying on it for my roadmap to a Windows CI.

thomthehound:
As of today, 2026-06-05, here is where we stand on the Windows lit tests with this patch:
```
Failed Tests (15):
  AIE_TEST :: npu-xrt/add_one_objFifo/run.lit
  AIE_TEST :: python-concurrency/jit_parallel_compilation.py
  AIE_TEST :: python/npu-xrt/test_algorithms.py
  AIE_TEST :: python/npu-xrt/test_cached_xrt_runtime.py
  AIE_TEST :: python/npu-xrt/test_cached_xrt_runtime_insts.py
  AIE_TEST :: python/npu-xrt/test_compile_cache_functionality.py
  AIE_TEST :: python/npu-xrt/test_iron_object_fifo_init_values.py
  AIE_TEST :: python/npu-xrt/test_jit_compilation.py
  AIE_TEST :: python/npu-xrt/test_jit_extern_functions.py
  AIE_TEST :: python/npu-xrt/test_jit_extern_functions_inside_jit.py
  AIE_TEST :: python/npu-xrt/test_jit_placed_style.py
  AIE_TEST :: python/npu-xrt/test_jit_trace.py
  AIE_TEST :: python/npu-xrt/test_jit_two_extern_functions.py
  AIE_TEST :: python/npu-xrt/test_jit_utils.py
  AIE_TEST :: python/npu-xrt/test_tensor.py


Testing Time: 189.67s

Total Discovered Tests: 1306
  Unsupported      : 430 (32.92%)
  Passed           : 850 (65.08%)
  Expectedly Failed:  11 (0.84%)
  Failed           :  15 (1.15%)
  ```
  
  If #3100 is also applied, these fall to:
  ```
Testing Time: 181.06s

Total Discovered Tests: 1309
  Unsupported      : 430 (32.85%)
  Passed           : 867 (66.23%)
  Expectedly Failed:  12 (0.92%)
  ```

No failures.
   
The six tests that touch the device firmware are marked as `XFAIL` on Windows because the signed driver flat-out rejects their execution (they build perfectly fine, though).

I will try to continue maintaining this patch as long as I can, but I'm not going to be able to move on without it.

**Review by erwei-xilinx — APPROVED** [review]

**Comment by thomthehound** [comment]
Something strange seems to be happening with Chess vs. Peano detection now. I will fix this tomorrow. I think the position of a guard in the Python tests must have moved.

**Comment by thomthehound** [comment]
The problem was a little bit more subtle than I expected, but it ultimately did boil down to a guard issue. I had to reposition `DefaultNPURuntime` in `test_compilabledesign.py` and also change when it was called in `test_compile_mixed_explicit_paths_raises`. Software CI was getting confused trying to query non-existent NPUs.

Also, `test_kernels_chess.py / test_kernels_specs.py` were rejecting aie2p, even though they appear to have been written to have `aiep2` fallback to `aie2`, and the scalar kernel was written in such a way to be compatible with it. So I just added an `aie2p` entrypoint. Works fine.

**Comment by thomthehound** [comment]
oy. Something is wrong with CI. It's pulling the wrong `nanobind`. Something changed in the last few hours.

**Comment by thomthehound** [comment]
hmm. Not a very helpful failure log. I think `test_kernels_specs.py` is just a collection of a huge number of tests and a 600 second timeout might not be enough for that on CI hardware. Investigating further...

Edit: no, I think I see the real problem now. It's still phantom NPU attempts on the software runner.

Edit2: `get_current_device()` is the problem. Working on a patch.

**Comment by thomthehound** [comment]
Fixed death by a thousand cuts.

`get_current_device()` was causing metadata-only tests to attempt to initialize the NPU. Failed attempt made software-only tests waste ~10 seconds looking for hardware that does not exist in that particular CI container. `test_kernels_specs.py` contains over 280 tests. 280 x 10 >> 600 seconds.

The timeout is resolved with a "just don't do that, bro" path for metadata-only device lookups.

File: test/npu-xrt/dynamic_object_fifo/sliding_window_conditional/aie2.py [hunk 1]
```diff
@@ -5,12 +5,12 @@
 #
 # (c) Copyright 2025 AMD Inc.
 
-# REQUIRES: ryzen_ai, valid_xchess_license
+# REQUIRES: ryzen_ai_npu1, valid_xchess_license
```
hunhoffe, line 8: Why doesn't this work on npu2?

test/npu-xrt/loadpdi/run.lit
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // REQUIRES: ryzen_ai_npu2, peano
+// XFAIL: system-windows
```
hunhoffe: Can you remind me why this would be xfailed?

test/npu-xrt/nd_memcpy_linear_repeat/aie2.py
```diff
@@ -5,11 +5,11 @@
 #
 # (c) Copyright 2024 AMD Inc.
 
-# REQUIRES: ryzen_ai, valid_xchess_license
+# REQUIRES: ryzen_ai_npu1, valid_xchess_license
```
hunhoffe, line 8: Why only npu1?

test/npu-xrt/nd_memcpy_transforms/aie2.py
```diff
@@ -5,13 +5,13 @@
 #
 # (c) Copyright 2024 AMD Inc.
 
-# REQUIRES: ryzen_ai, valid_xchess_license
+# REQUIRES: ryzen_ai_npu1, valid_xchess_license
```
hunhoffe, line 8: Why only npu1?

test/npu-xrt/neighbor_tile_memory_access/aie2.py
```diff
@@ -7,11 +7,11 @@
 # (c) Copyright 2024 Advanced Micro Devices, Inc. or its affiliates
 
 # Adapted from vector_scalar_add/aie2.py but with link between ComputeTiles
-# REQUIRES: ryzen_ai, chess
+# REQUIRES: ryzen_ai_npu1, chess
```
hunhoffe, line 10: Why only npu1?

test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // REQUIRES: ryzen_ai_npu2, peano
+// XFAIL: system-windows
```
hunhoffe: Is loadpdi really incompatible with windows?

test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
```diff
@@ -2,7 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
 // REQUIRES: ryzen_ai_npu2, peano
+// XFAIL: system-windows
```
hunhoffe: Same -- not clear to me why this is ok to fail? I'm happy to accept some intermediate failures that can be addressed in future work, but I'd like the limitations to be well-documented.

test/npu-xrt/sync_task_complete_token/aie2.py
```diff
@@ -5,12 +5,12 @@
 #
 # (c) Copyright 2024 AMD Inc.
 
-# REQUIRES: ryzen_ai
+# REQUIRES: ryzen_ai_npu1
```
hunhoffe, line 8: why only npu1?

test/npu-xrt/sync_task_complete_token_bd_chaining/aie2.py
```diff
@@ -5,12 +5,12 @@
 #
 # (c) Copyright 2024 AMD Inc.
 
-# REQUIRES: ryzen_ai
+# REQUIRES: ryzen_ai_npu1
```
hunhoffe, line 8: Why only npu1?

test/npu-xrt/vec_mul_event_trace/test.py
```diff
@@ -9,12 +9,12 @@
 #
 # ===-----------------------------------------------------------------------===#
 #
-# REQUIRES: ryzen_ai_npu2, xrt_python_bindings
+# REQUIRES: ryzen_ai_npu2, xrt_python_bindings, chess
```
hunhoffe, line 12: Good catch!

test/parse-trace/test1/aie_test1.mlir
```diff
@@ -1,11 +1,9 @@
 // (c) Copyright 2023 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// RUN: mkdir -p test
-// RUN: cd test
-// RUN: make -f %S/Makefile clean
-// RUN: make -f %S/Makefile 
-// RUN: make -f %S/Makefile diff
```
hunhoffe, line 8: Can you help me understand the change here?

test/parse-trace/test2/aie_test2.mlir
```diff
@@ -1,11 +1,9 @@
 // (c) Copyright 2023 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 //
-// RUN: mkdir -p test
-// RUN: cd test
-// RUN: make -f %S/Makefile clean
-// RUN: make -f %S/Makefile 
-// RUN: make -f %S/Makefile diff
```
hunhoffe, line 8: And here too!

test/python/test_callable_design_unit.py
```diff
@@ -25,6 +25,12 @@
 from aie.utils.jit import _JIT_CONFIG_KEYS, jit
 from aie.iron.kernel import ExternalFunction, Kernel
 
+
+def _path_text(path) -> str:
```
hunhoffe: Is there a python library function we can use instead here that is platform independent?

test/python/test_compilabledesign.py
```diff
@@ -28,6 +28,12 @@
 from aie.utils.compile.jit.markers import CompileTime, In, InOut, Out
 from aie.utils.hostruntime import set_current_device
 
+
+def _path_text(path) -> str:
```
hunhoffe: Not a huge fan of multiple versions of the same helper... a python library function would be preferred here too.

test/python/test_compileconfig.py
```diff
@@ -20,6 +20,12 @@
 from aie.utils.compile.jit.compileconfig import compileconfig
 from aie.utils.compile.jit.markers import CompileTime, In, Out
 
+
+def _path_text(path) -> str:
```
hunhoffe: Ack definitely don't want this function in 3 places!

hunhoffe (commented):
A few comments; I'll keep working with you to get this merged!

test/npu-xrt/loadpdi/run.lit
thomthehound: It touches the firmware and the Windows driver hard-crashes it. I couldn't find a reasonable work-around, but I also don't know if there are plans or not to allow this in the future. The Windows drivers lag by a *very* long time after Linux drivers get updates, so it might flip to an XPASS in the future, which I want to see.

test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
thomthehound: I know this is annoying. But it really is, at least for the time being, a signed driver issue.

test/npu-xrt/dynamic_object_fifo/sliding_window_conditional/aie2.py
thomthehound: Honest answer? It is labeled `aie2`. I didn't even test it on npu2 when I first submitted this. I didn't realize I'd only get one pass at this. The "No1" in the name was to indicate that I wanted another look.

test/npu-xrt/nd_memcpy_linear_repeat/aie2.py
thomthehound: Same as above. But I will investigate further.

test/parse-trace/test1/aie_test1.mlir
thomthehound: Make doesn't work for this repo in Windows. Yes, git bash make exists, but that is... problematic. And our WSL support *really* complicates things.

test/parse-trace/test2/aie_test2.mlir
thomthehound: To continue, I don't even think these things are even needed to make the test work.

test/python/test_callable_design_unit.py
thomthehound: hmm. Nothing came to mind at the time, but I will investigate this. Perhaps there is a better way.

test/python/test_compileconfig.py
thomthehound: I'll try to come up with a better way to handle this.

test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
thomthehound: I could leave them as failures, but a Windows CI would just fail then. This seemed rational and noticeable. But I read your point: I will add comments.

thomthehound:
@hunhoffe Thank you for that review! It's given me several things to think about. I'll try to get this all addressed tomorrow!

test/npu-xrt/nd_memcpy_transforms/aie2.py
thomthehound: I think I might need to expand coverage to make these real on npu2.

thomthehound:
Resolved: duplicate path helpers

I was able to remove the duplicated test-local helpers by just modifying a few lines in `CompilableDesign`’s JSON serialization machinery. `compilabledesign.py` did not exist when I first submitted this PR, but using it here makes the tests much cleaner.

Thank you for pointing this out!

thomthehound:
Partial resolution: 
```
sliding_window_conditional/aie2.py
nd_memcpy_transforms/aie2.py
nd_memcpy_linear_repeat/aie2.py
neighbor_tile_memory_access/aie2.py
```

These are Chess-only runtime tests and are explicitly written for NPU1/NPU1_1col. I realize now that I did not test them on NPU2 because I cannot run them with Peano. The changes I made were just requirement tightening to stop them from firing where they do not belong. I believe it would be counter to the spirit of this PR if I were to alter them further or attempt to expand coverage to NPU2. However, I can revert my changes, if that is preferred.

thomthehound:
Resolved: reminder why firmware-touching full-ELF tests are marked XFAIL on Windows

I added a two-line comment that says exactly this to each affected file:
``` 
// The Windows driver stack fails this firmware-touching full-ELF path at runtime.
// Keep XFAIL so lit reports XPASS if driver support changes.
```

thomthehound:
Resolved: `parse_trace` lit cleanup

Removed the gigantic `python -c ...` one-liner and centralized the output check in a small, reviewable Python script at the `parse_trace/` test root. This looks much cleaner now and makes the test path, including its checks, pure Python without relying on POSIX-oriented `make`.

thomthehound:
@hunhoffe Thank you again for taking the time to review this. I know everyone has been extremely busy, so I especially appreciate your help getting this polished up.

test/npu-xrt/dynamic_object_fifo/sliding_window_conditional/aie2.py
hunhoffe: You can have No<1>-No<N> PRs to get this fully functional, I'll just be trying to keep the code base in a reasonable state between each one!

python/aie_lit_utils/lit_config_helpers.py
```diff
@@ -620,6 +612,152 @@ def setup_host_target_triplet(
...
+                probe_env[key] = LitConfigHelper._prepend_env_paths(
+                    probe_env.get(key, ""), value
+                )
+            else:
+                probe_env[key] = value
+
+        probe = (
+            "import importlib; "
+            f"importlib.import_module({module_name!r}); "
+            "print('ok')"
```
hunhoffe: Final nit (this is really testing my python knowledge, it's a good excuse to learn new things!)

In can_import_python_module, consider using `importlib.util.find_spec` instead of `importlib.import_module `for the probe:

```python
probe = f"import importlib.util, sys; sys.exit(0 if importlib.util.find_spec({module_name!r}) else 1)"
```
`find_spec` checks whether the module is importable without actually executing its top-level code, so it's faster and has no side effects. It also lets you drop the `print('ok')` and simplify the result check to just `result.returncode == 0`.

hunhoffe:
@thomthehound Thanks for being so responsive to my remarks and doing the research on various changes! Hopefully by getting my questions answered I'll be more prepared for enablement No2 whenever it may come.

thomthehound:
> @thomthehound Thanks for being so responsive to my remarks and doing the research on various changes! Hopefully by getting my questions answered I'll be more prepared for enablement No2 whenever it may come.

Of course! I enjoy working with you. And there is *definitely* going to be a No2 soon. I've missed a few things.

Embarrassingly, I had accidentally disabled my local NPU detection for these tests a few weeks ago, so I've been getting false 'green' sweeps lately. I needed a custom XRT compile for another project with a different directory layout than the SDK, and just forgot to change it back because `mlir-aie` was still compiling and running kernels just fine. Now that I've switched back, I'm seeing real errors.

I'm hoping I can get this fixed up by the end of the day, but right now I am unsure.

Edit: I'm not saying I didn't test this on real NPU before. I'm just saying that I haven't been testing on real NPU for the last two weeks, during which some major changes were made that I need to properly update around.

thomthehound:
The problem during amdhx370 isn't a transient. I've tracked it down. There is a small bug in `CallableDesign.__call__`. Falling back to `aie2` isn't always safe in the way that it was intended. I'm working on a fix now. It's just going to take some time.

Edit: OK, this goes a little bit deeper than I was expecting. I might need a separate PR to fix this. I have a good idea of the the shape of it now, though. It just might take me a bit more time.

hunhoffe:
@thomthehound no worries! There's no hurry at all from our end -- I just don't want to lose all your momentum if we leave it here stuck too long.

thomthehound:
> @thomthehound no worries! There's no hurry at all from our end -- I just don't want to lose all your momentum if we leave it here stuck too long.

Word.

Honestly, I'm laughing right now at the irony of it all. But having my eyes on the prize is very motivating again, so it's all good. As you know, I'm just a chatty Cathy.

thomthehound:
Now depends: #3184

thomthehound:
I can't believe this, but in the last couple days the Windows drivers **did** update. The firmware-touching tests now XPASS. I will revert the XFAIL tomorrow.

(This is precisely the reason why I wanted to XFAIL in the first place. I had a feeling this would happen. Just... not with quite such impeccable timing.)

thomthehound:
In addition to reverting the XFAILs and removing the comments, the last commit also adds a probe of `pyxrt.pyd` during testing. This is needed because the XRT SDK is now older than the new drivers. So, while C++ tests get the support for the features they need, the Python bindings will need an update to prevent them from failing. Since we know that is going to happen eventually, the probe just checks for the needed features and if it does not find them, it moves the tests to "unsupported". This should be self maintaining.

python/aie_lit_utils/lit_config_helpers.py
thomthehound: Thanks! Sorry about not responding to this sooner. This was actually a bit of a red herring. It was written that way in anticipation of what I needed to do in `a73767c`. To get a proper probe of `pyxrt.pyd`’s capabilities, I need to load it fully rather than merely check whether Python can find it.

But I totally agree that `print('ok')` is both redundant and repetitive.

thomthehound:
Now that #3184 is merged, this patch *should* pass CI, in which case it will be ready to go.

thomthehound:
This is interesting: no jobs are being executed on any of the hardware runners, even though many jobs are enqueued. The same thing happened yesterday around this time. I think a third data-point would be needed to establish this as a trend, but something seems to be be happening between Noon Mountain Time and Noon Pacific Time. The coincidence is interesting.

Regardless, I am going to play Nostradamus and say this merge is getting bumped.

Edit: ... I say as I enjoy the taste of my foot in my mouth. This diagnosis was not quite right. Something is running now. Still, strange...

hunhoffe:
When the CI runners get overloaded it appears they sometimes have some sort of recovery period where almost nothing runs? I don't have an explanation for this, but I've seen some cases of close to no throughput on/off.

thomthehound:
~~Yeah. I don't want to waste too much of your mindshare with this. I have a bad habit of thinking out loud. It just seems strange to me that they sometimes *both* go down at almost exactly the same time. Just one of them makes perfect sense but, both... it makes my science-sense tingle.~~

Ah, I shouldn't be wasting anybody's time with it. Sorry about that. I'm just looking for patterns where there probably are none to be found aside from load. What I see can easily be explained by just by the fact that we only have one merge group at a time, and the runners take turns flaking out until the queue is so jammed that whichever is the weakest one never has a chance to recover. That is probably what is happening.

PR #3077, opened 2026-05-18
objectfifo ArrayAttr-depth + initValues + iter_count cycling

hunhoffe:
- `printObjectFifoInitValues` segfaulted when `$elemNumber` was an `ArrayAttr` (list of per-endpoint depths); fixed to mirror the parser.
- IRON `ObjectFifo` now takes `init_values=`, passed through to the underlying op's `initValues` attribute.
- `iter_count > 1` on a static-init MemTile producer with no upstream link used to deadlock on the second BD-chain iteration (per-BD locks consumed by first pass, never refilled). The stateful transform now omits source-side `use_lock` ops for that exact configuration; the BD chain re-runs cleanly via the channel's `task_count`.

## Tests

- `test/dialect/AIE/objectfifo_array_depth_init_values_roundtrip.mlir` (lit; crashed pre-fix)
- `test/python/objFifo_iron_init_values.py` (MLIR-emit + FileCheck)
- `test/python/npu-xrt/test_iron_object_fifo_init_values.py` (pytest, passes on NPU Strix)
- `test/objectFifo-stateful-transform/bd_chain_on_memtile/init_values_no_link_skips_source_locks.mlir` (lit; verifies the source BDs have no `use_lock` ops)

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
@@ -396,6 +396,12 @@ struct AIEObjectFifoStatefulTransformPass
     std::vector<LockOp> locks;
     if (op.getDisableSynchronization())
       return locks;
+    // Static-init no-link producer cycled via iter_count: source side needs
+    // no sync; skip allocation to free the lock IDs.
+    if (op.getInitValues().has_value() && op.getIterCount().has_value() &&
+        op.getIterCount().value() > 1 && !getOptionalLinkOp(op).has_value() &&
+        static_cast<int>(op.getInitValues().value().size()) == numElem)
+      return locks;
     auto dev = op->getParentOfType<DeviceOp>();
     auto &target = dev.getTargetModel();
     // if shimTile external buffers are collected from input code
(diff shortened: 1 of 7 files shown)


lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -854,7 +854,18 @@ struct AIEObjectFifoStatefulTransformPass
...
+    // skip source-side locks. The BD chain restarts via the channel's
+    // task_count, but the per-BD lock state never gets replenished (no
+    // upstream S2MM refills the buffers) so the chain would deadlock on
+    // the second pass. Back-pressure to the downstream consumer is handled
+    // by the DMA stream's flow control; source-side locking is unnecessary
+    // for correctness in this configuration.
+    bool isCycledStaticInitProducer =
+        channelDir == DMAChannelDir::MM2S && op.getInitValues().has_value() &&
+        op.getIterCount().has_value() && op.getIterCount().value() > 1 &&
+        !getOptionalLinkOp(op).has_value();
```
andrej: Why does this problem with the locks only happen on the second iteration, i.e., what locks/unlocks them in the first pass? Asking because if there truly is no need for synchronization due to static value initialization on the producer side, shouldn't we not allocate any locks altogether (so those lock ID resources can remain free)?

andrej (approved):

jgmelber (approved):

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
hunhoffe: Thanks for catching this. First-pass mechanics: in createObjectFifoLocks, the cons lock for an initValues fifo is pre-charged with initValues * repeatCount units. Each source BD does acquire(consLock, 1) / release(prodLock, 1), so pass 1 drains the pre-charge cleanly. Pass 2 starves on consLock = 0 — with no upstream S2MM in the no-link case, nothing ever refills it.                                                                 

And you're right that the locks shouldn't be allocated at all here. I had been going for a minimal change, but I should be more correct here. I pushed a follow-up (8f206b1) that adds an early-return in createObjectFifoLocks mirroring the existing disableSynchronization path, drops the now-redundant gate in createBdBlock, and extends the lit test to assert no aie.lock ops are emitted for the fifo — so the regression is "lock IDs are freed," not just "no use_lock ops."

andrej: Nice, yeah I like that better. This seems clearer. :)

PR #3078, opened 2026-05-18
Unplaced IRON: extend SequentialPlacer + strip (most) residual `tile=` hints

hunhoffe:
Extends `SequentialPlacer` so that most IRON examples that previously needed `tile=Tile(c, r)` / `tile=AnyMemTile` hints compile cleanly with every placement argument stripped. Drops those hints from the Python sources, except ResNet, which still doesn't place correctly. Adds lit tests under `test/place-tiles/` that minimally cover the failure modes the placer used to hit.

## Changes

**SequentialPlacer** 
- **Pinned-first ordering**: stable-sort logicalTiles into pinned / partially-constrained / unpinned buckets before Phase 3 so a pinned coordinate is reserved before any unpinned tile can claim it.
- **Linked-fifo channel accounting**: stop blanket-skipping `aie.objectfifo.link` source/dest fifos in `buildChannelRequirements` — the off-link-tile endpoint (shim producer of a linked source, core consumer of a linked dest) still needs its DMA channel counted.
- **Load-balanced non-core placement**: `findTileWithCapacity` replaces first-match with lex-min `(distance, load, col, row)`; Phase 4 worklist is sorted by descending channel demand so heavy joins place before lighter pass-throughs crowd them out.
- **Compute-tile DMA budget for high-fanin Workers**: new `computePeerAdjacency` graph + per-LTO `needNeighbor{In,Out}` derived from non-peer DMA load + new `satisfiesComputePeer` pre-placement check (self-side, forward-look on free physical neighbor slots, symmetric on already-placed peers) + per-LTO candidate ordering that prefers interior rows when demand > 0 + sort key that lifts a heavy Worker's peers to the front so they grab its neighbor slots. Diagnostic mentions "compute-peer DMA budget" with peer notes when no candidate fits.

**Trace packet ID allocation**                                                                                                          
  - `TracePacketOp.id` becomes optional. `-aie-insert-trace-flows` auto-allocates absent ids in `(col, row)` order over the active trace tile set, making the id-to-tile
   binding a pure function of the placement rather than of worker-iteration order. Fixes a Strix routing-pipeline rejection in `vector_reduce_max/multi_column_designs` 
  where the new placer's worker→tile permutation aliased packet id 4 against a column-0 switch port rule.                                                               
  - IRON's `configure_trace` stops baking an id into emitted `trace_packet` ops (still passes `type` for the core-vs-mem distinction on shared tiles); the pass owns    
  allocation.                                                                                                                                                           
  - Overflow past the 5-bit hardware packet-id field is now a hard error in `-aie-insert-trace-flows` (previously wrapped silently and aliased traces onto the same id).
  - User-supplied explicit ids continue to be honored unchanged.


erwei-xilinx (approved):

yenjames:
@hunhoffe Glad to see the placer is becoming more and more robust! My only comment is that we may need a check in `--aie-insert-trace-flows` to make sure auto-assigned and user-declared packet IDs don't collide, but this can be added as a follow-up PR.

hunhoffe:
@erwei-xilinx Thanks for the review
@yenjames Fixed! Good catch.

PR #3100, opened 2026-05-20
Fix Windows crashes in Python XRT cached runtime

thomthehound:
This fixes a high-severity issue in the Windows Python XRT runtime path. The cached runtime previously destroyed XRT objects in an unsafe order, which caused access violations on Windows and made the Python wrapper unreliable on that platform.

This was an instruction BO lifetime problem. `hostruntime.py` could (and sometimes deliberately) let an instruction BO outlive the `kernel` / `hw_context` state it depended upon. Linux tolerates that ordering, but Windows does not. This patch makes the ownership explicit and destroys cached XRT objects in the same dependency order that is already enforced by RAII when using C++. This correctness/hygiene fix does not change observable Linux behavior, but it is required for reliable Windows Python execution.

It also tightens the cache bookkeeping around context eviction, removes platform-specific assumptions, and makes runtime-owned instruction BOs use the runtime's existing XRT device instead of creating a fresh `xrt.device(0)` so ownership can be tracked correctly.

### Changes

- Track which cached runtime entry owns each instruction BO.
- Destroy instruction BOs before their owning `kernel` / `hw_context`.
- Allocate runtime-owned instruction BOs from the runtime's existing XRT device.
- Retry `hw_context` creation after evicting cached contexts when context slots are exhausted.
- Update the cache-fill test so it tests cache policy, not the machine's practical `hw_context` limit.
- Use `np.asarray(...)` instead of `np.array(..., copy=False)` for NumPy 2.x compatibility.

### Tests

I added C++ lit coverage for the lifetime-ordering issue:

- normal RAII destruction order passes
- stale input/output BOs also pass
- stale instruction BO destruction reproduces the Windows access violation and is marked `XFAIL` on Windows but should pass on Linux.

These tests depend on #3075; without it, they fail due to lit syntax drift.


jgmelber (approved):

thomthehound:
Being "fortunate" enough to straddle this problem with complementary pulls (yesterday v. today), I can say with confidence that the latest CI failures are due to an issue exposed by the nightly build of Peano. This has been reproduced locally with virtually identical results to CI. 

There is an easy "just don't do that bro" fix to get the test running, but the operations it is lowering *should* be legal, which is problematic. I am too tired to investigate this further.

thomthehound:
@hunhoffe Obviously subject to your approval, this one is ready to go. It brings Windows lit errors down to zero.

Given all the recent changes in that area, it's going to take me a bit of time to rebase the Cmake programming_examples follow-up to all of this, but I think I can finally have that out after the weekend.

hunhoffe:
@thomthehound excellent! Queued!

thomthehound:
Seems like just my typical bad CI luck. It knows I've been talking about it.

hunhoffe:
@thomthehound agreed! Thanks for the reminder to push it back into the queue. There are now 2 more CI runners than before (thanks to Andre) which should relieve some of the pressure. In the process of setting them up there have been some backend failures here and there.

thomthehound:
> @thomthehound agreed! Thanks for the reminder to push it back into the queue. There are now 2 more CI runners than before (thanks to Andre) which should relieve some of the pressure. In the process of setting them up there have been some backend failures here and there.

That's great to hear! It's been an exciting week to say the least.

thomthehound:
That still looks like a ~~hardware~~ runner issue from what I am seeing. I can't immediately make sense of it otherwise. But certainly nothing in this patch could be causing it.

PR #3104, opened 2026-05-21
ci: serialize ATB chess builds and extend their lit timeout

hunhoffe:
The new gemm_asymmetric_tile_buffering chess builds peak at ~8 GB RSS and ~9 min wall time alone on 32 GB. Under -j4 on amdhx370 (32 GB runner), four-way concurrency exhausts memory and pushes wall time past the suite's 600 s lit timeout, producing flaky failures (e.g. run 26189907468).

- Register a new `atb_chess` lit parallelism group with concurrency=1.
- Opt the three configs into it via a lit.local.cfg in the ATB directory.
- Split the CI check-reference-designs invocation into two: main run skips ATB via LIT_FILTER_OUT, then a second ATB-only run with --timeout 1200.

Suite-wide 600 s timeout is unchanged for the other 278 tests. Local single-config measurement: 537 s wall, 8.0 GB peak RSS on a quiet 32 GB box, so 1200 s provides ~2x headroom over the slowest plausible CI run.

If this doesn't work, the next step is to disable them in the lit tests entirely until we find a better fix.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -215,8 +215,17 @@ jobs:
           mkdir $NPU_CACHE_HOME
 
           ninja install
-          ninja check-reference-designs
+
+          # gemm_asymmetric_tile_buffering chess builds peak ~8 GB RSS and
+          # ~9 min wall time alone on 32 GB. They run serialized via lit's
+          # atb_chess parallelism group and need a longer timeout than the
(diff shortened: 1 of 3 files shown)


jgmelber (approved):
Should we consider a nightly run for Chess tests rather than every run?

hunhoffe:
Yes, I think so although we'd want to trigger them for PRs that modify relevant files for spot testing. I'm going to be doing some CI work later today on other things, so if it's alright with you, I'll hold this thought for a later date.

PR #3109, opened 2026-05-22
[WIP] Share BD size/stride verification between AIE and AIEX dialects (#2566)

### LINKED ISSUE #2566: Missing and Wrong Data Layout Transformation Checks
The compiler currently (1) rejects some data layout transformations that are legal and (2) fails to reject some data layout transformations that are not realizable in hardware (i..e, sub-32-bit accesses).

For (1), this check, introduced in #2435, effectively disables use of the lowest-dimension stride:
https://github.com/Xilinx/mlir-aie/blob/d98dd252c141a94c67cd890adbcbb307e3d9e7c3/lib/Dialect/AIE/IR/AIEDialect.cpp#L1809
I don't know the background of why it was introduced, but I believe it effectively makes the lowest-dimension stride unusable.

For (2), as long as the lowest-dimension stride is 1 for smaller data types (e.g. int8) the compiler currently does not reject smaller-than-4-bytes transfers. Something like `[<size=3, stride=4>, <size=2, stride=1>]` should be rejected if the data type is `int8`, since the smallest unit of transfer here is 16 bits.

We have this function used elsewhere:
https://github.com/Xilinx/mlir-aie/blob/d98dd252c141a94c67cd890adbcbb307e3d9e7c3/lib/Dialect/AIEX/IR/AIEXDialect.cpp#L171
I think we should try to factor that out and use that for verification elsewhere as well. It probably contains mistakes too but not the above two.

hunhoffe:
Closes #2566.                                             

`AIEX::verifyStridesWraps` already had a careful granularity-based check for BD sizes and strides — this PR pulls the size/stride/granularity portion into a shared helper `AIE::verifyBDSizesStrides` so `aie.dma_bd`'s verifier can use the same logic. Two follow-on behavior changes fall out:
- **`aie.dma_bd` now accepts inner-dim strides > 1 on sub-32b types** when `stride × elemWidth` is a multiple of the address granularity (e.g. `memref<128xi16>, [<size=32, stride=2>]` — 2 × 16b = 32b, perfectly word-aligned). The previous blanket check rejected these even though hardware can encode them.
- **`aie.dma_bd` now rejects sub-word innermost contiguous runs** (e.g. `memref<128xi8>, [<size=3, stride=4>, <size=2, stride=1>]` — innermost run = 2 bytes). The AIEX path already caught these; the AIE path didn't.                                                                                                                                                                                                            
   
The `elemWidth > granularity` constraint added in #2435 (innermost stride must be 1 for bfp / wide types) is preserved in the shared helper, matching the silent `strides[0] = 0` behavior in `getHardwareStridesWraps`.

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -2215,73 +2215,132 @@ void DMABDOp::print(::mlir::OpAsmPrinter &printer) {
   printer.printOptionalAttrDict((*this)->getAttrs(), elidedAttrs);
 }
 
-// A BDDimLayoutAttr array (outermost-first) describes a contiguous row-major
-// scan when the innermost stride is 1 and each outer stride equals the product
-// of all inner sizes.  Used by both DMABDOp verification and canonicalization.
-mlir::LogicalResult xilinx::AIE::verifyBDSizesStrides(
-    mlir::Operation *forOp, unsigned elemWidthBits,
-    uint32_t addressGranularityBits, llvm::ArrayRef<int64_t> inputSizes,
-    llvm::ArrayRef<int64_t> inputStrides) {
+// Shared verifier for a DMA buffer descriptor's data-layout transform and its
+// explicit iteration. Used by DMABDOp::verify and by the AIEX runtime-sequence
+// lowerings (dma_memcpy_nd, dma_task). See the header comment for the contract.
+//
+// `inputSizes`/`inputStrides` are the pure data-layout dims, innermost-first, in
+// element-width units; all strides are positive (iteration is NOT folded in as a
+// magic stride-0 outer dim). `iterSize`/`iterStride` describe iteration in
+// logical, element-width units (iterSize == 0 means no iteration).
+mlir::LogicalResult xilinx::AIE::verifyBDDataLayoutAndIteration(
+    mlir::Operation *forOp, const AIETargetModel &targetModel,
+    AIETileType tileType, unsigned elemWidthBits,
+    llvm::ArrayRef<int64_t> inputSizes, llvm::ArrayRef<int64_t> inputStrides,
+    int64_t iterSize, int64_t iterStride, bool skipTransformationChecks) {
   assert(inputSizes.size() == inputStrides.size());
   const int n = static_cast<int>(inputSizes.size());
+  const uint32_t granularity = targetModel.getAddressGenGranularity();
+  const uint32_t maxDims = targetModel.getDmaBdMaxDims(tileType);
+  const uint32_t wrapBits = targetModel.getDmaBdWrapBits(tileType);
+  const uint32_t stepBits = targetModel.getDmaBdStepBits(tileType);
+  const uint32_t iterBits = targetModel.getDmaBdIterBits(tileType);
+
+  if (static_cast<uint32_t>(n) > maxDims)
+    return forOp->emitOpError("Cannot give more than ")
+           << maxDims << " dimensions for step sizes and wraps in this tile "
+           << "(got " << n << ").";
+
   if (n == 0)
     return success();
 
   for (int i = 0; i < n; ++i) {
     if (inputSizes[i] <= 0)
       return forOp->emitOpError("Size ") << i << " must be a positive integer.";
+    // A size-1 dimension's stride is never applied (the loop runs once), so any
+    // stride value is allowed for it. For any larger size the stride must be a
+    // positive integer (iteration/repeat is expressed separately, not as a
+    // stride-0 outer dim).
+    if (inputSizes[i] > 1 && inputStrides[i] < 1)
+      return forOp->emitOpError("Stride ")
+             << i << " must be a positive integer.";
   }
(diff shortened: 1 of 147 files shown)


lib/Dialect/AIEX/IR/AIEXDialect.cpp
andrej: Would it make sense to absorb these checks into `verifyBDSizesStrides` as well, and get rid of `verifyStridesWraps` entirely, just use the new `verifyBDSizesStrides`?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2196,6 +2196,83 @@ void DMABDOp::print(::mlir::OpAsmPrinter &printer) {
...
+  }
+
+  // For element widths larger than the granularity (e.g. bfp blocks, i64),
+  // the hardware cannot encode a non-1 innermost stride; getHardwareStrides-
+  // Wraps would silently drop the stride. Force innermost stride == 1.
+  if (elemWidthBits > addressGranularityBits && inputStrides[innermost] != 1)
+    return forOp->emitOpError(
+               "For element widths larger than the address granularity (")
+           << (addressGranularityBits / 8)
+           << " bytes), innermost dim stride must be 1.";
```
andrej: So I'm not allowed to, e.g., pick every odd uint64_t element of an array? Why? I don't think a check like this existed in the old verifyStridesWraps.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2196,6 +2196,83 @@ void DMABDOp::print(::mlir::OpAsmPrinter &printer) {
...
+        << " elements at " << (elemWidthBits / 8) << " bytes each equal "
+        << (inputSizes[innermost] * elemWidthBits / 8)
+        << " bytes, which is not divisible by " << (addressGranularityBits / 8)
+        << ". ";
+    return forOp->emitOpError(msg.str());
+  }
+
+  // Non-repeat dim strides must be positive when the corresponding size > 1
+  // (the repeat dim, if present as the outermost, may have stride 0).
+  const int repeatDim = n - 1;
```
andrej: I don't think this is correct.

- For the `dma_memcpy_nd` op, which goes through the `AIEDmaToNpu.cpp` lowering, it is always the size at hard-coded index `3` that is [used as both the `Repeat_Count` of the task](https://github.com/Xilinx/mlir-aie/blob/main/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L393) and the [`Iteration_Wrap` and `Iteration_Stepsize`](https://github.com/Xilinx/mlir-aie/blob/3c8c81a5b97dc57f3b3c10b2a777ab64e4214727/lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp#L382-L383). Since for `dma_memcpy_nd`, there always is exactly one BD per task, and because the BD does not toggle any locks, this effectively makes this dimension at index 3 behave the same as the other dimensions except it also allows stride zero.
- For `dma_task`s in the runtime_sequence, which go through the `AIEDMATasksToNpu.cpp` lowering, the size and stride at hard-coded index `3` are used as the [`Iteration_Wrap` and `Iteration_Stepsize`](https://github.com/Xilinx/mlir-aie/blob/3c8c81a5b97dc57f3b3c10b2a777ab64e4214727/lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp#L407-L408) of that BD. Since this syntax allows chaining multiple BDs or BDs which toggle locks, this does not set the `Repeat_Count` of the task (which one would we use if there are multiple BDs? .. it wouldn't make sense). The repeat count is instead set on the DMA task itself using an attribute.
- For all other BDs in the static part of the design (inside of `aie.mem`, `aie.memtile_dma`, etc. ops), which go through the `AIERT.cpp` lowering, I was unable to find any special treatment for any of the higher dimensions. The dimensions get [assigned here](https://github.com/Xilinx/mlir-aie/blob/3c8c81a5b97dc57f3b3c10b2a777ab64e4214727/lib/Targets/AIERT.cpp#L424). Repeat counts are explicitly set on the entire DMA task as well here.

Long story short, only hard-coded index 3 gets special treatment, and only sometimes. For those special cases, you have to pad out your dimensions up to four to make use of this behavior. I think the `repeatDim = n - 1` here does not capture that.

andrej (changes requested):
Fairly certain about the inconsistency with `repeatDim` introduced here

andrej:
Additional note regarding the repeat_count, I noticed that we do not call `XAie_DmaSetBdIteration` anywhere in our codebase, which I believe is the aie-rt interface for setting `Iteration_Stepsize` and `Iteration_Wrap`. This leads me to believe we currently do not support some of the iteration features in the static part of the design (the runtime sequence part manually lowers without going through aie-rt). I think repeat_count might still work if the Iteration_Stepsize and Iteration_Wrap registers default to zero if you don't call `XAie_DmaSetBdIteration`, but you couldn't add a stride, etc., in statically-configured BDs through that path right now with our compiler.

PR #3110, opened 2026-05-23
Add host-side bfloat16 helpers for platforms without std::bfloat16_t

thomthehound:
Under the C++23 standard, `std::bfloat16_t` support is optional. The current MSVC toolchain has chosen *not* to exercise that option, while the Linux toolchain does. Therefore, host-side example code that manipulates bfloat16 data cannot use `std::bfloat16_t` safely when compiled with MSVC, as is the intended route on native Windows.

This patch adds `test_utils::bfloat16_t` (and related helpers) to `test_utils.h`. When the standard type is available, `test_utils::bfloat16_t` resolves directly to `std::bfloat16_t`, preserving native execution. Otherwise, it uses `std::uint16_t` storage and performs host-side conversion/math through `float` (rounding back to 16 bits by the nearest-even method).

In the trivial case, the type selection works as so:

```cpp
#if defined(__STDCPP_BFLOAT16_T__)
using bfloat16_t = std::bfloat16_t;
#else
using bfloat16_t = std::uint16_t;
#endif
```

These changes are host-side only; device code is unaffected. The examples have been updated to use `test_utils::bfloat16_t` consistently for host-side bfloat16 data (including those that are not obviously Windows-portable) so there are not two competing vocabularies.

This workaround is not intended to permanently solve host-side `bfloat16` portability for all user code. It is intended to draw attention to this limitation, and to prepare the programming examples for cross-platform conversion. In the future, it may be reasonable to provide a better API boundary for this, but doing so would create a new public convention and is therefore a separate consideration.


hunhoffe:
We've had some degree of consternation with numerical approximation/precision/rounding errors in the past; we are pretty behind in our windows CI infrastructure so I devised a quality test as a follow on PR: https://github.com/Xilinx/mlir-aie/pull/3137

@thomthehound short of us getting better windows CI infra, do you see any other avenues for ensuring quality for Windows users with this change?

hunhoffe (approved):

thomthehound:
> We've had some degree of consternation with numerical approximation/precision/rounding errors in the past; we are pretty behind in our windows CI infrastructure so I devised a quality test as a follow on PR: #3137
> 
> @thomthehound short of us getting better windows CI infra, do you see any other avenues for ensuring quality for Windows users with this change?

Fortunately, in this case, the approximation is *very* minor, and actually goes through a *more* precise path than bfloat16 itself. For add/sub, the worst it could be off is by 1/2 LSB in the significand (and slightly worse in the transcendentals due to compounding of errors). It really works in our favor that bfloat16 is already designed to map 1:1 to float32 in terms of the important exponent portion. If this had been fp16 math causing the problem instead, I would have just disabled the test for Windows rather than risking trouble with bad approximations. I hear you on that, and I fully agree with your concern.

Checking the math on Windows (host versus device returned values) turned out fine on my end, so I don't anticipate any problems there (the device is already/still using bfloat16, after all so... "In Soviet Russia, device tests host!"). However, I think the test you came up with is exactly the right kind of way to double-check/reproduce my work on Linux.

I'm glad you brought up better Windows CI. That's what I've been working towards with this latest tranche of patches. Once we have a real CI (and I update the terrible and out-of-date user guide for Windows), I'm hoping Windows support can become "alpha" instead of just "experimental". I've been working on getting us ready for that day, and it doesn't seem too far off now.

PR #3111, opened 2026-05-24
Resolve JIT compiler tools portably

thomthehound:
The Python JIT path currently looks up `aiecc` directly from `PATH`, which is flaky on native Windows.

This tiny patch routes `aiecc` lookup through `python/utils/config.py` and makes the configured compiler-tool names use `.exe` on Windows.


hunhoffe (approved):
Thank you for another contribution!

jgmelber (approved):

PR #3118, opened 2026-05-27
Update `xdna-driver` tag

andrej:
The latest tag on the `1.7` branch; contains the scratchpad memory Python bindings in PyXRT that will be required for #3061


fifield:
NPU runners are updated to the driver hash in the commit. ok to merge.

PR #3120, opened 2026-05-28
Select Shim Tile Used for Trace Packet Extraction

Victor-Jung:
@jgmelber 

Trace packet flows currently land on the shim at column 0 (hard-coded target of the `routing = single` strategy in `AIEInsertTraceFlows`). On a busy column whose shim channels are already taken by compute flows, this causes the design to be unrootable.

This PR adds a small IR-level knob to select which column to use (default is 0).

MLIR:
```mlir 
  aie.trace.host_config buffer_size = 8192 egress_shim_col = 4
```

Python:
```python
  trace_host_config(buffer_size=8192, egress_shim_col=4)
```


fifield (commented):
How does this relate to and interact with the lateral trace routing feature?

Victor-Jung:
> How does this relate to and interact with the lateral trace routing feature?

This feature is the equivalent to using the `--lateral-target-col` flag to set a column, but instead it's in the IR (cleaner IMO).

Currently the `--lateral-routing` pass can redirect the column set by `egress_shim_col`. I think this is the intended behavior of the `--lateral-routing` option. Maybe we could add an `emitRemark` when `--lateral-routing` changes which ShimDMA is used so that it's more obvious for the user.

jgmelber (approved):

PR #3137, opened 2026-06-03
Add fallback-path test for test_utils::bfloat16_t 

hunhoffe:
* Adds `test/CppTests/bfloat16_fallback_test.cpp`, a small standalone executable that exercises the `uint16`-storage fallback path of `test_utils::bfloat16_t` from #3110.                                                                                                                                                                                                      
* Built with `-U__STDCPP_BFLOAT16_T__` (or `/U` on MSVC) so the fallback branch is selected even on Linux toolchains that have native `std::bfloat16_t`.                                                                                                                                         
* Wired into the existing `check-aie-cpp` target — picked up automatically by `ninja check-aie` in CI.                                                                                                                                                                                         
                                                                                                                                                                                                                     
## Why                                                                                                                                                                                                             
The fallback only activates when the compiler doesn't define `__STDCPP_BFLOAT16_T__` (today: MSVC). Linux CI exercises the native branch and would not notice if the fallback's round-to-nearest-even arithmetic or specials handling broke. This protects against that.                                                                                                                                                    
                                                                                                                                      
Depends on #3110 — will rebase to drop those commits after it merges.


andrej (approved):
note: I only looked at [this commit](https://github.com/Xilinx/mlir-aie/pull/3137/changes/b1e684f8c6568620a56cbdaae785b17cfdb755ff), I think the rest of the changes in the changeset shown on GitHub are from #3110

hunhoffe:
Thanks @andrej ! you are correct, it's just the one commit.

PR #3151, opened 2026-06-08
wheels: fix nightly publish + #3140 follow-ups + pin vulnerable example deps

hunhoffe:
Three threads since #3072.

1. Fix nightly publish (release blocker) buildRyzenWheels.yml's publish job installs ci-tools.lock and runs pypi_rehearsal.sh but has no checkout step, so every nightly since #3072 fails at Could not open requirements file. A tagged-release push would build all 16 wheels then fail before Release. Adds a sparse-checkout matching mlirAIEDistro.yml.

2. Re-land #3140
- Strip mlir_aie/lib/*.lib on Windows — MSVC equivalent of the Linux *.a cleanup from #3072.
- Prune unused dialect bindings (spirv, omp, smt, shard, sparse_tensor, x86, amdgpu, shape, emitc, async). Kept pdl / irdl / nvgpu (pulled in transitively).
- Override egg_info so top_level.txt says mlir_aie (the installed package) instead of _mlir_aie (the CMakeExtension placeholder).
- Drop Python :: 3.10 classifier; python_requires is already >=3.11.
- Remove the dead dataclasses; python_version < "3.7" requirement and the build-time dataclasses dep.
- Linux size flags: -ffunction-sections -fdata-sections (compile) and -Wl,--gc-sections (link). MSVC's /OPT:ICF equivalent (lld --icf=safe) is unavailable in the manylinux container's GNU ld, so we keep just the dead-section strip.

3. Pin example requirements (scorecard hygiene)
OSV-scanner treats unbounded reqs like Pillow / torch / notebook as matching every version ever, inflating the public scorecard to "118 vulnerabilities." These files aren't part of the wheel build — golden generation and notebook demos only — so pinning is low impact.
- mobilenet/requirements_gen_golden.txt: pin torch / torchvision / tqdm / Pillow / brevitas / onnx
- python/requirements_ml.txt: pin torch
- python/requirements_notebook.txt: pin notebook
- magika/py/requirements.txt: bump onnx==1.16.1 → 1.21.0 (drops 12 GHSAs)


hunhoffe:
@erwei-xilinx this PR prunes some dialects which are unused at the mlir-aie level from the wheel in order to decrease wheel size. Before this lands, I want to make sure these dialects are not assumed by mlir-air.

erwei-xilinx (approved):
Thanks for pinning. I can confirm that AIR does not consume those dialects from mlir-aie wheel.

PR #3162, opened 2026-06-10
Add skills / collection with aie-kernel-opt guide

hunhoffe:
Introduce a top-level skills/ directory holding Agent Skills: self-contained, on-demand guides that capture measured, codebase-specific knowledge for coding agents and humans alike.

The first skill, aie-kernel-opt, is a standalone guide to optimizing AIE / Peano-compiled kernels: a measure-first methodology (baseline, bit-exact gate, ablation, verify-in-.o) plus a priority-ordered catalog of concrete optimization levers (loop hints, compile-time constants, divide elimination, branch-splitting, epilogue vectorization, operand-layout pre-pack, wide packing, wider mmul, DMA layout offload) and the constraint each one carries. Mechanisms are written generically across AIE generations; performance deltas are framed as illustrative optimization headroom on example kernels.

I derived this skill from my work in https://github.com/Xilinx/mlir-aie/pull/3142 I'm sure it could be improved when applied to a wider variety of kernels.

### DIFF
--- a/skills/README.md
+++ b/skills/README.md
@@ -28,20 +28,23 @@ We welcome contributions, additions, and amendments to existing skills.
 
 ## Using a skill
 
-**With Claude Code.** Symlink or copy a skill into a discovered skills
-location so it auto-loads:
+**With Claude Code.** Symlink this whole `skills/` directory into a
+discovered skills location once, and every skill here (current and future)
+auto-loads — no per-skill step:
 
 ```shell
 # Project-scoped (this repo only):
-mkdir -p .claude/skills
-ln -s ../../skills/aie-kernel-opt .claude/skills/aie-kernel-opt
+mkdir -p .claude
+ln -s ../skills .claude/skills
 
 # Or personal (all your projects):
-ln -s "$(pwd)/skills/aie-kernel-opt" ~/.claude/skills/aie-kernel-opt
+ln -s "$(pwd)/skills" ~/.claude/skills
 ```
 
 Claude Code scans `.claude/skills/` (walking up to the repo root) and
-`~/.claude/skills/` at startup and loads each `SKILL.md` it finds.
+`~/.claude/skills/` at startup and loads each `SKILL.md` it finds. To pull
+in just one skill instead of the whole collection, symlink that single
+subdirectory (e.g. `mkdir -p .claude/skills && ln -s ../../skills/aie-kernel-opt .claude/skills/aie-kernel-opt`).
 
 **With any other agent, or by hand.** A `SKILL.md` is plain Markdown — read
 it directly, or point your agent's instructions file at the relevant skill.
--- a/skills/README.md
+++ b/skills/README.md
@@ -1,0 +1,59 @@
+<!--
+This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+See https://llvm.org/LICENSE.txt for license information.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+Copyright (C) 2026, Advanced Micro Devices, Inc.
+-->
+
+# Agent Skills
+
+This directory is a collection of [Agent Skills](https://code.claude.com/docs/en/skills):
+self-contained guides that capture knowledge about
+working in this codebase, packaged so that coding agents (and humans) can
+load them on demand.
+
+Each skill lives in its own subdirectory containing a `SKILL.md` file with
+YAML frontmatter (`name` + `description`) followed by the guide itself. The
+`description` is what an agent reads to decide whether a skill is relevant
+to the task at hand.
+
+We welcome contributions, additions, and amendments to existing skills.
+
+## Available skills
+
+| Skill | What it covers |
+|-------|----------------|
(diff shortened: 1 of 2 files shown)


skills/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Using a skill
+
+**With Claude Code.** Symlink or copy a skill into a discovered skills
+location so it auto-loads:
+
+```shell
+# Project-scoped (this repo only):
+mkdir -p .claude/skills
+ln -s ../../skills/aie-kernel-opt .claude/skills/aie-kernel-opt
```
andrej: Are we going to have to extend this for each added skill or could we just symlink the whole directory?

skills/aie-kernel-opt/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+The deltas cited are illustrative results from hand-optimizing example
+INT8 kernels (convs, 1x1 projections, depthwise, GEMM, softmax-style
+attention) — they show the headroom a given lever can unlock relative to a
+straightforward first implementation, not a benchmark of any shipping
+product. The *mechanisms* are general to AIE/Peano; the conv-shaped
+examples are just illustrations — read "inner reduction loop", "output
+loop", "epilogue" generically for your kernel. Concrete vector widths and
+`mmul<...>` shapes in the examples reflect one AIE generation (the numbers
+differ across AIE/AIE-ML/AIE2P); take them as worked examples and check the
+aie-api headers for your target's actual widths.
```
andrej, line 30: Would pointing to the AIE API path here help the agent spend less tokens going to look for where said API is? I think we clone it in third_party.

skills/aie-kernel-opt/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+
+The deltas cited are illustrative results from hand-optimizing example
+INT8 kernels (convs, 1x1 projections, depthwise, GEMM, softmax-style
+attention) — they show the headroom a given lever can unlock relative to a
+straightforward first implementation, not a benchmark of any shipping
+product. The *mechanisms* are general to AIE/Peano; the conv-shaped
+examples are just illustrations — read "inner reduction loop", "output
+loop", "epilogue" generically for your kernel. Concrete vector widths and
+`mmul<...>` shapes in the examples reflect one AIE generation (the numbers
+differ across AIE/AIE-ML/AIE2P); take them as worked examples and check the
```
andrej, line 29: Would an agent benefit from a table about the different architecture names or are they easily inferable from the code?

````
internal arch. name(s) | marketing arch name | chip codenames with this architecture
aie | AIE | Versal VCK5000
aie2 == NPU1 | AIE-ML == XDNA | Phoenix
aie2p == NPU2 | XDNA2 | Strix Point, Strix Halo, Krackan
aie2ps | AIE-MLv2 | Telluride
```

skills/aie-kernel-opt/SKILL.md
```diff
@@ -0,0 +1,356 @@
...
+5. **Verify the mechanism in the `.o` before believing your cycle-model.**
+   In practice, *both* "obvious" explanations for a given regression
+   (register spill; doubled scalar gather) have turned out wrong on
+   objdump inspection. A regression you can't see in the codegen is one you don't
+   understand yet. Useful checks:
+   - `llvm-nm --print-size build/X.o` — spill ≈ noticeably bigger function.
+   - `llvm-objdump -d build/X.o | grep -coE '\bv(lda|ldb|mac|mul|st)\b'` —
+     confirm vector ops actually appear (scalar-only `mac`/`lda`/`st` means
+     the kernel didn't vectorize at all — fix that before anything else).
+   - `llvm-nm build/X.o | grep __div` — should be empty (see lever #3).
```
andrej, line 79: Disassembly would arguably be most helpful, at least to me as a human, and it's not listed as one of the commands here. Things to look at, especially: inserted no-ops and total number of instructions in the zero-overhead loop. The beautiful thing about our architecture is that there's no stalling, so the number of instructions in the assembly divided by clock frequency corresponds to execution time (as far as I understand). So if we're compute bound, there's no need for hardware testing, just looking at the instructions should be enough to predict execution time.  Here's something on how to get the assembly out from my notes:

> When compiling using peano, you can disassemble the generated file using ```$PEANO_INSTALL_DIR/bin/llvm-objdump -d -r <file.o>```. The -r will print relocations, which is important for identifying the start/end of zero-overhead loops.

and

> Using peano, instead of compiling and then disassembling, you can instead also directly compile into assembly:
> ```${PEANO_INSTALL_DIR}/bin/clang++ -O2 -std=c++20 -I${MLIR_AIE_INSTALL_DIR}/include --target=aie2-none-unknown-elf -c my_kernel.cc -S -o kernel.s```
> And then, if needed, compile the assembly into the final object file:
> ```${PEANO_INSTALL_DIR}/bin/clang++ -O2 -std=c++20 -I${MLIR_AIE_INSTALL_DIR}/include --target=aie2-none-unknown-elf -c kernel.s -o kernel.o```

An explanation of the concept of a zero-overhead loop might also be useful. I'm not sure if that's general knowledge the agent already has or if it is AIE specific.

andrej (approved):

PR #3168, opened 2026-06-12
[WIP][Draft] Dynamic runtime sequences via standalone TXN encoding + EmitC C++ generation

hunhoffe:
Lets an NPU control program (runtime sequence) be **compiled once and run at different problem configurations**, by emitting the transaction (TXN) program as C++ that builds the binary at runtime instead of baking fixed sizes into it.

**Primary: dynamic TXN generation.**
- DMA buffer-descriptor sizes/strides/offsets lower to **SSA operands** instead of compile-time constants. `npu.dma_memcpy_nd` and `aie.dma_bd` gain a dynamic lowering path; static and dynamic paths share BD-word emission and TXN col/row encoding so the two stay byte-identical.
- A standalone, dependency-free **TXN encoding library** (`TxnEncoding.h`) plus an **AIEX→EmitC** conversion emit the control program as compilable C++; the TXN binary is assembled at run time from runtime scalar arguments. Verified byte-equivalent to the binary path (`cpp_static_vs_dynamic_txn`).
- `npu.sync` / `npu.write32` / `npu.maskwrite32` / `npu.address_patch` / `npu.rtp_write` converted to SSA-only operands (attributes deprecated), enabling the dynamic path end-to-end.
- Demonstrated by a **dynamic GEMM** (M/K/N as runtime values) and dynamic passthrough, driven through the high-level IRON compile flow.

**Secondary: IRON improvements.**
- *Eager runtime sequences.* The runtime sequence is now an eagerly-executed Python callable (like `Worker.core_fn`) rather than a replayed flat task list, enabling `range_`/`if_` and SSA scalar args in the body — the frontend half of the dynamic-sizing story.
- *Redesigned hardware trace API.* `TileTrace` (per-tile sources, with the hardware unit inferred from each event's type) + `TraceBuffer` (the single shared output buffer, renamed from `TraceConfig`), declared on Worker / `Program(trace_tiles=, trace=)`. Replaces `Runtime.enable_trace`, whose in-body state-setting ran after the trace-emitting code read it — dead config that silently produced empty trace buffers. The new ownership makes that ordering bug structurally impossible.

  ## Notable details
  
- "A tile may be traced at most once per hardware unit" is enforced in `-aie-insert-trace-flows` (authoritative across all frontends).
- `static_assert`s pin the TXN opcodes shared between `AIETargetNPU.cpp` and `TxnEncoding.h` so they can't drift silently.
- `TileTrace`/`TraceBuffer` are exported from `aie.iron` with no new dependency (iron already imported `aie.utils`; its pyxrt import is soft-guarded, so iron stays importable without pyxrt).

  ## Testing
  
- FileCheck coverage for static-vs-dynamic BD lowering, EmitC TXN C++ generation, blockwrite fusion, and static/dynamic TXN equivalence.
- Unit tests for `TileTrace` inference/validation + a compiler test for the duplicate-unit check.
- Dynamic GEMM / passthrough lit tests on NPU1 and NPU2.
- Trace verified on NPU2 hardware (forced fresh compile): non-empty buffer, all four trace units (core/core-mem/mem-tile/shim) emitted and parsed.

  ## Migration (breaking)

- npu ops (`sync`/`write32`/`maskwrite32`/`address_patch`/`rtp_write`) take SSA operands instead of attributes.
- `Runtime.enable_trace(...)` → `Worker(trace=TileTrace(...))` + `Program(trace=TraceBuffer(...), trace_tiles=[...])`; `TraceConfig` → `TraceBuffer`; `Worker(trace=<int>)` → `Worker(trace=TileTrace())`.
- `with rt.sequence(...) as (...)` / `rt.start(w)` → `rt.sequence(fn, [types])` with workers passed to `Program(workers=[...])`.


programming_examples/basic/passthrough_kernel/passthrough_kernel.py
```diff
@@ -59,18 +61,36 @@ def core_fn(of_in, of_out, pass_through_line):
...
+        # Runtime-parameterized transfer length: `buffer_length` is an SSA
+        # value of the runtime sequence (not baked in at compile time), so a
+        # single compiled design serves any size up to `n`.  Consumed by the
+        # aiecc TXN-C++ flow (see passthrough_kernel_dynamic.py); this path
+        # emits MLIR for that flow rather than executing on the NPU.
+        def sequence(a_in, b_out, buffer_length):
+            of_in.prod().fill(a_in, sizes=[1, 1, 1, buffer_length])
+            of_out.cons().drain(b_out, sizes=[1, 1, 1, buffer_length], wait=True)
+
+        rt.sequence(sequence, [vector_type, vector_type, T.i32])
```
hunhoffe: Shouldn't only the rt.sequence() line change? Cant' you always def but then only use based on flag?

programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config1/n32_core.py
```diff
@@ -229,49 +230,41 @@ def core_fn(in_a, in_b, out_c, zero, matmul):
...
             if slot_idx == 3:
-                rt.finish_task_group(slots[0])
-                rt.finish_task_group(slots[1])
+                slots[0].resolve()
+                slots[1].resolve()
 
         # Drain the two slots still pending at the end of the rotation.
-        rt.finish_task_group(slots[2])
-        rt.finish_task_group(slots[3])
+        slots[2].resolve()
```
hunhoffe, line 257: I don't like there being an API where the user has to call resolve on ANYTHING other than the final program. That's too manual!

programming_examples/ml/magika/group0.py
```diff
@@ -143,14 +145,19 @@ def group0b_body(of_di, of_do, lut_a, lut_b, kernel):
...
-        rt.fill(of_din_L3L2.prod(), A)
-        rt.drain(of_dout_L2L3.cons(), C, wait=True)
-
-    return Program(iron.get_current_device(), rt).resolve_program()
+
+    def sequence(A, C, _unused0):
+        of_din_L3L2.prod().fill(A)
+        of_dout_L2L3.cons().drain(C, wait=True)
+
+    rt.sequence(sequence, [transfer_in_ty, transfer_out_ty, scalar_ty])
```
hunhoffe: Can we use typed function declarations for the sequence to get the types instead of having to do this? Do we ever do anything other than rt = Runtime().... rt.sequence()... This seems overly verbose for a simple runtime registration.

python/dialects/aie.py
```diff
@@ -10,6 +10,68 @@
...
+    bd_id=None,
+    packet=None,
+    burst_length=None,
+    offset_parameter=None,
+    offset_state_table_idx=None,
+    next_bd_id=None,
+    loc=None,
+    ip=None,
+):
+    """Backward-compatible wrapper for the auto-generated `aie.dma_bd` op.
```
hunhoffe: Why do we need a backwards compatible wrapper? Can't we just support the current version?

python/dialects/aiex.py
```diff
@@ -39,7 +45,28 @@
...
 
-npu_sync = partial(npu_sync, column_num=1, row_num=1)
+# The generated npu_sync builder takes its six parameters as SSA i32 operands.
+# This wrapper accepts plain Python ints (materialized as arith.constant) or
+# existing SSA Values, so call sites can pass compile-time-known values
+# directly. column_num/row_num default to 1.
+_npu_sync_gen = npu_sync
+
+
+def _as_i32(v):
```
hunhoffe: Do we really need this? Can't we do auto conversion somehow?

python/dialects/aiex.py
```diff
@@ -54,6 +81,40 @@ def dma_wait(*args: ObjectFifoCreateOp | str):
         npu_dma_wait(str_name)
 
 
+def _cast_to_i64(v):
```
hunhoffe: This seems silly. I think there are already registered type converters for mlir/dialects. Can't we reuse something that already exists?

python/dialects/aiex.py
```diff
@@ -145,6 +214,55 @@ def __init__(
...
+# SSA values. These wrappers accept plain Python ints (materialized as
+# arith.constant) or existing SSA Values, so call sites can pass
+# compile-time-known values directly or runtime sequence values for
+# runtime-parameterized sequences.
+_npu_write32_gen = npu_write32
+_npu_maskwrite32_gen = npu_maskwrite32
+_npu_address_patch_gen = npu_address_patch
+
+
+def npu_write32(
```
hunhoffe, line 213: Same. I don't like adding more wrappers unless we REALLY need them.

python/iron/dataflow/objectfifo.py
```diff
@@ -528,6 +528,142 @@ def all_of_endpoints(self) -> list[ObjectFifoEndpoint]:
             is_prod=True
         ) + self._object_fifo._get_endpoint(is_prod=False)
 
+    def fill(
```
hunhoffe, line 531: It seems sillly to specify tile for fill/drain... shouldn't you specify the tile of the endpoint instead?

python/iron/runtime/_context.py
```diff
@@ -0,0 +1,43 @@
...
+``Runtime``, which keeps the call sites free of plumbing and avoids importing
+``Runtime`` into the dataflow / buffer modules (circular).
+"""
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+
+_ACTIVE_SEQUENCE: ContextVar = ContextVar("active_runtime_sequence", default=None)
+
+
```
hunhoffe, line 25: Are you sure this is the right way to do this? Do we REALLY need a sequence context? Couldn't we rely on the mlir module context during the call to program.resolve_program()?

python/iron/runtime/data.py
```diff
@@ -60,3 +61,46 @@ def op(self, op: MemRefValue):
         if self._op:
             raise ValueError("Cannot set operation for RuntimeData more than once.")
         self._op = op
+
+    def reset_op(self) -> None:
+        """Clear the bound SSA value so the handle can be rebound on a fresh build."""
+        self._op = None
+
+
+class RuntimeScalar:
```
hunhoffe: I thought we had methods for auto converison. Why do we need a new type???

python/iron/runtime/dmatask.py
```diff
@@ -4,28 +4,31 @@
...
 class DMATask(RuntimeTask):
     def __init__(
         self,
         object_fifo: ObjectFifoHandle,
         rt_data: RuntimeData,
-        tap: TensorAccessPattern,
-        task_group: RuntimeTaskGroup | None = None,
+        tap: TensorAccessPattern | None = None,
         wait: bool = False,
+        offset=None,
```
hunhoffe: Why can't we use tap here? sizes/strides is unnecessarily adding optional arguments.

python/iron/runtime/dmatask.py
```diff
@@ -65,20 +78,121 @@ def fifo(self) -> ObjectFifoHandle:
...
         ip: ir.InsertionPoint | None = None,
+        bd_id: int | None = None,
     ) -> None:
+        if self.uses_direct_npu_dma():
+            if bd_id is None:
+                raise ValueError("Direct NPU DMA lowering requires an assigned bd_id.")
+            self._bd_id = bd_id
+
+            if self._tap is not None:
+                self._task = npu_dma_memcpy_nd(
```
hunhoffe: Why are we lowering via npu_dma_memcpy_nd instead ofthe task lists?????

python/iron/runtime/dmatask.py
```diff
@@ -65,20 +78,121 @@ def fifo(self) -> ObjectFifoHandle:
...
+                    self._rt_data.op,
+                    tap=self._tap,
+                    issue_token=self._wait,
+                )
+                return
+
+            sizes = self._resolve_runtime_values(self._sizes)
+            if sizes is None:
+                raise ValueError(
+                    "Direct NPU DMA lowering requires explicit sizes or a tap."
```
hunhoffe: We should still allow defaults!!!

python/iron/runtime/runtime.py
```diff
@@ -5,11 +5,32 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # (c) Copyright 2024-2026 Advanced Micro Devices, Inc.
-"""Runtime: orchestrates host-side data movement and worker execution for an IRON program."""
+"""Runtime: orchestrates host-side data movement and worker execution for an IRON program.
+
+A :class:`Runtime` owns a single runtime *sequence* — the host-side program that
```
hunhoffe: If a single runtime owns a single sequence, they seems unecessarily complex.

python/iron/runtime/runtime.py
```diff
@@ -21,65 +42,224 @@
...
-    def __init__(
-        self,
-        strict_task_groups: bool = True,
-    ) -> None:
-        """Initialize a runtime object.
+    def __init__(self, runtime: "Runtime", device, device_ip: ir.InsertionPoint):
+        self._runtime = runtime
+        self._device = device
+        self._device_ip = device_ip
+        self._bd_id_allocator = _RuntimeBdIdAllocator()
```
hunhoffe: This is duplicated work the compiler does for us if we use the DMATask API.

python/iron/runtime/runtime.py
```diff
@@ -21,65 +42,224 @@
...
+
+    # -- transfer emission (called by ObjectFifoHandle.fill/drain) -------------
+
+    def record_touched_fifo(self, handle: ObjectFifoHandle) -> None:
+        """Record that the runtime touches this fifo, for resolution after the body."""
+        of = handle._object_fifo
+        if of not in self._touched_fifos:
+            self._touched_fifos.append(of)
+
+    def emit_transfer(self, task, group) -> None:
```
hunhoffe: Why isn't this just a special resolve() method? Why do we need separate "emit_transfer" functions???

python/iron/program.py
```diff
@@ -16,28 +16,95 @@
 
 from .device import Device
 from .runtime import Runtime
+from .dataflow import ObjectFifoHandle
 from .scratchpad_parameter import ScratchpadParameter
 from .resolvable import Resolvable
 from ..utils import trace as trace_utils
 
 
+def _fifo_endpoints_ready(handle, _seen=None) -> bool:
```
hunhoffe, line 52: I think this is so complicated because the Python wrappers were written before Logical tiles exist. If we remove the IRON placer API, and just emitted logical tiles (sometimes with coordinates) would that make program resolution simpler?

python/iron/worker.py
```diff
@@ -184,6 +188,22 @@ def buffers(self) -> list[Buffer]:
         """
         return self._buffers.copy()
 
+    def resolve_barrier_locks(self) -> None:
```
hunhoffe, line 192: Why is this needed again? Is this something the compiler can handle for us instead of the Python frontend?

python/utils/trace/parse.py
```diff
@@ -430,19 +430,28 @@ def parse_mlir_trace_events(mlir_module_str, colshift=None):
         target_model = aiedialect.get_target_model(device)
         events_module = get_events_for_device(str(device))
 
+    def const_of(ssa_value):
```
hunhoffe: Why is this needed? I don't want to be doing this kind of compiler work in the python frontend.

python/utils/trace/tiletrace.py
```diff
@@ -0,0 +1,159 @@
...
+    if tile_op.is_core_tile():
+        return "core"
+    if tile_op.is_mem_tile():
+        return "mem"
+    if tile_op.is_shim_tile():
+        return "shim"
+    raise ValueError(f"Cannot trace tile {tile_op}: unknown tile class")
+
+
+class TileTrace:
```
hunhoffe, line 82: Does this belong in IRON? Does it need to be resolvable?

python/utils/regdb.py
```diff
@@ -551,25 +551,47 @@ def get_address_from_op(
...
+                attr = owner_op.value
+                if hasattr(attr, "value"):
+                    return int(attr.value)
+            return None
+
+        # Get address - SSA operand on write32/maskwrite32, else 'address'/'addr'
+        # attribute on the other ops.
         if hasattr(op, "address") and op.address is not None:
-            address = int(op.address.value)
+            if hasattr(op.address, "value"):
```
hunhoffe: Can't we, like, figure out how to do this more simply? The if/else everywhere seems.... bad.

runtime_lib/CMakeLists.txt
```diff
@@ -133,6 +133,20 @@ foreach(target ${AIE_RUNTIME_TARGETS})
       TEST_BEFORE_INSTALL true
       TEST_EXCLUDE_FROM_MAIN true
     )
+  else()
+    # Even when the full test_lib ExternalProject is disabled (e.g. wheels
+    # build without VITIS or LibXAIE), the header-only payload that the
+    # programming examples rely on (cxxopts.hpp, test_utils.h,
+    # xrt_test_wrapper.h) must still be installed so host-side test
+    # executables can compile against the install tree.
+    if (NOT WIN32 OR ${target} STREQUAL "x86_64")
```
hunhoffe, line 147: I don't fully understand this guard. Why not for windows?

test/aiecc/Inputs/static_vs_dynamic_txn/compare_main.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <algorithm>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <vector>
+
+std::vector<uint32_t> static_txn();
+std::vector<uint32_t> dynamic_txn(int32_t n);
+
+int main(int argc, char **argv) {
```
hunhoffe, line 28: What is this file? Is this just a test file that was accidentally committed?

test/aiecc/Inputs/static_vs_dynamic_txn/gen_dynamic.cpp
```diff
@@ -0,0 +1,24 @@
...
+// more static wrappers.
+//
+//===----------------------------------------------------------------------===//
+
+#include "dynamic_txn.h"
+
+#include <cstdint>
+#include <vector>
+
+std::vector<uint32_t> dynamic_txn(int32_t n) {
```
hunhoffe, line 22: why wrapper?

hunhoffe (commented):
Initial review.

hunhoffe:
Succeeded by discussion https://github.com/Xilinx/mlir-aie/discussions/3222 and associated PRs.

PR #3169, opened 2026-06-12
aie_api: move submodule to public Xilinx/aie_api 2026.1 release

hunhoffe:
Repoint the aie_api submodule from the jgmelber/aie_api stopgap mirror to the now-public Xilinx/aie_api repository and bump to the 2026.1 release (bec000f). The fork was introduced in #1900 only because no public upstream existed at the time; that is no longer the case.

The 2026.1 headers are a strict superset of the pinned mirror (no files removed) and add AIE2PS support plus all fixes since Jan 2025. The fork's only local carry was commenting out the ADF header include under __AIENGINE__, which is dead code in the Peano flow since __AIENGINE__ is never defined, so it is dropped by moving to clean upstream.


jgmelber:
It would be good to bump this but I think we still need my work with the `adf.h` patch

hunhoffe:
Closed in favor of: https://github.com/Xilinx/mlir-aie/pull/3260

PR #3171, opened 2026-06-13
pyright type checking

hunhoffe:
Generally, open source python projects use:
* black
* pyright
* ruff

We already use black for formatting, this integrates the pyright type checker. This is a step towards a more disciplined code base where the code style + type declarations of contributions can be validated automatically in the CI instead of via manual review. As you can see in this PR, pyright found a lot of small things to fix!


python/iron/dataflow/flow.py
```diff
@@ -26,16 +26,16 @@
 from dataclasses import dataclass
 from typing import Sequence
 
-from ... import ir  # type: ignore
-from ...dialects._aie_enum_gen import (  # type: ignore
+from ... import ir  # pyright: ignore[reportMissingImports]
+from ...dialects._aie_enum_gen import (  # pyright: ignore[reportMissingImports]
```
andrej, line 30: Are these ignores so local builds don't report them as errors? I see that you added some code in the GitHub workflow that, as I understand it, is supposed to make it so these imports should resolve.

python/iron/device/device.py
```diff
@@ -108,87 +111,6 @@ def get_compute_tiles(self) -> list[Tile]:
...
-        if tile.row == 0:
-            if output:
-                return self.get_num_source_shim_mux_connections(tile)
-            else:
-                return self.get_num_dest_shim_mux_connections(tile)
-        if output:
-            return self.get_num_source_switchbox_connections(tile)
-        else:
-            return self.get_num_dest_switchbox_connections(tile)
-
```
andrej, line 191: Assuming these are leftovers from when we had the SequentialPlacer in Python and are now no longer needed?

andrej (approved):

python/iron/dataflow/flow.py
hunhoffe: The imports at the top level kinda resolve but the generated MLIR bindings do some `import *` type things within the module that pyright (and all type checkers, really) don't like.

python/iron/device/device.py
hunhoffe: That is correct. They were found to be essetially dead (and also not type compliant) so I decided to remove them rather than fix the types.

PR #3174, opened 2026-06-15
[iron] Expose multi-dim DMA strides (Bd.dimensions) and cross-core L1 buffer sharing

hunhoffe:
Two small, independent fixes to IRON. Both expose capabilities the `aie.*` dialect already supports but that IRON silently dropped or rejected.

**1. `Bd.dimensions` — multi-dim (strided) DMA access patterns.** `aie.dma_bd` accepts a `dimensions` attribute (list of `(size, stride)` tuples), but `Bd` had no field for it and `TileDma.resolve` never forwarded one, so IRON could only express contiguous transfers. Adds the field; `None` (default) keeps the existing behavior.

**2. Worker — allow an explicitly-tiled Buffer to be shared (read) across Workers.** AIE cores can read a neighbor tile's L1 directly, so a producer core's buffer can be a read-only input to a consumer on an adjacent tile — and the SequentialPlacer already models this (`buildBufferAdjacency` + `isLegalMemAffinity`). IRON's Worker guard rejected it anyway. Now a Buffer pinned to an explicit tile may be referenced by later Workers (first owns/places it, rest are non-owning readers); auto-placed buffers are still rejected to avoid ambiguous placement. Adds `Buffer._explicit_tile` to track user intent, since the owner auto-pins `_tile` and so `_tile` alone can't signal it.

### DIFF
--- a/python/iron/buffer.py
+++ b/python/iron/buffer.py
@@ -62,6 +62,12 @@ class Buffer(Resolvable):
             self._name = f"buf_{next(Buffer._gbuf_index)}"
         self._use_write_rtp = use_write_rtp
         self._tile = tile
+        # Whether the user pinned this Buffer to an explicit tile at
+        # construction.  A Worker may auto-pin ``_tile`` later as a
+        # convenience, so ``_tile is not None`` is not a reliable signal of
+        # user intent; this flag is.  Only explicitly-placed Buffers may be
+        # shared (read) across Workers — see :class:`Worker`.
+        self._explicit_tile = tile is not None
         self._owner_worker = None
 
     @property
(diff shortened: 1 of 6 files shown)


erwei-xilinx (approved):

andrej (approved):
LGTM. Question about the shared buffer thing: In the future, are we going to want to allow shared buffers on unplaced Workers as well? It would just be another constraint on placement, but not sure if it's supported to express that and/or if we are interested in supporting this use case.

hunhoffe:
@andrej Good question! There's already some modeling of adjacency constraints between tiles, so it might just "work" with a little more plumbing but I wanted to start with the simple case for now.

PR #3181, opened 2026-06-18
requirements_dev: pin nanobind==2.12.0 (ABI-coupled to mlir distro)

hunhoffe:
The Ryzen AI Software workflow installs the loose requirements_dev.txt (cp310-only SDK can't use the py>=3.11 lock), so nanobind>=2.12.0 auto-floated to 2.13.0 when it published, mismatching the prebuilt mlir distro wheel's nanobind ABI and breaking Value-subclass recognition (TypeError in arith.py at runtime across all Python tests). Pin exact so the source matches the lock and only moves alongside a distro rebuild.

### DIFF
--- a/python/requirements_dev.txt
+++ b/python/requirements_dev.txt
@@ -13,7 +13,10 @@ cibuildwheel
 patchelf; platform_system == "Linux"
 pre-commit>=4.6.0,<5.0
 black[jupyter]
-nanobind>=2.12.0
+# nanobind ABI-couples to the prebuilt mlir distro wheel (built against this
+# same version); a mismatch breaks Value-subclass recognition at runtime. Pin
+# exact and bump only alongside a distro rebuild.
+nanobind==2.12.0
 lit
 matplotlib
 psutil


thomthehound:
bingo.

hunhoffe:
@thomthehound I'll try to get this merged soon so I can get all the other PRs unblocked!

thomthehound:
Word.

I think this might be a warning to us, though; CI might need a migration to a newer RAI.

hunhoffe:
Yeah, my thought too but there are some complications with the packaging from what I understand, that might make the upgrade more difficult than anticipated.

thomthehound:
> Yeah, my thought too but there are some complications with the packaging from what I understand, that might make the upgrade more difficult than anticipated.

Yup. I'm not excited about it. Whenever somebody tells me that they hate Python, my answer is always "I agree with you." 

But I think CI is due for an overhaul, regardless. Maybe by the end of the summer we can figure something out. At least one of the the things working in our favor is that we don't even formally support 3.10 anyway.

jgmelber (approved):

Issue #3180, opened 2026-06-18
[RFC] Code Cleanup


hunhoffe:
Time for some spring cleaning! Before I delete anything, though, I'd like to give a chance for feedback.

The two cleanup PRs that I am proposing so far are:
* https://github.com/Xilinx/mlir-aie/pull/3190
* https://github.com/Xilinx/mlir-aie/pull/3192
* https://github.com/Xilinx/mlir-aie/pull/3193
* https://github.com/Xilinx/mlir-aie/pull/3179

Please feel free to add PR's to this effort or suggest further candidates for cleanup.

Edit: A few additions
* https://github.com/Xilinx/mlir-aie/pull/3189
* https://github.com/Xilinx/mlir-aie/pull/3183
* https://github.com/Xilinx/mlir-aie/pull/3195

hunhoffe:
And one more: https://github.com/Xilinx/mlir-aie/pull/3183

jsetoain:
It's worth noting that some of these tests (like the ones in `Integration/Dialect/AIEVec`) can be converted to versions for Peano and the LLVM IR backend, which is preferable to leave whole workflows untested; our test suite is already severely lacking as it is.

To facilitate this, I've created the following tag: [last-legacy-vectorize](https://github.com/Xilinx/mlir-aie/tree/last-legacy-vectorize), right before any of these changes landed. Before closing this issue, we must open a new one pointing here, to keep a record of the need for that effort.

hunhoffe:
@jsetoain Would you be willing to create that issue now, and/or work on porting tests to peano as makes sense? Do those tests belong in this repo and not the peano repo? It would be great to have the test gaps clearly articulated.

jsetoain:
> [@jsetoain](https://github.com/jsetoain) Would you be willing to create that issue now, and/or work on porting tests to peano as makes sense? Do those tests belong in this repo and not the peano repo? It would be great to have the test gaps clearly articulated.

- Happy to create the issue myself, but won't monitor it
- I won't be working on it
- These are mlir-aie tests, but in order to perform integration tests you need e2e compilation, and that requires a backend compiler (Peano, in this case)

hunhoffe:
And another: https://github.com/Xilinx/mlir-aie/pull/3307

hunhoffe:
@jsetoain Yes, if you could create the issue with as much detail as possible as to what solution/capabilities you would see as a desirable solution in this area, that would be very helpful!

[...]

PR #3183, opened 2026-06-18
[python] Code cleanup: remove dead files and tests (#3180)

hunhoffe:
Part of the spring cleaning effort tracked in #3180. Removes four dead things from the `python/` tree:
  
- **`test/python/aiex_ops.py`**: Drop the `getTileOp` test and its two unused imports (`arith`, `T`). Mirrors the removal of `aiex.getTile` in #3179 — no lowering path, no callers.
- **`python/ILP_solution.py`**: 2021-era experimental Gurobi ILP solver for tiling space exploration. Never imported, not installed, not tested.
- **`python/aiert.py`**: Broken stub — imports from `_mlir_libs._aiert`, but the backing `AIERTModule.cpp` was deleted in #2405, so `_aiert.so` is never built. Also removes the dead `declare_mlir_python_sources(AIEPythonSources.AIERT ...)` entry from `python/CMakeLists.txt`.
- **`python/xrt.py`**: Re-exports `XCLBin` from `_mlir_libs._xrt`, but no Python code in the repo imports `aie.xrt` or uses `XCLBin`. Removes it from `CMakeLists.txt`; the `_xrt.pyi` stub is kept since it documents the C extension's interface.


thomthehound:
I think we might be able to get rid of the python-bindings build target entirely, actually.

hunhoffe:
Good call @thomthehound — done in `1123e5f`. The in-tree `_xrt` binding (`XRTModule.cpp` / `XCLBin`) was the original "run on NPU from Python" path from #897, before the official `pyxrt` package existed. Nothing imports `aie.xrt` / `_xrt` / `XCLBin` anymore (verified in both mlir-aie and mlir-air, which now use `pyxrt`), so I removed the module, the `_xrt.pyi` stub, the extension/symlink machinery, and the orphaned `xrt_coreutil` imported target. 
  
One nuance: I kept the `AIE_ENABLE_XRT_PYTHON_BINDINGS` flag rather than removing it entirely — it still gates the live `_parameter_scratchpad` host module and the `pyxrt`-backed NPU test suite (~26 tests via the `xrt_python_bindings` lit feature). The `XRT_` prefix also leaves room for a future second runtime.

@erwei-xilinx — one externally-visible change to flag for you: this also drops the wheel-shipping of `libxrt_coreutil.so.2` (the copy/RENAME hack that was tied to the deleted `_xrt.dso` target). My understanding is `pyxrt` brings its own `libxrt` so this was redundant, but since it touches the wheel contents, wanted to confirm nothing downstream relies on that .so.2 being present in the wheel. Does that match your understanding?

erwei-xilinx (approved):

PR #3184, opened 2026-06-19
Bind IRON target device before JIT generation and caching

thomthehound:
This fixes a target-device ordering issue in the centralized IRON JIT path that was discovered while testing #3075.

IRON generators and external-kernel helpers choose different code paths depending on the active target device. The JIT path could discover the runtime device, but not bind it early enough for generation and cache lookup to see the **same** device that compilation and execution would later use. On NPU2 systems, that could leave part of the pipeline specializing for AIE2 while the rest targeted AIE2P.

This patch makes the ordering explicit:

* Resolve and bind the CLI-selected device before `run_design_cli()` callbacks.
* Bind the runtime-observed device in direct `@iron.jit` calls after argument validation, but before cache lookup, generation, and compilation.
* Cache generated MLIR per active device.
* Include the full IRON device identity in JIT cache keys.

With this change, architecture-sensitive generation, external-kernel selection, cache lookup, compilation, and execution all now agree on the same target device, and tests have been added to enforce that rule.

### DIFF
--- a/programming_examples/basic/packet_switch/packet_switch.py
+++ b/programming_examples/basic/packet_switch/packet_switch.py
@@ -27,7 +27,7 @@ packet ID, so use ``rt.inline_ops`` as the escape hatch there.
 Two invocation modes:
 
   * compile-only: ``... --op add --xclbin-path=PATH --insts-path=PATH``
-  * emit-MLIR:    ``... --op add --emit-mlir``
+  * emit-MLIR:    ``... --op add --dev npu --emit-mlir``
 """
 
 import argparse
--- a/programming_examples/basic/vector_vector_modulo/README.md
+++ b/programming_examples/basic/vector_vector_modulo/README.md
@@ -14,7 +14,7 @@ A simple binary operator: a single AIE core computes `c = a % b` element-wise on
 
 ## Source Files Overview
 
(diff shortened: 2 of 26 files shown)


thomthehound:
OK, that CI failure was actually extremely informative: it proved that the patch is working, but that it was not yet complete.

The overarching purpose of this update is to make the compilation/runtime contract IRON-clad, such that the intended code executes on the intended device. Previously, there were several leaks in this contract that caused fallback to `device=none` and, consequently, `npu1`, even when doing so was unsafe. These changes aim to plug those leaks without injuring cross-compilation or offline/static workflows.

Given how complicated this all is, I can now understand why there are rumors out of the Python Steering Council that broader CPython JIT support may not survive beyond 3.15 without a very convincing plan...

thomthehound:
@hunhoffe If you have the time, could you please kick off CI again on this? Sorry to bother you on the weekend. No worries if you don't.

hunhoffe:
"IRON-clad" -- I see what you did there!

hunhoffe (approved):
Thanks for the bug fix! When the tests/linting are stable, I'll work on getting this merged!

thomthehound:
> Thanks for the bug fix! When the tests/linting are stable, I'll work on getting this merged!

Always a pleasure! 
I think I just addressed the lint issue. All other tests came back green, so it's just the home stretch now!

thomthehound:
I have a new side-quest: it appears that our CI ccache mechanism is broken or outdated. `MLIR Distro` is a huge hog of these resources, and wastes time attempting to execute the broken mechanism. I learned a lot about GitHub workflows back in January/February, and I don't want that knowledge to go stale, so I am going to attempt a fix for this.

PR #3185, opened 2026-06-22
[wheels] Restore C++ dev artifacts needed by mlir-air

hunhoffe:
@erwei-xilinx flagged in #3072 that mlir-air builds against the mlir-aie wheel's C++ artifacts — specifically the AIE dialect headers (`include/aie/`), CMake config (`lib/cmake/`) for find_package(`AIE`), and static libraries (`*.a`) for `AIE`, `AIETransforms`, and `AIEX`.

This restores those three while keeping the removal of `include/aie-c/`, `include/bootgen_c_api.h`, and `include/xaienginecdo_static/` — confirmed absent from mlir-air's source tree.

Closes #3072 (comment): https://github.com/Xilinx/mlir-aie/pull/3072#issuecomment-4760813630

### DIFF
--- a/utils/mlir_aie_wheels/setup.py
+++ b/utils/mlir_aie_wheels/setup.py
@@ -321,22 +321,13 @@ class CMakeBuild(build_ext):
                 check=True,
             )
 
-            # C++-developer-only content: anyone pip installing this wheel
-            # reaches the toolchain through Python, never through native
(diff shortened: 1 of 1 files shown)


erwei-xilinx (approved):
Thanks for the swift response, @hunhoffe!

PR #3177, opened 2026-06-17
CI: detect newly skipped lit tests on PRs; remove dead test suites

hunhoffe:
### Unsupported Test Check

Adds a baseline-vs-current diff to the Run Tests and Run Examples jobs in buildAndTestRyzenAI.yml. On each PR, each job fetches the UNSUPPORTED list from the last successful run on the base branch, diffs against the current run, and posts a PR comment listing any tests that flipped.

False positives filtered out:
- npu-xrt/.*/test.py — always REQUIRES: dont_run by design; the real entry point is the sibling run.lit
- REQUIRES: aiesimulator — legitimately UNSUPPORTED on npu runners but passing on the Vitis runner

### Remove dead test suites
- test/Integration/ — AIEVec+chess+simulator pipeline, superseded by Peano, disabled in every CI job. Removed tests and the AIE_INCLUDE_INTEGRATION_TESTS cmake option.

PR #3194, opened 2026-06-22
CI: detect newly UNSUPPORTED lit tests on PRs

hunhoffe:
Adds a baseline-vs-current diff to the Run Tests and Run Examples jobs in `buildAndTestRyzenAI.yml`. On each PR, each job fetches the UNSUPPORTED list from the last successful run on the base branch, diffs against the current run, and posts a PR comment listing any tests that flipped.

False positives filtered out:
- `npu-xrt/.*/test.py` — always `REQUIRES: dont_run` by design; the real entry point is the sibling `run.lit`
- `REQUIRES: aiesimulator` — legitimately UNSUPPORTED on npu runners but passing on the Vitis runner

Split out from #3177 as requested by @jsetoain.


hunhoffe:
I like this idea but the implementation is currently too brittle/complicated. Closing for now.

PR #3211, opened 2026-06-23
aiecc: lower NPU instructions once for all devices on the full-ELF path

atassis:
## What
On the full-ELF / no-expand-load-pdis path, `generateNpuInstructions()` is
called once per device, and each call clones the whole module and re-runs the
module-level runtime-sequence materialize + NPU lowering pipeline, which makes
this phase O(devices^2).

Because the materialize pass and PDI-id assignment operate on the whole module
(independently of which device is later extracted), the module can be lowered
once after the per-core compile loop, then each device's runtime sequence
translated from that single lowered module via `AIETranslateNpuToBinary` (the
only device-specific step). This is done in a new `generateAllNpuInstructionsOnce`.

## Why
On large multi-device designs this redundant re-lowering dominates aiecc wall
time. Measured on a B=128, single-layer design (frozen MLIR, `-j16`):

- aiecc-only wall: **855.6s -> 362.6s (~58% reduction)**

## Behavior
Only the full-ELF / no-expand-load-pdis path is rerouted; the gate
(`generateFullElf && !expandLoadPdis && !generateCtrlPkt`) mirrors the original
per-device call's conditions exactly, so the expand-load-pdis (Phase B),
control-packet, and non-full-ELF paths are unchanged. Device selection
(`--device-name`), runtime-sequence filtering (`--sequence-name`), `patchPdiIds`,
output filenames, and `--keep-loc` sidecars are all preserved. The emitted
per-device instruction binaries are byte-identical to the per-device path.

## Testing
- Verified per-device output is bit-for-bit identical to the previous per-device
  path via a frozen-MLIR byte gate on the B=128 design (PASS).

Follow-up to #3178 (same super-linear compile-scaling pattern, different site).


andrej (commented):
This makes sense to me and could be a good performance improvement, but I wonder if we could achieve the same thing with less code duplication? At a glance, `generateAllNpuInstructionsOnce` and `generateNpuInstructions` seem very similar, and merging this would further blow up aiecc.cpp and require any future changes to the NPU lowering pipeline to be made in two places. Is there a reason not just to make `devName` in `generateNpuInstructions` optional, and adapt that function to emit it for all devices if no devName? (Genuinely asking, if that would be complicated to implement we can consider this.)

atassis:
@andrej agree with your call, the file is really too big to keep that as is. Made a change already, if you don't mind

jgmelber (approved):

atassis:
It looks like the tests are already broken on main

PR #3216, opened 2026-06-24
aiecc: build per-core compile slices from a stripped base module

atassis:
# aiecc: build per-core compile slices from a stripped base module

## What

`compileCores` fans out per-core compilation across worker threads. Each worker
re-parses the module text in its own `MLIRContext` and the per-core lowering
keeps only its own core, discarding the rest. The previous code serialized the
whole multi-core, multi-device module once and handed that same string to every
worker, so peak memory grew with (workers x full module) and every worker
re-parsed the runtime sequences, which the per-core lowering never uses but which
dominate the module text once the host DMA program is unrolled.

This builds a stripped base module once (drop the runtime sequences and every
other device), then derives each per-core slice by cloning that small base and
erasing the other cores. The per-core slice strings are byte-identical to the old
path, so the emitted object/ELF is unchanged; only the parse and clone cost
changes.

## Why it matters

Measured against upstream `main` at `1a5eed4` on a large multi-device design
(a batch-128 decode design: ~212 cores across ~21 devices), cold `aiecc`, `-j16`,
on an AMD Ryzen AI 9 465 (10 cores / 20 threads) with 32 GB RAM (Linux). The ELF
is byte-identical across all three rows (sha256 `370686d...`).

| aiecc configuration | aiecc wall (cold) | peak RSS |
|---|---|---|
| current upstream | did not finish within 30 min | ~18.7 GB |
| upstream + #3211 + #3212 | 139 s | 18.7 GB |
| upstream + #3211 + #3212 + this PR | 67 s | 1.7 GB |

On current upstream this design did not finish compiling within 30 minutes on
our hardware (it was still in the per-core phase, ~56 of ~212 cores after 27
minutes), which is why the companion compile-scaling fixes (#3211 lower-once,
#3212 getOrCreateDataMemref dedup) are needed to make it tractable at all. Its
peak RSS is the same ~18.7 GB as the second row, because the per-core memory
blowup this PR addresses is independent of those two fixes.

This PR's specific contribution is the last row: on top of #3211 + #3212 it
**halves the total aiecc wall (139 s -> 67 s)** and **cuts peak RSS ~11x (18.7 GB
-> 1.7 GB)**. The memory drop is the important part: without slicing, peak RSS
scales with worker-count x full-module-size, so it grows with both the design
size and the build parallelism.

## Correctness

The per-core slice is the lowering of that core's body plus its device-level
tiles/buffers/locks/objectfifos; it never references the runtime sequence
(host-side DMA orchestration) or any other device. The emitted per-core
object/ELF is therefore unchanged: the final whole-array ELF is byte-identical
(sha256 unchanged) before and after this change on the design above.

### DIFF
--- a/test/aiecc/Inputs/per_core_slice_lowering_equivalence.divergent.mlir
+++ b/test/aiecc/Inputs/per_core_slice_lowering_equivalence.divergent.mlir
@@ -1,0 +1,33 @@
+//===- per_core_slice_lowering_equivalence.divergent.mlir -----*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// Negative fixture for per_core_slice_lowering_equivalence.mlir. Same as the
+// stripped slice, except the target core stores a different value (123 instead
+// of 7), so its lowering diverges. A "not FileCheck" RUN asserts this does NOT
+// match the golden, which proves the golden actually pins the target core's
+// computation (guards the equivalence check against silently going vacuous).
+module @slice_divergent {
+ aie.device(npu1_2col) @device1 {
+  %t02 = aie.tile(0, 2)
+  %t12 = aie.tile(1, 2)
+  %l02 = aie.lock(%t02, 0)
+  %b02 = aie.buffer(%t02) { sym_name = "a02" } : memref<16xi32>
+  %l12 = aie.lock(%t12, 0)
+  %b12 = aie.buffer(%t12) { sym_name = "a12" } : memref<16xi32>
+  %c02 = aie.core(%t02) {
+    aie.use_lock(%l02, Acquire, 0)
+    %v = arith.constant 123 : i32
+    %i = arith.constant 3 : index
+    memref.store %v, %b02[%i] : memref<16xi32>
+    aie.use_lock(%l02, Release, 1)
+    aie.end
+  }
+ }
+}
--- a/test/aiecc/Inputs/per_core_slice_lowering_equivalence.slice.mlir
+++ b/test/aiecc/Inputs/per_core_slice_lowering_equivalence.slice.mlir
@@ -1,0 +1,33 @@
+//===- per_core_slice_lowering_equivalence.slice.mlir ---------*- MLIR -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2026, Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// Stripped slice for core (0, 2), produced the way compileCores produces it:
+// @device1 only (@device2 removed), the runtime sequence removed, and the other
+// core's aie.core op removed. The other core's tile, buffer, and lock are kept,
+// because compileCores erases only the aie.core ops. Driven by the RUN lines in
+// per_core_slice_lowering_equivalence.mlir.
+module @slice {
+ aie.device(npu1_2col) @device1 {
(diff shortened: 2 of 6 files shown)


jgmelber:
Is there some way to add tests for this?

atassis:
@jgmelber So far I've only tested this end-to-end, but a dedicated test makes sense here. My idea is a lowering-equivalence test: take one core and run the per-core lowering twice (aie-standard-lowering with its tile coords), once on the full multi-device module with the runtime sequences, and once on the stripped slice. The lowered core should come out identical both ways. That's really the property this PR relies on: stripping only removes stuff the per-core lowering was going to drop anyway, so the emitted core can't change. It's pure aie-opt, no hardware needed, so it won't be flaky. I can put that together if you think it's a good direction.

jgmelber:
> @jgmelber So far I've only tested this end-to-end, but a dedicated test makes sense here. My idea is a lowering-equivalence test: take one core and run the per-core lowering twice (aie-standard-lowering with its tile coords), once on the full multi-device module with the runtime sequences, and once on the stripped slice. The lowered core should come out identical both ways. That's really the property this PR relies on: stripping only removes stuff the per-core lowering was going to drop anyway, so the emitted core can't change. It's pure aie-opt, no hardware needed, so it won't be flaky. I can put that together if you think it's a good direction.

I think that sounds reasonable

hunhoffe:
Nice result re: memory — this could be a big deal for the large multi-device designs!
  
One suggestion on test coverage. The new `per_core_slice_lowering_equivalence.mlir` is a great check of the premise (a correctly-stripped slice lowers identically to the full module via aie-standard-lowering) but it validates the lowering invariant in aie-opt, not the new C++ stripping in `compileCores` itself. The existing e2e tests (`cpp_parallel_compilation`, `cpp_multi_core`, …) exercise the -j path, but they only assert smoke-level outcomes (Generated ELF, Compilation completed successfully) — none of them pin the headline claim that the emitted object is unchanged, and none compare the serial vs. parallel output. So a future regression in the slicing (wrong col/row predicate, an over-aggressive erase, etc.) that still produced some ELF would slip through.

A cheap way to close that: have `cpp_parallel_compilation.mlir` compile the same design both ways into separate tmpdirs and diff the per-core objects. Maybe something like this?

```diff
// RUN: aiecc --no-xchesscc --no-xbridge --verbose -j 2 %s | FileCheck %s

// CHECK: Successfully parsed input file
// CHECK: Device 'main' with 2 core(s)
// CHECK: Compiling 2 core(s) in parallel (2 threads)
// CHECK: Compiling core
// CHECK: Compiling core
// CHECK: Generated ELF
// CHECK: Generated ELF
// CHECK: Compilation completed successfully
+
+// Parallel per-core slicing must emit the same object as serial compilation.
+// Compile the same design both ways into separate tmpdirs and diff the per-core
+// objects: byte-identical objects pin the slicing equivalence claim.
+// RUN: aiecc --no-xchesscc --no-xbridge --tmpdir=%t.ser %s
+// RUN: aiecc --no-xchesscc --no-xbridge -j 2 --tmpdir=%t.par %s
+// RUN: diff %t.ser/main_core_0_2.o %t.par/main_core_0_2.o
+// RUN: diff %t.ser/main_core_1_2.o %t.par/main_core_1_2.o
```

I tried this locally against main + this branch and it holds up. What are your thoughts @atassis ?

atassis:
@hunhoffe makes sense, thanks! Added your diff already

atassis:
Also, while you're in this area - the companion fix referenced here, #3212, is still open without a review. It's the last of the compile-scaling set now that #3211 landed, so it'd be great to get it in alongside this.

hunhoffe:
@atassis I think I'm almost out for the day. I'll try to get this one in, but #3212 will probably need to wait until next week. It is not forgotten, though! 

As a heads up, since you seem pretty active in this space, most of my focus is going towards https://github.com/Xilinx/mlir-aie/discussions/3222 so I may be a bit slower to respond to PRs until I get the bulk of that work done. It will likely include some breaking changes, so please do keep an eye out. I've done some minor releases to try to give some good stopping points before the breaking changes happen, but I haven't managed to get all of your PRs in, what with the CI chaos this week.

atassis:
This looks really useful for the front of work I'm doing in my own project, so when I get a chance I'll read through it properly and keep in mind there might be some breaking changes coming. And no worries on #3212, next week is totally fine - appreciate you fitting these in around the CI chaos.

PR #3221, opened 2026-06-25
[ci] Pin llvm-aie to last-good nightly (2026062301) to unblock CI

atassis:
## Problem

The scheduled "Build and Test with AIE tools on Ryzen AI" workflow has been red since 2026-06-25. CI installs llvm-aie (Peano) from the unpinned `nightly` channel, and the 2026-06-24 nightly miscompiles `int -> float -> int` at `-O2`: the argument-defining load is scheduled into the soft-float call's delay slots, so on AIE2 (in-order, no interlocks, ~7-cycle load latency vs 5 delay slots) the callee reads a stale argument register. The result is a silent wrong answer, which the on-hardware tests catch as failures (e.g. `test_algorithms.py::test_transform_different_datatypes_extern[int32-int]`).

Root cause, a hardware-free reproducer, the bisect (a single inter-block scheduler commit), and on-device confirmation (good PASS, bad FAIL, `-O0` PASS) are in Xilinx/llvm-aie#1053.

## Fix

Pin llvm-aie to the last-good nightly `21.0.0.2026062301+cb664e8c` at every install site (the Ryzen AI and Vitis workflows plus the three `utils` setup scripts), so CI goes green again. This is a temporary stopgap to be reverted once the upstream codegen fix lands.

Verified the pinned wheel resolves from the nightly find-links index.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -101,7 +101,8 @@ jobs:
           python -m pip install --require-hashes -r python/requirements_dev.lock
 
           # Install llvm-aie (Peano) which is needed for unittests requiring compiling NPU code
-          pip install -I llvm-aie -f https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly
+          # Temp pin: latest llvm-aie nightly miscompiles int->float->int at -O2 (Xilinx/llvm-aie#1053). Revert once fixed upstream.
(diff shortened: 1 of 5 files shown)


hunhoffe:
Thank you for the good work with the PR here and the peano issue @atassis (and thank you also to @thomthehound for pointing out it's likely a peano issue in a comment!)

hunhoffe (approved):

Issue #3198, opened 2026-06-22
CI/CD Overhaul


thomthehound:
The CI has gradually become expensive in ways that no longer seem reasonable. Some long-running jobs are, obviously, always going to be necessary, but others appear to be regularly rebuilding large parts of the toolchain without producing any artifacts or giving us feedback that we could not be getting much more cheaply. On top of this, our `ccache` implementation has aged into obsolescence and now, instead of saving build time, it only wastes it.

In the spirit of the other cleanup work we have been doing lately, I propose a CI modernization campaign with the goal of improving efficiency and reliability while maintaining testing integrity.

This will mean, at least: auditing cache correctness, container visibility, workflow triggers, matrix policy, PR validation boundaries, wheel production, and LLVM-main canary-ish coverage, as well as probably other things that I'm sure are not immediately coming to mind.

hunhoffe:
@thomthehound I'm working incrementally on setting up a few more physical machines. Improvements to CI in other ways would be welcome, but will need to be done incrementally. I think ccache would be a good place to start.

thomthehound:
> [@thomthehound](https://github.com/thomthehound) I'm working incrementally on setting up a few more physical machines. Improvements to CI in other ways would be welcome, but will need to be done incrementally. I think ccache would be a good place to start.

That's great to hear about the sorely-needed extra hardware! It certainly seems that our poor-abused machines have been struggling lately under the load.

And I agree; incremental is the way to go. It would be a disaster to accidentally break CI in some inscrutable way. When we rely on something this much, each patch should be something readily understandable.

Unlike the Windows work, there is no reason to get this all done at once. I really don't want to end up having a reputation for hard-to-review mega-patches. I hate that, too. But I think it is understandable why introducing support for a new platform has required me to touch a lot of things at once, even if I have tried to keep the core concepts of each individual patch disciplined. I look forward to the day when this is no longer the case.

Ccache is a good first target, but I think there might be a lower-hanging fruit worth grabbing before that, although perhaps I am misunderstanding something. Why is it that we are rebuilding MLIR every six hours just to throw it away? Is that some sort of upstream canary? If so, doesn't it seem like that is something we could be doing on the weekend or, if needed more often, once nightly at UTC 0:00 (or some other dead-zone time)? It is perhaps not as critical as I initially thought, given most of our problems these days are hardware-based, but this does seem wasteful to me.

hunhoffe:
[...]

PR #3223, opened 2026-06-25
[ci] Pin Peano via utils/peano-version.txt + scheduled auto-bump

hunhoffe:
## Problem
  
CI installed llvm-aie (Peano) from the rolling `nightly` channel at every install site, so a bad nightly flowed straight into scheduled `main` with no PR to catch it — e.g. the 2026-06-24 `int -> float -> int` miscompile (Xilinx/llvm-aie#1053) that turned the CI jobs red. Dependabot can't guard this: Peano is a GitHub-release `--find-links` channel, not a PyPI package.
  
## Fix
  
- Pin the version in `utils/peano-requirements.txt` — a real pip requirements file that bundles the find-links channel URL with the pin, mirroring the eudsl pin style already in `python/requirements.txt`. All six install sites (the two NPU workflows + `env_setup.sh`, `quick_setup.sh`, `build-mlir-aie-from-wheels.sh`, `iron_setup.py`) now install with a single `pip install -r utils/peano-requirements.txt`, so the nightly channel URL is no longer duplicated inline at each site. An explicit `--llvm-aie <url>` to `iron_setup.py` still overrides and stays unpinned.
- Add a **weekly** `update-peano` workflow (mirrors the existing `update-llvm.yml`) that bumps the pin via `update_peano_version.py` and opens a PR with a rotated assignee. Scheduled Sunday night US Mountain (Monday 05:53 UTC); the bump always targets the newest available nightly.
    
## Why this works
  
The bump PR is meant to run the on-device NPU test matrix. A green run merges and keeps CI on a fresh, *tested* Peano; a red run quarantines the bad nightly in an un-merged PR instead of breaking `main` — exactly the failure that prompted this.
  
# Note on triggering the bump PR's checks
  
The workflow creates the PR with the default `GITHUB_TOKEN`, matching the existing `update-llvm.yml`. GitHub does not trigger `on: pull_request` / `on: push` workflows for PRs opened by `GITHUB_TOKEN` (recursion guard), so the NPU/Vitis checks won't start automatically. The assignee should kick them off by closing and immediately reopening the PR (or pushing an empty commit). Switching the `token:` to a repo-scoped PAT or a GitHub App token would make the checks fire on creation, but that needs an org-level secret we don't currently have.
  
## Relationship to #3221
  
#3221 is the immediate stopgap (static pin to last-good). This makes that pin the permanent mechanism and keeps it moving forward automatically. Seeded at the same last-good `21.0.0.2026062301+cb664e8c`. 

# Note on triggering the bump PR's checks                       

The workflow creates the PR with the default `GITHUB_TOKEN`, matching the existing `update-llvm.yml`. GitHub does not trigger `on: pull_request` / `on: push` workflows for PRs opened by `GITHUB_TOKEN` (recursion guard), so the NPU/Vitis checks won't start automatically. The assignee should kick them off by closing and immediately reopening the PR (or pushing an empty commit). Switching the `token:` to a repo-scoped PAT or a GitHub App token would make the checks fire on creation, but that needs an org-level secret we don't currently have.      

## Relationship to #3221                                         

#3221 is the immediate stopgap (static pin to last-good). This makes that pin the permanent mechanism and keeps it moving forward automatically. Seeded at the same last-good `21.0.0.2026062301+cb664e8c`.


thomthehound:
I think this is a good idea. When we get around to the cache-policy work in #3198, the pinned Peano install should be one of the things we make sure stays cached.

I'm just commenting here so I see it later in the Issue.

hunhoffe:
@erwei-xilinx I'm about ready to change this to "ready for review" once the CI tests pass -- I wanted to highlight this because it might be nice to keep mlir-air and mlir-aie's llvm-aie version in sync and I wanted to get your input.

andrej (approved):
Seems good. Any particular reason to put `requirements-peano.txt` in `utils` rather than `python` with the other requirements.txts though?

jgmelber (approved):

hunhoffe:
@andrej and @jgmelber I added you to the rotation to be assigned to update peano and/or llvm. If the rotation works ok, we can add Jeff and Erwei too (if they agree!)

PR #3225, opened 2026-06-25
[dyn-seq P1.2] Carry npu scalar op fields as SSA operands

hunhoffe:
## Description:

PR 2 of the dynamic-runtime-sequences rebuild ([#3222](https://github.com/Xilinx/mlir-aie/discussions/3222)).

Changes the npu scalar ops — `npu.write32`, `npu.maskwrite32`, `npu.sync`, `npu.address_patch`, `npu.rtp_write` — to carry their integer fields as SSA `i32` operands instead of attributes, so runtime-sequence values can flow into them end to end. Today every value is an `arith.constant` (byte-identical to the old attribute form); later PRs let runtime scalars feed the same operands.
  
- Op defs: integer attrs → `I32` operands, new operand-form assembly syntax (`aiex.npu.write32(%addr, %val) : i32, i32`). `buffer`/`column`/`row` (placement) and `address_patch`'s `addr`/`arg_idx` stay attributes.
- New helpers in the AIEX dialect: `getConstantIntOperand` (fold an SSA value to a compile-time `uint32_t`) and `createConstantI32` (the inverse, for lowering passes feeding compile-time-known values).
- All producers migrated to materialize `arith.constant` operands; all consumers (static TXN emitter, CERT lowering, AIE↔configuration roundtrip, trace utilities) fold the operand back and emit a clean diagnostic on a non-constant value rather than miscompiling.
- Python: `dialects/aiex.py` / `aie.py` builders preserve the historical int-argument API by materializing constants on the user's behalf.
- AIECoreToStandard: clone device-level constants into the cores that use them before outlining. The new SSA operands let the constant folder hoist constants out of the (non-isolated) runtime_sequence into the device body, where CSE can merge them with a core's own constants; without this, such a constant is erased while still used by a core during device removal.


lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -746,6 +747,32 @@ struct AIECoreToStandardPass
...
     declareAIEIntrinsics(targetModel.getTargetArch(), builder);
 
+    // A core may use a constant defined in the parent device body (the folder
+    // hoists constants out of non-IsolatedFromAbove ops like
+    // aie.runtime_sequence, then CSE merges them with a core's own constants).
+    // The core is outlined into a func below while the device is erased, so
+    // such a device-level constant would be erased while still used by the
+    // core's body. Clone it into the core first so the outlined func is
+    // self-contained.
+    {
```
hunhoffe: This is the best solution I could come up with for this problem, but I'm not sure if it's actually a good solution.

python/dialects/aiex.py
```diff
@@ -39,7 +47,74 @@
...
+    through unchanged. The npu scalar ops (write32/maskwrite32/sync/
+    address_patch/rtp_write) carry their integer fields as SSA i32 operands;
+    these wrappers preserve the historical int-argument API by converting
+    constants on the user's behalf."""
+    if isinstance(v, int):
+        return constant(v, T.i32())
+    return v
+
+
+def npu_write32(address, value, buffer=None, column=None, row=None, **kwargs):
```
hunhoffe, line 62: This seems a little silly? Are you sure we can't do this automagically somehow? eudsl has a lot of coercive/automatic type interpretations....

python/dialects/aiex.py
```diff
@@ -39,7 +47,74 @@
 # Comes from _aie
 register_dialect(get_dialect_registry())
 
-npu_sync = partial(npu_sync, column_num=1, row_num=1)
+
+def _as_i32(v):
```
hunhoffe, line 51: This seems a little silly? Are you sure we can't do this automagically somehow? eudsl has a lot of coercive/automatic type interpretations....

python/dialects/aiex.py
```diff
@@ -39,7 +47,74 @@
...
+        _as_i32(address),
+        _as_i32(value),
+        buffer=buffer,
+        column=column,
+        row=row,
+        **kwargs,
+    )
+
+
+def npu_maskwrite32(address, value, mask, buffer=None, column=None, row=None, **kwargs):
```
hunhoffe, line 73: This seems a little silly? Are you sure we can't do this automagically somehow? eudsl has a lot of coercive/automatic type interpretations....

python/dialects/aiex.py
```diff
@@ -39,7 +47,74 @@
...
+        _as_i32(value),
+        _as_i32(mask),
+        buffer=buffer,
+        column=column,
+        row=row,
+        **kwargs,
+    )
+
+
+def npu_sync(column, row, direction, channel, column_num=1, row_num=1, **kwargs):
```
hunhoffe, line 85: This seems a little silly? Are you sure we can't do this automagically somehow? eudsl has a lot of coercive/automatic type interpretations....

python/utils/trace/parse.py
```diff
@@ -402,19 +402,31 @@ def parse_mlir_trace_events(mlir_module_str, colshift=None):
         target_model = aiedialect.get_target_model(device)
         events_module = get_events_for_device(str(device))
 
+    # write32 carries address/value as SSA i32 operands (materialized via
+    # arith.constant); row/column remain optional attributes. Fold the operand
+    # back to its constant integer.
+    def fold_constant_operand(operand):
+        defining = operand.owner
+        if defining is None or not hasattr(defining, "value"):
+            return None
```
hunhoffe: Is returning "None" here and below a valid path?

python/utils/regdb.py
```diff
@@ -553,25 +553,39 @@ def get_address_from_op(
         value = None
         mask = None
 
-        # Get address - could be 'address' or 'addr' attribute
+        # write32/maskwrite32 carry address/value/mask as SSA i32 operands
+        # (materialized via arith.constant); address_patch's 'addr' and the
+        # 'row'/'column' placement fields remain attributes. Fold the operand
+        # back to its constant integer where applicable.
+        def fold_constant_operand(operand):
```
hunhoffe: It seems like I've seen this fucntion before. I don't like duplicated code. If it's useful, take a clean and correct version of it and put it into a common utilities location.

test/aiecc/buffers_xclbin.mlir
```diff
@@ -47,13 +47,31 @@ module {
     aie.runtime_sequence(%arg0: memref<1024xi32>, %arg1: memref<1024xi32>, %arg2: memref<1024xi32>, %arg3: memref<1024xi32>, %arg4: memref<1024xi32>, %arg5: memref<1024xi32>) {
       aiex.npu.dma_memcpy_nd (%arg0[0, 0, 0, 0][1, 1, 1, 1024][0, 0, 0, 1]) {id = 0 : i64, metadata = @in0} : memref<1024xi32>
       aiex.npu.dma_memcpy_nd (%arg1[0, 0, 0, 0][1, 1, 1, 1024][0, 0, 0, 1]) {id = 1 : i64, metadata = @out0} : memref<1024xi32>
-      aiex.npu.sync {channel = 0 : i32, column = 0 : i32, column_num = 1 : i32, direction = 0 : i32, row = 0 : i32, row_num = 1 : i32}
+      %cst_npu_0 = arith.constant 0 : i32
```
hunhoffe, line 45: It looks like a lot of the same constants are defined here. Is this input or output to the test? If it's input, is it really correct to have that type of duplication?

hunhoffe (commented):
Needs a bit more work.

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
fifield: I think `AIEDialectFoldInterface` supposed to gate the hoisting. Can we update that class to make the problem go away?

hunhoffe: Thanks for the tip! I believe that is correct. I'm working on implementing this path now...

erwei-xilinx (approved):

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -748,15 +771,18 @@ static std::optional<uint32_t> getAbsoluteAddress(T *op) {
 }
```
andrej, line 771: I guess future code for the dynamic lowering will have to emit actual IR for folding the column, row, offset into the absolute address.

andrej (approved):
Lookin' great!

lib/Dialect/AIEX/IR/AIEXDialect.cpp
hunhoffe: Yep yep!   The runtime folding you're describing is the job of the AIEX→EmitC conversion / C++ TXN target — Phase 1, PR #3 in the plan (#3222). That path will emit IR/generated-C++ that computes (col << colShift) | (row << rowShift) | (addr ▎ & 0xfffff) at runtime instead of doing it here in C++. This PR (#3225) deliberately keeps the static path static and just gates the non-constant case off with an error.

hunhoffe:
@andrej Thanks for the review!

PR #3226, opened 2026-06-25
[dyn-seq P0a] Unroll-first BD-ID allocation for runtime sequences

hunhoffe:
Delivers rolled ping-pong runtime sequences: a constant-bound `scf.for` whose DMA task hands off between physical BD ids across iterations, compiled to straight-line hardware.

**Approach:** flatten control flow before allocating, so the allocator only sees straight-line IR. Constant `scf.for` is unrolled (new `aie-unroll-runtime-sequence-loops` pass); constant `scf.if` is folded (`--canonicalize`). Both run before `aie-assign-runtime-sequence-bd-ids`, so no scf op reaches the allocator in the static path. A ping-pong over N tiles becomes N configures with ids alternating 0, 1 — two physical BDs for any tile count.

**Allocator:** allocates straight-line IR by ordinary liveness reuse; rejects any surviving (runtime-valued) `scf.for`/`scf.if`/`scf.while` with a pointer to the dynamic path (Phase 2). Keeps double-free detection.

**Also:** `HasParent`→`HasAncestor` on the DMA task ops; the NPU test is written in high-level IRON (`ObjectFifo`+`Worker`) with the ping-pong via `rt.inline_ops`, which now registers `ObjectFifo` args like it already does `Buffer` args.


hunhoffe:
Depends on https://github.com/Xilinx/mlir-aie/pull/3243

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +49,368 @@ using namespace mlir;
...
+        // Preserve the helpful "lower this first" notes for unlowered task ops.
+        (void)emitUnresolvedTask(op, task);
+        return WalkResult::interrupt();
+      }
+      return WalkResult::advance();
+    });
+    if (wr.wasInterrupted())
+      return failure();
+
+    // 4. Peak simultaneous BD liveness must fit the tile's BD pool.
```
fifield: nit: counting is 1,2,4,3

include/aie/Dialect/AIE/IR/AIEDialect.h
```diff
@@ -44,6 +44,24 @@ template <typename ConcreteType>
...
+// direct child or anywhere inside structured control flow such as scf.for /
+// scf.if in the sequence body). Replaces HasParent<RuntimeSequenceOp> for ops
+// that may legally appear inside control flow within the sequence.
+template <typename ConcreteType>
+struct NestedInRuntimeSequence
+    : mlir::OpTrait::TraitBase<ConcreteType, NestedInRuntimeSequence> {
+  static mlir::LogicalResult verifyTrait(mlir::Operation *op) {
+    return verifyNestedInRuntimeSequence(op);
+  }
+};
```
fifield: Would `OpTrait::HasAncestor<RuntimeSequenceOp>` from MLIR work here?

hunhoffe: Yes, thank you for pointing this out. I found a few other wheel-reinventing areas and addressed them in the commits I just pushed.

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
hunhoffe: Fixed

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -990,6 +990,20 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
...
+        // `bd_id` stay valid; absent on every non-rotating BD.
+        OptionalAttr<DenseI32ArrayAttr>:$bd_id_window,
+        // Internal handshake between --aie-assign-runtime-sequence-bd-ids (which
+        // stamps a unique id per rotation group) and
+        // --aie-unroll-runtime-sequence-loops (which round-robins the window per
+        // group and then removes both attributes). Distinct groups can share
+        // identical `bd_id_window` contents -- different tiles allocate ids
+        // independently, and sequential loops on one tile reuse the freed pool --
+        // so the window contents alone cannot key the round-robin counter.
+        OptionalAttr<AIEI32Attr>:$bd_id_window_group,
```
andrej: Understanding question -- This gets populated by the assign-runtime-bd-ids pass? And then when you unroll the loop, you sequentially pick IDs out of this list for every iteration? It's a bit confusing to me, conceptually couldn't we unroll the loop _first_, and then just assign BD IDs so nothing conflicts liveness wise?

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -990,6 +990,20 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", []> {
         OptionalAttr<BDPadLayoutArrayAttr>:$pad_dimensions,
         DefaultValuedOptionalAttr<AIEI32Attr, "0">:$pad_value,
         OptionalAttr<AIEI32Attr>:$bd_id,
+        // For a buffer descriptor whose physical BD id must rotate across loop
```
hunhoffe: Cleanup: Comment seems excessive.

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +46,368 @@ using namespace mlir;
...
+      // reports backEdgesCrossed > 0 but is validated via its body, not here.
+      // Surface the rotation-specific rejections (mismatched chain lengths, a
+      // prologue that does not trace to a configure) on the body.
+      if (range.backEdgesCrossed > 0 && range.enclosingLoop) {
+        LoopRotationGroup group = resolveLoopRotationGroup(configure);
+        if (group.status == LoopRotationGroup::ChainLengthMismatch) {
+          configure.emitOpError(
+              "rotating buffer-descriptor chain length differs from the "
+              "prologue tasks it rotates with; a rolled ping-pong must rotate "
+              "chains of equal length");
```
andrej: I really can't wrap my head around what this means ... maybe we can discuss offline

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -59,115 +421,194 @@ struct AIEAssignRuntimeSequenceBDIDsPass
...
+    SmallVector<DMAConfigureTaskOp> targets;
+    if (!resolveFreeTargets(task, targets) || targets.empty())
+      return emitUnresolvedTask(op, task);
+    for (DMAConfigureTaskOp t : targets)
+      if (failed(recycle(t)))
+        return failure();
+    return success();
+  }
 
-    // First, honor all the hard-coded BD IDs
```
hunhoffe, line 112: To double check: Do we still honor hard-coded IDs? Is this tested in a regression test?

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +46,368 @@ using namespace mlir;
...
+            "form");
+        return WalkResult::interrupt();
+      }
+      if (range.leaked && range.enclosingLoop) {
+        configure.emitOpError(
+            "buffer descriptor configured in a loop is never completed (no "
+            "aiex.dma_await_task / aiex.dma_free_task reachable); its BD would "
+            "not be reusable across iterations");
+        return WalkResult::interrupt();
+      }
```
andrej: Does this apply to the whole program or just the loop? Asked differently, can a BD allocation span around a loop without triggering this error (i.e. I allocate before the top of the loop for a single task, then run a unrelated loop that may not use that BD ID for any of its iterations, then after the loop I free)?

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +46,368 @@ using namespace mlir;
...
+      Value task;
+      if (auto f = dyn_cast<DMAFreeTaskOp>(op))
+        task = f.getTask();
+      else if (auto a = dyn_cast<DMAAwaitTaskOp>(op))
+        task = a.getTask();
+      else
+        return WalkResult::advance();
+      SmallVector<DMAConfigureTaskOp> targets;
+      if (!resolveFreeTargets(task, targets) || targets.empty()) {
+        // Preserve the helpful "lower this first" notes for unlowered task ops.
```
andrej:
```suggestion
        // Emit "lower this first" notes for unlowered task ops.
```

test/bd-chains-and-dma-tasks/assign-runtime-sequence-bd-ids/bad-rotation-bd-count-mismatch.mlir
```diff
@@ -0,0 +1,42 @@
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
```
hunhoffe: Is this truly invalid, or just not handled?

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +46,368 @@ using namespace mlir;
...
+      auto peaks = computePeakBdLiveness(seq);
+      const AIETargetModel &tm =
+          seq->getParentOfType<AIE::DeviceOp>().getTargetModel();
+      for (auto &kv : peaks) {
+        int col = kv.first.first, row = kv.first.second;
+        uint32_t pool = tm.getNumBDs(col, row);
+        if (kv.second > pool) {
+          seq.emitOpError("peak simultaneous buffer-descriptor liveness ")
+              << kv.second << " on tile(" << col << "," << row
+              << ") exceeds the tile's " << pool << " buffer descriptors";
```
andrej: Clarity suggestion
```suggestion
          seq.emitOpError("Too many simultaneously active buffer descriptors -- attempted to use ")
              << kv.second << " buffer descriptors on tile(" << col << "," << row
              << "), which only supports up to " << pool << " simultaneously active buffer descriptors";
```

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -59,115 +421,194 @@ struct AIEAssignRuntimeSequenceBDIDsPass
...
+    WalkResult result = task_op.walk<WalkOrder::PreOrder>([&](AIE::DMABDOp bd) {
+      if (!bd.getBdId().has_value()) {
+        bd.emitOpError("Free called on BD chain with unassigned IDs.");
+        return WalkResult::interrupt();
       }
-      if (llvm::isa<DMAConfigureTaskForOp>(task_op)) {
-        err.attachNote(task_op->getLoc())
+      // Ignore double-frees.
+      if (gen.bdIdAlreadyAssigned(bd.getBdId().value()))
+        gen.freeBdId(bd.getBdId().value());
```
andrej: Would an error be better? Double-freeing seems to point to user confusion for me

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -59,115 +421,194 @@ struct AIEAssignRuntimeSequenceBDIDsPass
...
     return success();
   }
 
   void runOnOperation() override {
-
-    // This pass currently assigns BD IDs with a simple linear pass. IDs are
-    // assigned in sequence, and issuing an aiex.free_bds or aiex.await_bds op
-    // kills the correspondings IDs use. If in the future we support
-    // branching/jumping in the sequence function, a proper liveness analysis
-    // will become necessary here.
```
andrej, line 141: Thanks for cleaning up my mess

lib/Dialect/AIEX/Transforms/AIERuntimeBdLiveness.cpp
```diff
@@ -0,0 +1,391 @@
+//===- AIERuntimeBdLiveness.cpp ---------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
```
andrej: This passes checks because since the Copyright line matches our format and it has the SPDX-License identifier, but we've dropped the extra "This file is licensed under" lines in the license header refactor and moved the copyright above the license.

Perhaps I should add a negative check to the CI linter to make sure the "This file is licensed under..." line triggers a red X...

lib/Dialect/AIEX/Transforms/AIERuntimeBdLiveness.cpp
andrej, line 1: Could any of the infrastructure in [`mlir::Liveness`](https://mlir.llvm.org/doxygen/classmlir_1_1Liveness.html#details) reduce the amount of custom code here? I can't claim to understand it well enough to know the answer.

hunhoffe (commented):
TODO: double check tests for nested loops with various ping-pong patterns and nested loops/conditionals

andrej (commented):
Good progress -- it's cool that this works. 

I do think I lack some of the higher level understanding for this. From a 10,000 ft view, it does seem like quite a lot of code, and some of it seems quite specialized on the "ping-pong rotating BD IDs" concept. I don't know how well this generalizes over complex control flow (nested loops, etc.). (Not saying it doesn't, I just don't understand it.)

My mental model approaching this before looking at this PR was:

1. For the static case (this PR), unroll everything first. Then liveness should become much easier, as there are no back-edges at all. With no back-edges, you're looking at a tree (with `scf.if`s splitting into multiple sub-nodes), where you can just traverse through the tree and as long as each path to a leaf doesn't collide with BD IDs, you should be fine. (Correct me if I'm missing something.) Anything that you can't unroll due to dependence on runtime values gets rejected in the static path.
2. For the dynamic case (not in this PR), don't even attempt to do any allocation in the compiler -- it'd be undecidable anyways. Instead, a tiny runtime libary/emitC-generated-code keeps track of a set of available BD IDs. Whenever the runtime encounters a new configure op, pick the next available one out of the set (error if set empty). Add IDs back when encountering free/await. No special handling/analysis of loops. Low cost, and should generalize across all control flow.

Can you explain from a high-level how your approach with the loop rotations differs and why it is preferrable over the above? 

I can guarantee that I am missing some of the finer points since I haven't spent any time on this and am only looking at it from afar, but I would really like to understand the motivation for complexity. I really did try to stare at it for a while and think I still don't really understand which cases it covers.

My goal is also not to block quick progress, so you can also tell me I don't have to understand everything and just push this forward. In that case, my only blocker would be making the error messages a bit more actionable for users, as I've stared at some of them and as a user wouldn't know what I would need to do to fix it, especially the "rotating buffer-descriptor chain length differs from the prologue tasks it rotates with"). :)

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
hunhoffe: Since I now unroll first, this is somewhat trivially handled.

lib/Dialect/AIEX/Transforms/AIERuntimeBdLiveness.cpp
hunhoffe: I really tried, but it doesn't quite fit due to region/block scoping at least that I could find.

include/aie/Dialect/AIE/IR/AIEOps.td
hunhoffe: Updated to unroll first!

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
hunhoffe: Double free now explicitly caught

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
hunhoffe: I believe removed now.

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -241,6 +241,30 @@ def AIEAssignRuntimeSequenceBDIDs : Pass<"aie-assign-runtime-sequence-bd-ids", "
...
+  let description = [{
+    Fully unrolls `scf.for` loops with compile-time-constant trip counts that
+    appear inside `aie.runtime_sequence` bodies.
+
+    Runs before `aie-assign-runtime-sequence-bd-ids` so BD-ID allocation sees
+    straight-line IR: a rolled ping-pong (a task freed a later iteration than it
+    is configured) becomes several distinct configures whose ids ordinary
+    liveness-based allocation recycles, with no rolled-loop rotation analysis.
+    Runtime-bound loops (non-constant trip count) are left in place for the
+    dynamic EmitC lowering path.
```
andrej, line 256:
```suggestion
    Runs before `aie-assign-runtime-sequence-bd-ids` so BD-ID allocation sees
    straight-line IR: configure operations in each loop iteration can receive 
    individual BD IDs after unrolling.
    
    Runtime-bound loops (non-constant trip count) are left in place for the
    dynamic EmitC lowering path.
```

Without knowing the history of the code, it would not be clear what a "rolled-loop rotation analysis" even is, and it is not relevant since that code no longer exists.

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -22,30 +39,52 @@ using namespace mlir;
...
+    WalkResult wr = seq.walk([&](Operation *op) -> WalkResult {
+      if (isa<scf::ForOp, scf::IfOp, scf::WhileOp>(op)) {
+        op->emitOpError(
+            "runtime-valued control flow in a runtime sequence is not "
+            "supported "
+            "by static BD-ID allocation. A constant-trip scf.for is unrolled "
+            "and a constant-predicate scf.if is folded before this pass; a "
+            "surviving scf op has a runtime bound/predicate and must be "
+            "lowered "
+            "with the dynamic EmitC path");
```
andrej:
```suggestion
            "Runtime-valued control flow in a runtime sequence is not "
            "supported by this static BD-ID allocation pass. Either pass "
            "only constant-valued predicates to scf.for and scf.if so "
            "`aie-unroll-runtime-sequence-loops` can unroll/fold them, or "
            "use the dynamic EmitC path instead.");
```

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -59,115 +98,142 @@ struct AIEAssignRuntimeSequenceBDIDsPass
...
+          // Runtime sequences configure BDs on shim (and compute) tiles only,
+          // which are channel-agnostic (always accessible), so passing 0 is
+          // correct here.
           std::optional<int32_t> next_id = gen.nextBdId(/*channelIndex=*/0);
           if (!next_id) {
+            const AIETargetModel &tm =
+                tile->getParentOfType<AIE::DeviceOp>().getTargetModel();
             op.emitOpError()
-                << "Allocator exhausted available buffer descriptor IDs.";
+                << "Too many simultaneously active buffer descriptors on tile("
```
andrej:
```suggestion
                << "Too many simultaneously active buffer descriptors on tile ("
```

lib/Dialect/AIEX/Transforms/AIEAssignRuntimeSequenceBDIDs.cpp
```diff
@@ -59,115 +98,142 @@ struct AIEAssignRuntimeSequenceBDIDsPass
...
-        return signalPassFailure();
-      }
+    WalkResult wr = device.walk([&](AIE::RuntimeSequenceOp seq) -> WalkResult {
+      if (failed(validate(seq)))
+        return WalkResult::interrupt();
+      gens.clear();
+      awaitedConfigures.clear();
+
+      // Straight-line walk. Collect frees to erase after (recycling reads the
+      // configure the free points at, so erase only once the walk is done).
```
andrej, line 206:
```suggestion
      // Straight-line walk. Collect frees to erase after (recycling reads the
      // configure the free points at, so erase only once the walk is done).
      // This allocation strategy works only for straight-line IR without
      // branching or conditionals; the verifier of this pass ensures this is
      // the case on the input IR.
```

andrej (approved):
Great!

PR #3229, opened 2026-06-26
aiecc: emit the real host BO count in kernels.json instead of hardcoding 5

atassis:
## Problem

`aiecc`'s `generateKernelsJson` writes the `<device>_kernels.json` ABI manifest that XRT uses to launch a DPU kernel. It emitted the host-buffer (`boN`) argument list with a loop hardcoded to exactly 5:

```cpp
for (int i = 0; i < 5; i++) { ... "bo" + Twine(i) ... }
```

The 5 is unrelated to the kernel; it never looks at how many host buffers the kernel actually takes. A kernel whose runtime sequence takes more than 5 host buffers gets a `kernels.json` that under-declares its arguments: every `boN` past index 4 is dropped, so the generated DPU kernel metadata no longer matches the kernel.

Minimal repro: a `runtime_sequence` with 6 or more host buffer arguments produces a `kernels.json` that stops at `bo4`.

## Fix

Pass the host-buffer count into `generateKernelsJson` and loop to it. The count is the runtime-sequence argument count, computed at the call site in `generateCdoArtifacts` with the same `getNumArguments()` idiom this file already uses for the full-ELF artifact paths.

`5` is kept as a floor, so kernels with `<=5` host BOs produce byte-identical `kernels.json` (no change for existing kernels; only the previously-broken `>5` case changes). I used the floor rather than an exact count specifically to avoid changing the output of every existing small kernel. Happy to switch `generateKernelsJson` to the exact runtime-sequence count (matching the full-ELF paths exactly) instead, if you prefer that consistency.

## Test

`test/aiecc/buffers_xclbin.mlir` already drives a 6-host-buffer `runtime_sequence` through the xclbin path, but only checked `bo0..bo4` (which is why the truncation went unnoticed). I extended it to also check that `bo5` is emitted at its expected offset. Without this fix the 6th host buffer is silently dropped and the new check fails; with it, it passes. The test runs in the `test/aiecc` lit suite.

I have not run a full local toolchain build (the multi-hour LLVM build); the change is verified by inspection against the three existing `getNumArguments()` call sites in this file and by the strengthened lit test, which CI builds and runs.

## Note

`python/compiler/aiecc/main.py` is now a thin wrapper that delegates to this C++ binary, so the Python `aiecc.py` xclbin path is covered by the same fix and the same test.

### DIFF
--- a/test/aiecc/buffers_xclbin.mlir
+++ b/test/aiecc/buffers_xclbin.mlir
@@ -30,7 +30,7 @@
 // The runtime_sequence below has 6 host buffer arguments; bo5 must be emitted.
 // A previously hardcoded 5-arg loop dropped it, under-declaring the kernel ABI.
 // CHECK: "name": "bo5"
-// CHECK: "offset": "0x3c"
+// CHECK: "offset": "0x3C"
 // CHECK: "dpu_kernel_id": "0x901"
 // CHECK: "name": "MLIRAIE"
 // CHECK: "name": "MLIR_AIE"
--- a/test/aiecc/buffers_xclbin.mlir
+++ b/test/aiecc/buffers_xclbin.mlir
@@ -27,6 +27,10 @@
 // CHECK: "name": "bo2"
 // CHECK: "name": "bo3"
 // CHECK: "name": "bo4"
+// The runtime_sequence below has 6 host buffer arguments; bo5 must be emitted.
+// A previously hardcoded 5-arg loop dropped it, under-declaring the kernel ABI.
+// CHECK: "name": "bo5"
+// CHECK: "offset": "0x3c"
 // CHECK: "dpu_kernel_id": "0x901"
 // CHECK: "name": "MLIRAIE"
 // CHECK: "name": "MLIR_AIE"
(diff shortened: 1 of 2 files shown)


hunhoffe:
@atassis did you check to see if there is a max? I'm wondering if aiecc needs to fail gracefully in that case.

atassis:
Good question. I have looked into it right now.
So, the hardcoded `5` came in with the initial C++ aiecc driver. It doesn't track any hardware or ABI limit I can find - it just reads like a port placeholder rather than a real bound.
I couldn't find a documented per-kernel host-buffer maximum in mlir-aie. The full-ELF artifact paths in this same file already derive the bo/arg count from the unbounded runtime-sequence `getNumArguments()` (info.argCount, no upper-bound check), so this PR mostly makes kernels.json consistent with how the rest of the file already treats the count, rather then adding a new unbounded path.
In practice the count is small - the breakage shows up at 6+ host buffers, and any real ABI ceiling (the XRT command-packet argument capacity) sits well above that, so the old cap of 5 was already below any hardware limit.

That said, I can add a graceful failure if there's a max we should enforce. If you know the actual DPU/XRT host-buffer limit, I can add an emitError when the count exceeds it - and it might make sense to put this guard somewhere shared, so the full-ELF paths get it too, not only kernels.json. Or if you want to pick a conservative bound, I can do that, too. Let me know if you have any preference

PR #3233, opened 2026-06-28
Peano lit enablement No2

thomthehound:
This is both a follow-up to #3075 and a foundation for a universal cross-platform path for building the designs in `programming_examples` and `programming_guide` and executing their tests. It removes the remaining directory-level feature gates I missed in "Peano lit enablement No1" while introducing a new `makefile_examples` file-level gate for Make-driven lit tests.

The purpose of `makefile_examples` is to prevent failure noise as individual tests become cross-platform. Files with that gate, in addition to being gated on the feature level, are supported only when the system is non-Windows and `make` is on `PATH` (that is to say: it marks them `unsupported` on Windows). Eventually, almost all of these files are going to need to be remade, with their `RUN:` lines rewritten, using something more portable than `make`. This will almost certainly be Python JIT in the vast majority of cases, but some host-native examples will need new CMake wrappers.

Additionally, a small handful of AIE2 unit tests that I missed in #3075 are now covered.

With this patch, all of the following testing build targets are now runnable in a pure Peano/llvm-aie environment, without CHESS/Vitis (it should be possible to verify this locally by disabling, or simply not enabling, Vitis in the Linux shell environment):

```text
check-aie
check-aie-concurrency
check-reference-designs (*)
check-programming-guide
```

> `mlir_exercises` is not supported here because that bundle very much wants to be XCHESS-shaped, and does not look appropriate for a Ryzen AI desktop environment.

(*) This work uncovered a Python JIT cache lifetime/ownership issue that will be addressed shortly in a follow-up patch. Due to a coincidence, the present patch should still pass on CI, but `test_dma_compression.py` fails locally (behavior unchanged).

I find it irritating that this patch has to be so loud for the tiny amount of work that it is actually doing. But this was the best way I could think of to carry on the cross-platform conversion process in smaller, more digestible, patches without everything looking like a huge error-dump in the off-chance that somebody actually tries running something like `check-programming-guide` under Windows. I strongly considered gating by filename pattern instead, but that seemed like an even bigger maintenance hazard that this.

As far as portability is concerned, I look back at my older attempt to do this with CMake-only, and Percy Shelley comes to mind:
```
I met a traveler from an antique land
Who said: Two vast and trunkless legs of stone
Stand in the dessert. Near them, on the sand,
Half sunk, a shattered visage lies, whose frown,
And wrinkled lip, and sneer of cold command,
Tell that its sculptor well those passions read
Which yet survive, stamped on these lifeless things,
The hand that mocked them and the heart that fed.

And on the pedestal these words appear:
"My name is Ozymandias, king of kings:
Look on my works, ye Mighty, and despair!"

Nothing beside remains. Round the decay
Of that colossal wreck, boundless and bare
The lone and level sands stretch far away.
```

That just isn't going to be the way anymore. JIT is the future.


thomthehound:
I hate this, but I think it is the right thing to do. When the test file is `make`-only, the system needs to know that.

hunhoffe:
I'll review this shortly; but I wanted to give you a heads up that I'm working on some (other) breaking changes, discussed at https://github.com/Xilinx/mlir-aie/discussions/3222 ; and then afterwards, I'll be looking into https://github.com/Xilinx/mlir-aie/discussions/3239

Would you be willing to create a discussion on the windows work you are doing? I think that will help us sync efforts!

thomthehound:
> I'll review this shortly; but I wanted to give you a heads up that I'm working on some (other) breaking changes, discussed at #3222 ; and then afterwards, I'll be looking into #3239
> 
> Would you be willing to create a discussion on the windows work you are doing? I think that will help us sync efforts!

You know... that is a really great idea! I don't know why I didn't think of doing that.

I think I want to take a few days to let my brain reset on this, though. I have some strong leads for moving forward, but I haven't given it all enough time to re-coalesce into something that would make a well-grounded RFC. Ultimately, however, this is all for the best. The old CMake system I had running was more about "getting the job done" than being elegant. I think that, with the smaller, more deliberate CMake + `.cpp` surface area that JIT affords us, I can spare more individual attention for the pedagogy of individual examples, instead of just forcing them all down the same over-engineered pipe.

hunhoffe (approved):
When I first saw the quote and Shelley I thought for sure you were going to say something along the lines of JIT being Frankenstein's monster... but then I realized, wrong Shelley.

Looks good to me! Thanks for more good work. I'll look forward to your discussion/issue RFC and please do participate as you see fit in the existing discussions!

PR #3235, opened 2026-06-29
Fix unsafe JIT artifact-cache cleanup in dma_compression

thomthehound:
There was some pathological cache behavior going on with this example/test. In short, when run locally, it was nuking *its own* `xclbin` while it was still running. The problem is that the paths were still in memory, causing `CallableDesign` to try to reuse a cached kernel whose backing no longer existed, so... crash.

This wasn't showing up during CI because the hard-coded nuke target did not match the cache root in the runner environment. So, I started by removing the hard-coded cleanup and having `CallableDesign` attempt to recompile missing artifacts if needed. But then, while testing with parallel workers, it turned out the nuke actually had MIRVs, and recursively destroyed *every* JIT artifact directory under the cache root. Just the whole thing, regardless of hash or `.lock` (!!!).

Again, that did not show up in CI because the thing was mostly pointed at the wrong cache tree. Anyway, I think I have it fixed up now. The recursion is removed, I neutered the thing's ability to meddle in cache management that should be handled by the compiler, and I replaced it with runtime cleanup that should not wake XRT and, I think, fulfills the needs of the example better than nukes. I also added a few tests around the new behavior.

### DIFF
--- a/programming_examples/basic/dma_compression/test_dma_compression.py
+++ b/programming_examples/basic/dma_compression/test_dma_compression.py
@@ -8,14 +8,13 @@ python3 test_dma_compression.py --skip both,core_both
 
 import argparse
 import hashlib
-import shutil
 import sys
 import time
-from pathlib import Path
 
 import numpy as np
 
 import aie.iron as iron
+from aie.utils import cleanup_npu_runtime
 
 from dma_compression import dma_compression, CONFIGS, RATIOED_N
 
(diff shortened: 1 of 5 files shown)


hunhoffe (approved):
Thank you for the fix! I left a couple of small, non-blocking follow-up ideas separately in https://github.com/Xilinx/mlir-aie/issues/3236

PR #3237, opened 2026-06-29
[test] Remove unit tests for aievec-to-cpp backend

jgmelber:
## Summary

- Removes 4 unit test directories that rely on `aie-translate --aievec-to-cpp` (and the `-aie2=true` flag), which is being removed in #3217
- On the `remove-cpp-backend` branch CI, these tests produce hard failures because `--aievec-to-cpp` and `-aie2=true` are no longer recognized arguments to `aie-translate`
- PR #3217 already deleted the companion `.cc` and data files from these directories; this removes the remaining `.mlir` test drivers and any leftover files

**Intended merge order**: after #3217 lands.

## Test plan

- [ ] Vitis CI (`Run Tests on Vitis`) should no longer show failures for these 4 tests once #3217 + this PR are both merged


hunhoffe (approved):
Nice catch!

PR #3242, opened 2026-06-29
[ci][bugfix] Test examples against the same-commit wheel, not a mutable tag 

hunhoffe:
A combination of more CI runners (higher concurrency) + https://github.com/Xilinx/mlir-aie/pull/3234 + https://github.com/Xilinx/mlir-aie/pull/3235 revealed some fragility in how we test wheels. These PRs are correct; the errors are pre-existing.

"Run Examples on Ryzen AI" installed `mlir_aie` from the **mutable** `latest-wheels-4` tag (via `utils/quick_setup.sh`), so it tested the last-published wheel rather than the commit under test. When source added a symbol the wheel lacked (e.g. `cleanup_npu_runtime` from #3235), the example failed with `ImportError` — and since every PR overwrites the tag, failures were nondeterministic.                                             
                                       
Now the example installs the wheel built from the **same workflow run**, so it always tests this commit's `mlir_aie`.
                                       
- **Same-run handoff:** new `build-examples-wheel` job builds the one needed wheel (py3.12, RTTI ON) via `workflow_call`; `build-quick-setup` pulls it with `needs:` + `download-artifact` and installs it through a new `MLIR_AIE_WHEEL_DIR` seam in `quick_setup.sh` (falls back to `latest-wheels-4` locally).                                                                  
- **Cost-gated:** the ~19 min build runs only in the **merge queue** or on-demand via the `test-wheels` label; ordinary PR pushes keep the cheap source/lit/unit tier.                                                      
- **Publish hygiene:** PRs no longer overwrite `latest-wheels-4` (publish gated to `merge_group`/`schedule`/`workflow_dispatch`); reusable calls never republish or build Windows.          
- **Guided failure:** `if: failure()` explains the likely skew cause and points to the `test-wheels` label.
                                       
Requires the `test-wheels` label (already created). Follow-ups: build-once-test-publish-if-green pipeline (kills the duplicate py3.12 build + gates publish on the test); ccache persistence to speed the cold build.          
                                       
## Test plan                         
                                                                             
- [ ] Merge queue: `build-examples-wheel` uploads `mlir_aie_rtti_ON-3.12`; `build-quick-setup` installs it and passes
  `basic/dma_compression/run_jit.lit`.                                       
- [ ] `test-wheels`-labeled PR: same path runs on the branch.
- [ ] Unlabeled PR push: wheel jobs skipped, cheap tier still runs.        
- [ ] No publish to `latest-wheels-4` on `pull_request` or reusable calls.

PR #3245, opened 2026-06-30
[CI] Clear up hardware source-build tests and add wheel smoke tests

andrej:
The CI jobs for building and testing on Ryzen AI hardware had become a bit confusing. The "examples" tests especially were confusing because they both installed wheels and built from source, but then only tested the code that was built from source. In this PR:

- Add testing for the porgramming examples and programming guide test suite at the end of the existing run tests workflow, after MLIR-AIE has been built. This consolidates the tests and examples steps into one, meaning we only build from source once and then test everything.
- Use the `build-mlir-aie-from-wheels.sh` script in the build-from-source job to make sure that script is actually tested. Removes duplication of CMake configure and ninja install commands in the GitHub workflow script.
- Factor out common prerequisite installation into a common `env_install.sh` script. There will be further opportunities for code deduplication, but this is a start.
- Add a "wheel smoke test." Previously, wheels were untested. Although the previous job for the example testing downloaded the wheel using `quick_setup.sh`, it **also** built MLIR-AIE from source and the actual `lit` tests it ran were form that from-source build. Since lit tests and the source build are somewhat tightly coupled, I opted to forgo `lit` and running the whole test suite on the wheels for now. Instead, we do a smoke test and run a couple programming examples on new wheel builds.
- Absorbs @hunhoffe's PR #3242 that added the logic for re-building and testing wheels for all PRs in the merge queue as well as PRs tagged with the `test-wheels` tag. The above smoke tests only run under these conditions.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -180,7 +180,7 @@ jobs:
     uses: ./.github/workflows/buildRyzenWheels.yml
     with:
       python_version: "3.12"
-      enable_rtti: ON
+      enable_rtti: "ON"
       reusable_call: true
     permissions:
       id-token: write
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -12,6 +12,9 @@ on:
       - main
       - test-ryzen-ai
   pull_request:
+    # Default activity types plus labeled/unlabeled so toggling the
+    # "test-wheels" label re-evaluates the gated wheel-example jobs.
+    types: [opened, synchronize, reopened, labeled, unlabeled]
   merge_group:
   schedule:
     # Runs at midnight (00:00) every day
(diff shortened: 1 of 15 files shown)


andrej:
Investigating the failing programming examples uncovered that these examples are currently reported as "unsupported" in the CI flow and are getting [skipped](https://github.com/Xilinx/mlir-aie/actions/runs/28413424264/job/84191131594#step:5:8627):
image
After the refactor, they actually run:
image

The root cause is that these tests have a `// REQUIRES: torch` which is not satisfied in the old flow but is in the new.

For now, I will disable the tests, but the fact that these examples are failing when torch is available probably means we should either remove or fix them.

The failing tests are:

Phoenix:
  AIE_PROGRAMMING_EXAMPLES :: ml/magika/run_phoenix.lit

Strix:
  AIE_PROGRAMMING_EXAMPLES :: ml/bottleneck/run.lit
  AIE_PROGRAMMING_EXAMPLES :: ml/conv2d_14x14/run.lit
  AIE_PROGRAMMING_EXAMPLES :: ml/resnet/layers_conv2_x/run_strix_makefile.lit

All of these have the `// REQUIRES: torch` which caused them to get skipped in the old CI.

andrej:
Contrary to what GitHub says, this passes all tests. That's because I renamed some of the tests (now "Run Tests **and Examples**" and **"Smoke-Test Wheels"** (new)).

I will change the branch protection rules to use these new tests as soon as we merge this.

This uncovered an issue in the smoke-test wheels build of an undocumented dependency we have: llvm-objcopy. I've decided for now to bundle llvm-objcopy with our mlir-aie wheel, which should make for a seamless user experience. We can always revert that decision later and decide to instead add it as a documented dependency in our readme (and install it system-wide on all runners).

jgmelber (approved):

hunhoffe (approved):
Thanks for doing this! It looks ready to merge from my end.

andrej:
@hunhoffe Just rebased. I think you have some plans for the merge queue so I'll let you hit merge when you think is appropriate.

PR #3232, opened 2026-06-27
[aiecc] Downgrade LLVM 23 f0x float literals for Peano compatibility

erwei-xilinx:
## Summary

LLVM 23 introduced a compact `f0x<8hex>` syntax for float32 constants in textual LLVM IR, where the 8 hex digits directly encode the IEEE-754 bit pattern. Peano's LLVM 21 `opt` cannot parse this format, causing aiecc to fail with:

```
opt: ...peano-compat.ll: error: expected value token
  %r = fadd float %x, f0x3727C5AC
```

This broke softmax and layernorm tests in [mlir-air#1693](https://github.com/Xilinx/mlir-air/pull/1693) when bumping mlir-aie to v1.3.3 (which uses a newer mlir-distro / LLVM 23 build that emits `f0x`).

## Fix

Extend `downgradeIRForPeano` in `tools/aiecc/aiecc.cpp` to rewrite each `f0x<8hex>` literal to the double-widened `0x<16hex>` form that LLVM 21 accepts:

- Scan the IR string for the `f0x` prefix followed by exactly 8 hex digits
- Decode the 32-bit bit pattern as IEEE-754 float32
- Widen to double and emit as `0x<16 uppercase hex digits>`

Example: `f0x3727C5AC` → `0x3EE4F8B580000000` (the float32 value 9.99999974E-6 widened to double).

## Test

`test/aiecc/peano_compat_f0x_float.mlir` — a minimal AIE core that uses `arith.constant 9.99999974E-6 : f32` (bits 0x3727C5AC, the layernorm epsilon), which LLVM 23 emits as `f0x3727C5AC`. The test checks:
- `--implicit-check-not="f0x"` — no `f0x` literals remain in the peano-compat.ll
- `CHECK-HEX: 0x3EE4F8B580000000` — the double-widened replacement is present

## Test plan

- [ ] `test/aiecc/peano_compat_f0x_float.mlir` passes (REQUIRES: peano)
- [ ] `test/xrt/43_triton_layernorm` passes after bumping mlir-air to include this fix
- [ ] `test/xrt/42_triton_softmax_bf16` passes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

PR #3247, opened 2026-06-30
[aiecc] Downgrade bare inf/nan literals in phi instructions for Peano compatibility

erwei-xilinx:
## Summary

Third fix in the series (after #3232 f0x floats, #3241 decimal bfloat16) for the same upstream LLVM 23 commit `41c214f0b115` (2026-05-07, https://github.com/llvm/llvm-project/pull/190649).

In phi instructions LLVM 23 emits infinity/NaN without a type prefix:
```
phi float [ %a, %entry ], [ -inf, %loop ]
```
Peano's LLVM 21 `opt` rejects the bare `-inf` keyword:
```
error: expected value token
  %7 = phi float [ %16, %10 ], [ -inf, %2 ]
```

The previous fixes handled the type-prefixed forms (`float -inf`, `bfloat 1.445e+00`). This PR handles the untyped form in phi/select operand lists.

## Fix

Extend `downgradeIRForPeano` with `replaceTypedLiteral` calls for `[ -inf`, `, -inf`, `[ inf`, `, inf`, `[ nan`, `, nan` → corresponding `0xFFF...`/`0x7FF...` double-widened hex forms that Peano's LLVM 21 accepts.

## Verification

Reproduced on local NPU1 (Phoenix): `programming_examples/mnist_fc/argmax` failed with the bare `-inf` error; **PASS!** after applying the fix.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/test/aiecc/peano_compat_phi_inf_nan.mlir
+++ b/test/aiecc/peano_compat_phi_inf_nan.mlir
@@ -1,0 +1,62 @@
+//===- peano_compat_phi_inf_nan.mlir - downgradeIRForPeano phi inf/nan ---===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+//
+//===----------------------------------------------------------------------===//
+
+// Regression test: downgradeIRForPeano must rewrite *bare* inf/nan literals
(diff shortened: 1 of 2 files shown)


hunhoffe (approved):
The others in this series had some testing attached, but this one doesn't seem to. Is there a reason why? Might be nice to have a follow on PR with tests if appropriate but no need to block this one.

erwei-xilinx:
> The others in this series had some testing attached, but this one doesn't seem to. Is there a reason why? Might be nice to have a follow on PR with tests if appropriate but no need to block this one.

Good catch. MLIR test added.

PR #3249, opened 2026-06-30
[aiecc] Parallelize unified per-core link loop + default -j to 4

jgmelber:
## Summary

On the chess (`xchesscc`) backend, `--unified` is the standard recipe: it collapses the 32× per-core compile into a single `xchesscc` process. But the per-core **link** that follows (BCF/ldscript + `xchesscc_wrapper`, or `clang`) was a plain serial `for`-loop — so `--unified` removed the 32× compile only to reintroduce a 32× *serial* link. The `-j` flag did nothing for it.

This PR makes three coordinated changes:

1. **Parallelize the unified per-core link loop.** Split into two phases:
   - *Phase 1 (serial):* everything that touches the shared MLIR module/context — `AIETranslateToBCF` / `AIETranslateToLdScript`, `link_with` file copies — building each core's final link argv into a self-contained `LinkJob`. The `MLIRContext` is not safe for concurrent access, so this stays single-threaded.
   - *Phase 2 (parallel):* run only the link subprocess per `LinkJob` via a bounded `llvm::DefaultThreadPool`. Each job has its own argv and a distinct output ELF, so they're independent. Concurrency honors the `-j` cap (important for chess, which oversubscribes past ~4 concurrent processes); `nThreads <= 1` keeps the serial path.

2. **Default `-j` to 4** (was 1/sequential). IRON passes `--unified` for chess but does **not** pass `-j`, so without a non-trivial default the parallel link never engages for the common path. `-j4` is the chess sweet spot; `-j1` remains available for memory-constrained hosts.

3. **Fix a latent chesshack temp-filename collision** in `compileCore()`. The `.chesshack.ll` name was built from `col + "_" + col + "_" + row` (column twice, no distinct row slot), so two cores in the same column but different rows produced the same path and could clobber each other under concurrent compilation. Corrected to `col_row`, matching every other per-core artifact — needed once `-j` defaults to parallel.

## Results

Measured on `whole_array` 8×4 (32 cores, npu2/aie2p, chess backend):

| config | before | after |
|---|---|---|
| bare `--unified` (default `-j`) | 149 s | **~74 s (~2×)** |

The win comes from the link parallelism; the `-j4` default is what lets IRON's bare `--unified` actually engage it. (Verified that on `main`, `--unified -j4` is still 149 s — the old serial loop ignored `-j` entirely.)

## Correctness

Output is **byte-identical** to the prior serial result:
- All 32 per-core link ELFs and `insts.bin` match the serial (`-j1`) output exactly.
- Verified on both chess and peano backends.
- End-to-end on NPU: 8×4 design **PASS**.
- A clean IRON chess build produces **0 compiler/toolchain warnings**; `aiecc.cpp` itself compiles warning-free.

## Test plan

- [x] Builds clean (no new warnings) on the from-wheels LLVM 23 toolchain
- [x] chess `--unified` 8×4 produces byte-identical `insts.bin` + per-core ELFs vs serial
- [x] peano default path produces byte-identical output vs baseline
- [x] 8×4 design runs PASS on NPU2 (Strix)
- [x] CI regression suite


hunhoffe (approved):
Is there a way to test parallelism/serialization regressions? Do we have parallel tests?

jgmelber:
Closing in favor of #3261, which takes just the `-j4` default (plus the chesshack filename fix and the cpp_multi_core CHECK-DAG fix it requires) and drops the parallel-unified-link rewrite that was hanging the Build and Test with Ryzen AI Software CI job.

PR #3252, opened 2026-06-30
[aiecc] Support >5 host buffers (fixes #3248 segfault)

### LINKED ISSUE #3248: IRON programs with more than five host buffers can segfault in XRT argument setup
## Summary

For the NPU/DPU-style xclbin path, `aiecc` appears to emit metadata for only five host buffer objects (`bo0` through `bo4`). Passing more than five IRON tensors to `DefaultNPURuntime.run(...)` can segfault in XRT/pyxrt argument setup before device execution.

## Minimal Structure That Triggers It

A two-worker IRON design can naturally exceed five host tensors:

```python
with rt.sequence(
    input_ty_a, output_ty_a, params_ty_a, debug_ty_a,
    input_ty_b, output_ty_b, params_ty_b, debug_ty_b,
) as (input_a, output_a, params_a, debug_a,
      input_b, output_b, params_b, debug_b):
    ...
```

and the host call passes eight tensors:

```python
aie.utils.DefaultNPURuntime.run(
    kernel_handle,
    [in_a, out_a, params_a, debug_a, in_b, out_b, params_b, debug_b],
)
```

The actual kernel compute does not seem to matter. Replacing the sixth or later tensor with a small dummy tensor still triggers the host-side crash.

## Observed Crash

```text
Segmentation fault (core dumped)
```

`gdb` shows the crash in XRT argument setup:

```text
#0 xrt::run_impl::validate_bo_at_index(unsigned long, xrt::bo const&)
#1 xrt::run::set_arg_at_index(int, xrt::bo const&)
#2 pyxrt...
```

## Metadata Investigation

Dumping embedded metadata:

```sh
xclbinutil \
  --dump-section EMBEDDED_METADATA:RAW:/tmp/metadata.raw \
  --input build/final.xclbin \
  --force
```

shows only:

```xml
<arg name="opcode" id="0" .../>
<arg name="instr"  id="1" .../>
<arg name="ninstr" id="2" .../>
<arg name="bo0"    id="3" .../>
<arg name="bo1"    id="4" .../>
<arg name="bo2"    id="5" .../>
<arg name="bo3"    id="6" .../>
<arg name="bo4"    id="7" .../>
```

There is no `bo5`.

## Apparent Hardcoded Limit

The limit appears to be hardcoded in `tools/aiecc/aiecc.cpp`:
https://github.com/Xilinx/mlir-aie/blob/afc887fcc9eaf46757901df3064e7a746636645e/tools/aiecc/aiecc.cpp#L3796-L3804

The test `test/aiecc/buffers_xclbin.mlir` also checks only `bo0` through `bo4`.
https://github.com/Xilinx/mlir-aie/blob/afc887fcc9eaf46757901df3064e7a746636645e/test/aiecc/buffers_xclbin.mlir#L26-L29

## Workaround

Packing/reducing the runtime ABI to five tensors avoids the segfault, e.g.:

```text
input_a, output_a, shared_params, input_b, output_b
```

## Questions

1. Is the five-host-BO limit intentional for this xclbin path?
2. If yes, could it be documented for IRON users?
3. Could `aiecc` diagnose `aie.runtime_sequence` with more than five host memrefs for this target/path?
4. Could `DefaultNPURuntime.run(...)` raise a Python exception instead of letting XRT segfault?
5. Is increasing or deriving the number of generated `boN` entries from the runtime sequence feasible?

hunhoffe:
Supersedes #3229. Thanks to @atassis for identifying the root cause and the initial fix in #3229/#3248.                            
                                                                                                                                     
## Problem                                                                                                                         
                                                                                                                                     
`generateKernelsJson` hardcoded exactly 5 host buffer (`boN`) entries in `kernels.json` regardless of the actual `aie.runtime_sequence` argument count. Kernels with more than 5 host BOs got an under-declared ABI, causing XRT to segfault in argument setup (#3248).                                                                                                            
                                                                                                                                     
Investigating the fix surfaced a deeper, related issue: even with a correctly-declared ABI, kernels with more than 5 host buffers silently produced wrong results: the 6th (and later) buffer's DMA never ran, with no error.
                                  
## Fix                                                           
                                  
**Host-BO count = the runtime_sequence operand list.** The lowered `aie.runtime_sequence` argument list is the host-call ABI: argument *i* maps to host BO *i* (after the fixed opcode/instr/ninstr scalars). `generateKernelsJson` now derives the count from `getNumArguments()` instead of hardcoding 5. Every host buffer — data, trace, control-packet — is an operand of that list.
                                  
**Support >5 host buffers.** For arg indices ≥ 5, `AIETargetNPU` now folds an offset into the DDR address patch, so buffers beyond the 5th are DMA'd from the correct address. Verified on Strix up to 16 host buffers.
                                                                   
**Guard the ceiling.** `aiecc` rejects designs above `kMaxHostBOs` (16) — a conservative, hardware-verified ceiling — with a clear compile-time error, rather than silently miscompiling.           
                                                                   
**Trace buffer is an appended operand.** Trace lowering now appends the trace buffer as a `runtime_sequence` operand (landing at the tail) instead of pinning it to a hardcoded XRT arg index (`arg_idx=4`) with host-side filler padding. The former `arg_idx=-1` "append after last output" mode becomes an explicit `reuse_output_buffer` flag. Enabling trace no longer perturbs the data arguments' indices, and the offset translation above applies to trace/control-packet buffers uniformly.
                                                                   
**Unify the C++ test harness.** The two positional `setup_and_run_aie` overloads in `xrt_test_wrapper.h` are replaced by a single variadic wrapper that passes exactly the declared host BOs (data, then optional control-packet/trace). All call sites migrated.

## Follow-ups (not in this PR)                                   
- The ctrl-packet reconfig "base" xclbin has no `runtime_sequence` of its own (its ABI comes from a separate "main" compile); it uses a conservative fallback count. Making it self-describing is tracked separately.
- The firmware's 5-argument translation asymmetry is worked around here; an FYI to the firmware team is optional since designs now compile and run correctly.                                       
                                                                   
Fixes #3248.                                                     
Co-authored-by: Taimuraz Kaitmazov <atassikay38@gmail.com>   


## Update: Control-packet reconfig: correct base xclbin host-buffer count
  
The base xclbin (loaded on hardware) has no runtime_sequence of its own — its host-buffer ABI is defined by the sibling @main device. aiecc previously fell back to a fixed count (5) for it, which silently under-declared the ABI (XRT segfault) for designs needing more than 5 buffers. It now derives base's exact kernels.json host-buffer count from @main's sequence (data args + control-packet buffer). Verified on npu1 across the ctrl_packet_reconfig tests plus a new aiecc lit test.


python/utils/trace/__init__.py
```diff
@@ -2,161 +2,72 @@
...
-        This one allows us to control the size, offset, and inout buffer mapping.
-
-        To better appreciate what this wrapper function does, we need to delve more deeply into the details on how trace units are configured.
-
-* Additional helper functions can be found in the `trace.py` and are documented in the source directly.
-
-Trace Mechanisms and Explanations
-
-The basic concept for trace configuration as summarized in [section-4b](../../programming_guide/section-4/section-4b/). MOre details about the trace hardware can be found for AIE-ML/AIE2 at [am020](https://docs.amd.com/r/en-US/am020-versal-aie-ml/Trace).
+Trace utilities
```
andrej, line 5: Are these documentation changes intended?

test/dialect/AIE/trace/test_insert_trace_flows.mlir
```diff
@@ -301,7 +301,7 @@ module @trace_after_last_tensor {
     }
     aie.runtime_sequence(%arg0: memref<16xi32>, %arg1: memref<1024xi32>) {
       // CHECK: aiex.npu.address_patch {{{.*}}arg_idx = 1{{.*}}arg_plus = 4096{{.*}}}
-      aie.trace.host_config buffer_size = 8192 arg_idx = -1
+      aie.trace.host_config buffer_size = 8192 reuse_output_buffer
```
andrej: Doesn't need to be addressed in this PR ... but why does aie.trace.host_config need its own parser with this custom syntax? Why not just a regular attr-dict?

test/npu-xrt/many_buffers/test.cpp
```diff
@@ -0,0 +1,108 @@
...
+  auto device = xrt::device(0);
+  auto xclbin = xrt::xclbin(vm["xclbin"].as<std::string>());
+  std::string kernel_name = vm["kernel"].as<std::string>();
+
+  auto xkernels = xclbin.get_kernels();
+  auto xkernel = *std::find_if(xkernels.begin(), xkernels.end(),
+                               [&](xrt::xclbin::kernel &k) {
+                                 return k.get_name().rfind(kernel_name, 0) == 0;
+                               });
+  auto kernelName = xkernel.get_name();
```
andrej: I know this is just copied over from other test.cpps, but I don't know why we have this block in all our test.cpps and I'd prefer not to add it to new ones. At the end of all this, all we have is `kernelName == kernel_name` (if it succeeds) or a hard segfault if the kernel didn't exist (which is a bad error vs. what XRT would produce later on in the `xrt::kernel` instantiation). `xkernel` isn't otherwise used, it's just to get the same `kernel_name` string into `kernelName`.

tools/aiecc/aiecc.cpp
```diff
@@ -5090,6 +5090,60 @@ generateFullElfArtifact(ArrayRef<DeviceElfInfo> deviceInfos,
...
+/// The NPU firmware only pre-translates the first 5 host buffer addresses from
+/// the host address space into the AIE address space. Buffers beyond that would
+/// be DMA'd from the wrong address -- except AIETargetNPU folds the translation
+/// offset into the DDR address patch for those args (see kDDRAIEAddrOffset), so
+/// designs with more than 5 host buffers work correctly. This cap is therefore
+/// a conservative, verified ceiling rather than a hard hardware limit: host BO
+/// counts up to this value are validated on hardware (see the many_buffers
+/// test). Raise it (and extend the hardware tests) if larger counts are needed
+/// and verified.
+static constexpr int kMaxHostBOs = 16;
```
andrej, line 5139: Why not test all the way up to 20, which is the firmware limit on number of total BOs?

tools/aiecc/aiecc.cpp
```diff
@@ -5090,6 +5090,60 @@ generateFullElfArtifact(ArrayRef<DeviceElfInfo> deviceInfos,
...
+
+/// Fallback host-buffer (boN) count for a device that has no runtime_sequence.
+/// Such a device has no operand list to read, yet may still be launched by the
+/// host with host buffers (e.g. the ctrl-packet reconfig "base" xclbin, whose
+/// ABI is defined by a separate "main" compile that is not present when base is
+/// built). XRT tolerates a kernel declaring more host BOs than the host passes;
+/// under-declaring segfaults (#3248). Declaring this many slots keeps that case
+/// working until base can be made self-describing (tracked follow-up).
+static constexpr int kNoSequenceHostBOs = 5;
```
andrej: Do we have tests that break without this fallback functionality?

python/utils/trace/__init__.py
hunhoffe: Yes — the previous docstring documented the old procedural trace API (configure_coretile_tracing_aie2, configure_packet_tracing_aie2, configure_simple_tracing_aie2, etc.), which this PR removes in favor of the declarative aie.trace ops. Left unchanged it would describe functions that no longer exist, so I rewrote it to cover the current public surface (configure_trace, start_trace, TraceConfig, parse_trace) and the operand-list-as-ABI model.

test/dialect/AIE/trace/test_insert_trace_flows.mlir
hunhoffe: Absolutely no good reason. Nice catch! Will push a fix.

test/npu-xrt/many_buffers/test.cpp
hunhoffe: Agreed — dropped the find_if/xkernel block. The test now passes kernel_name directly to xrt::kernel, which surfaces XRT's own error if the kernel is missing instead of segfaulting.

tools/aiecc/aiecc.cpp
hunhoffe: `kMaxHostBOs` is set to the count we've actually verified on hardware, not the firmware's theoretical ceiling. The many_buffers test exercises 16 host BOs (8 in + 8 out), which is the most we've run on silicon — and on a 4-column part that's the single-round maximum(4 columns × 2 shim channel-pairs). 

Easy to bump later: raise the constant and extend many_buffers once >16 is verified. I left a note to that effect in the comment above the constant.

hunhoffe: @andrej -- are you ok with me leaving at 16 for now?

tools/aiecc/aiecc.cpp
hunhoffe: Yes — test/npu-xrt/ctrl_packet_reconfig. Its @base device is just tiles + packet flows with no runtime_sequence, so there's no operand list to derive the count from; without the fallback computeNumHostBOs returns 0 and the base xclbin under-declares its host-BO ABI (the same under-declare that segfaults in #3248). The base's real ABI is actually defined by the separate @main compile, which isn't present when base is built — hence the fixed fallback until base can be made self-describing (noted as a follow-up in the comment).

Are you ok with me leaving as-is for now?

hunhoffe:
Notes on updated scope:
- Runtime.enable_trace(ddr_id=...) and the aie.trace.host_config/writebd ddr_id index are removed. Trace no longer occupies a fixed host arg slot — it's appended at the tail. To share an output buffer instead of a dedicated one, pass reuse_output_buffer=True.
- Out-of-tree C++ host harnesses that passed the trace buffer at a fixed group_id must pass it after the data buffers (the in-tree harnesses are migrated to the new variadic setup_and_run_aie).

runtime_lib/test_lib/xrt_test_wrapper.h
fifield, line 1: Not really related to this PR, but it might be nice if this library had its own namespace.

fifield (approved):

runtime_lib/test_lib/xrt_test_wrapper.h
hunhoffe: I'll defer this for now. But yes, that makes sense to me.

tools/aiecc/aiecc.cpp
andrej: Sure!

tools/aiecc/aiecc.cpp
andrej: Sure. Maybe we can change that test to something sensible (add an empty runtime sequence decl, since it's just a single test) at a later point and drop all this code, as to me it seems a design that doesn't specify a runtime interface shouldn't be able to assume any.

tools/aiecc/aiecc.cpp
```diff
@@ -4324,6 +4324,35 @@ static LogicalResult generateTransactionOutput(ModuleOp moduleOp,
...
+/// host-buffer contract before any ctrl-packet buffer is appended (the ctrlpkt
+/// operand is added later by AIECtrlPacketToDma, outside this count).
+static std::optional<int>
+getRuntimeSequenceArgCount(xilinx::AIE::DeviceOp dev) {
+  for (auto seqOp : dev.getOps<xilinx::AIE::RuntimeSequenceOp>()) {
+    if (!seqOp.getBody().empty())
+      return static_cast<int>(seqOp.getBody().front().getNumArguments());
+  }
+  return std::nullopt;
+}
```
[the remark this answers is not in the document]
hunhoffe: done

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -257,6 +257,23 @@ def run(
...
+        # Validate BO count against xclbin metadata before calling into XRT.
+        # XRT's validate_bo_at_index segfaults if the index exceeds the count
+        # declared in kernels.json. The fixed args (opcode, instr, ninstr) are
+        # the first 3; the rest are host BOs. Passing fewer BOs than declared is
+        # fine (the kernel may declare a minimum ABI width); passing more is the
+        # fatal case.
+        xclbin_kernels = kernel_handle.xclbin.get_kernels()
+        if xclbin_kernels:
+            declared_bo_count = xclbin_kernels[0].get_num_args() - 3
+            if len(buffers) > declared_bo_count:
```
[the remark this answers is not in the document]
hunhoffe: done

PR #3254, opened 2026-06-30
[aiex] Reject deadlocking task-completion-token imbalances in runtime sequences

hunhoffe:
An `aiex.dma_await_task` with no matching `issue_token` `dma_start_task` on its channel blocks the runtime sequence on an empty task-completion-token (TCT) FIFO forever. This adds `verifyTokenBalance` to the BD-ID allocator to catch that at compile time.

Since the allocator runs after runtime control flow is rejected and constant loops are unrolled, the check sees only straight-line IR: a per-channel count (`+1` per `issue_token` push, `-1` per await, error if negative). Over-production stays legal, so awaiting the last of N pushes and letting the FIFO cover the rest is not flagged.

The BD-ID recycle path can't catch this — it tracks descriptor-pool reuse and tolerates repeated awaits (the await-then-free idiom), which is exactly the under-production signature.
  
Tests: `good-tct-fire-and-forget` (push-N/await-1 anti-false-positive), `good-tct-nontoken-await-ignored`, `bad-tct-imbalance` (three deadlock cases).


andrej (approved):
I think this will be a great debuggability improvement for deadlocks!

PR #3224, opened 2026-06-25
[dyn-seq P1.1] Extract TXN instruction encoding into a standalone header

hunhoffe:
PR 1 of the dynamic-runtime-sequences rebuild ([#3222](https://github.com/Xilinx/mlir-aie/discussions/3222)).

Moves TXN binary word-encoding out of `AIETargetNPU.cpp` into a new dependency-free header `include/aie/Runtime/TxnEncoding.h` (`<cstdint>` only), so it can be the single source of truth for the firmware format — used by the compiler today and `#include`d by generated host C++ in later PRs.

- New header: `aie_runtime` namespace, `TxnOpcode` enum, `TxnDeviceInfo`, `txn_init` / `txn_prepend_header` / `txn_append_*`.
- `AIETargetNPU.cpp` `append*` become thin wrappers delegating to the header; the hand-copied enum is removed in favor of the header's (one in-tree definition — can't drift). Block-write cache and byte layout preserved.

**Pure refactor: emitted binaries are unchanged.**

## Placement — milestone #3222

Phase 1 (TXN/EmitC foundation), PR 1 of 4.
- **Depends on:** nothing (independent of Phase 0).
- **Unblocks:** PR 3 (AIEX→EmitC, includes this header), PR 4 (equivalence harness).

## Notes for reviewers

- "Pinning" is **structural**, not `static_assert`: one in-tree definition, so nothing to assert against. Opcodes intentionally don't match the submodule's`xaie_txn.h` (older numbering) — see the `DRIFT WARNING` in the header.
- No CMake change needed — `include/aie/**/*.h` is packaged recursively (ships in wheel).
- Scratchpad ops left inline (referencing the header enum); behavior unchanged.
- Out of scope: SSA-operand ops, EmitC generation, dynamic BD encoder (later PRs).

PR #3256, opened 2026-07-01
[dyn-seq P1.3] AIEX→EmitC conversion + C++ TXN target  

hunhoffe:
## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
Adds the runtime-parameterizable mirror of the compile-time binary TXN emitter (`AIETargetNPU.cpp`). A straight-line `aie.runtime_sequence` of npu ops lowers to a \standalone `generate_txn_<seq>()` C++ function that assembles the transaction (TXN) stream by calling `TxnEncoding.h` (the standalone header from #3224). Scalar operands — constant or runtime — flow through as C++ literals / function parameters.
  
This is Phase 1 PR #3 of the dynamic-runtime-sequences milestone (#3222): the compiler foundation the dynamic path builds on.
  
## What's included
  
- **New pass `--convert-aiex-to-emitc`** — converts the straight-line npu ops (`write32`, `maskwrite32`, `sync`, `address_patch`, `blockwrite` + the feeding `memref.get_global`) to `emitc` calls on `TxnEncoding.h`. Device-relative addresses (buffer/column/row) are resolved to literals; other scalar fields pass through as SSA. Runs upstream `ArithToEmitC` so `arith.constant` **and** derived runtime values (e.g. `%p + 1`) lower to EmitC uniformly. `op_count` and the device-info header are statically emitted.
- **New `aie-translate` target `--aie-npu-to-cpp`** — runs the pass + `translateToCpp`, producing the C++ builder.
- **Scope boundary:** `scf.for`/`scf.if` inside a runtime sequence are rejected with a clean diagnostic (control flow is later-PR work); unsupported ops `emitOpError` (never silently dropped). 
    
## Correctness
  
For **constant** operands, the generated C++ (compiled + run) produces a **byte-identical** TXN stream to `aie-translate --aie-npu-to-binary` — verified across all five op types on npu1 and npu2. This is the milestone's "static ≡ dynamic" criterion, checked in compiled C++ (which `aie-translate` alone can't do).
  
For **runtime** (non-constant) scalar operands, the value becomes a `generate_txn_<seq>(int32_t ...)` parameter — one artifact, many values — which is the foundation the dynamic runtime-sequence work needs. (The binary path correctly rejects these; a runtime *address*, which selects a hardware register, is also rejected, matching the binary emitter.)

## Out of scope (later milestone PRs)
  
Control flow (scf → unroll / dynamic); the automated static-vs-dynamic equivalence harness; the dynamic BD-word encoder (runtime transfer extents); the host bridge that dispatches a runtime-scalar TXN on hardware.


lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -0,0 +1,427 @@
...
+  // The compile-time-resolved absolute address (folding buffer/col/row) as a
+  // uint32_t literal, or null if it could not be resolved statically. Address
+  // resolution depends on the parent device, so it is precomputed before the op
+  // is cloned out (see recordDeviceResolved).
+  Value resolvedAddr(OpBuilder &b, Location loc, Operation *clone) {
+    auto it = resolved.absoluteAddr.find(clone);
+    if (it == resolved.absoluteAddr.end())
+      return {};
+    return u32Literal(b, loc, it->second);
+  }
```
andrej, line 209: Any reason the address must be compile-time-resolved? Although I can't think of a good use case, I'm just curious whether it would be possible to run-time calculate these as well, and pass them in as SSA values just like elsewhere.

Not a blocker since there's probably no use case for this, just asking out of curiosity.

andrej (approved):
Looking good!

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
hunhoffe: I definitely missed some values that should be SSA/prepared for dispatch-time resolution. I'm working on a PR to address that now -- I'll make sure to include this. Thanks for pointing this out!

Issue #2586, opened 2025-09-17
Problem Implementing Broadcast with skip objectfifo for GRU network


iuwizeyimana:
I am currently implementing a GRU network on the NPU using the IRON API.

The hardware mapping used in this setup is shown in the picture [here](https://drive.google.com/file/d/1quabjDmI497TclCF4SYaLatUigAlu0JL/view?usp=sharing). 

For runtime data movement, I am following the pattern from the [single-core matrix-multiplication example](https://github.com/Xilinx/mlir-aie/blob/main/programming_examples/basic/matrix_multiplication/single_core/single_core.py), which serves as the baseline for my matrix multiplication compute tiles (A–F in the picture). No memory transformations are applied to the objectfifos transferring data between compute tiles.

The issue arises with compute tile **J**. As shown in the diagram, tile J participates in a broadcast with skip connection pattern as described [here](https://github.com/Xilinx/mlir-aie/tree/main/programming_guide/section-2/section-2b/02_Broadcast#broadcast-with-skip-connection). The tile consumes vector **H** from the mem tile a few cycles after other consumers of H, on which J depends for its second input (**Z**).

A FIFO of depth 2 between the mem tile and compute tiles works correctly in a non-tiled configuration (where N=n, M=m, K=k for inputs of size M×K and weights of size K×N), but fails in the tiled setup. Adjusting the FIFO depth such that each tile of the objectfifo that transfers H from the mem tile to the compute tiles has its own FIFO (as suggested in the broadcast-with-skip example) does not resolve the problem.

Interestingly, the issue persists as long as tile J is included as a consumer of the objectfifo even if it does not actually consume the vector in its core computation.

Any help with resolving this issue would be appreciated.

AndraBisca:
Hello @iuwizeyimana and thank you for submitting this detailed issue!
Could you please share your IRON design code so I may get a better understanding of how the ObjectFifos are set up?
Also when you say the design fails, do you get wrong results, or is there an error either during compilation or at runtime?

iuwizeyimana:
Thank you for the follow up!
I have condensed all the files needed to reproduce the error in [this repository](https://github.com/iuwizeyimana/GRU_on_NPU).

The IRON design is in the [gru_full.py](https://github.com/iuwizeyimana/GRU_on_NPU/blob/main/gru_full.py) file. 

By "design fails" I mean I am getting wrong results (all the output values are 0).

[...]

Issue #2463, opened 2025-07-31
Seeking help: certain behaviours in core_body() causes Kernel returned ert_cmd_state.ERT_CMD_STATE_ABORT


RonZ13:
Hi there,

I was trying to create a CNN example using the placed level IRON. I found a weird behaviour with my program. The code below compiles, but it causes exceptions: Kernel returned ert_cmd_state.ERT_CMD_STATE_ABORT. In the **_core_body_** function, If I comment out the acquire&release of the objectFIFO that passes in the weight values and reduce the inner loop interations from 14 to 7. Then it runs without causing exceptions. I am puzzled with why this causes the kernal to crash. For example, the **edge_detect** example in the programing_examples, the **_core_body_** function that handles the filter2D can handle a inner loop with 1080 iterations. However, somehow for my case the inner loop cannot handle anything greater than 7 iterations. I might be misunderstanding some concepts here, so please feel free to correct me if I'm mistaken.

I have also attached my entire example files in a zip file at the bottom (Makefile + host program + kernel + placed-level IRON code that is shown here).

Your help is highly appreciated,
Ron

```
import numpy as np
import sys
import argparse

from aie.dialects.aie import *
from aie.dialects.aiex import *
from aie.helpers.dialects.ext.scf import _for as range_
from aie.extras.context import mlir_mod_ctx

import aie.utils.trace as trace_utils
from aie.utils.trace import PortEvent, ShimTilePortEvent
from aie.utils.trace_events_enum import CoreEvent, ShimTileEvent, MemTileEvent, MemEvent

dtype_map = {
    "i8": np.int8,
}

def CNN(opts):
    image_width = opts.image_width
    image_height = opts.image_height
    num_rows = opts.num_rows
    num_cols = opts.num_cols
    trace_size = opts.trace_size

    layer1_padding = 1
    layer1_stride = 1
    layer1_kernel_width = 3
    layer1_kernel_height = 3
    layer1_per_core_input_width = 16
    layer1_per_core_input_height = 16
    layer1_input_channel = 1
    layer1_per_core_output_width = 14
    layer1_per_core_output_height = 14
    layer1_output_channel = 8
    layer1_per_core_after_maxpool_width = 7
    layer1_per_core_after_maxpool_height = 7
    layer1_W_total = layer1_input_channel * layer1_kernel_width * layer1_kernel_height * layer1_output_channel
    layer1_X_total = layer1_input_channel * image_width * image_height
[...]

PR #3257, opened 2026-07-01
[AIE] Add aie-objectfifo-liveness: a static under-buffering check for coupled multicast objectFIFOs

atassis:
Scope up front so nobody is surprised: this catches ONE class of objectFIFO
hang, a coupled cyclic multicast that gets replayed and runs out of buffer slack.
It is not a general deadlock detector. Everything it does not cover is listed
under "Explicitly NOT in this PR", with tests showing it stays quiet on those. I
would rather land a small check I can prove correct than a broad one that
guesses, so the rest is planned as follow-ups (see Roadmap).

## Motivation

ObjectFIFO designs can compile completely clean and then silently hang the array
at runtime. There is no diagnostic today: the lowering, routing, and BD
assignment all succeed, the xclbin builds, and the NPU simply deadlocks when you
run it. Debugging that is brutal: no error to grep, no failing pass, just a hang.

We hit exactly this on a 16-core whole-array GEMM-to-GEMM FFN (`wa_gemm2`). The
design fans a weight objectFIFO out to a row of compute cores as a coupled
broadcast and replays it across multiple trips. With output-tiling P_div_p = 8
and inner trip-count N_div_n = 2 it deadlocks; drop either axis (P_div_p or
N_div_n down to 1, or the fan-out) and the identical design runs. The failure is
structural: the coupled broadcast's outstanding tokens outrun the objectFIFO's
round-trip buffer slack, so the producer can never re-acquire a slot. It took
days to localize because nothing in the toolchain flagged it.

This pass turns that silent class of hang into a compile-time error.

### Where it runs

`--aie-objectfifo-liveness` runs once per `aie.device` (like `aie-assign-bd-ids`)
on the high-level objectFIFO IR, before `--aie-objectfifo-stateful-transform`
lowers it. It is a pure analysis: it emits diagnostics and does not mutate the
IR. It is opt-in and is NOT added to any default pipeline, so it changes no
existing flow -- you invoke it explicitly (or wire it into a design's own check
step) and it either stays silent or errors out before you build an xclbin that
would hang.

### What it does

`--aie-objectfifo-liveness`:

1. Walks every `aie.objectfifo` and records producer tile, consumer tiles,
   depth, and repeat_count.
2. Builds the objectFIFO data + back-pressure dependency graph: fifos that share
   a tile (one produces into it, another consumes from it) get a bidirectional
   edge (data forward, back-pressure backward).
3. Finds cyclic strongly-connected components with Tarjan's algorithm (keeping
   components of size > 1).
4. Applies the validated static-dataflow (SDF) buffer model to the cyclic
   coupled multicasts and errors when the allocated depth is below the
   deadlock-free minimum.

### The model

For the cyclic multicast objectFIFOs:

```
demand   = array_fan * T          array_fan = name-base-summed multicast
                                  fan-out within the cycle (the coupled
                                  consumer count); T = max objectFIFO
                                  repeat_count (the replay trip-count)
slack    = 2 * depth              round-trip buffer slack
DEADLOCK iff  T >= 2  AND  depth > 0  AND  demand > slack
```

Where the two terms come from. Each replay trip the producer has to hand the same
buffer to `array_fan` coupled consumers, and it repeats that `T` times before it
can retire the slot, so up to `array_fan * T` tokens are outstanding at once (the
demand). The buffer that absorbs in-flight tokens is the round trip: the depth of
slots the producer can fill ahead plus the depth of slots the consumers are
holding, i.e. `2 * depth` (the slack). Once demand exceeds slack the producer
cannot re-acquire a slot for trip t+1 while trip t's tokens are still in flight,
and the coupled broadcast is a circular wait.

When it fires, the required minimum depth reported is `ceil(demand / 2)`.

## Soundness stance

Sound but not complete, same idea as a borrow checker. It is sound for the
static SDF class it models: a coupled, cyclic, replayed multicast with a fixed
fan-out and a single integer depth. Within that class it does not miss a
deadlock, and it does not false-positive (it only errors inside a proven cyclic
under-buffered multicast SCC).

Outside that class it stays silent rather than guess. Specifically:

- The `T >= 2` replay guard. A single-trip (T == 1) broadcast fans out once and
  drains monotonically; it never forms the re-acquire cycle regardless of
  fan-out. The bundled whole-array matmul examples broadcast to the entire array
  at depth 2, T == 1, and run on device -- the guard keeps the pass quiet on
  them. The cycle only exists when the producer must re-acquire a slot on trip
  t+1 while trip t's tokens are still outstanding.
- Array-form (per-tile) objectFIFO depths are skipped (only single-integer
  depths are modeled), so they are never flagged.
- Data-dependent / dynamic dataflow is out of scope.

## Example diagnostic

On the `n8_nd2` reproducer (fan-out 8, depth 4, repeat_count 2):

```
error: objectFIFO @memW1 in a cyclic dependency requires depth >= 8 for
deadlock-free execution; allocated depth = 4. (Coupled broadcast fan-out 8 x
trip-count 2 = 16 outstanding tokens exceeds round-trip slack 2*depth = 8;
raise depth or reduce coupled consumers / output tiles.)
```

The message names the offending objectFIFO, the required minimum depth, the
allocated depth, and the arithmetic, plus the two ways to fix it (raise depth or
cut coupled consumers).

## Validation

The model was checked against a measured device oracle: the same whole-array FFN
swept across fan-out (n_cols) and replay trip-count (N_div_n), each configuration
actually run on XDNA2. The pass errors on exactly the configuration that
deadlocks on silicon and stays silent on the ones that run:

| config              | fan (array_fan) | T | depth | demand | slack | pass     | device   |
|---------------------|-----------------|---|-------|--------|-------|----------|----------|
| n8_nd2 (reproducer) | 8               | 2 | 4     | 16     | 8     | error    | DEADLOCK |
| n4_nd2              | 4               | 2 | 4     | 8      | 8     | silent   | runs     |
| n2_nd2              | 2               | 2 | 4     | 4      | 8     | silent   | runs     |
| n1_nd2              | 0               | 2 | 0     | 0      | 0     | silent   | runs     |
| n8_nd1              | 8               | 1 | 4     | 8      | 8     | silent   | runs     |

5/5 agreement. (`repeat_count` is the oracle's own discriminator: n8_nd2 and
n8_nd1 differ only in T, and only n8_nd2 hangs -- which is what motivates the
`T >= 2` guard.) The pass is also clean across the bundled `programming_examples`
/ `test` objectFIFO designs (no false positives on the shipped known-good
examples), and stays silent on the out-of-scope reproducers listed below.

The hang/run ground truth was measured on XDNA2, so reproducing it needs that
hardware. What travels with this PR is checkable without a device: the bundled
`n8_nd2` reproducer, the lit tests asserting the pass behavior, and the table
above showing the boundary the model draws.

This check complements existing lowering-time depth heuristics in
`AIEObjectFifoStatefulTransform` (broadcast consumer-depth auto-adjust; the
cyclostatic acquire-in-loop peel), which are per-class fixes shown incomplete by
user-filed hangs such as #2586 and #2463. There is no existing static pass that
flags this under-buffering class before the run.

## Tests

Lit-style tests using `--verify-diagnostics`:

Positive (must flag):
- `under_buffered.mlir` -- the n8_nd2 reproducer (fan 8, depth 4, T = 2) with the
  `expected-error` annotation on `@memW1`.

Negative / true-negative (must stay silent -- these pin the soundness guards):
- `acyclic.mlir` -- two independent feed-forward fifos, no shared
  producer/consumer tile, no cycle.
- `sufficient.mlir` -- the full wa_gemm2 FFN at a depth that satisfies the model.
- `single_trip_wide_broadcast.mlir` -- a wide (fan-out 8) coupled broadcast in a
  cycle but SINGLE-TRIP (no repeat_count, T = 1): the replay guard keeps it
  silent, matching the known-good whole-array matmul examples that broadcast
  array-wide at low depth and run on device.
- `pointpoint_cycle.mlir` -- a replayed (T = 2) two-core cyclic dependency built
  from point-to-point fifos (fan-out 1, array_fan = 0): no coupled-broadcast
  demand, so the pass stays silent even though the cycle and the T >= 2 guard are
  both present.

Additionally validated silent (no false positive) against the in-tree
out-of-scope reproducers `dynamic_cyclostatic_acquire_in_loop.mlir` (#2463),
`link/link_test_broadcast_skip_connection.mlir` (#2586),
`broadcast_enforced_depths.mlir`, `broadcast_self_adjusted_depths.mlir`, and
`bd_chain_on_memtile/broadcast_with_iterations.mlir`, plus the full bundled
objectFIFO example/test corpus.

## Explicitly NOT in this PR (out-of-scope classes -- the pass stays silent, tested)

These are other ways an objectFIFO / AIE dataflow program can hang. This pass
does NOT model them and, by design, does not flag them (verified silent against
the in-tree reproducers noted). They are called out so the boundary is explicit,
not accidental:

- Cyclostatic single-fifo acquire-in-loop exhaustion (issue #2463). A per-
  iteration `acquire(N)` on a single fan-out-1 fifo, handled today by the
  stateful-transform peel rewrite. Different (temporal, single-consumer) class.
- Broadcast-with-skip-connection under-depth (issue #2586). The extra buffer a
  consumer needs comes from second-input (join) latency, not from fan-out, so the
  fan-based demand term does not see it. This is the most likely false-negative
  and is called out honestly.
- Cross-fifo mutual cycles (two cores each cyclostatic on a fifo the other
  produces). The circular wait spans two fifos; a per-fifo demand/slack test does
  not model it.
- Cascade-bus / lock re-arm across host re-dispatch (mlir-air #1694). Lives on the
  cascade bus, not on `aie.objectfifo`, and is sign-inverted vs. this model
  (single-trip is the hazard there, not multi-trip), so it wants its own check.
- DMA BD-chain / iter_count / repeat_count resource exhaustion (runtime-sequence
  level, below the objectFIFO SDF layer).
- Automatic buffer sizing (this pass diagnoses + reports the minimum depth; it
  does not rewrite the depth for you).
- Throughput / Max-Plus timing analysis -- this is a liveness check, not a
  performance model.
- The P_div_p (output-tiling) axis as an independent term. The current device
  data constrains the fan-out x replay product; it does not separately constrain
  P_div_p, so the model folds it into array_fan rather than modeling it on its
  own axis. Revisit if future data separates the two.

## Roadmap (planned follow-up PRs)

I intend to extend this check incrementally in later PRs, each gated on a
reproducer + (where it needs one) a device oracle point, rather than bolting
speculative heuristics onto this one. Likely order:

1. Skip-connection / join-latency demand term (#2586) -- turn the biggest known
   false-negative into a modeled, tested positive.
2. Cross-fifo cycle demand (mutual ping-pong across two fifos).
3. Cyclostatic single-fifo depth check (#2463), if not better left to the peel
   rewrite it currently relies on.
4. A separate cascade-bus liveness check (mlir-air #1694 class), different enough
   to want its own pass rather than overloading this one.

Landing the small correct core first keeps each follow-up reviewable on its own.


lib/Dialect/AIE/Transforms/AIEObjectFifoLiveness.cpp
```diff
@@ -0,0 +1,254 @@
+//===- AIEObjectFifoLiveness.cpp --------------------------------*- C++ -*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// (c) Copyright 2026 Advanced Micro Devices Inc.
```
jgmelber:
```suggestion
// Copyright (C) 2026 Advanced Micro Devices, Inc.
```

jgmelber (commented):
Thanks! This is cool, and I appreciate the roadmap.

Triggering CI.

Tests need copyright headers.

atassis:
@jgmelber I am continuing to catch deadlocks from time to time even now during work. I have now fixed all the mentioned issues per reviews.

Also, I'll try to bump the other PRs done by me across the toolchain repos. I have to juggle them locally, and it makes it a bit difficult to land new ones without breaking the flow. Would you and the team mind if I ping here and there?

jgmelber:
> Would you and the team mind if I ping here and there?

No problem!

hunhoffe:
@atassis I'm really happy to see progress in this area, thanks for contributing! If we don't make it part of the default pipeline, we'll need to find the correct place to clearly document how/when to run this so other users are made aware of these types of checks.

atassis:
@hunhoffe I understand this can't be landed onto default path now, including the fact that there are many places to track deadlocks down. 
I want to cover all cases first, including ones I meet during development, and only then to think about what is a proper way to land this onto default path (it might be useful, but needs a proper DX look from above). I still study, research the work you do, so this might take some time to land properly, and I am happy to get any feedback on any suggested PR.
One of my intents is to make the toolchain to break early, so there are less situations, where the build is clean, but runtime fails.

atassis:
I have read through again and understood that I have answered without answering.
Per documentation- can we merge this and I bump little later the documentation? I have to think a little bit more on that, to land proper text, etc.
And per landing properly- I meant to finish a feature with small iterations (series of PRs) and then to decide how to land that properly.

jgmelber:
I'm supportive of that plan

hunhoffe:
Sounds good to me!

jgmelber (approved):

hunhoffe:
Really enjoyed reading this — nice to see a silent-hang class turned into a real compile-time error.

A few notes so far:
- The graph is built undirected, so I think the Tarjan SCC ends up equivalent to connected-components (backpressure makes every producer/consumer pair mutually dependent anyway). If that's right, a union-find might say the same thing more simply — curious whether cycles are doing something I'm missing.
- Group coupling keys off fifo names via nameBase(). Since array_fan drives the whole check, would aie.objectfifo.link / shared producer tile be a more rename-proof signal?
- under_buffered.mlir and sufficient.mlir are quite large, but the pass only walks ObjectFifoCreateOp — could they shrink to just the objectfifo/link skeleton? 

All relatively minor but I'm not a big fan on regex-on-names -- a more robust solution there would be a nice to have before merge.

jgmelber:
Is this close to ready to merge? @atassis @hunhoffe

hunhoffe:
I was hoping to get a response to my last comment, but since this isn't on the critical path, I'm ok with it merging as-is

atassis:
Hi there! I got sick last week and slowly coming back to normal. I might change some minor changes, but if you merge- I can later create a small PR per @hunhoffe notes- they are reasonable, but not a blocker. So it is your call, I can easily land fixes later, that's, maybe, might be a more propriate way to land things right now

hunhoffe:
Sounds like a good plan!

PR #3270, opened 2026-07-01
ci: sync nanobind across wheel/test locks and guard against drift

hunhoffe:
Pin nanobind to 2.12, as 2.13 doesn't seem compatible with the eudsl version we are using currently.

To guard against further version drift:
- Collapse the three pip Dependabot entries into one multi-directory entry so a shared-dependency bump lands as a single cross-directory PR and can't drift.
- Add utils/check_shared_dep_versions.py (wired into the lockfiles CI job) to assert ABI-coupled deps resolve to the same version across all locks.
- Align mlir_aie_wheels cmake upper bound with the dev manifest.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -31,6 +31,12 @@ updates:
     schedule:
       interval: "cron"
       cronjob: "0 6 8,22 * *"
+    # nanobind is held at 2.12.0: 2.13.0 rejects passing an OpResult where a
+    # Value is declared, breaking eudsl-python-extras' ArithValue path. Bumping
+    # requires a compatible eudsl/mlir-distro rebuild, so let dependabot skip it
+    # until that lands (revisit alongside the eudsl/llvm bump).
+    ignore:
+      - dependency-name: "nanobind"
     groups:
       pip-minor-patch:
         patterns: ["*"]
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,37 +14,29 @@ updates:
       github-actions-major:
         update-types: ["major"]
 
+  # All pip manifests share a build toolchain (nanobind, pybind11, cmake,
+  # setuptools, numpy, ...). They live in one entry using the plural
+  # `directories` key so a bump to a shared dependency lands as a SINGLE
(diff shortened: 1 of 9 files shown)


.github/dependabot.yml
```diff
@@ -14,37 +14,29 @@ updates:
...
 
+  # All pip manifests share a build toolchain (nanobind, pybind11, cmake,
+  # setuptools, numpy, ...). They live in one entry using the plural
+  # `directories` key so a bump to a shared dependency lands as a SINGLE
+  # cross-directory PR instead of independent per-directory PRs. Previously
+  # these were three separate entries; a per-directory bump moved one manifest
+  # and left the others behind, drifting nanobind out of ABI lockstep and
+  # breaking wheels at import time. One entry makes that divergence impossible
+  # by construction (the check_shared_dep_versions.py CI guard is the backstop
+  # for hand edits).
```
andrej, line 25: nit, you might disagree: I feel like these AI comments are always extremely verbose and explain the whole history of the code, which blows up files. I know more documentation is usually better but it uses up my human context window to read it all. Maybe just the first part of this (up to the "previously") would be enough?

python/requirements_dev.txt
```diff
@@ -14,10 +14,13 @@ patchelf; platform_system == "Linux"
...
-# same version); a mismatch breaks Value-subclass recognition at runtime. Pin
-# exact and bump only alongside a distro rebuild.
-nanobind==2.12.0
+# nanobind ABI-couples to the prebuilt mlir distro wheel (built against the
+# same version); a mismatch breaks Value-subclass recognition at runtime. Keep
+# this constraint identical to utils/mlir_wheels/requirements.in and
+# utils/mlir_aie_wheels/requirements.txt — the abi-deps CI guard
+# (utils/check_shared_dep_versions.py) fails the build if the resolved versions
+# diverge across those locks.
+nanobind>=2.13.0
```
andrej: Is `>=` safe? The previous comment mentioned that the match should be exact and raising the version would require a rebuild. My (very uninformed) intuition agrees: the C++ side will need to be recompiled if we use a newer nanobind.

andrej (approved):

jgmelber (approved):

PR #3272, opened 2026-07-02
Revisit setup scripts

andrej:
- Resilience against mismatched wheel version vs. repo checkout. New install script picks the matching wheel version for the currently checked out repository and errors if none can be found. Running example code from a newer repository with an older compiler version between breaking API changes has been a frequent source of errors in the past.
- Split `env_setup.sh` into `env_install.sh` (first time, one time install / update script) and `env_setup.sh` (environment setup only).
- Remove `quick_setup.sh` script in favor of `env_install.sh`.
- Add instruction to README for using new scripts and checking out the repository at a tagged release.


hunhoffe (commented):
I like it! This seems like a solid rework that will help users avoid known pitfalls.

hunhoffe (approved):

jgmelber (approved):
I appreciate the clean up!

PR #3273, opened 2026-07-02
Revert "Update nanobind requirement from >=2.12.0 to >=2.13.0 in /utils/mlir_wheels"

jgmelber:
Reverts Xilinx/mlir-aie#3202

Use nanobind <= 2.12 for compatibility with eudsl for now.


jgmelber:
It seems this is already accomplished by #3270

PR #3231, opened 2026-06-26
Harmonize license headers everywhere and add mechanical license header checks

andrej:

PR #3279, opened 2026-07-02
Fix REUSE pre-push hook on Windows

thomthehound:
#3231 added the `copyright-format pre-push` hook, but its pre-commit environment installs only `reuse==6.2.0`, without fallback.

On Windows, REUSE cannot use `python-magic` because the `libmagic` backend does not normally exist. Without an alternate encoding detector, `reuse lint` fails with the following before performing any checks:

```text
reuse.exceptions.NoEncodingModuleError
```

as discovered while attempting to push a commit from Windows. While it is possible to make `libmagic` available using `vcpkg` or installing from source, that seems an unnecessary external dependency and a complication that is not trivial to automate.

Instead, this patch just adds a fallback by changing the line to install `reuse[charset-normalizer]==6.2.0` everywhere the repository installs REUSE (so... the pre-commit hook, development requirements and lockfile, license workflow, and the checker’s install hint).

`charset-normalizer` provides REUSE a 'good enough' pure-Python encoding-detection fallback when `libmagic` is not present. The copyright check remains enabled and is unchanged; this just makes it work on Windows. The fact that this patch is appearing here at all is, in fact, its own verification.


andrej:
Thanks for the fix! Testing it, I realized the pre-push hook also produced very unfriendly output when not REUSE compliant, because the script exited with non-zero exit code. I took the liberty of tacking on two changes to your PR to fix that, so there's a nicer error when the copyright notices don't match the format or there are non-REUSE compliant files.

andrej (approved):

thomthehound:
> Thanks for the fix! Testing it, I realized the pre-push hook also produced very unfriendly output when not REUSE compliant, because the script exited with non-zero exit code. I took the liberty of tacking on two changes to your PR to fix that, so there's a nicer error when the copyright notices don't match the format or there are non-REUSE compliant files.

Awesome and appreciated!

PR #3287, opened 2026-07-06
ci: drop merge_group trigger from Ryzen AI wheel builds

jgmelber:
## Summary

`main` shows a red ✗ even though all tests and required checks pass. The cause is the **`Build wheels for Ryzen AI`** workflow (`.github/workflows/buildRyzenWheels.yml`):

- It triggers on `merge_group` with `concurrency: cancel-in-progress: true`.
- When a merge-queue run is superseded, its in-flight wheel jobs are **cancelled** mid-build.
- GitHub attaches those `cancelled` check-runs to the `main` HEAD sha and treats `cancelled` as non-success, so the commit renders with a ✗.

On the current `main` HEAD (`04a172cc4dd4`), 57 checks are `success` and the only non-green results are **11 `cancelled`** wheel jobs — all from this workflow, none of them test failures. None of the wheel jobs are required status checks (verified against branch protection), so dropping the `merge_group` trigger does not affect the merge queue.

## Changes

- Remove the `merge_group:` trigger from `buildRyzenWheels.yml`. Wheels still build on `pull_request`, tag pushes (`v*.*.*`), the nightly `schedule`, and `workflow_dispatch`.
- Update the two "Release latest wheels" publish steps: they previously used `merge_group` as the "represents main" signal for refreshing the rolling `latest-wheels-*` tags. That refresh now happens via the nightly `schedule` (and manual dispatch), so the `merge_group` entry is removed from their `if:` conditions.
- Refresh the now-stale concurrency and publish-step comments.

## Test plan

- [ ] Confirm the wheel workflow no longer runs on `merge_group` (no cancelled wheel check-runs land on `main` after merge).
- [ ] Confirm the nightly scheduled run still builds and publishes the `latest-wheels-*` rolling tags.
- [ ] Confirm PR-triggered wheel builds still run.

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -40,7 +40,6 @@ on:
   push:
     tags:
       - 'v*.*.*'
-  merge_group:
   schedule:
     # At 04:00. (see https://crontab.guru)
     - cron: '0 4 * * *'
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):

andrej (approved):
Looks good to fix the cancellation errors. I think the effect of this will be that latest-wheels will only update once daily, rather than on every merge to main, which I think is fine. If not, as an alternative, we could consider `concurrency: cancel-in-progress: false` for merges to main. But let's try this.

PR #3290, opened 2026-07-06
[ci] Add retry logic for 'Run Tests and Examples on Ryzen AI'

andrej:


.github/workflows/buildAndTestRyzenAI.yml
```diff
@@ -74,6 +74,15 @@ jobs:
 
       - name: Run commands
         run: |
+          # Transient NPU hardware faults (e.g. DRM_IOCTL_AMDXDNA_EXEC_CMD
```
hunhoffe: Do we need to find a way to at least log transient errors or ensure only NPU hardware faults are the issue (e.g., through grepping for DRM_IOCTL_AMDXDNA_EXEC_CMD in the output or similar)? I'm a little concerned about this hiding other flaky behavior that might be addressable.

andrej: Yes, that is a really good idea. I'll make sure the retry only happens if this particular error is observed in the output, and only on aie2-4col. All others should still fail on first try.

andrej: Done

hunhoffe (approved):

PR #3291, opened 2026-07-06
[cleanup] Remove redundant aie-python-bindings wheel build

thomthehound:
The main `mlir_aie` wheel provides the `aie` package and `aiecc`, and has done so for quite some time. XRT provides our Python bindings. Local wheel builds no longer need a separate `aie-python-bindings` wheel. 

#3183 already nuked the `AIE_ENABLE_XRT_PYTHON_BINDINGS` CMake build target. This just does some mop-up work to prevent local builds from wasting disk writes on something totally unnecessary. 

Validated by building and installing only `mlir_aie`: `import aie`, `aiecc --help`, and `pip check` all succeed and tests still run fine (as expected).


hunhoffe (commented):
Good cleanup! One thing to fold in before merge: `utils/iron_setup.py` isn't touched here but still installs the old wheel in wheelhouse mode:

```
  # utils/iron_setup.py:525-527
  print(f"[setup] Installing mlir_aie + aie_python_bindings from wheelhouse: {wheelhouse_dir}")
  for pkg in ("mlir_aie", "aie_python_bindings"):
      pip_pkg(pkg, wheelhouse=wheelhouse_dir, no_deps=True, no_index=True)
```

Dropping `aie_python_bindings` from that loop (and updating the log string to just mlir_aie) should be sufficient.
  
As bonus points -- if you are up for it, it may also be worth a quick pass on the `build_local.py` header docstring — it still describes the two-wheel staging steps.

thomthehound:
> Good cleanup! One thing to fold in before merge: `utils/iron_setup.py` isn't touched here but still installs the old wheel in wheelhouse mode:
> 
> ```
>   # utils/iron_setup.py:525-527
>   print(f"[setup] Installing mlir_aie + aie_python_bindings from wheelhouse: {wheelhouse_dir}")
>   for pkg in ("mlir_aie", "aie_python_bindings"):
>       pip_pkg(pkg, wheelhouse=wheelhouse_dir, no_deps=True, no_index=True)
> ```
> 
> Dropping `aie_python_bindings` from that loop (and updating the log string to just mlir_aie) should be sufficient.
> 
> As bonus points -- if you are up for it, it may also be worth a quick pass on the `build_local.py` header docstring — it still describes the two-wheel staging steps.

Funny you should mention `iron_setup.py`: it was actually the thing that reminded me that the `python_bindings` still existed in the first place! I'm almost done with the rewrite on that. I got held up trying to incorporate some of Andrej's ideas about keeping the allowed wheels tied to `HEAD`, but I'm not sure they're going to make the final cut because it turns out to be *frustrating* to actually use. Otherwise, that would have been first.

I'm having trouble seeing the stale mention in `build_local.py`, though. I'll take another look.

hunhoffe (approved):

PR #3293, opened 2026-07-07
[dyn-seq P5] Dynamic BD-word encoder: dma_memcpy_nd + dma_task, shim-NOC 

hunhoffe:
## Summary

Phase 2 of the dynamic-runtime-sequences milestone (#3222). Lets a `runtime_sequence` DMA carry **runtime (SSA) sizes/strides/len** on the shim-NOC path, so one xclbin can run many transfer shapes.

Covers both lowering paths:

- `aiex.npu.dma_memcpy_nd` (`--aie-dma-to-npu`)
- `aiex.dma_configure_task` / `aie.dma_bd` (`--aie-dma-tasks-to-npu`) — the path `rt.fill`/`rt.drain` use, and the one dynamic GEMM (whole_array) needs

A runtime BD lowers to the same static-template `blockwrite` **plus per-word `npu.write32` overrides** from the runtime operands. A runtime value equal to a constant reproduces the exact static BD words.

## What's in scope

- **Shared encoder.** One policy-templated size/stride algorithm (`BdLowering.h`) drives both the constant and dynamic paths, so they can't drift. Both DMA paths emit their runtime BD words through one shared helper.
- **`dma_task` runtime BDs keep their locks and `next_bd` chaining** — those are compile-time and ride in the blockwrite template; only sizes/strides/len can be runtime.
- **Runtime `repeat_count`** (runtime outer/tile-count dim): the queue-push command word is built with arith instead of rejected. Subsumes #3222 PR 6.
- **Host-side overflow guard.** A runtime value that would overflow a narrow BD field (d0/d1 wrap 10-bit, iteration 6-bit) emits `aiex.npu.assert_bd_field` → generated-C++ `if (v > max) return std::nullopt;`. Builders return `std::optional`. (The TXN stream has no on-device trap, so the check runs host-side at build time.)
- Contiguous transfers take linear mode (count in full-width `buffer_length`), dodging the 10-bit d0_size field.

**Scope boundary:** shim-NOC only, innermost stride constant 1, no padding. Anything else is a clean diagnostic, never a silent miscompile.

## Verification

The `static_vs_dynamic` harness previously never ran the dynamic encoder — its runtime arg drove only `rtp_write`, all DMA sizes stayed constant. Now:

- **Register-replay comparator** (`-DDYN_STRUCTURAL`): a runtime size isn't byte-identical to the static bake (zero-template blockwrite + write32 overrides vs one baked blockwrite) but programs the same BD registers. The comparator replays both TXN streams into an addr→value map and compares final state; golden-vs-static stays byte-exact.
- Host-run equivalence tests for a runtime **size** (both paths) and a runtime **repeat_count**; FileCheck tests pin the BD-word addresses and the reject diagnostics.

Register-replay caught two real encoder bugs the prior structure-only checks missed (wrong end of the size list for the repeat dim; raw vs biased hw repeat value). Byte-identity gate (AIEGenerateXAIE + DmaToNpu goldens) unchanged — static paths refactored onto the shared helpers with no output change.


include/aie/Dialect/AIEX/Utils/BdLowering.h
```diff
@@ -0,0 +1,193 @@
...
+  };
+
+  // d0_size, d0_stride
+  sizes[0] = p.div(p.mul(inputSizes[0], elemWidth), addressGranularity);
+  // The innermost stride is the sub-granularity / wide-element special case.
+  // The static path folds this from constants; the dynamic path requires
+  // inputStrides[0] == 1 (verified), so in both domains d0_stride resolves to a
+  // compile-time constant and the branch is decided by the policy on constants.
+  if (p.d0StrideIsZero(inputStrides[0], elemWidth, addressGranularity)) {
+    // First reason: the hardware cannot transfer less than addressGranularity
```
hunhoffe: These comments are excessive.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -312,6 +323,10 @@ struct ConvertAIEXToEmitCPass
                              /*is_standard=*/true);
     emitc::IncludeOp::create(builder, moduleOp.getLoc(), "vector",
                              /*is_standard=*/true);
+    // The builder returns std::optional: a runtime scalar that would overflow a
+    // narrow BD field yields std::nullopt instead of a truncated stream.
```
hunhoffe, line 342: shouldn't it throw an error?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -113,60 +114,16 @@ void AIEX::getHardwareStridesWraps(const AIE::AIETargetModel &targetModel,
...
-    // Note that the iteration_stride must be positive, just like the other
-    // dimensions. However, one can encode a zero-stride "repeat" of the same
-    // transfer by setting a positive repeat_count on the pushToQueue instr,
-    // and setting the size here to 1. This causes the BD to "wrap" at every
-    // single iteration, effectively never adding the specified stride, in turn
-    // equalling a repeat without stride.
-    if (inputStrides[3] > 0) {
-      strides[3] = inputStrides[3] * elemWidth / addressGranularity - 1;
-    }
+  ConstStridePolicy policy;
```
hunhoffe, line 117: shouldn't dimensions be fetched from targetmodel?

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -525,6 +482,96 @@ checkBurstLength(const xilinx::AIE::AIETargetModel &targetModel,
...
+  // SSA value while others stay constant. Only the runtime-dependent fields
+  // need runtime handling; the encoder produces the same word for a constant
+  // operand either way, so the static ≡ dynamic byte-identity holds field by
+  // field. (This matches the milestone's GEMM target, whose tiling dimensions
+  // are constant while the problem dimensions are runtime.)
+  auto sizes = getMixedSizes();
+  auto strides = getMixedStrides();
+
+  // Innermost stride must be a compile-time 1 (contiguous innermost scan). The
+  // encoder relies on this to resolve the d0 special case statically.
```
hunhoffe: why? this sounds limiting? maybe the dynamic path just shouldn't do the special casing, at least initially? And then eventually emitC can do special case detection on-the-way.

test/Conversion/DmaToNpu/bad_dma_to_npu_dynamic.mlir
```diff
@@ -0,0 +1,66 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Rejected cases for dynamic (runtime SSA size/stride) dma_memcpy_nd lowering.
+
+// RUN: aie-opt --split-input-file --aie-dma-to-npu --verify-diagnostics %s
+
+// Runtime offsets are not supported (the buffer pointer is set by the address
+// patch from a compile-time offset).
```
hunhoffe: why are runtime offsets not supported? That seems like a missing feature.

hunhoffe (commented):
For AI

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -651,6 +657,69 @@ def AIE_NpuWrite32Op: AIEX_Op<"npu.write32", []> {
...
+  let results = (outs );
+  let assemblyFormat = [{
+    `(` $value `)` attr-dict `:` type($value)
+  }];
+  let description = [{
+    Asserts, at TXN-build time, that a runtime value assigned to a buffer-
+    descriptor field fits its hardware field: `value <= max`. Emitted by the
+    dynamic DMA lowering for a runtime scalar landing in a narrow field (e.g. the
+    10-bit d0/d1 wrap or 6-bit iteration wrap), where masking would silently
+    truncate an out-of-range value.
```
andrej, line 676: Please add a description of what happens with this op in the static lowering path (no-op, rejected, ...)

include/aie/Dialect/AIEX/Utils/BdLowering.h
```diff
@@ -0,0 +1,237 @@
...
+// for realizability: d0 size and every non-unit stride must be a whole number
+// of granules (a unit innermost stride is the exempt contiguous case), and a
+// stride must be positive where its size > 1. Runtime operands are skipped
+// (guarded at lowering by assert_bd_divisible). Shared by both dynamic paths;
+// emits a diagnostic on `op` and fails on the first violation.
+inline mlir::LogicalResult
+verifyConstBdRealizability(mlir::Operation *op,
+                           llvm::ArrayRef<mlir::OpFoldResult> sizes,
+                           llvm::ArrayRef<mlir::OpFoldResult> strides,
+                           uint64_t elemWidth, uint32_t gran) {
```
andrej, line 86: Can you confirm these checks match what we've had in the compiler so far, just moved?

include/aie/Dialect/AIEX/Utils/BdLowering.h
```diff
@@ -0,0 +1,237 @@
...
+               << " bytes each is not a multiple of the " << (gran / 8)
+               << "-byte address-gen granule.";
+  for (int i = 0; i < (int)strides.size(); i++) {
+    auto s = mlir::getConstantIntValue(strides[i]);
+    if (!s)
+      continue;
+    // A unit innermost stride is the contiguous sub-granule case (collapses to
+    // hardware stride 0); every other stride must be granule-aligned.
+    if (i == 0 && *s == 1)
+      continue;
```
andrej: I think this is just moved and we already have this check, so it's probably fine to leave as is.

That being said, I'm not so sure special-casing `1` is entirely correct. This "sub-granule" is less than one hardware unit:
```
wraps=[ ..., 2]
strides=[ 4, 1]
```

Also the "collapses to hardware stride 0" is true but confusing to bring up here. All strides in hardware are encoded as `actual - 1`, and I think it's more confusing than relevant here.

include/aie/Dialect/AIEX/Utils/BdLowering.h
```diff
@@ -0,0 +1,237 @@
...
+  }
+  return mlir::success();
+}
+
+// Shared hardware size/stride encoder. The BD encodes each dimension's wrap
+// ("size") scaled to address-gen granules and step ("stride") biased by -1 (a
+// stored 0 means one granule). One algorithm drives both lowerings via a
+// Policy: ConstStridePolicy (int64 math) for the static path, SsaStridePolicy
+// (arith ops) for runtime operands -- keeping them bit-identical. The Policy
+// supplies cst/mul/div/sub and selectGT1/selectGT0/selectLt. Inputs/outputs are
```
andrej, line 136: Elegant!

andrej (approved):
Looks good. Did a light pass. I'm going to trust that our test coverage for data layout transformations would catch any egregious errors here because it is too hard to reason through.

include/aie/Dialect/AIEX/IR/AIEX.td
hunhoffe: Good call — expanded the description (both `assert_bd_field` and `assert_bd_divisible`) to spell out all three cases:
- It only ever appears on the dynamic (runtime-value) lowering path; a fully static sequence never produces it.
- When the value is a constant, the op verifier hard-errors if it's out of range — caught at compile time.
- On the binary (`--aie-npu-to-binary`) path the value is always a verified constant, so the op is emitted as a no-op (the `AIETargetNPU.cpp` `TypeSwitch` has no case for it and no `.Default`, so it's skipped, contributing no TXN words). The only place a runtime guard has effect is `--aie-npu-to-cpp`, which lowers it to an early return `std::nullopt`.

include/aie/Dialect/AIEX/Utils/BdLowering.h
hunhoffe: Mostly moved verbatim from verifyStridesWraps — the divisibility/granule check and the "d0 unit stride is exempt" carve-out are identical. But confirming this surfaced one real divergence, now fixed in 64b3ce8:

The old verifyStridesWraps allowed a zero stride on the d3 iteration dim (it checked inputStrides[3] < 0, i.e. 0 is legal — the pure-repeat case) while requiring positive strides on d0–d2. The shared helper was applying the < 1 rule to all dims, so it would wrongly reject a constant pure-repeat outer dim (size > 1, stride 0) paired with a runtime inner dim — which is exactly the whole-array GEMM A/B-tile fetch pattern (strides [0, 64, 512, 1]). Restored the d3 exemption to match, and added a @const_repeat_rt_size regression in dma_to_npu_dynamic.mlir (confirmed it fails without the fix).

TL;DR -- everything should be faithful but we still really need a unifying cleanup/refactor to reduce complexity, so I'm not 100% certain. Hopefully at some point I'll get back to https://github.com/Xilinx/mlir-aie/pull/3109

include/aie/Dialect/AIEX/Utils/BdLowering.h
hunhoffe: I believe this matches the prior verifyStridesWraps behavior.. but yeah, worth revisiting in the future.

PR #3294, opened 2026-07-07
[Windows] Update documentation and setup script

thomthehound:
## Executive Summary

This PR begins advertising native Windows as an **explicitly supported** path alongside Linux and WSL, and it includes the documentation improvements and script updates to back up that claim.

The main Windows onboarding document, `buildHostWinNative.md`, has been completely overhauled and streamlined. The installation/setup flow is now dead-simple:

```text
1) install MSVC tooling
2) install Python 3.13
3) install XRT SDK
4) clone `mlir-aie`
5) run `iron_setup.py`
6) run `iron_env.cmd` (generated by `iron_setup.py`)
7) test with a real Python JIT example
```

This can all be done without assuming a preconfigured developer environment, using instructions that should be intelligible even to a novice. The guide no longer treats building XRT from source as something any normal user should ever be expected to do, and custom wheel builds are left only as a final addendum for the curious.

To complement this change, the supported host choices are now visible in the top-level `README.md` so users can navigate directly to the correct onboarding path from the landing page:

| Host                      | Setup path             |
| ------------------------- | ---------------------- |
| Ubuntu Linux              | Existing Linux setup   |
| Other Linux distributions | Non-Ubuntu Linux guide |
| Native Windows 11         | Native Windows guide   |
| Windows via WSL           | WSL guide              |

### Make host-platform choice visible

* Prepend the Linux-only top-level onboarding entry point with a Ryzen AI host-setup section that presents Linux, non-Ubuntu Linux, native Windows, and WSL together.
* Clearly label the older WSL workflow and link native Windows users to the native path to prevent confusion.

### Vastly simplify Windows setup

* Rewrite the native Windows guide around the Windows XRT SDK and our distributed Windows wheels instead of manual source builds.
* Document the expected Visual Studio, CPython 3.13, NPU driver, XRT SDK, and Native Tools prompt workflow.
* Use a real Python JIT run as the end-to-end validation step.

### Fulfill `iron_setup.py`'s promise as a checkout bootstrap

* Replace the separate install/update/environment-command workflow with one idempotent setup command that creates or refreshes the checkout-local IRON environment.
* Detect and report a Windows `pyxrt` / CPython ABI mismatch during setup rather than leaving users to discover it at import time.
* Select `mlir_aie` wheels that match how the checkout is being used:

  * exact release assets for a tagged checkout,
  * the rolling wheel channel for development checkouts,
  * an explicit local wheelhouse for local wheel development.
* Install the repository-pinned LLVM-AIE toolchain and repair the Windows wheel layout needed by the AIE linker.
* Resolve the installed `mlir_aie`, Peano, and XRT locations once, then write activation helpers for the current host platform instead of requiring users to reconstruct environment variables in every shell.
* Generate `iron_env.cmd` and `iron_env.ps1` on Windows, or `iron_env.sh` on Linux and WSL. These are added to `.gitignore`.

### `iron_setup.py` is intended to be truly cross-platform

I realize there is strong institutional momentum around the existing setup scripts, but even from its inception, `iron_setup.py` was capable of setting up and activating `mlir-aie`/IRON on Linux machines. This area could still benefit from more testing, but my intent is that we can unify on one cross-platform script in order to prevent drift and make maintenance easier. I believe it is much more usable now than it was before, and it includes its own `--help` output for more information.

### `iron_setup.py` includes an experimental "lite" version of #3272

I really liked @andrej's idea behind that concept, so I initially attempted to incorporate it here almost verbatim. However, I found myself somewhat frustrated as a developer during my (admittedly limited) testing.

The conclusion I came to is that we should probably steer average users toward our genuine, version-labeled release builds, because those artifacts are all in one distribution channel and the matching wheel is guaranteed to exist for each one. Therefore, `iron_setup.py` enforces exact matching wheels for that case only.

Otherwise, it treats a direct clone as a developer checkout and stops getting in the way. Documentation changes were not made around that philosophy, as that would be a different sort of change, but I think that would be the easier direction to go. The relevant code section around this in `iron_setup.py` should be readily portable, and even removable upon request.


hunhoffe (approved):
Nice work! We would really like to continue to aim to streamline install for new users as much as possible, and I appreciate that this work integrates well with the `iron_setup.py` method @andrej introduced!

No updates on windows CI yet, unfortunately -- not due to lack of interest, just due to lack of time.

PR #3295, opened 2026-07-07
Allow balanced conditionals along with unconditional ObjectFIFO acquire/releases

andrej:
A "balanced" ObjectFIFO acquire/release block is a block which has a net-zero effect on the number of acquired objects in an ObjectFIFO. For example, if both the then and the else branch of a conditional acquire and ObjectFIFO `n` times and release it `m` times, the net effect is `n-m` for both branches, so this is balanced.

Such balanced blocks can be lowered with static lock allocations without issue, as anything that comes after the block is unaffected by the outcome of the conditional.

This PR allows such conditionals, whereas the previous implementation rejected any ObjectFIFO usage that had mixed conditional and non-conditional usage.

This fixes the MHA example in IRON.


hunhoffe (approved):
Nice work!

PR #3297, opened 2026-07-08
[ci] Move wheel smoke-test into wheel build workflow to avoid cancellations

andrej:
The wheel-build workflow steps have been getting insta-cancelled on most runs, typically after just one second. The suspected reason for this is that they are triggered _twice_ for each PR: once on the PR event itself, and once as a dependency of the smoke-test wheels step. Since they have the same identifier and `concurrency: cancel-in-progress: true`, the slower one then cancels the faster one.

There's really no need to build the wheels twice, and these cancellations cause ugly red Xs. Conceptually, testing the wheels belongs more with the wheel build workflow anyways, so this PR moves that step there, which should hopefully fix these issues.


jgmelber (approved):

hunhoffe (approved):
Awesome! To sanity check -- wheels still get uploaded/published to the right spots, correct? E.g., main is the only one to publish to latest-wheels-4?

andrej:
Yes, the publishing is unchanged by this. This only adds the smoke test steps without touching the publishing steps.

PR #3305, opened 2026-07-09
ci: close three CI gaps that let breakage reach main

hunhoffe:
## Summary
  
Three independent CI-hygiene fixes, all in the "a check that should have caught something didn't" family:
  
### 1. Lockfile drift check (original)
- `reviewdog/action-suggester` uses `filter_mode: diff_context`, so it only flags drift in lines already in the PR diff. A Dependabot bump that edits `requirements.txt` but not the compiled `.lock` passes silently — how PR #3302 slipped through with `setuptools==82.0.1` still in the lock.
- Adds an explicit `git diff --exit-code` step that fails on any drift in any file; keeps the reviewdog suggester as best-effort inline annotation.
    
### 2. MLIR distro wheel download has no retry
- The four build-and-test workflows fetch the 1.1 GB MLIR distro wheel with a single un-retried `pip download`. A transient GitHub release-assets hiccup makes pip fall back to the dead PyPI `mlir` 0.0.1 package (`No matching distribution found for mlir==...`) and fails the whole job.
- Wraps the download in the same 3-attempt retry loop already used in `buildRyzenWheels.yml`.
    
### 3. Silent `[: missing ']'` in buildAndTestMulti.yml
- `[ x"..." == x"OFF"]` (missing space before `]`) has errored since #868 (Dec 2023). It fell through to `else`, silently downloading the RTTI-on wheel for `ENABLE_RTTI=OFF` matrix entries — never a hard failure, so it went unnoticed for 2.5 years. Adds the space.
    
### Also: unbreak main
- Migrates the 7 objectfifo-liveness files from #3257 to the canonical copyright-then-SPDX header. Their old LLVM-style boilerplate was banned by #3296's forbidden-strings list; #3257 was approved before #3296 landed and merged without rebasing, so the header check only failed once both were on main.

### DIFF
--- a/.github/workflows/buildAndTestMulti.yml
+++ b/.github/workflows/buildAndTestMulti.yml
@@ -116,13 +116,23 @@ jobs:
         run: |
 
           VERSION=$(utils/clone-llvm.sh --get-wheel-version)
-          if [ x"${{ matrix.ENABLE_RTTI }}" == x"OFF"]; then
+          if [ x"${{ matrix.ENABLE_RTTI }}" == x"OFF" ]; then
             WHEEL=mlir_no_rtti
           else
             WHEEL=mlir
           fi
-          pip -q download $WHEEL==$VERSION \
-            -f https://github.com/Xilinx/mlir-aie/releases/expanded_assets/mlir-distro
+          for attempt in 1 2 3; do
+            rm -f mlir-*.whl
+            if pip -q download $WHEEL==$VERSION \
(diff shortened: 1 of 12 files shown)


andrej (approved):
Thanks for fixing. Who actually uses the no-RTTI wheel? If it was untested for 2.5 years, maybe we could just save ourselves the effort and remove it from the matrix.

hunhoffe:
@andrej I have no idea... maybe mlir-air at some point?

PR #3298, opened 2026-07-08
Make ObjectFIFO fully dynamic -- remove custom unrolling logic, and add new unrolling pass

andrej:
The PR removes the existing custom unrolling logic from the stateful transform pass. This means this pass now always emits _dynamic_ code, with aforementioned runtime bookkeeping. 

Doing the bookkeeping at runtime might have a small performance impact. To counter this, I add a new unrolling pass after the ObjectFIFO transform. This pass reuses existing MLIR infrastructure to unroll the loops containing the dynamic code. After unrolling, the compiler can statically reason about the generated bookkeeping code and, in the common, simple cases, fold them back into fully constant lock acquire/release values. This erases the runtime overhead for the common case. 

This sort of "roundabout" way of first emitting dynamic code, then unrolling this code and folding constants back where possible has the advantage of simplifying the passes, and supporting _all_ control flow (code that can't be reasoned about statically just remains dynamic and takes the potential performance hit of a couple extra instruction the core).

This should drastically simplify the code, as the ObjectFIFO lowering pass no longer performs any static analysis of loops and conditionals, and it allows arbitrary control flow. The static implementation had to reject cases it couldn't reason about.

Issue #3283, opened 2026-07-06
[AIE2][Bug] ObjectFifo gather load reads at wrong base


wenguang1111:
## Bug
On a single AIE2 tile (AMD Phoenix / NPU1), reading an **ObjectFifo**
buffer with a **runtime (data-dependent) index** returns the wrong element.
Instead of `in[q]` the load delivers `in[q + C]` with a **constant offset
`C = 232`** — the gather reads ~232 elements too far into the buffer.

The values are not garbage and not zeros: every wrong probe is shifted by
exactly the same `C`. Reads become correct again only in the tail, once
`q + C ≥ N`, where the over-shot address runs off the end of the buffer.

The **same buffer, copied sequentially into a local array and then indexed the
same way, is always correct** — so the data arrives on the tile fine; only the
direct gather load off the ObjectFifo buffer is miscompiled.

## Environment

| component | version |
|-----------|---------|
| mlir-aie  | `0.0.1.2026051205+614e8d8` |
| llvm-aie (Peano) | `21.0.0.2026051201+3fbeee11` (clang 21, commit `3fbeee11`) |
| XRT       | `2.23.0` |
| Device    | AMD Phoenix NPU1, single AIE2 compute tile (`NPU1Col1`) |
| Build     | `aiecc.py --no-xchesscc --no-xbridge --peano=<llvm-aie>` , kernel `-O2 --target=aie2-none-unknown-elf -std=c++20` |

## Design of the reproducer

Single AIE2 tile, one worker, three ObjectFifos (2 inputs + 1 output), all
`float32`:

```
in  : [N=1280]     float32   host fills in[i] = i
idx : [NPROBE=32]  float32   runtime probe indices q (separate buffer)
out : [NPROBE*3]   float32   {direct, staged, expected} per probe
```

`design.py` (IRON) — the worker acquires one object from each fifo, calls the
kernel, releases:

```python
of_in  = ObjectFifo(in_ty,  depth=2, name="in")
of_idx = ObjectFifo(idx_ty, depth=2, name="idx")
of_out = ObjectFifo(out_ty, depth=2, name="out")

def core_fn(of_in, of_idx, of_out, kern):
    a = of_in.acquire(1)
    b = of_idx.acquire(1)
    c = of_out.acquire(1)
    kern(a, b, c)          # a, b, c are acquire-relative pointers into the fifo buffers
    of_in.release(1); of_idx.release(1); of_out.release(1)
```

[...]

Issue #3281, opened 2026-07-04
[AIE][ObjectFifo] unrollForLoops fabricates an ID-less lock when unrolling an objectfifo.acquire in an scf.for nest (aie2p)


xynexus:
### Summary

`aie-opt --aie-objectFifo-stateful-transform` aborts with an assertion when lowering a design that **fuses two kernels** (each driving its own objectFIFO loop) onto the same compute cores. The failure surfaces during `AIEObjectFifoStatefulTransformPass::unrollForLoops`: a `LockOp` reaches `getLockIDValue()` without an assigned lock ID.

```
aie-opt: .../include/aie/Dialect/AIE/IR/AIEOps.h.inc: int xilinx::AIE::LockOp::getLockIDValue():
  Assertion `getLockID().has_value() && "Lock has no ID value"' failed.
```

### Reproduce

```console
$ aie-opt --aie-objectFifo-stateful-transform aie_objectfifo_unroll_lock_repro.mlir
Aborted (core dumped)
```

Self-contained input (2502 lines, `aie2p` / `npu2`):
https://gist.github.com/xynexus/8a09c8d22c0bcc9bf26be92af272ea66

Abbreviated backtrace (from the full `aiecc` run, before isolating the pass):

```
AIEObjectFifoStatefulTransformPass::unrollForLoops(DeviceOp&, OpBuilder&, ...)
  ::'lambda'(scf::ForOp)::operator()  ::'lambda'(scf::ForOp)::operator()   [recursive, nested loops]
AIEObjectFifoStatefulTransformPass::runOnOperation()
```

### Environment

Reproduced **identically** on two mlir_aie wheel releases:
- `0.0.1.2026033104+e4f35d6` (2026-03-31) — asserts at `AIEOps.h.inc:6998`
- `0.0.1.2026052005+886d932` (2026-05-20) — asserts at `AIEOps.h.inc:7056`

`git log` on `AIEObjectFifoStatefulTransform.cpp` shows no change to the unroll / lock-ID path in `main` through `d0988194` (2026-07-02). Peano/llvm-aie `21.0.0`; device `npu2` (aie2p, Strix Halo).

### Where it comes from

[...]

PR #3308, opened 2026-07-09
Enable dynamic lock acquisition for ObjectFIFO acquire/release in cores

andrej:
The --dynamic-objFifos option -- now the default -- so far made the buffer selection dynamic in the ObjectFIFO. This means the lowering injects some code into the IR which resolves which buffer to read from/write to at runtime instead of reasoning about this at compile-time and hard-coding the buffers.

This PR does the same for locks, allowing acquiring and releasing objects within arbitrary control flow paths, while a small runtime component keeps track of the number of objects that the core holds. This means we keep a runtime-computed value for the lock acquire/release counts in a register or in memory on the core.

To enable this, we have to refactor the IR to use SSA values for lock counts rather than static attributes -- this is the reason this PR touches many tests. Existing passes that rely on constant values for lock acquire/release emit arith.constant (or trace the SSA value back to a defining arith.constant and error if the value is not constant).


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1378,34 +1378,49 @@ def AIE_UseLockOp: AIE_Op<"use_lock", []> {
...
         LockAction:$action,
-        OptionalAttr<AIEI32Attr>:$value,
+        I32:$value,
         OptionalAttr<LockBlocking>:$blocking,
         DefaultValuedOptionalAttr<BoolAttr, "true">:$acq_en
   );
 
   let assemblyFormat = [{
-    `(` $lock `,` $action (`,` $value^)? (`,` $blocking^)? `)` attr-dict
+    `(` $lock `,` custom<UseLockValue>($action, $value, $blocking) `)` attr-dict
```
andrej: Why do we need a custom parser? Can we get away without it?

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -73,7 +75,15 @@ struct AIEDialectFoldInterface : DialectFoldInterface {
...
     //    referencing a device-level value that is erased when the core is
     //    outlined.
-    return isa<CoreOp, RuntimeSequenceOp>(region->getParentOp());
+    //  - aie.mem/aie.memtile_dma/aie.shim_dma bodies carry constant lock values
+    //    (aie.use_lock operands). If these hoisted to the device, CSE would
+    //    merge them with the identical constant kept local to a core, and since
+    //    the device-level constant dominates the core, the core's use would be
+    //    rewritten to reference it -- reintroducing the cross-region reference
+    //    that breaks core outlining. Keeping them local avoids the dominating
+    //    device-level duplicate entirely.
```
```suggestion
    //  - Make sure SSA values for aie.use_lock operands in 
    //    aie.mem/aie.memtile_dma/aie.shim_dma bodies do not get
    //    hoisted.
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2679,6 +2689,17 @@ LogicalResult LockOp::verify() {
   return success();
 }
 
+// Centralized lookup for the compile-time constant lock value: returns the
+// integer if `op`'s value operand is defined by an arith.constant, otherwise
+// std::nullopt. This is the single place that knows how a static lock value is
+// recovered from the SSA operand.
```
```suggestion
// Look up for compile-time constant lock values, if any.
// Returns std::nullopt if lock value does not reference an `arith.constant`.
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2737,6 +2763,15 @@ LogicalResult UseLockOp::verify() {
     return (*this)->emitOpError(
         "AcquireGreaterEqual is not supported in AIE1.");
 
+  // Locks used inside a DMA/BD block are configured via MMIO and therefore
+  // require a compile-time constant value (an arith.constant operand).
```
```suggestion
  // Locks used inside a DMA/BD block are configured via static register writes
  // and therefore require a compile-time constant value.
```

stephenneuendorffer:
I'm glad to see this Andrej, it's a big improvement on code size!

hunhoffe:
Out of curiosity, does this address either of the two recently opened issues?

andrej:
> Out of curiosity, does this address either of the two recently opened issues?

@hunhoffe I think #3298, which builds on top of this, will address #3281. Not sure about #3283.

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -2243,6 +1672,38 @@ struct AIEObjectFifoStatefulTransformPass
     }
   }
 
+  /// Emit a UseLockOp whose lock value is a runtime-computed SSA i32. Used by
+  /// the dynamic objectFifo lowering, where the number of locks to
+  /// acquire/release is not known at compile time. Only valid for AIE2
+  /// semaphore locks (the dynamic lowering targets AIE2/AIE2p).
```
andrej:
```suggestion
  /// Emit a UseLockOp whose lock value is a runtime-computed SSA i32.
```

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -2861,6 +2327,62 @@ struct AIEObjectFifoStatefulTransformPass
...
 
+      //===----------------------------------------------------------------===//
+      // Dynamic lock bookkeeping setup
+      //===----------------------------------------------------------------===//
+      // For tiles using the dynamic objectFifo lowering, the number of locks
+      // to acquire is computed at runtime from a per-(fifo,port) "held" counter
+      // stored in a local buffer. This replaces the static (cyclostatic)
+      // accounting that cannot see through runtime-dependent control flow.
+      // Runtime lock counts require semaphore locks (AIE2+); on architectures
+      // with binary locks (AIE1) the dynamic lowering keeps static lock counts.
```
andrej:
```suggestion
      // stored in a local buffer. 
      // Runtime lock counts require semaphore locks (AIE2+); on architectures
      // with binary locks (AIE1) the dynamic lowering keeps static lock counts.
```

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -2861,6 +2327,62 @@ struct AIEObjectFifoStatefulTransformPass
...
+        if (!slotOrder.empty()) {
+          builder.setInsertionPoint(coreOp);
+          auto heldTy =
+              MemRefType::get(SmallVector<int64_t>{(int64_t)slotOrder.size()},
+                              builder.getI32Type());
+          heldBuffer = BufferOp::create(
+              builder, coreOp.getLoc(), heldTy, coreOp.getTile(),
+              /*sym_name*/ nullptr, /*address*/ nullptr,
+              /*initial_value*/ nullptr, /*mem_bank*/ nullptr,
+              /*aligned*/ nullptr);
```
andrej: Wow ... dumb?! Why not use a register / SSA value threaded through the control flow that keeps track of the currently held number of objects? Memory makes things completely opaque and then the compiler can't reason about it any more. Threading SSA values through would be strongly preferrable, though if that's hard, memref.alloca + mem2reg could also work.

andrej (commented):
Can we add a custom parser that allows immediate values in the IR instead of forcing arith.constant everywhere to reduce the size of the diff / having to rewrite all tehse tests?

programming_examples/ml/bottleneck/bottleneck.py
```diff
@@ -130,6 +130,7 @@ def conv1x1_fn(of_wts, of_act_in, of_act_out, conv1x1):
         Worker(
             conv1x1_fn,
             fn_args=[wts_buf_00.cons(), of_act_l3l2.cons(), of_act_2_3_5.prod(), conv1],
+            stack_size=0x1000,
```
hunhoffe, line 133: Interesting! Are you sure the changes to this design are necessary? I'm surprised it was passing before this!

hunhoffe (approved):
Looks good! This is a good time to do this type of change, since we're already doing a lot of type conversions to SSA

programming_examples/ml/bottleneck/bottleneck.py
andrej: Yes, they definitely are. This was hard to debug, along with another missized buffer. Keep in mind we are now keeping lock count bookkeeping data in memory. If that gets overridden by a buffer override, it causes the whole design to misbehave/timeout. Previously, it might just have overridden other data which might have passed some of our fuzzy output tests.

hunhoffe: Nice catch then!

PR #3312, opened 2026-07-10
[AIE] objectfifo-liveness: union-find cycles + structural coupling gate + slim tests

atassis:
Follow-up to #3257 addressing @hunhoffe's review notes. Three changes: the cycle
detection is simpler, the name-base coupling is now structurally gated so it can
no longer false-positive, and the two big lit tests are cut down to the skeleton
the pass actually walks. No behavior change on the merged cases; the gate only
removes a false-positive class.

## Union-find instead of Tarjan

The dependency graph is undirected: a shared tile couples its producer and
consumer fifos with data and back-pressure edges in both directions, so "in a
cycle" degenerates to "in a connected component of size > 1". Tarjan was doing
connected components the hard way. Replaced it with a union-find -- same result,
less code, no recursion. The header comment now spells out why the two are
equivalent here so the next reader does not re-derive it.

## Structurally-gated coupling

I looked at keying the coupling off `aie.objectfifo.link` or a shared producer
tile, and neither identifies it: the two halves of one logical broadcast
(`memW1_0`, `memW1_1`) sit on different, unlinked MemTiles with different
producer tiles, so there is no link edge and no shared tile between them. Only
the shared name base ties them together, so nameBase stays -- but it is now
gated. Before summing a group's fan-out I require the halves to structurally
agree: identical element type, depth and repeat_count, and disjoint consumer
tiles. A pair that merely collides on the base name is analyzed on its own
instead of folded, which is exactly the false-positive you flagged.

`namebase_gate.mlir` pins this: two unrelated broadcasts share the base name
`memX` but differ in element type (32x32xbf16 vs 16x16xbf16). Gated, each is
checked alone (fan 2, T 2 -> demand 4 == slack 4, silent); ungated they fold to
fan 4 -> demand 8 > 4 and the pass emits a false error. The file must stay
silent.

The durable fix is a first-class coupling attribute on the frontend ops so the
analysis does not lean on a name at all. That is a tracked follow-up, flagged in
a comment; it is out of scope here.

## Slimmer tests

`under_buffered.mlir` 916 -> 45 lines, `sufficient.mlir` 467 -> 35. The pass only
walks `ObjectFifoCreateOp`, so the compute cores, kernels and DMA program are
elided and only the objectFIFO declarations remain. Assertions are unchanged --
`under_buffered` still expects the same `depth >= 8` error, `sufficient` still
verifies silence -- and the delta between the two files is now exactly the
coupled-group summation, which is much easier to read.

## Tests

All of `test/Passes/objectfifo-liveness` passes. `namebase_gate.mlir` is a new
regression guard for the gate: it errors on the pre-gate pass and is silent with
it.


jgmelber (approved):

hunhoffe (approved):
Thank you for the follow up!

atassis:
@hunhoffe hi there!
Sorry, I had to push another commit to fix the CI lint error. It should be fine now

PR #3317, opened 2026-07-10
AIEObjectFifoStatefulTransform: diagnose unplaced tiles instead of crashing

atassis:
## Summary

`AIEObjectFifoStatefulTransform` assumes every objectFifo producer/consumer
tile is a placed `aie.tile` and reaches them via `getProducerTileOp()` /
`cast<TileOp>(...)`. If a design reaches this pass before placement -- an
`aie.objectfifo` whose tile is still an unplaced `aie.logical_tile` -- the
unconditional `cast<TileOp>` dereferences a null and the pass **SIGSEGVs**
instead of reporting the problem.

Minimal repro (crashes on current main):
```mlir
aie.device(npu1) {
  %prod = aie.logical_tile<ShimNOCTile>(0, ?)
  %cons = aie.logical_tile<CoreTile>(?, ?)
  aie.objectfifo @of (%prod, {%cons}, 2 : i32) : !aie.objectfifo<memref<64xi16>>
}
```
`aie-opt --aie-objectFifo-stateful-transform` -> SIGSEGV in
`AIEObjectFifoStatefulTransformPass::runOnOperation`.

## Fix

Add an early `verifyObjectFifoTilesArePlaced()` check (sibling of the existing
`verifyObjectFifoLinks()`) that emits an op error and fails the pass cleanly
when a producer/consumer tile is not a placed `aie.tile`:

```
error: 'aie.objectfifo' op producer tile is not a placed aie.tile;
       run --aie-place-tiles before this pass
```

## Tests

Adds two `--verify-diagnostics` cases (unplaced producer, unplaced consumer) to
`test/objectFifo-stateful-transform/check_errors.mlir`. All existing
objectFifo-stateful-transform lit tests pass unchanged.

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
@@ -2385,6 +2385,34 @@ struct AIEObjectFifoStatefulTransformPass
     return hasError ? failure() : success();
   }
 
+  /// This pass assumes every objectFifo producer/consumer tile is a placed
+  /// `aie.tile`; the rest of the pass reaches them via getProducerTileOp() /
+  /// cast<TileOp>(...), which dereferences a null if a tile is still an
+  /// unplaced `aie.logical_tile` (or any non-TileOp). Emit a diagnostic and
+  /// fail cleanly instead of crashing when a design reaches this pass before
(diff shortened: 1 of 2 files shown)


jgmelber (approved):

atassis:
@jgmelber I have rebased this and next my PRs onto  latest commit, looks like it is a test error from the commit in upstream

PR #3319, opened 2026-07-10
cmake: allow building the compiler standalone (AIE_INCLUDE_TESTS_AND_EXAMPLES)

atassis:
The `programming_examples`, `programming_guide`, and `test` subdirectories' `check-*` lit targets depend on `AIEPythonModules`, so configuring them fails when `-DAIE_ENABLE_BINDINGS_PYTHON=OFF`. That makes it awkward to build just the compiler (`aiecc` and friends) without the Python bindings.

Gate those three subdirectories behind a new `AIE_INCLUDE_TESTS_AND_EXAMPLES` option (default `ON`, so the normal build is unchanged), mirroring the existing `AIE_INCLUDE_DOCS` option a few lines up. With `-DAIE_INCLUDE_TESTS_AND_EXAMPLES=OFF` the compiler can be configured and built standalone (e.g. slimmer packaging or a compiler-only CI job).

If you'd prefer, I'm happy to instead express this as a `cmake_dependent_option` tied to `AIE_ENABLE_BINDINGS_PYTHON` (force-off when bindings are off) rather than an independent flag - whichever fits the project's conventions better.


CMakeLists.txt
```diff
@@ -344,9 +344,16 @@ endif()
 
 
 # Last because each of these has its own CMakeLists.txt which reloads/re-finds LLVM, thus resettings globals.
-add_subdirectory(programming_examples)
-add_subdirectory(programming_guide)
-add_subdirectory(test)
+# The test/example/tutorial subdirectories' check-* lit targets depend on AIEPythonModules, so they
```
andrej, line 347: So then would it be best to error at configure time when AIE_ENABLE_PYTHON_BINDINGS=OFF but AIE_INCLUDE_TESTS_AND_EXAMPLES=ON? Or make toggling the Python bindings off automatically also make the examples toggle off?

atassis: Yeah, you're right - I configured this around my own workflow (I always pass both flags for a standalone build), so I didn't think about the common case where someone just flips `-DAIE_ENABLE_BINDINGS_PYTHON=OFF` and still trips the original error. Auto-forcing it off is the better default.

Switched to a `cmake_dependent_option` tied to `AIE_ENABLE_BINDINGS_PYTHON` (forced OFF when bindings are off), matching how `AIE_ENABLE_BINDINGS_PYTHON` itself is defined just above. Now `-DAIE_ENABLE_BINDINGS_PYTHON=OFF` alone is enough, you can still turn tests off with bindings on, and the broken combo just can't be selected - without a hard configure error. Happy to make it error explicitly instead if you'd prefer that.

andrej (approved):
thanks!

PR #3320, opened 2026-07-12
many_buffers: re-sync and re-read before trusting a mismatch

atassis:
## What

test/npu-xrt/many_buffers intermittently fails in CI on the aie2p-8col
runners, with output like:

    pair 2 idx 1: got 3 expected 3
    failed.

The reported values are equal. The mismatch branch prints "got N expected N",
which means out[i] read back stale (not equal to expected) at compare time,
then settled to the correct value by the time it was re-read for the
diagnostic print.

## Why it happens

The test uses host-only output BOs (XRT_BO_FLAGS_HOST_ONLY). After run.wait()
returns COMPLETED it calls bo.sync(XCL_BO_SYNC_BO_FROM_DEVICE) before reading.
On this stack a just-completed run can briefly leave the last DMA line not yet
visible to the host, so sync() can return before that line has settled.
Because the kernel is a pure DMA passthrough (out[i] == in[i]), a stale read
returns the pre-DMA pattern, which matches the expected value once it settles.
That is why the failure prints matching values. The data is never actually
wrong, only read a moment too early. The root cause looks to sit below the
test, in the sync/driver visibility path; this change is a test-side
workaround for it.

## The fix

Move the readback into a verify_pair() helper. sync(FROM_DEVICE) still does the
cache invalidate. On a mismatch, re-sync and re-read the pair, up to
VERIFY_ATTEMPTS times. A genuine miscompute is deterministic and survives every
attempt, so this only absorbs the transient window and cannot hide a real
regression.

## Validation

Reproduced and fixed on an XDNA2 NPU (aie2p, npu2_4col), 500 runs each:

| build     | failures / 500                            |
| --------- | ----------------------------------------- |
| unpatched | 75 (15%), all "got N expected N" at idx 1 |
| patched   | 0                                         |

This matches the signature seen intermittently on the aie2p-8col CI runners,
including on main and on unrelated PRs.

## Scope

Other npu-xrt tests use the same run.wait(), sync(FROM_DEVICE), read sequence,
but many_buffers is the one that actually hits it, because 8 output buffers
plus passthrough maximize the exposure. This change is limited to that one
test. A shared helper could generalize it later if other tests start flaking.

### DIFF
--- a/test/npu-xrt/many_buffers/test.cpp
+++ b/test/npu-xrt/many_buffers/test.cpp
@@ -23,6 +23,36 @@
 
 constexpr int BUF_SIZE = 64;
 constexpr int NUM_PAIRS = 8;
+constexpr int VERIFY_ATTEMPTS = 5;
+
+// Read back and verify one input/output pair. bo.sync(FROM_DEVICE) does the
+// cache invalidate, but a just-completed run can still leave the last DMA line
+// briefly unread on a host-only BO, so re-sync and re-read up to
+// VERIFY_ATTEMPTS times. A genuine miscompute is deterministic and survives
+// every attempt, so this absorbs only the transient coherency window; it never
+// hides a regression. Returns the count of still-mismatching elements after the
+// final attempt.
+static int verify_pair(xrt::bo &bo, int p) {
+  int errors = 0;
+  for (int attempt = 0; attempt < VERIFY_ATTEMPTS; attempt++) {
+    bool last = attempt == VERIFY_ATTEMPTS - 1;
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):
Thank you for this workaround to the flaky nature of this test -- maybe one day we'll find a better root cause of it.

atassis:
No trouble!
I also poked at a couple of the other flaky tests (vector_scalar_mul among them) and had some broader ideas on the retry and harness side, but nothing clean enough to put up as a PR yet. The cases I tried did not reproduce on my device, serial or under load, so I do not have a solid fix to offer there.

PR #3332, opened 2026-07-13
Add drop-in replacement of `load_pdi` with control-packet-based configuration

andrej:
Adds support for a mode of `--expand-load-pdis` that replaces all `load_pdi` instructions with

1. A `load_pdi` instruction that loads a minimal column control overlay -- the routes for the control packets.
2. A set of instructions configuring a DMA to stream in the control packets.

Plumbs in the control packet data into the full ELF format.

Reworks the control packet overlay generation for the multi-device scenario; each device gets an identical control packet overlay, which covers the area of the union of all devices. The idea behind this is that one control packet overlay is capable of configuring all devices in the module, so if we move to smarter reconfiguration (e.g., diff based), that overlay can be reused.

[This new test](https://github.com/andrej/mlir-aie/blob/c3d5b1bf7868793078574c49f9084fe0cfc8bd55/test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit) shows how the functionality is intended to be used.

This is how the control packet data and the DMA sequence for streaming them in makes it into the full ELF:

image


lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+        continue;
+      sourceDevices.push_back(dev);
+    }
+
+    // Baseline-preserving fast path: unless a standalone `@ctrl_pkt_overlay`
+    // device was explicitly requested (the load-pdi-to-ctrl-pkt / reconfigure
+    // flow), apply the overlay in place to every device exactly as the legacy
+    // per-device pass did -- no extra attributes and no standalone device.
+    // This keeps both the default `route-shim-to-tct` flow and direct
+    // `route-shim-to-tile-ctrl=true` invocations byte-identical.
```
andrej: References to old code ("baseline") in comments are undesirable. The reader will not have the context of how this code evolved. Instead, document what actually happens here: if the pass user requests the pass be emitted _into_ the device, emit it into the device, otherwise emit it as a standalone device.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+    if (!clEmitStandaloneOverlay) {
+      for (auto dev : sourceDevices)
+        if (failed(applyOverlayToDevice(dev)))
+          return signalPassFailure();
+      return;
+    }
```
andrej: I don't like that this is special-cased. There should be two cases: (a) emit overlay into all devices but no standalone device, (b) emit ovleray into all devices AND a standalone device. So the "emit overlay into all devices" path is always active (unless opted out, which this path ignores currently), no need to duplicate it here.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+        dev->setAttr("has_ctrl_pkt_overlay", boolFalse);
+        continue;
+      }
+      participating.push_back(dev);
+    }
+
+    // Step 1: apply the overlay in-place to each participating device, unless
+    // the user asked us to suppress the in-place application via
+    // `standalone-device-only`. Either way, tag each participating source
+    // device with a `has_ctrl_pkt_overlay` boolean reflecting what happened.
```
andrej: Comment needlessly restates code
```suggestion
    // Apply the overlay in-place to participating devices.
```

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
```diff
@@ -119,6 +119,15 @@ class DMAChannelAnalysis {
...
                            flowOp.getDestChannel()}] = 1;
     }
+    // Scan ShimDMAAllocationOps so that channels already claimed (e.g. by
+    // the control packet overlay) are not reused by objectFIFO lowering.
+    for (auto allocOp : device.getOps<ShimDMAAllocationOp>()) {
+      auto tile = allocOp.getTileOp();
+      if (!tile)
+        continue;
+      channelsPerTile[{tile.getResult(), allocOp.getChannelDir(),
+                       (int)allocOp.getChannelIndex()}] = 1;
```
andrej, line 131: What is `channelsPerTile` used for downstream? I assume this is to mark this channel used, even though it is not used for an ObjectFIFO.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+    llvm::SmallSet<AIE::TileID, 8> seenTiles;
+    builder.setInsertionPointToStart(body);
+    for (auto dev : participating) {
+      for (auto tOp : dev.getOps<AIE::TileOp>()) {
+        AIE::TileID id{tOp.colIndex(), tOp.rowIndex()};
+        if (!seenTiles.insert(id).second)
+          continue;
+        AIE::TileOp::create(builder, tOp.getLoc(), id.col, id.row);
+      }
+    }
```
andrej: This should be factored out into its own function (createOverlayDevice or similar). The logic for cloning tile ops should also be factored out, as it is shared here for the standalone device as well as above for the overlay-in-device case.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+      if (failed(applyOverlayToDevice(dev)))
+        return signalPassFailure();
+      dev->setAttr("has_ctrl_pkt_overlay", boolTrue);
+    }
+
+    // Step 2: always emit a separate `@ctrl_pkt_overlay` device that contains
+    // only the overlay (and the union of tiles it references). This gives
+    // downstream consumers a single, isolated device to compile when they
+    // need to ship a reconfigure-only PDI, regardless of whether the overlay
+    // was also applied in place.
```
andrej: Comment too verbose. 
```suggestion
    // Emit standalone `@ctrl_pkt_overlay` device.
```

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+      for (auto tOp : dev.getOps<AIE::TileOp>())
+        existing.insert({tOp.colIndex(), tOp.rowIndex()});
+      OpBuilder b = OpBuilder::atBlockBegin(dev.getBody());
+      for (auto id : unionTiles) {
+        if (existing.contains(id))
+          continue;
+        auto proto = prototypeTile[id];
+        auto cloned = cast<AIE::TileOp>(b.clone(*proto.getOperation()));
+        (void)cloned;
+      }
```
andrej: factor this out

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,155 @@ struct AIEGenerateColumnControlOverlayPass
...
+          return signalPassFailure();
+      return;
+    }
+
+    // Ctrl-pkt flow: apply the overlay in place to each participating device
+    // and emit a separate standalone `@ctrl_pkt_overlay` device.
+
+    // Devices that participate in the standalone overlay (i.e. have not opted
+    // out via `needs_ctrl_pkt_overlay = false`). Opted-out devices are tagged
+    // immediately with `has_ctrl_pkt_overlay = false`.
```
andrej: Stop it with the needless comments that restate code.

```suggestion
```

tools/aiecc/aiecc.cpp
```diff
@@ -580,26 +582,63 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  bool npuTransactionsNeedCoresLowered = expandLoadPdis.getValue() ||
+                                         generateTxn.getValue() ||
+                                         loadPdiToCtrlPkt.getValue();
   EdgeWithTypedOutput<ModRef> &npuLoweringInput =
       npuTransactionsNeedCoresLowered
           ? static_cast<EdgeWithTypedOutput<ModRef> &>(physicalWithElfs)
           : static_cast<EdgeWithTypedOutput<ModRef> &>(physical);
-  auto &npuLowered = npuLoweringInput.map<ModRef>(
+  // NPU lowering, decomposed so each pass pipeline is its own edge (one
+  // pipeline per edge) and every intermediate module is a first-class artifact.
```
```suggestion
  // NPU instruction sequence lowering
```

tools/aiecc/aiecc.cpp
```diff
@@ -580,26 +582,63 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  auto &ctrlPktExpanded = ctrlPktMaterialized.map<ModRef>(
+      "ctrlpkt_expanded.mlir",
+      PassPipeline{&context, [](mlir::MLIRContext *ctx, mlir::ModuleOp) {
+                     return getExpandLoadPdiPipeline(ctx, /*ctrlPkt=*/true);
+                   }});
+  auto &ctrlPktDmaLowered = ctrlPktExpanded.map<ModRef>(
+      "ctrlpkt_to_dma.mlir", PassPipeline{getCtrlPktToDmaPipeline(&context)});
+  auto &ctrlPktNpuLowered = ctrlPktDmaLowered.map<ModRef>(
+      "ctrlpkt_npu_lowered.mlir",
+      PassPipeline{getPerDeviceDmaLoweringPipeline(&context)});
```
andrej: Do we need these defined as intermediates, or can we chain the second `.map` directly on the first? Simplify if possible. If you define it as a separate variable, it indicates to me that the output is used elsewhere downstream (fan-out). If this is just linear, just do `.map().map()`.

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1064,104 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+                return mlir::success();
+              });
+
+  // Combined ELF: all PDIs + NPU insts bundled into one aie2_config ELF. The
+  // full-ELF NPU insts (unfolded DDR offset) plus, when the
+  // load-pdi-to-ctrl-pkt reconfigure flow is active, the per-device
+  // control-packet binary and patch-info edges are bundled in so their file
+  // dependency is captured in the graph; the join reads their materialized
+  // paths and hands them to the config builder, which attaches them to the
+  // owning device's runtime-sequence instances.
```
```suggestion
  // Combined ELF: all PDIs + NPU insts bundled 
  // + control packet data, if any
```

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1064,104 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+                item.get().module.get(), words, d.getSymName(), "");
+          }));
+  auto &fullElfCtrlpkt = ctrlPktBinaries.filter(
+      "fullElfCtrlpktNonEmpty",
+      [](const std::vector<char> &bin) { return !bin.empty(); });
+
+  // Patch-info for each control-packet-carrying device: the runtime-argument
+  // slot the control-packet buffer occupies (the sequence's pre-lowering
+  // argument count) and the binary's byte size. Keyed by the non-empty ctrl-pkt
+  // devices; the per-device module supplies the argument slot.
```
andrej: Make these comments actually understandable if you insist on writing so much.
```suggestion
  // When control packets are enabled, the control data gets passed into the runtime sequence as an argument. `fullElfPatchInfo` captures which argument index contains this control data, and the size of the control data buffer.
```

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1064,104 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+      emitBinary<OpInModule<DeviceOp>>(
+          [](const Item<OpInModule<DeviceOp>> &item,
+             std::vector<uint32_t> &words) -> mlir::LogicalResult {
+            DeviceOp d = item.get().op;
+            return xilinx::AIE::AIETranslateControlPacketsToUI32Vec(
+                item.get().module.get(), words, d.getSymName(), "");
+          }));
+  auto &fullElfCtrlpkt = ctrlPktBinaries.filter(
+      "fullElfCtrlpktNonEmpty",
+      [](const std::vector<char> &bin) { return !bin.empty(); });
```
andrej: Can probably chain `.filter` right after `.map`, unless ctrlPktBinaries is really consumed elsewhere?

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1064,104 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+                noCtrlPktDevices);
+
+  // The overlay/reset devices synthesized by expand-load-pdi carry no runtime
+  // sequence and thus no control packets; skip them so the translation only
+  // runs on real devices.
+  auto &ctrlPktDevices =
+      ctrlPktSource.filter("ctrlPktDevices", [](const OpInModule<DeviceOp> &x) {
+        auto seqs = DeviceOp(x.op).getOps<RuntimeSequenceOp>();
+        return seqs.begin() != seqs.end();
+      });
```
andrej: Again -- let's avoid defining another variable `ctrlPktDevices`, making it harder to trace through the graph. Call .filter directly on the ctrlPktSource definition. Drop the comment.

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1064,104 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  auto &ctrlPktExpandedPerDevice = splitPerDevice(
+      ctrlPktExpanded, "ctrlpkt_expanded_{0}.mlir", "ctrlPktExpandedMatching");
+  auto &noCtrlPktDevices = staticPerDevice.filter(
+      "noCtrlPktDevices", [](const OpInModule<DeviceOp> &) { return false; });
+  EdgeWithTypedOutput<OpInModule<DeviceOp>> &ctrlPktSource =
+      loadPdiToCtrlPkt.getValue()
+          ? static_cast<EdgeWithTypedOutput<OpInModule<DeviceOp>> &>(
+                ctrlPktExpandedPerDevice)
+          : static_cast<EdgeWithTypedOutput<OpInModule<DeviceOp>> &>(
+                noCtrlPktDevices);
```
andrej: That's kind of ugly. I think we need this because some downstream edge of this feeds into the full ELF bundle downstream -- but could we make this just an explicit "empty" edge in the non-load-pdi-to-ctrl-pkt case? If you need to add a supporting function to Graph.h for this, that's fine. It'll express the intent more clearly than an always-false filter.

tools/aiecc/aiecc.cpp
```diff
@@ -580,26 +582,63 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  // control-packet ops (before DMA lowering), so it is the extraction point for
+  // the control-packet binary.
+  auto &ctrlPktMaterialized = npuLoweringInput.map<ModRef>(
+      "ctrlpkt_materialized.mlir",
+      PassPipeline{getMaterializeRuntimeSeqPipeline(&context)});
+  auto &ctrlPktExpanded = ctrlPktMaterialized.map<ModRef>(
+      "ctrlpkt_expanded.mlir",
+      PassPipeline{&context, [](mlir::MLIRContext *ctx, mlir::ModuleOp) {
+                     return getExpandLoadPdiPipeline(ctx, /*ctrlPkt=*/true);
+                   }});
```
andrej: Assume --expand-load-pdis and --load-pdis-to-ctrl-pkts can never be passed at the same time. Error in the command line parsing section / header if both are passed.

Then, based on that assumption, I think we can share this code with the couple lines above. No duplicate `getExpandLoadPdiPipeline` edges that only differ in the ctrlPkt=true flag. Just one `npuExpanded`.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
andrej: (Claude) Removed the baseline/legacy references. The top of `runOnOperation` now just describes the two cases: apply the overlay in-place to participating devices, and — when `emit-standalone-overlay` is set — additionally emit the standalone `@ctrl_pkt_overlay` device.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
andrej: (Claude) Restructured. The "apply overlay to all participating devices" loop is now always active (and respects the `needs_ctrl_pkt_overlay = false` opt-out). `emit-standalone-overlay` only gates the extra standalone-device emission plus the `has_ctrl_pkt_overlay` tagging — no duplicated apply loop.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
andrej: (Claude) Factored out `createOverlayDevice(...)` for the standalone device, plus shared `collectTileUnion(...)` / `cloneMissingTiles(...)` helpers used by both the in-device path and the standalone device. The standalone device now populates its tiles through the same `cloneMissingTiles` helper (so it copies tile attributes too, matching the in-device case).

lib/Dialect/AIE/Transforms/AIEObjectFifoStatefulTransform.cpp
andrej: (Claude) Correct. `channelsPerTile` is the used-channel map that `getDMAChannelIndex` consults when auto-assigning a free DMA channel (and `reservePinnedChannel` for pinned ones). Marking the overlay's `ShimDMAAllocationOp` channels as used keeps objectFIFO lowering from handing out a channel the control-packet overlay already claimed. I expanded the comment to say this explicitly.

tools/aiecc/aiecc.cpp
andrej: (Claude) Chained the purely-linear intermediates: `ctrlpkt_materialized` now feeds straight into `ctrlPktExpanded` (`.map().map()`), and `ctrlpkt_to_dma` into `ctrlPktNpuLowered`. I kept `ctrlPktExpanded` and `npuDefaultLowered` as named edges because they genuinely fan out — `ctrlPktExpanded` is also the extraction point for the control-packet binary (`splitPerDevice` below), and `npuDefaultLowered` is both the no-expansion result and the expand-load-pdi input.

tools/aiecc/aiecc.cpp
andrej: (Claude) Done — `ctrlPktBinaries` was only consumed here, so it's now `.map(...).filter(...)` chained directly into `fullElfCtrlpkt`.

tools/aiecc/aiecc.cpp
andrej: (Claude) Dropped the `ctrlPktSource` intermediate — the source select is now inlined directly into the `.filter(...)` that produces `ctrlPktDevices`. I kept `ctrlPktDevices` as a named edge because it fans out: it feeds both the control-packet binary (`fullElfCtrlpkt`) and the patch-info bundle (`fullElfPatchInfo` reads `ctrlPktDevices.out` for the runtime-arg slot).

tools/aiecc/aiecc.cpp
andrej: (Claude) Added a `Graph::empty<T>(name)` helper (an `EmptyEdge` source that yields zero items) and used it for `noCtrlPktDevices` in place of the always-false filter, so the non-`--load-pdi-to-ctrl-pkt` case is now an explicit empty edge.

tools/aiecc/aiecc.cpp
andrej: (Claude) Added the mutual-exclusivity check: `--expand-load-pdis` + `--load-pdi-to-ctrl-pkt` now error out in the command-line validation, and I dropped the stale "implies --expand-load-pdis" from the `--load-pdi-to-ctrl-pkt` help text.

On collapsing to a single `npuExpanded`: the two expand edges sit at different points in the pipeline, so they can't fully merge. The default flow expands *after* full NPU lowering (`getNpuLoweringPipeline` -> expand), whereas the ctrl-pkt flow must expand *after materialization but before DMA lowering* — the control packets are extracted from `ctrlPktExpanded` (pre-DMA-lowering) and only then lowered (`ctrlpkt-to-dma` -> per-device DMA->NPU). I did de-duplicate everything that was safely linear (see the chaining above), so there's now a single expand edge per flow. I can reorder the default flow to unify further if you'd prefer, but that changes the tested `--expand-load-pdis` path, so I left it as-is.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -424,6 +424,8 @@ def AIEGenerateColumnControlOverlay : Pass<"aie-generate-column-control-overlay"
             "Flag to generate TCT routing between tile CTRL and shim SOUTH ports. Available options: ['shim-only', 'all-tiles', 'disable'].">,
     Option<"clRouteShimDmaToTileCTRL", "route-shim-to-tile-ctrl", "bool", /*default=*/"false",
             "Flag to generate routing between shim dma DMA and tile CTRL ports, for configuration.">,
+    Option<"clEmitStandaloneOverlay", "emit-standalone-overlay", "bool", /*default=*/"false",
+            "Emit a standalone @ctrl_pkt_overlay device (and tag source devices) for the load-pdi-to-ctrl-pkt reconfigure flow. When false, the pass applies the overlay in place only (legacy behavior).">,
```
andrej:
```suggestion
            "Emit a standalone @ctrl_pkt_overlay device (and tag source devices) for the load-pdi-to-ctrl-pkt reconfigure flow. When false, the pass applies the overlay in place only.">,
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -365,9 +367,10 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
       Port destPort = conn.dst;
       auto sourceFlow =
           std::make_pair(std::make_pair(tileId, sourcePort), flowID);
-      if (ctrlPktFlows[{{tileId, destPort}, flowID}])
+      if (ctrlPktFlows[{{tileId, destPort}, flowID}]) {
         ctrlPacketFlows[sourceFlow].push_back({tileId, destPort});
-      else
+        ctrlPktOverlayMasterPorts.insert({tileId, destPort});
+      } else
         packetFlows[sourceFlow].push_back({tileId, destPort});
```
andrej:
```suggestion
      } else {
        packetFlows[sourceFlow].push_back({tileId, destPort});
      }
```

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,145 @@ struct AIEGenerateColumnControlOverlayPass
...
+      sourceDevices.push_back(dev);
+    }
+
+    // Devices that receive the overlay: those that have not opted out via
+    // `needs_ctrl_pkt_overlay = false`.
+    SmallVector<DeviceOp> participating;
+    for (auto dev : sourceDevices) {
+      if (deviceOptedOut(dev)) {
+        if (clEmitStandaloneOverlay)
+          dev->setAttr("has_ctrl_pkt_overlay", builder.getBoolAttr(false));
```
```suggestion
        if (clEmitStandaloneOverlay) {
          dev->setAttr("has_ctrl_pkt_overlay", builder.getBoolAttr(false));
        }
```

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,145 @@ struct AIEGenerateColumnControlOverlayPass
...
+      if (failed(applyOverlayToDevice(dev)))
+        return signalPassFailure();
+      if (clEmitStandaloneOverlay)
+        dev->setAttr("has_ctrl_pkt_overlay", builder.getBoolAttr(true));
+    }
+
+    // Emit standalone `@ctrl_pkt_overlay` device.
+    if (clEmitStandaloneOverlay)
+      if (failed(createOverlayDevice(module, builder, participating)))
+        return signalPassFailure();
```
```suggestion
    if (clEmitStandaloneOverlay) {
      if (failed(createOverlayDevice(module, builder, participating))) {
        return signalPassFailure();
      }
    }
```

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,145 @@ struct AIEGenerateColumnControlOverlayPass
...
+      return success();
+
+    // All participating devices must share the same target.
+    DeviceOp firstDev = participating.front();
+    auto refDevice = firstDev.getDevice();
+    for (auto dev : llvm::drop_begin(participating))
+      if (dev.getDevice() != refDevice)
+        return dev->emitOpError(
+            "cannot generate a single standalone ctrl_pkt_overlay device: "
+            "participating devices have mismatched target architectures.");
```
```suggestion
    for (auto dev : llvm::drop_begin(participating)) {
      if (dev.getDevice() != refDevice) {
        return dev->emitOpError(
            "cannot generate a single standalone ctrl_pkt_overlay device: "
            "participating devices have mismatched target architectures.");
      }
    }
```

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -144,12 +145,145 @@ struct AIEGenerateColumnControlOverlayPass
...
+
+    if (module.lookupSymbol("ctrl_pkt_overlay"))
+      return module.emitOpError(
+          "a symbol named `ctrl_pkt_overlay` already exists in the module; "
+          "cannot create a standalone ctrl_pkt_overlay device.");
```
```suggestion
    if (module.lookupSymbol("ctrl_pkt_overlay")) {
      return module.emitOpError(
          "a symbol named `ctrl_pkt_overlay` already exists in the module; "
          "cannot create a standalone ctrl_pkt_overlay device.");
    }
```

lib/Dialect/AIEX/Transforms/AIEExpandLoadPdi.cpp
```diff
@@ -48,47 +92,91 @@ static LogicalResult transformLoadPdi(NpuLoadPdiOp loadPdiOp, ModuleOp moduleOp,
...
-    emptyDevice.getRegion().emplaceBlock();
-    Block *deviceBlock = &emptyDevice.getRegion().front();
-    builder.setInsertionPointToEnd(deviceBlock);
-    AIE::EndOp::create(builder, loc);
+  FlatSymbolRefAttr preloadRef;
+  if (ctrlPkt) {
+    // Step 1 (ctrl-pkt): emit a load_pdi for the ctrl_pkt_overlay device so
+    // that the NPU is configured to stream further configuration as control
+    // packets.
+    //
```
andrej:
```suggestion
    // Overlay device PDI
```

lib/Dialect/AIEX/Transforms/AIEExpandLoadPdi.cpp
```diff
@@ -48,47 +92,91 @@ static LogicalResult transformLoadPdi(NpuLoadPdiOp loadPdiOp, ModuleOp moduleOp,
...
+    } else if (!moduleOp.lookupSymbol<AIE::DeviceOp>(kCtrlPktOverlayName)) {
+      loadPdiOp.emitError("ctrl-pkt mode requires a `@")
+          << kCtrlPktOverlayName << "` device in the module";
+      return failure();
+    }
+    preloadRef = FlatSymbolRefAttr::get(builder.getContext(), overlayName);
+  } else {
+    // Step 1 (default): create a unique empty device for this reset to avoid
+    // PDI address caching, then emit a load_pdi for it (this triggers a
+    // device reset, needed even for the first device configuration).
```
andrej:
```suggestion
    // Empty device PDI (triggers firmware reset)
```

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1067,96 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  // the downstream control-packet edges are empty and the config omits the
+  // control-packet fields.
+  auto &ctrlPktExpandedPerDevice = splitPerDevice(
+      ctrlPktExpanded, "ctrlpkt_expanded_{0}.mlir", "ctrlPktExpandedMatching");
+  auto &noCtrlPktDevices = g.empty<OpInModule<DeviceOp>>("noCtrlPktDevices");
+
+  // The overlay/reset devices synthesized by expand-load-pdi carry no runtime
+  // sequence and thus no control packets; skip them so the translation only
+  // runs on real devices. `ctrlPktDevices` fans out to the control-packet
+  // binary and its patch-info.
```
```suggestion
  // `ctrlPktDevices` contains all devices that have control packet data
```

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1067,96 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
       "npu_insts_full_elf_{0}.bin",
       [](const NpuProgram &p) { return p.insts; });
 
-  // Combined ELF: all PDIs + NPU insts bundled into one aie2_config ELF.
+  // Per-device control-packet artifacts for the load-pdi-to-ctrl-pkt
+  // reconfigure flow. The control-packet binary is extracted directly from the
+  // ctrl-pkt-expanded module (control-packet ops still present, before they are
+  // lowered to DMA). Outside the flow the source is an explicit empty edge, so
+  // the downstream control-packet edges are empty and the config omits the
+  // control-packet fields.
```
```suggestion
  // Full ELF: all PDIs + NPU insts + control packet data if applicable
  // If the control packet lowering is not enabled, the empty 
  // `noCtrlPktDevices` edge is fed into the full ELF assembly bundle.
```

tools/aiecc/aiecc.cpp
```diff
@@ -1147,7 +1269,13 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
     outputs.push_back(&pdi);
   if (generateTxn)
     outputs.push_back(&txn);
-  if (generateCtrlpkt) {
+  // The standalone control-packet artifacts (partial ELF + DMA sequence, with
+  // the DDR-aperture offset folded for the xclbin / instruction-buffer runtime)
+  // are the xclbin-path delivery mechanism. In the --load-pdi-to-ctrl-pkt
+  // reconfigure flow targeting a full ELF they are superseded: the control
+  // packets are baked into aie.elf directly (fold-free, via the fullElfCtrlpkt
+  // / patch-info edges), so the standalone partial outputs must not be emitted.
```
```suggestion
  // Only emit standalone control-packet artifacts (partial ELF + DMA sequence)
  // in the non-full-ELF flow. In the full-ELF-flow, the control packet data is
  // contained in the ELF and patched into a runtime sequence argument.
```

tools/aiecc/aiecc.cpp
```diff
@@ -1296,10 +1424,22 @@ int main(int argc, char **argv) {
...
+
+  // The standalone control-packet DMA sequence targets the xclbin /
+  // instruction-buffer runtime, never the full-ELF runtime, so the two must not
+  // be requested together (see the hard-coded fold in the ctrlpktDmaSeq edge).
+  // The --load-pdi-to-ctrl-pkt reconfigure flow is the exception: there the
+  // control packets are baked into the full ELF itself (fold-free, via the
+  // fullElfCtrlpkt / patch-info edges), so full-ELF + ctrl-pkt is valid.
+  if (generateFullElf && generateCtrlpkt && !loadPdiToCtrlPkt) {
     llvm::errs() << "aiecc: --generate-full-elf and --aie-generate-ctrlpkt are "
                     "mutually exclusive\n";
```
```suggestion
  // Disambiguate the full-ELF control packet flow and the standalone
  // artifact flows.
  if (generateFullElf && generateCtrlpkt && !loadPdiToCtrlPkt) {
llvm::errs() << "aiecc: --generate-full-elf and --aie-generate-ctrlpkt "
"together also requires --load-pdi-to-ctrl-pkt\n";
```

tools/aiecc/aiecc.cpp
```diff
@@ -1019,22 +1061,90 @@ std::vector<EdgeBase *> buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+                  [](const Item<OpInModule<DeviceOp>> &item,
+                     std::vector<uint32_t> &words) -> mlir::LogicalResult {
+                    DeviceOp d = item.get().op;
+                    return xilinx::AIE::AIETranslateControlPacketsToUI32Vec(
+                        item.get().module.get(), words, d.getSymName(), "");
+                  }))
+          .filter("fullElfCtrlpktNonEmpty",
+                  [](const std::vector<char> &bin) { return !bin.empty(); });
+
+  // When control packets are enabled, the control data is passed into the
```
hunhoffe, line 1119: I think this might already be tested with tracing, but might be worth a second look to make sure it's compatible.

hunhoffe (approved):
From what I can tell, this all seems reasonable and clean!

tools/aiecc/aiecc.cpp
andrej: Oof, good thought. I don't think there currently are any tests for full ELF + tracing. Created issue #3383.

PR #3340, opened 2026-07-14
Unconditionally run 'Smoke-test wheels'

andrej:
Whenever the wheel gets built, it should get smoke-tested. Testing is cheap. Remove the old conditional.


jgmelber (approved):
We can always gate this again if it becomes a load on CI

PR #980, opened 2024-02-06
Cleanup calls to dyn_cast

AndraBisca:
Cleanup unchecked calls to `llvm::dyn_cast` by replacing them with `llvm::cast`

Issue #721, opened 2023-11-06
Clean up/refactor


makslevental:
- [ ] remove `using namespace` in headers
- [ ] fix `I32Attr`s 
- [ ] remove `class NetlistAnalysis` entirely
- [ ] `ConfinedAttr<I32Attr, [IntMinValue<0>, IntMaxValue<1>]>:$AB` to EnumAttr
- [ ] `AIE_DimTupleAttr` to EnumAttr

jsetoain:
Shall I take a couple of these?

makslevental:
> Shall I take a couple of these?

You don't have to but of course I'd be glad 😊.

hunhoffe:
Got some valuable feedback from @jsetoain on https://github.com/Xilinx/mlir-aie/pull/3344 which includes some other code-style issues which should probably be added to this list at TODOs.

PR #3344, opened 2026-07-14
Replace unchecked dyn_cast with cast

hunhoffe:
Succeeds #980, adapted to current `main`

## Summary
  
Converts `dyn_cast` to `cast` at call sites where the result is dereferenced without a null check, or where a preceding `isa`/`isa_and_nonnull`/`assert` already guarantees the type.

`cast` documents the invariant and traps at the cast site (in asserts builds) instead of deferring to a downstream null dereference. No behavior change — these sites already assumed the cast succeeded.


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -750,7 +750,7 @@ static ParseResult parseObjectFifoInitValues(OpAsmParser &parser,
   if (parser.parseAttribute(initValues, tensorType))
     return failure();
   for (int i = 0; i < depth; i++) {
-    auto initialValues = llvm::dyn_cast<mlir::ArrayAttr>(initValues);
+    auto initialValues = llvm::cast<mlir::ArrayAttr>(initValues);
     if ((int)initialValues.size() != depth)
       return parser.emitError(parser.getNameLoc())
              << "initial values should initialize all objects";
```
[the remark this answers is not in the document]
hunhoffe: addressed

andrej (approved):

jsetoain (changes requested):
Hold your horses!

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -711,14 +711,14 @@ static void printObjectFifoInitValues(OpAsmPrinter &p, ObjectFifoCreateOp op,
     // is the first entry of the ArrayAttr.
     int depth;
     if (isa<ArrayAttr>(numElem)) {
-      depth = llvm::dyn_cast<mlir::IntegerAttr>(
-                  llvm::dyn_cast<mlir::ArrayAttr>(numElem)[0])
-                  .getInt();
+      depth =
+          llvm::cast<mlir::IntegerAttr>(llvm::cast<mlir::ArrayAttr>(numElem)[0])
+              .getInt();
```
jsetoain, line 716: Two things to consider: 1) `cast` should only be used when a different constraint guarantees the type; 2) `dyn_cast` performs better than `isa` + `cast`.

Things like:
```c++
if (isa<Type>(op))
  typedOp = dyn_cast<Type>(op);
```
Would be better fixed as:
```c++
typedOp = dyn_cast<Type>(op);
if (!typedOp)
    fail_gracefully();
```

In this specific case, where you are going to be checking several possible types, you should use the `TypeSwitch` construct.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -751,11 +751,14 @@ static ParseResult parseObjectFifoInitValues(OpAsmParser &parser,
   Type tensorType = mlir::memref::getTensorTypeFromMemRefType(memrefType);
   if (parser.parseAttribute(initValues, tensorType))
     return failure();
+  auto initialValues = llvm::dyn_cast<mlir::ArrayAttr>(initValues);
+  if (!initialValues)
+    return parser.emitError(parser.getNameLoc())
+           << "initial values should be an array attribute";
+  if ((int)initialValues.size() != depth)
+    return parser.emitError(parser.getNameLoc())
+           << "initial values should initialize all objects";
```
jsetoain: This change is unrelated to the description of the PR. It's annoying, but maybe consider moving it to its own PR.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -799,7 +802,7 @@ ObjectFifoCreateOp ObjectFifoAllocateOp::getObjectFifo() {
     if (parent->hasTrait<OpTrait::SymbolTable>()) {
       if (auto *st = SymbolTable::lookupSymbolIn(parent, getObjFifoName());
           isa_and_nonnull<ObjectFifoCreateOp>(st))
-        return dyn_cast<ObjectFifoCreateOp>(st);
+        return cast<ObjectFifoCreateOp>(st);
```
jsetoain, line 802: This code should be arranged differently. `if (<statement>; <statement>)` is not good code style. There are a bunch of them.

hunhoffe:
@jsetoain sure thing! Let me know what's up?

jsetoain (commented):
Here it goes 🙂

lib/Dialect/AIE/IR/AIEDialect.cpp
hunhoffe: This seems slightly off topic, although I agree with you. Are you ok with deferring `if (<statement>; <statement>)` cleanups as a future task? Perhaps it could be added on as another TODO to https://github.com/Xilinx/mlir-aie/issues/721 which hopefully myself or someone else will get around to finishing eventually.

lib/Dialect/AIE/IR/AIEDialect.cpp
hunhoffe: This seems slightly off topic, as in this PR I didn't intend to change control flow. I tried to keep it narrowly scoped. 

One question, though, for my own edification --
The options are `isa` + `dyncast` OR `isa` + `cast` -- I didn't change control flow. Which of those choices would be optimal?

Are you ok with deferring code flow changes to a different PR/follow-up/add-on to https://github.com/Xilinx/mlir-aie/issues/721?

lib/Dialect/AIE/IR/AIEDialect.cpp
hunhoffe: This was a scope creep - nice catch! I will remove it from this PR.

lib/Dialect/AIE/IR/AIEDialect.cpp
hunhoffe: Side note: Happy to revert this change, if you found the original form preferable.

lib/Dialect/AIE/IR/AIEDialect.cpp
jsetoain: Of course, that's absolutely fine. I mentioned it because I'd expect the need for the switch from `dyn_cast` to `cast` to be gone after the cleanup.

lib/Dialect/AIE/IR/AIEDialect.cpp
jsetoain: The preferable form, in this case, is the `TypeSwitch` form 🙂 And the optimal one for anything else would be `dyn_cast` +  `nullptr` check, leaving `cast` (alone) for cases when the type is verified by either type constraints in the op, the pattern, or wherever else. `isa` + `dyn_cast` is definitely something that shouldn't happen.

jsetoain (approved):

PR #3345, opened 2026-07-14
[AIETargetNPU] Gate DDR-patch aperture offset by runtime (fix full-ELF >5 buffers)

erwei-xilinx:
## Problem

`appendAddressPatch` (in `AIETargetNPU.cpp`, added in #3252) folds a `0x80000000`
AIE-aperture offset into `arg_plus` for host arguments with `arg_idx >= 5`.

That fold is **runtime-specific**:

- **xclbin + instruction-buffer runtime** (`HOST_ONLY` BOs launched as
  `kernel(opcode, insts_bo, ninsts, ...)`): the NPU firmware only
  address-translates the first 5 host buffers, so the offset is **required** for
  args ≥5. Exercised (and kept passing) by `npu-xrt/many_buffers`,
  `npu-xrt/add_multi_arg_runlist`, `python/npu-xrt/test_jit_many_args`.

- **full-ELF runtime** (`xrt.elf` + `xrt.ext.kernel` + `run.set_arg`): the driver
  assigns NPU-space device addresses to **all** host buffers, so folding the
  offset **double-translates** the 6th+ buffer — its shim DMA lands ~2 GiB off and
  reads/writes garbage.

mlir-aie's own full-ELF path (`full_elf_42`) only ever used a single buffer, so
the full-ELF + >5-buffer case was never exercised and the bug went unnoticed.

## Fix

Make the fold conditional on the target runtime instead of unconditional:

- `AIETranslateNpuToBinary()` gains a `foldDDRAddrOffset` parameter, **default
  `true`** — so the xclbin/insts TXN is byte-identical to today.
- `aie-translate` exposes `--aie-npu-fold-ddr-addr-offset` (default `true`).
- `aiecc` passes `foldDDRAddrOffset = !generateFullElf` at the three sites that
  emit the full-ELF TXN — the single-sequence, control-packet, and multi-PDI
  (`expandLoadPdis`) generation paths — so no `--generate-full-elf` output folds
  the offset. (The separate `--aie-generate-elf` / insts.elf site is left
  unchanged; see Note.)

Net: xclbin builds are unchanged; full-ELF builds with >5 buffers are fixed.

## Tests

- `test/Targets/NPU/npu_address_patch_ddr_offset.mlir` — unit test asserting the
  `arg_plus` word gets the offset for `arg_idx >= 5` in fold mode (xclbin) and is
  left verbatim without it (full-ELF).
- `test/python/npu-xrt/full_elf_many_args/` — e2e full-ELF design with 8 host
  buffers. On the pre-fix compiler, args 5-7 read back 0; after the fix all 8 are
  correct. (This is the first full-ELF test in the tree with >5 buffers.)

## Verified on NPU2 hardware

- `full_elf_many_args` (8 buffers, full-ELF): args 5-7 = 0 before, all correct after.
- A downstream design with 6 host buffers built through the `--generate-full-elf`
  multi-PDI path: all-zeros before, correct after.
- xclbin `aie-translate` output for `arg_idx=5` is unchanged (still `0x80000100`).

## Note

The `--aie-generate-elf` (insts.elf) site is left folding by default; it has no
>5-buffer test today. If that path is also an `ext.kernel` runtime, it may want
`foldDDRAddrOffset=false` too — happy to extend if maintainers confirm.

### DIFF
--- a/include/aie/Targets/AIETargets.h
+++ b/include/aie/Targets/AIETargets.h
@@ -65,11 +65,10 @@ mlir::LogicalResult AIETranslateShimSolution(mlir::ModuleOp module,
                                              llvm::StringRef deviceName = "");
 mlir::LogicalResult AIETranslateGraphXPE(mlir::ModuleOp module,
                                          llvm::raw_ostream &, llvm::StringRef);
-mlir::LogicalResult
-AIETranslateNpuToBinary(mlir::ModuleOp, std::vector<uint32_t> &,
-                        llvm::StringRef deviceName = "",
-                        llvm::StringRef sequenceName = "",
-                        std::vector<TxnLocEntry> *locmap = nullptr);
+mlir::LogicalResult AIETranslateNpuToBinary(
+    mlir::ModuleOp, std::vector<uint32_t> &, llvm::StringRef deviceName = "",
+    llvm::StringRef sequenceName = "",
+    std::vector<TxnLocEntry> *locmap = nullptr, bool foldDDRAddrOffset = true);
 mlir::LogicalResult AIETranslateNpuToCpp(mlir::ModuleOp module,
                                          llvm::raw_ostream &output);
 mlir::LogicalResult AIETranslateToUcDma(mlir::ModuleOp module,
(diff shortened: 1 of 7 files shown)


hunhoffe (approved):
Thanks for figuring out this bug!

PR #3346, opened 2026-07-15
[python] aiex: fix shim_dma_single_bd_task deadlock on sub-4-dim taps

atassis:
## Problem

`shim_dma_single_bd_task` hoists the leading tap dimension into the shim DMA's
hardware repeat/iteration dimension:

```python
repeat_count = 0
if sizes and sizes[0] > 1:
    repeat_count = sizes[0] - 1
```

but the transferred extent (in `shim_dma_bd`) is `transfer_len = prod(sizes[-3:])`,
and the full `sizes`/`strides` are handed to the BD. That is only self-consistent
for a 4-dimensional tap, where `sizes[0]` is the dedicated iteration dim and is
therefore excluded from `transfer_len`.

For a tap with fewer than 4 dims and `sizes[0] > 1`, `sizes[0]` is counted twice:
once as a real access dimension (it is in `sizes[-3:]`, so it is in `transfer_len`,
and it appears in the BD dimensions) and once as `repeat_count`. The shim then
re-issues the whole transfer `sizes[0]` times and waits on the objectfifo for
`sizes[0]x` the number of objects the producer emits, so `dma_await_task` never
returns. It compiles cleanly and hangs on the device.

Taps produced by `TensorTiler2D.simple_tiler` dodge this because they are already
4-dim (leading unit dims), so `sizes[0] == 1` and no repeat is emitted. A
hand-built 3-dim `TensorAccessPattern` (for example a strided/reshaping drain)
hits it.

## Concretely

A 3-dim drain tap `sizes=[64, 4, 1024]`, `strides=[1024, 524288, 1]` currently
lowers to:

```
aie.dma_bd(%out, 0, 262144, [<64,1024>, <4,524288>, <1024,1>]) {repeat_count = 63}
```

`len = 262144 = 64*4*1024` already covers all 64 rows, and `repeat_count = 63`
asks the shim to run that full transfer 64 times. With the fix it lowers to the
canonical 4-dim one-shot:

```
aie.dma_bd(%out, 0, 262144, [<1,0>, <64,1024>, <4,524288>, <1024,1>])
```

`repeat_count = 0`, same shape as the working 4-dim fill path.

## Fix

Normalize the tap to the canonical 4-dim form (left-pad with unit dims) before
computing `repeat_count`, so `sizes[0]` is only ever the iteration dimension.
Also raise on `len(sizes) > 4` instead of silently dropping dimensions from
`transfer_len` and emitting a wrong BD.

## Test

Adds `test/python/shim_dma_task_subdim_tap.py`: builds a `shim_dma_single_bd_task`
from a 3-dim tap and FileChecks that the emitted BD is the canonical 4-dim form
with no `repeat_count`, and that a 5-dim tap is rejected. It fails on `main`
(emits the full-extent BD plus `repeat_count`) and passes with this fix.

## Validation

Hit this on a chunk-major deinterleave drain (3-dim output tap). Rebuilding the
same generator with this patch flips `repeat_count` 63 -> 0 on each output BD
(the compiled instruction stream changes by exactly one byte per BD). On the
device the sole-context dispatch that hung deterministically now completes, and
a real workload (the deint running co-resident with a matmul, 408 dispatches)
runs clean with bit-identical results.

### DIFF
--- a/python/dialects/aiex.py
+++ b/python/dialects/aiex.py
@@ -356,6 +356,27 @@ def shim_dma_single_bd_task(
         # so here we make sure it is evaluated and properly is seen as an integer.
         offset = int(tap.offset)
 
+    # The shim DMA BD has 3 access dimensions plus a hardware repeat/iteration
+    # dimension. The repeat_count below hoists sizes[0] into that iteration
+    # dimension, but the transferred extent is prod(sizes[-3:]) (see shim_dma_bd),
+    # so sizes[0] is left out of it only when there are 4 dimensions. With fewer
+    # than 4 dims and sizes[0] > 1, sizes[0] is counted both as a real access dim
+    # (in transfer_len and in the BD dimensions) and as repeat_count, so the shim
+    # re-issues the whole transfer sizes[0] times, waits on the objectfifo for that
+    # many objects, and dma_await_task never returns. Normalize to the canonical
+    # 4-dim form (left-pad with unit dims) so sizes[0] is only ever the iteration
+    # dimension, and reject taps with more than 4 dims instead of silently emitting
+    # a wrong BD.
+    if sizes is not None:
+        if len(sizes) > 4:
+            raise ValueError(
+                f"shim DMA BD supports at most 4 dimensions, got {len(sizes)}"
+            )
+        while len(sizes) < 4:
+            sizes = [1] + list(sizes)
+            if strides is not None:
+                strides = [0] + list(strides)
+
     repeat_count = 0
     if sizes and sizes[0] > 1:
         repeat_count = sizes[0] - 1
--- a/test/python/shim_dma_task_subdim_tap.py
+++ b/test/python/shim_dma_task_subdim_tap.py
@@ -51,7 +51,10 @@ def case(name, dims, sizes, strides, use_tap=True):
                         )
                     else:
                         shim_dma_single_bd_task(
-                            of, out, sizes=list(sizes), strides=list(strides),
+                            of,
+                            out,
+                            sizes=list(sizes),
+                            strides=list(strides),
                             issue_token=True,
                         )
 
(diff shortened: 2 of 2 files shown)


python/dialects/aiex.py
```diff
@@ -356,6 +356,27 @@ def shim_dma_single_bd_task(
...
+    # a wrong BD.
+    if sizes is not None:
+        if len(sizes) > 4:
+            raise ValueError(
+                f"shim DMA BD supports at most 4 dimensions, got {len(sizes)}"
+            )
+        while len(sizes) < 4:
+            sizes = [1] + list(sizes)
+            if strides is not None:
+                strides = [0] + list(strides)
```
jgmelber, line 378: It's nice to add more checks, however these sorts of limits could be architecture dependent and are definitely tile dependent. The TargetModel in MLIR can be queried for these sorts of constraints

atassis: You're right, and digging into it I agree the clean fix is to source this from the TargetModel instead of a literal. I traced what that would take.

Right now DMABDOp::verify infers the limit from the parent op type (maxNDims = 3, or 4 under a MemTileDMAOp) rather than from the tile, and AIETargetModel doesn't expose a per-BD max-dimensions accessor (only getNumBDs, which is the BD count). So a real query means adding that accessor and overriding it across the target models.

It also wouldn't currently cover the path this PR touches. shim_dma_single_bd_task lowers through dma_configure_task_for and aie-dma-tasks-to-npu, and the BD never lands under a ShimDMAOp, so DMABDOp::verify is skipped for it. The tile is recoverable through the shim DMA allocation symbol, but on that path the BD carries the iteration dim in its own sizes list and hoists it into repeat_count, so the verifier would need to account for that or it would reject the legitimate 4-dim taps.

So doing it properly is a TargetModel plus verifier plus task-path change, with the test churn that comes with it, which is quite a bit deeper than this deadlock fix is meant to be. I'm happy to take it on as a follow-up: a new TargetModel accessor, DMABDOp::verify querying it by tile type, and extending the check to the runtime-sequence path. In the meantime I'd keep the small Python guard here, since it's currently the only thing catching this on that path. Would you be OK merging the deadlock fix as it is and letting me open the TargetModel-backed verifier rewrite separately? If you'd rather have them together I can fold it into this PR instead.

hunhoffe: Yes, the deeper fix as a follow up sounds good! I'm ok with merging as-is for now. I've created an issue to track it: https://github.com/Xilinx/mlir-aie/issues/3349

hunhoffe (approved):

PR #3347, opened 2026-07-15
Add HRX (amdxdna) host-runtime backend and XRT-free xclbin packaging

Abhishek-Varma:
Add HRX (amdxdna) host-runtime backend and XRT-free xclbin packaging

Purpose: give mlir-aie a direct HRX (`amdxdna`) host-dispatch path as an alternative to XRT, for both the IRON/Python flow and the C++ test flow, and remove the remaining XRT/aiebu build-time dependencies so an mlir-aie design can be built and run end-to-end on an XRT-free box.

## Motivation

The XRT host stack (and the aiebu control-code assembler it pulls in) is the last hard dependency between an mlir-aie design and running it on the NPU. This change adds a thin `hrx` runtime layer on top of `libhrx` so device/buffer/kernel/run calls go straight to `amdxdna`, and packages a self-contained `xclbinutil` so the `.xclbin` can be built without a system XRT install. The existing XRT path is untouched and remains the default.

## Technical Details

- IRON/Python backend (`python/utils/hostruntime/hrxruntime/`): a `HRXHostRuntime` + `HRXTensor` runtime over `libhrx`, selected with `IRON_RUNTIME=hrx`. NPU control code is wrapped into an amdxdna XADX "direct executable" via libhrx's native `hrx_amdxdna_executable_create` (libhrx builds the XADX flatbuffer and derives the patch table internally) and loaded via the `amdxdna-xclbin-fb` HAL format.
- C++ test backend: `runtime_lib/test_lib/hrx_test_wrapper.h` mirrors the variadic `setup_and_run_aie` API of `xrt_test_wrapper.h`, gated by `-DUSE_HRX=ON`, and calls `hrx_amdxdna_executable_create` directly. `programming_examples/common.cmake` + `makefile-common` add a `RUNTIME=xrt|hrx` switch that wires this in.
- Multi-dispatch / chain (runlist) support: `HRXHostRuntime.run_chain` (Python) and the C++ equivalent submit several dispatches as a single `ERT_CMD_CHAIN`, so a producer->consumer buffer dependency is honored within one batched submit.
- libhrx discovery: `cmake/modules/FindHRX.cmake` for C++ and `hrxruntime/discovery.py` for Python resolve the shipped `hrx::hrx` CMake package / install-prefix layout, probe standard locations and a sibling `../hrx` checkout, and honor `HRX_DIR`/`LIBHRX_DIR` hints.
- Bundled XRT-free `xclbinutil`: `third_party/hrx-xclbinutil` (a trimmed, Boost-free copy of XRT's `xclbinutil` with only the NPU section handlers) builds under `-DAIE_BUILD_HRXXCLBINUTIL=ON` and installs as `xclbinutil` alongside `aiecc`. `aiecc` takes an explicit `--xclbinutil-path` (and reads `AIE_XCLBINUTIL`), and `makefile-common`'s `RUNTIME=xrt|hrx` switch points it at the bundled tool for `hrx` and XRT's tool for `xrt` — so packaging is deterministic per-runtime and a `hrx` build never silently uses XRT's tool (it errors if the chosen tool is missing). This drops the last packaging-time XRT dependency.
- Pinned HRX provisioning: `hrx-integration/` (`hrx-release.env` + `fetch-hrx-release.sh`) downloads, checksum-verifies, and extracts a pinned `libhrx` release (`.tar.zst`) and synthesizes an `env.sh` (`HRX_DIR`/`LD_LIBRARY_PATH`/`CMAKE_PREFIX_PATH`) consumed by both discovery paths — so HRX can be used without building it from source.
- CI: a pure-HRX job (`.github/workflows/buildAndTestRyzenAI.yml`, `aie2p-8col`) fetches the pinned HRX release, builds with the bundled `hrx-xclbinutil` and no XRT bindings, runs the `hrx_xclbin_sections` lit check, and runs designs on the NPU via both the Python and C++ HRX paths — the multi-dispatch chain/runlist path (`vector_scalar_add`) plus the single-dispatch path (`vector_vector_add` IRON + `vector_reduce_max` C++).
- Regression coverage: `test/aiecc/hrx_xclbin_sections.mlir` builds a real `.xclbin` with the bundled tool and asserts it carries `MEM_TOPOLOGY`, `AIE_PARTITION`, and `EMBEDDED_METADATA`; gated on a new `hrxxclbinutil` lit feature so it only runs when the tool is built.
- Example wiring: `vector_scalar_add` gains `test_chain_hrx.py` (Python HRX chain test) and `test_runlist_hrx.cpp` (C++ HRX runlist testbench).
- Hardening: all path expansions in the `hrx-xclbinutil` CMake are quoted (so a source dir with spaces/semicolons can't break the commands). The host no longer shells out to `aiebu-asm`: libhrx derives the patch table from the raw transaction, so the Python and C++ paths hand the `insts.bin` transaction (or an ELF's `.ctrltext`) straight to `hrx_amdxdna_executable_create`.

## Test Plan

- Build mlir-aie from the prebuilt MLIR distro wheel + Peano with `-DAIE_BUILD_HRXXCLBINUTIL=ON`.
- Provision `libhrx` from the pinned release via `hrx-integration/fetch-hrx-release.sh` and source the emitted `env.sh`.
- Run the `hrx_xclbin_sections.mlir` lit test.
- Build a design's `.xclbin` through the IRON path with `RUNTIME=hrx` (bundled `xclbinutil` selected via `AIE_XCLBINUTIL`) and run it on the NPU via both the Python and C++ HRX backends, covering both the chain/runlist and the single-dispatch paths.

## Test Result

- HRX release: `jtuyls/hrx@flm-hrx-amdxdna-v2026.07.20` (commit `fcaecde`), which exports `hrx_amdxdna_executable_create`.
- HRX asset: `hrx-amdxdna-2026.07.20-amdxdna-hal-native-rel-fcaecde-linux-x86_64.tar.zst`.
- `xclbinutil --version` reports `Build Version Branch: mlir-aie-hrx`, and `RUNTIME=hrx` selects it via `AIE_XCLBINUTIL` (`RUNTIME=xrt` selects XRT's).
- lit: `PASS: AIE_TEST :: aiecc/hrx_xclbin_sections.mlir`; `--info` on the produced xclbin shows `MEM_TOPOLOGY, AIE_PARTITION, EMBEDDED_METADATA`.
- Python HRX chain (`IRON_RUNTIME=hrx test_chain_hrx.py`) on NPU: chain and depth-8 deep-chain results correct — `PASS!`.
- C++ HRX runlist (`make run_runlist_hrx`, `-DUSE_HRX=ON`, linking `libhrx`) on NPU: chained runs correct — `PASS!`.
- Single-dispatch on NPU: IRON `vector_vector_add` (`IRON_RUNTIME=hrx`) and C++ `vector_reduce_max` (`RUNTIME=hrx make run`) both `PASS!`.

## Submission Checklist

- [x] HRX backend for the IRON/Python path builds and runs on the NPU.
- [x] HRX backend for the C++ test path builds and runs on the NPU.
- [x] Bundled `hrx-xclbinutil` packages a valid NPU `.xclbin` (lit section check).
- [x] Existing XRT path remains the default and is unchanged.
- [x] Add NPU CI that exercises the HRX Python/C++ paths on hardware runners.


jgmelber:
Hey! @Abhishek-Varma is this ready to review? What are the next steps?

Abhishek-Varma:
> Hey! @Abhishek-Varma is this ready to review? What are the next steps?

Hi @jgmelber! Thank you! :)
Yes, it's now ready for review.
- Have addressed the review comments by Co-pilot so far and pushed.
- The current revision contains a bump in HRX and a few consequential changes (like removal of XADX helper).
- Have also updated PR description.

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
+<!---//===- README.md --------------------------*- Markdown -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//-->
+
+# Running MLIR-AIE IRON (Python) & C++ tests on the **HRX** runtime
```
jgmelber:
```suggestion
# Running IRON (Python) & C++ tests on the **HRX** runtime
```

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
...
+
+This package (`aie.utils.hostruntime.hrxruntime`) is the **HRX** host-runtime
+backend for MLIR-AIE. It dispatches IRON designs and the C++ testbench on the
```
jgmelber:
```suggestion
backend for IRON. It dispatches IRON designs and the C++ testbench on the
```

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
...
+| **MLIR-AIE installed from *this* branch** | So `import aie` works, `aiecc` can build examples, and the wheel includes this `hrxruntime` package + the `IRON_RUNTIME` switch. See §3. |
+| **A C++ toolchain + CMake ≥ 3.30** | Only for the C++ tests. (`pip install "cmake>=3.30"` into the venv if the system cmake is older.) |
+| **Peano / llvm-aie** | Needed by `aiecc` to build the example `.xclbin`/`insts.bin` (same as the XRT flow). |
+| **`xclbinutil`** (xclbin packaging) | `aiecc` packages the `.xclbin` with the tool named by **`AIE_XCLBINUTIL`** / **`--xclbinutil-path`** (explicit, deterministic — see §3b), falling back to a `PATH` lookup only if neither is set. Provided by XRT, **or** build the bundled XRT-free `hrx-xclbinutil` (§3b) — recommended on an XRT-free box. |
+| XRT userspace (pyxrt) | **Optional.** Only needed for the XRT baseline. HRX itself does not need it. |
+
+The NPU + driver + Peano requirements are exactly the same as the normal (XRT)
+MLIR-AIE flow — if XRT examples build/run on this box, HRX has what it needs too,
```
jgmelber:
```suggestion
IRON flow — if XRT examples build/run on this box, HRX has what it needs too,
```

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
...
+> `mlir-aie`** into `<parent-of-mlir-aie>/hrx-system` and install to
+> `hrx-system/build/hrx-install`. Auto-detection then finds it with no env vars.
+> (`~/hrx`, `/opt/hrx`, `/usr/local/hrx` install prefixes also work; anything
+> else needs the env hints in §4.)
+
+```bash
+# from the directory that contains your mlir-aie checkout:
+git clone https://github.com/ROCm/hrx-system.git hrx-system
+cd hrx-system
+git remote add jorn https://github.com/jtuyls/hrx.git
```
jgmelber: Should this be `ROCm/hrx` or this fork?

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
...
+cmake --install build/cmake --prefix build/hrx-install --component HrxPublicDist
+```
+
+> **libstdc++ note:** if the ROCm `clang++` can't find `libstdc++` (e.g. it
+> defaults to a gcc install dir that only ships the runtime `.so.6`), point it at
+> a full gcc toolchain by adding
+> `-DCMAKE_C_FLAGS=--gcc-install-dir=<dir> -DCMAKE_CXX_FLAGS=--gcc-install-dir=<dir>`
+> (e.g. `/usr/lib/gcc/x86_64-linux-gnu/13`) to the configure line.
+
+**Verify** the install produced what the MLIR-AIE HRX backend consumes — the
```
jgmelber:
```suggestion
**Verify** the install produced what the IRON HRX backend consumes — the
```

runtime_lib/test_lib/xrt_test_wrapper.h
```diff
@@ -1,6 +1,13 @@
 /* Copyright (C) 2025 Advanced Micro Devices, Inc.
```
jgmelber:
```suggestion
/* Copyright (C) 2025-2026 Advanced Micro Devices, Inc.
```

jgmelber (approved):
Overall LGTM. It would be nice to have an issue tracking the HRX integration plans and discussions. 

What is the plan for `xclbinutil`? Could that be shipped with HRX? Or still included as a third party submodule here? If the latter, that can be included in a follow up PR. I'd like to discuss alignment with the upstream `xclbinutil` rather than a personal fork.

hrx-integration/fetch-hrx-release.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
+##===- hrx-integration/fetch-hrx-release.sh -----------------*- Script -*-===##
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+##===----------------------------------------------------------------------===##
+#
+# Download, checksum-verify, and extract the pinned HRX (amdxdna) release so the
+# HRX runtime path (IRON_RUNTIME=hrx / RUNTIME=hrx) has a libhrx to dispatch
```
hunhoffe: Are there two environment variables for the same thing?

utils/fetch-hrx-release.sh
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env bash
```
hunhoffe, line 1: Has this been tested for windows integration by chance? That's not a blocker -- we don't have windows CI yet -- but we're working towards more native support for windows.

third_party/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
```
hunhoffe, line 1: Do we really need a new hrx-integration directory? Can we put these files in a mixture of scripts/third_party e.g., conform to the existing dependency organization?

programming_examples/basic/vector_scalar_add/test_chain_hrx.py
```diff
@@ -0,0 +1,109 @@
...
+    IRON_RUNTIME=hrx python3 test_chain_hrx.py \
+        --xclbin build/final.xclbin --instr build/insts.bin --kernel MLIR_AIE
+"""
+
+import argparse
+import sys
+
+import numpy as np
+
+from aie.utils.hostruntime.hrxruntime.hostruntime import HRXHostRuntime
```
hunhoffe: Why does this code need to be HRX specific? Can't we rely on setting the IRON_RUNTIME to ensure tensors, etc. are HRX? I'd like to keep runtime-specific details out of the higher-level Python code as much as possible. If you want to test the backend by being explicit, you could add it as a test to test/python/npu-xrt (or rename to test/python/npu or test/python/npu-hrx)

programming_examples/basic/vector_scalar_add/test_chain_hrx.py
```diff
@@ -0,0 +1,109 @@
...
+    stages = [HRXTensor((size,), dtype=np.int32) for _ in range(depth)]
+    chain = [(handle, [in_a, stages[0]])]
+    for k in range(1, depth):
+        chain.append((handle, [stages[k - 1], stages[k]]))
+    rt.run_chain(chain)
+
+    base = np.arange(1, size + 1, dtype=np.int32)
+    for k, st in enumerate(stages):
+        ref = base + (k + 1)
+        got = st.numpy()
```
hunhoffe: I think we have some helper functions for checking tensor equality.

python/utils/hostruntime/hrxruntime/__init__.py
```diff
@@ -0,0 +1,633 @@
...
+        ("flags", ctypes.c_uint32),
+        ("reserved", ctypes.c_uint32),
+        ("xclbins", ctypes.POINTER(HrxConstByteSpan)),
+        ("xclbin_count", ctypes.c_size_t),
+        ("entry_points", ctypes.POINTER(HrxAmdxdnaExecutableEntryPoint)),
+        ("entry_point_count", ctypes.c_size_t),
+    ]
+
+
+def _load_libhrx() -> ctypes.CDLL:
```
hunhoffe: This seems like a good start, so no changes needed now, but how platform dependent is this code?

python/utils/hostruntime/hrxruntime/__init__.py
```diff
@@ -0,0 +1,633 @@
...
+
+    Only needed for an ELF input (``aiecc --aie-generate-elf``): ``.ctrltext``
+    is the TXN stream verbatim, which is what libhrx patches from. A raw
+    ``insts.bin`` is already the transaction and needs no extraction. libhrx
+    derives the buffer patch table from the transaction itself inside
+    ``hrx_amdxdna_executable_create``, so no host-side patch table is produced.
+
+    Returns the control code as a ``numpy.uint32`` array.
+    """
+    import struct
```
hunhoffe: Style nit: we try to avoid on-the-fly imports.

python/utils/hostruntime/hrxruntime/__init__.py
```diff
@@ -0,0 +1,633 @@
...
+
+    for i in range(e_shnum):
+        if sname(i) == ".ctrltext":
+            coff = rd("<I", sh(i, 0x10))
+            csize = rd("<I", sh(i, 0x14))
+            return np.frombuffer(d[coff : coff + csize], dtype=np.uint32).copy()
+    raise HRXError("control ELF has no .ctrltext section")
+
+
+class HRXContext:
```
hunhoffe: This file is pretty big... maybe we could split it up into bindings and context modules or something like that?

python/utils/hostruntime/hrxruntime/hostruntime.py
```diff
@@ -0,0 +1,364 @@
...
+
+    Order:
+      1. ``IRON_HRX_DEVICE`` env override (always wins);
+      2. sysfs PCI device-id probe (amdxdna-bound device, else any AMD NPU id);
+      3. fall back to ``npu2`` (the common XDNA2 case) with a debug log.
+
+    Fully offline and XRT-free (no dlopen, no device init), so it is safe to call
+    from the runtime constructor.
+    """
+    env = os.environ.get("IRON_HRX_DEVICE")
```
hunhoffe, line 53: There seems to be a lot of environment variables... could you please ensure they are all documented here: https://github.com/Xilinx/mlir-aie/blob/main/programming_guide/iron_configuration.md

python/utils/hostruntime/hrxruntime/hostruntime.py
```diff
@@ -0,0 +1,364 @@
...
+
+        # Outputs were written on-device; the persistent host mapping is stale.
+        # Leave the tensors marked device="npu" so the next host read
+        # (numpy()/to("cpu")) invalidates the cache via _sync_from_device.
+        for a in args:
+            a.device = "npu"
+
+        return HRXKernelResult(stop - start, success=True)
+
+    def run_chain(self, runs, fail_on_error: bool = True) -> HRXKernelResult:
```
hunhoffe, line 308: How does this compare to the xrt runlist? I'm wondering if we could extract some sort of shared abstraction @andrej  -- do you have an opinion here?

python/utils/hostruntime/hrxruntime/hostruntime.py
```diff
@@ -0,0 +1,364 @@
...
+        except HRXError as e:
+            if fail_on_error:
+                raise HostRuntimeError(f"HRX chain dispatch failed: {e}") from e
+            stop = time.time_ns()
+            return HRXKernelResult(stop - start, success=False)
+        stop = time.time_ns()
+
+        # Mark every touched tensor device-resident so the next host read
+        # invalidates and observes the on-device results.
+        for a in touched:
```
hunhoffe, line 372: Does this do extra syncs for input buffers which don't need to be returned to the host?

python/utils/hostruntime/hrxruntime/hostruntime.py
```diff
@@ -0,0 +1,364 @@
...
+            return
+        while cache:
+            _, (exe, _) = cache.popitem(last=False)
+            try:
+                self._ctx.release_executable(exe)
+            except HRXError as e:
+                logger.debug("HRX executable release failed during cleanup: %s", e)
+
+
+class CachedHRXRuntime(HRXHostRuntime):
```
hunhoffe, line 417: One of the reasons we have cached vs. uncached is because on shared systems, you might not want a cache. Would it make sense to have cached and uncached for HRX too? Regardless, we shouldn't have a HRXHostRuntime and a CachedHRXRuntime if they are the same thing.

python/utils/hostruntime/hrxruntime/README.md
```diff
@@ -0,0 +1,434 @@
+<!---//===- README.md --------------------------*- Markdown -*-===//
```
hunhoffe, line 1: I'm very happy to see the documentation! We currently rely on two types of documentation:
* docstrings, which (for the most part) are rendered on the website https://xilinx.github.io/mlir-aie/dev/api/iron/
* markdown based docs served by the website: e.g., https://xilinx.github.io/mlir-aie/dev/programming_guide/iron_configuration/

In general, we have few embedded READMEs because they can be a bit hard to find for folks who install from wheels, etc. but do not clone source code. I noticed your docstrings don't include all the arg fields that the other docstrings do, and this README isn't available on the website (I don't think).

This could be a follow up, but I do think integrating HRX runtime documentation into the existing systems would be a very nice follow up.

python/utils/hostruntime/hrxruntime/tensor.py
```diff
@@ -0,0 +1,103 @@
...
+from aie.helpers.util import np_ndarray_type_get_shape
+from . import HRXContext
+
+
+class HRXTensor(Tensor):
+    """Tensor backed by an HRX persistent-mapped device buffer."""
+
+    def __init__(self, shape_or_data, dtype=np.uint32, device="npu", **kwargs):
+        super().__init__(shape_or_data, dtype=dtype, device=device)
+        self._ctx = HRXContext.get()
```
hunhoffe, line 52: Just to double check, does this work in a machine with multiple users/multiple contexts active an several python processes at once?

python/utils/__init__.py
```diff
@@ -35,7 +35,61 @@
...
+#   xrt   - force the XRT backend (error later if pyxrt is missing).
+#   hrx   - force the HRX backend (error here if libhrx is not found).
+#   auto  - (default) prefer XRT when present, else fall back to CPU.
+#
+# HRX is strictly opt-in: it is selected *only* when IRON_RUNTIME=hrx is set
+# explicitly. ``auto`` never selects HRX -- the product contract is "XRT remains
+# the default, HRX is opt-in", so an XRT-less host degrades to CPU rather than
+# silently switching to HRX. We import the selected backend lazily because the
+# HRX package dlopen()s libhrx on first use.
+_IRON_RUNTIME = os.environ.get("IRON_RUNTIME", "auto").lower()
```
hunhoffe: We should look for the IRON_RUNTIME before doing system probes; that way, if the runtime is hrx or xrt we don't unnecessarily probe for the other one.

python/utils/__init__.py
```diff
@@ -212,16 +266,28 @@ def set_tensor_class(cls):
...
+    """Release cached NPU runtime resources without initializing the runtime.
+
+    Works for both backends: ``CachedXRTRuntime`` releases hw contexts/insts
+    BOs and ``CachedHRXRuntime`` releases loaded XADX executables. If the
+    default runtime was never created, this is a no-op (it never forces
+    initialization).
+    """
     runtime = globals().get("DefaultNPURuntime", _DefaultNPURuntime)
-    if runtime is not None:
+    if runtime is not None and hasattr(runtime, "cleanup"):
```
hunhoffe: I'd like to avoid "hasattrs" -- I think it's generally better to force inheritance or a default implementation in the parent class.

python/utils/callabledesign.py
```diff
@@ -52,14 +52,18 @@ def _evict_xrt_context(xclbin_path: Path) -> None:
     """Evict a stale XRT hw_context after IOCTL EINVAL so the retry gets a fresh one."""
     from aie.utils import DefaultNPURuntime
 
-    if DefaultNPURuntime is None or not hasattr(DefaultNPURuntime, "_context_cache"):
+    # Only the XRT cached runtime keeps an evictable hw_context cache; under the
+    # HRX backend these attributes are absent, so this recovery is a no-op.
+    context_cache = getattr(DefaultNPURuntime, "_context_cache", None)
```
hunhoffe: We shouldn't need hasattr or getattr.

This breaks the inheritance/object oriented design of this library; if HRX is already cached, why can't the user manage it? Or, if it can't meet the inheritance contract to have these values, we need to refine the abstraction.

hunhoffe (commented):
This is great work -- really excited about this! I tried to design the python HostRuntime abstraction to easily allow a new runtime backend with minimal interference to users. Thank you for figuring out the hierarchy and trying to make the HRXRuntime fit in well! You are the first one to write a new HostRuntime backend, and I think you found a few minor snags.

I have some comments on structure, methodology, and style. In addition to the inline comments, I also want to notice that we do some fairly exhaustive tests for JIT and Tensors in test/python/npu-xrt. I'm thinking we either need to rename this test/python/npu, since really few of these tests really need XRT; they just need some runtime, and I'd like to see HRX test coverage at least as comprehensive as the current XRT test coverage.

Most of my comments can be addressed in follow ons, but please do consider them. We've worked hard to make a clean, tidy, and self-documenting Python frontend.

Please also make sure all dependencies (ctypes, etc.) are in requirements*.txt files as needed... I can't remember if ctypes, etc. are already documented dependencies.

python/utils/hostruntime/hrxruntime/README.md
jtuyls: Do we need to include the instructions to build from source or can we drop that for now? If not a hard requirement, I would leave it out for now.

jtuyls:
> What is the plan for `xclbinutil`? Could that be shipped with HRX? Or still included as a third party submodule here? If the latter, that can be included in a follow up PR. I'd like to discuss alignment with the upstream `xclbinutil` rather than a personal fork.

@jgmelber This is not really a fork, but a minimal component extracted and rewritten from XRT to avoid a dependency on it for creating the xclbins. I am not sure yet whether this can make its way into HRX as this is a bit of an odd component that we would ideally avoid altogether as xclbin is mostly just a necessary wrapper.. but dropping that won't work for current drivers (windows). Xclbinutul will have to be fairly stable though for backward compatibility so I don't think this will be hard to maintain and the cleanest way to avoid the XRT dep at this time is to bring it in as a submodule imo. In any case, it shouldn't stay in the personal repo and should really be a ROCm repo then if not in HRX, which I think would alleviate  some of your concerns?

python/utils/hostruntime/hrxruntime/README.md
Abhishek-Varma: Done - dropped the source build instructions for now. The README points at the pinned HRX release the fetch script consumes; we can add source-build docs later if there's demand.

python/utils/callabledesign.py
Abhishek-Varma: Done. evict_context is now a base-class no-op that HRX inherits, so the eviction path needs no hasattr/backend branching. Can you please check now?

python/utils/__init__.py
Abhishek-Varma: Done. Behavior now comes from a default implementation in the parent class instead of attribute probing.

python/utils/__init__.py
Abhishek-Varma: Done - IRON_RUNTIME is read before any probe now. Can you please take a look?

python/utils/hostruntime/hrxruntime/tensor.py
Abhishek-Varma: So I audited this - context/bindings use thread-safe lazy singletons, and each process gets its own HRX context. I have added a multiprocess test (test/python/npu-hrx/test_multiprocess_hrx.py) exercising concurrent dispatch case. Can you please take a look?

python/utils/hostruntime/hrxruntime/hostruntime.py
Abhishek-Varma: Done - split into HRXHostRuntime (uncached) and CachedHRXRuntime, mirroring the XRT cached/uncached split. They're no longer the same class. I've added tests for them as well.

python/utils/hostruntime/hrxruntime/hostruntime.py
Abhishek-Varma: Currently yes, the path is deliberately conservative - I can add direction-aware skipping as a follow-up optimization.

python/utils/hostruntime/hrxruntime/hostruntime.py
Abhishek-Varma: `run_chain` maps to the same producer→consumer ordering the XRT runlist provides, implemented via HRX dispatch + a single barrier synchronize. I agree a shared abstraction is worthwhile; I'd propose extracting it as a follow-up so we can align both backends deliberately.

python/utils/hostruntime/hrxruntime/hostruntime.py
Abhishek-Varma: Done. Can you please take a look?

python/utils/hostruntime/hrxruntime/__init__.py
Abhishek-Varma: Done - split into _bindings.py, context.py, and discovery.py. Can you please check now?

python/utils/hostruntime/hrxruntime/__init__.py
Abhishek-Varma: Done.

python/utils/hostruntime/hrxruntime/__init__.py
Abhishek-Varma: The ctypes signatures are portable. Only library discovery/loading is platform-specific which is handled. No code change needed here beyond that.

programming_examples/basic/vector_scalar_add/test_chain_hrx.py
Abhishek-Varma: Done.

programming_examples/basic/vector_scalar_add/test_chain_hrx.py
Abhishek-Varma: I've shifted it to test/python/npu-hrx for now as the test can't be made backend-neutral because there seems to be no XRT equivalent to exercise (I might be wrong but that's what I could infer so far - happy to adapt otherwise. :) )

third_party/.gitignore
Abhishek-Varma: Done. It lies under utils/ now. Does that look okay to you?

utils/fetch-hrx-release.sh
Abhishek-Varma: Yes. This has been tested on Windows. Added a fetch-hrx-release.ps1 for repro-ing it as well. Also this adds to CI a pure HRX build for Windows too.

hrx-integration/fetch-hrx-release.sh
Abhishek-Varma: Good catch! We now use just a single environment variable `NPU_RUNTIME`. Can you please check and let me know if it looks good to you?

python/utils/hostruntime/hrxruntime/README.md
Abhishek-Varma: Have addressed in this PR. Can you please take a look?

jgmelber:
> In any case, it shouldn't stay in the personal repo and should really be a ROCm repo then if not in HRX, which I think would alleviate some of your concerns?

Yes, if it is a major rewrite, rather than a patch that can be vendored here, it should be a ROCm repo. That would alleviate my concerns.

Abhishek-Varma:
>  I also want to notice that we do some fairly exhaustive tests for JIT and Tensors in test/python/npu-xrt. I'm thinking we either need to rename this test/python/npu, since really few of these tests really need XRT; they just need some runtime, and I'd like to see HRX test coverage at least as comprehensive as the current XRT test coverage.

Hi @hunhoffe - thank you for your comments. Regarding the one about runtime agonstic tests, I have made this change locally and have a test/python/npu now - is it okay to push in the same PR? Happy to adapt. :)

python/utils/hostruntime/hrxruntime/hostruntime.py
hunhoffe: Looks great, thank you for the documentation!

python/utils/hostruntime/hrxruntime/hostruntime.py
hunhoffe: I think there is even more support for runlist in amd/IRON repo (@andrej , is that true?) But yes, I think it makes sense as a follow on.

python/utils/hostruntime/hrxruntime/README.md
hunhoffe: Looks great, thank you!

utils/fetch-hrx-release.sh
hunhoffe: Awesome!

hunhoffe (approved):
@Abhishek-Varma Thank you for addressing my feedback so carefully! This looks great! I'll leave it up to you whether you include in this PR or as a follow up -- I'm fine with either path!

PR #3348, opened 2026-07-15
[ci] Fail loudly when OpenCV is absent

### LINKED ISSUE #2369: CI needs more OpenCV
Workflows appear to be skipping some tests requiring OpenCV.

related to https://github.com/Xilinx/mlir-aie/pull/2368

hunhoffe:
Vision tests have been silently skipped in CI because OpenCV was never checked for or passed into CMake, so the opencv lit feature was never registered. This PR surfaces that as a hard failure and fixes the wiring.

- Adds a pre-build "Check OpenCV availability" step that probes for opencv4 via pkg-config and fails with a clear error message if absent, instead of silently skipping vision tests
- Passes OpenCV_DIR into EXTRA_CMAKE_ARGS so find_package(OpenCV) succeeds at CMake configure time and the opencv feature is registered in the lit config

Closes #2369


andrej (approved):
Very good catch. Hopefully these examples didn't break in the meantime

PR #3351, opened 2026-07-15
[aie][aiex] back the per-BD dimension limit with AIETargetModel

### LINKED ISSUE #3349: [aiex] Back shim BD dimension limit with TargetModel instead of literal
Follow-up to #[PR number]. The deadlock fix there includes a small Python guard that rejects taps with more than 4 dimensions — currently the only thing catching this on the shim_dma_single_bd_task path.

The proper fix is a TargetModel-backed verifier:
1. Add a per-BD max-dimensions accessor to AIETargetModel — currently only getNumBDs (BD count) is exposed; the per-BD dimension limit is hardcoded as a literal in DMABDOp::verify (maxNDims = 3, or 4 under a MemTileDMAOp).
2. Have DMABDOp::verify query it by tile type — right now it infers the limit from the parent op type rather than from the tile's target model entry.
3. Extend the check to the runtime-sequence path — shim_dma_single_bd_task lowers through dma_configure_task_for and aie-dma-tasks-to-npu; the BD never lands under a ShimDMAOp, so DMABDOp::verify is skipped  entirely. The tile is recoverable through the shim DMA allocation symbol, but the verifier would also need to account for the 4-dim iteration-hoisting convention to avoid rejecting legitimate taps.

@atassis -- your contribution would be welcome here!

atassis:
Resolves #3349.

The maximum number of ND addressing dimensions a buffer descriptor may carry was hardcoded as a literal in `DMABDOp::verify` (`3`, or `4` under a `MemTileDMAOp`) and inferred from the parent op type. Separately, the runtime-sequence task path (`shim_dma_single_bd_task` -> `dma_configure_task_for` -> `aie-dma-tasks-to-npu`) had no BD-dimension check in the verifier at all -- the only guard was a Python-level check in `shim_dma_single_bd_task` (added in #3346).

## Changes

- Add `AIETargetModel::getBDMaxDims(tileType)` -- AIE2: `MemTile` 4, core/shim 3; AIE1: 3 -- mirroring `getNumBDs`, as the single source of truth for the per-BD ND limit.
- `DMABDOp::verify` now queries `getBDMaxDims` by tile type instead of the literal plus parent-op-type inference. This is behavior-preserving (an `aie.dma` on a memtile is always nested under a `MemTileDMAOp`, so the old and new checks are equivalent).
- Enforce the limit on the runtime-sequence path in the `DMAConfigureTaskOp` and `DMAConfigureTaskForOp` verifiers -- the latter recovers the tile through the shim DMA allocation symbol. A shim BD may carry one dimension beyond its ND limit because the leading tap dimension is hoisted into the hardware iteration/repeat register; this matches the uniform 4-dimension cap applied later by `aie-dma-tasks-to-npu` (`+1` for every tile except `MemTile`, whose ND width already spans the register).

## Tests

Cover the tile-type boundaries, the circuit-vs-task asymmetry (the same 4-dimension shim BD is rejected under `aie.shim_dma` but accepted under a task), and both the concrete-tile and alloc-symbol task paths, including graceful defer when the alloc symbol is unresolved.

## Notes

- The Python guard in `shim_dma_single_bd_task` is intentionally kept as a friendly frontend early-exit; the verifier is now the authoritative backstop for all paths (including hand-written MLIR).
- With the verifier catching over-dimension BDs at verification time, `AIEDMATasksToNPU`'s `"At most four data layout transformation dimensions"` guard is now unreachable for verify-clean IR (it remains as defense for `--no-verify`). `bad-2.mlir` was updated to expect the verifier diagnostic.


hunhoffe (approved):
Thank you for the follow up!

PR #3354, opened 2026-07-15
[python] iron.jit + XRTHostRuntime: full-ELF support

### LINKED ISSUE #3148: Enable full ELF for `iron.jit` and `XRTHostRuntime`
The former (`iron.jit`) currently [generates `insts.bin`s and `final.xclbin`s](https://github.com/Xilinx/mlir-aie/blob/d4f066008a75aacbcad4b7a4918ec8730280ec2b/python/utils/compile/utils.py#L109-L114).

The latter currently [ingests `insts.bin`s and `.xclbin`s](https://github.com/Xilinx/mlir-aie/blob/d4f066008a75aacbcad4b7a4918ec8730280ec2b/python/utils/hostruntime/xrtruntime/hostruntime.py#L183-L184).

Add an option for the former to instead emit only a single `.elf` and the latter to consume this and adapt the kernel call sites.

hunhoffe:
Makes the compiler's single self-contained **full ELF** (`aiecc --generate-full-elf`: PDIs + TXN control code) a first-class IRON path. Closes #3148.

Opt in three ways:
- `@iron.jit(full_elf=True)` — transparent compile + run
- `design.compile(full_elf_path="out.elf")` — AOT single-ELF emission

Also fixes a `specialize()` asymmetry surfaced by this work (see below).

## How full-ELF works
  
The full-ELF runtime differs from the xclbin+insts flow: the `hw_context` is built from `pyxrt.elf(...)`, the kernel is addressed by its `"<device>:<sequence>"` name, and args bind via `run.set_arg` + `run.start` + `run.wait2`.

Full-ELF designs must load their own PDI (no xclbin configures the device), so `Runtime.resolve` auto-injects `npu.load_pdi(@device)` at the start of the runtime sequence — gated by a compile-context flag, so the **default xclbin path is unchanged**. Trace works on the full-ELF path via the appended trace buffer.

## `specialize()`: config is now as retargetable as `CompileTime[T]`

Previously `specialize()` bound only `CompileTime[T]` kwargs and hand-copied a fixed list of config params, so it silently dropped any config not in that list — you could not flip an existing `@iron.jit` design onto the full-ELF path. Config was decoration-locked, while AOT `compile()` already honored `full_elf` per-call.
  
Now the config surface derives from the constructor signature via a shared `config_param_names()` helper (the single source of truth for `CompilableDesign`/`CallableDesign.specialize()` **and** `jit._JIT_CONFIG_KEYS`). `specialize()` splits overrides by name: config-named → replace that config; everything else → `CompileTime[T]` kwarg. New config params are honored automatically instead of being dropped.
  
```python
design.specialize(full_elf=True)   # re-aims an existing design at full-ELF,
                                   # symmetric with @iron.jit(full_elf=True)
```

Closes #3148.


python/utils/hostruntime/cli.py
```diff
@@ -165,10 +168,15 @@ def run_design_cli(
...
+    compile_only_requested = (
+        getattr(opts, "xclbin_path", None) is not None or full_elf_path is not None
+    )
     requested_dev = getattr(opts, "dev", None)
 
-    if compile_only_requested and not getattr(opts, "insts_path", None):
+    if getattr(opts, "xclbin_path", None) is not None and not getattr(
+        opts, "insts_path", None
+    ):
         sys.exit("--xclbin-path requires --insts-path (must be set together)")
```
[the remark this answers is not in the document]
andrej: Note that if the new `aiecc` works as it should, it should actually be possible to request both a full ELF and xclbin+insts in the same compilation invocation. Not sure why anyone would do this and it's untested though.

python/utils/compile/jit/compilabledesign.py
```diff
@@ -203,6 +265,10 @@ def compile(
         """
         from aie.iron.kernel import ExternalFunction
 
+        full_elf = self.full_elf or full_elf_path is not None
+        if full_elf:
+            return self._compile_full_elf(ExternalFunction, full_elf_path)
```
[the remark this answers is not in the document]
hunhoffe: (Claude) Good catch — fixed in 6fcfdf614cf. The root issue was that _generate_uncached keyed _iron_full_elf off self.full_elf, so calling compile(full_elf_path=...) on a design constructed without full_elf=True would emit MLIR without npu.load_pdi.

Fix: added full_elf to the generation cache key (so full-ELF and standard MLIR are cached separately), added a full_elf parameter to _generate_mlir/_generate_uncached/_generated_for, and _compile_full_elf now always passes full_elf=True regardless of self.full_elf.

python/utils/hostruntime/cli.py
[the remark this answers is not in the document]
hunhoffe: Fixed in 6fcfdf614cf — added a sys.exit guard that rejects --full-elf-path + --xclbin-path/--insts-path together. 

@andrej I've added a rejection for now since it's untested and the combined path isn't something we exercise yet. If we want to support it later (e.g. for debugging: dump the xclbin too), we can relax the guard then with an actual test.

atassis:
Was in the IRON 1.4 milestone and landed here via #3148. Possible gotcha on the resident case (might be specific to how I reuse buffers):

if an arg stays resident and gets re-authored in place between dispatches, the in-place write may not flip the parent host-dirty, so `[a.to("npu") for a in args]` no-ops and the dispatch runs on stale bytes. On aie2p that was nondeterministic output until I flipped residency on the write and re-synced output each dispatch.

It turned up in a per-token decode loop I've been running since mid-June (persistent `hw_context`, patched ELF rebound each token) from a Rust host: [rust/npu-xrt/src/lib.rs](https://github.com/atassis/xdna-engine/blob/main/rust/npu-xrt/src/lib.rs). Fix in amd/IRON#138 (andrej approved) if it's useful to you.

hunhoffe:
@atassis -- thank you for bringing this to my attention! I'm working to address it now in a separate PR.

python/utils/compile/jit/compilabledesign.py
```diff
@@ -352,6 +402,134 @@ def compile(
...
+                continue
+            device_sym = StringAttr(op.operation.attributes["sym_name"]).value
+            for inner in op.regions[0].blocks[0].operations:
+                if inner.operation.name == "aie.runtime_sequence":
+                    seq_sym = StringAttr(inner.operation.attributes["sym_name"]).value
+                    return f"{device_sym}:{seq_sym}"
+        raise RuntimeError(
+            f"Could not find an aie.device + aie.runtime_sequence in "
+            f"'{self.generator_name}' to derive the full-ELF kernel name."
+        )
```
andrej, line 531: Again I might be changing this in a future PR if we decide to add first-class `aiex.configure` support to IRON but for now this is likely sensible enough ...

python/iron/program.py
```diff
@@ -175,8 +176,14 @@ def device_body():
...
-                # In/Out Sequence
-                self._rt.resolve()
+                # In/Out Sequence.  On the full-ELF path the runtime sequence
+                # must load its own PDI (no xclbin configures the device), so
+                # pass the device symbol as the load_pdi reference.  The flag is
+                # injected into the compile context by CompilableDesign.
+                load_pdi_device_ref = (
+                    device_name if get_compile_arg("_iron_full_elf") else None
+                )
+                self._rt.resolve(load_pdi_device_ref=load_pdi_device_ref)
```
andrej, line 186: I might need to change this in a future PR if I add an example that uses `aiex.configure` from IRON, since that lowers to its own load_pdis, and there might then be multiple load_pdis in the sequence as we switch between kernels. But this is sensible for the single-kernel workload situation.

python/utils/hostruntime/xrtruntime/hostruntime.py
```diff
@@ -609,6 +698,88 @@ def _drain_for_phoenix_trace_quirk(self):
...
         gc.collect()
 
+    def _load_full_elf_cached(
+        self, npu_kernel, retry: bool = True
+    ) -> CachedXRTKernelHandle:
+        """Cached full-ELF load: reuse a hw_context keyed on ``(elf_path, mtime)``.
+
+        Full ELFs carry their own PDIs + control code, so the context is built
+        from ``pyxrt.elf`` (no xclbin, no instruction BO cache).  Context reuse
+        follows the same LRU + Phoenix-drain policy as the xclbin path.
```
andrej, line 708: What is the Phoenix-drain policy?

Note that Phoenix does not support full ELF.

andrej (approved):
Thanks for tackling this!!!

python/utils/hostruntime/xrtruntime/hostruntime.py
hunhoffe: Sometimes you have to fully drain (evict all contexts) to get phoenix to realize it can make new ones, IIRC. 

Yes, I do remember that -- but I figured I shouldn't unnecessarily put blockers if that there to change someday.

hunhoffe:
@andrej Thanks for the review! I'm very open to changes to this in the future!

PR #3356, opened 2026-07-16
[iron] Expose PDI artifact + demonstrate artifact-path control and BYO artifacts

hunhoffe:
## Summary
  
`@iron.jit` users had no way to obtain a PDI or run externally-built artifacts through the public API. This PR closes both gaps and documents the full artifact lifecycle. 
  
**Library**
- Thread `pdi_path` through `CompilableDesign.compile()`, `CallableDesign.compile()`, and the CLI helpers (`add_compile_args`/`run_design_cli`), mirroring the existing `elf_path` plumbing.
- Add `get_pdi_path()` so cache-mode users can locate the `main.pdi` that `aiecc` always emits into the cache directory.
    
**Example** (`basic/vector_scalar_add`)
- `--aot-dir` exports named `xclbin` / `insts` / `PDI` / `ELF` to a chosen directory (bypasses the cache).
- `--from-xclbin` / `--from-insts` runs a pre-built pair via `NPUKernel` — origin-agnostic, so artifacts from any toolchain work. 
    
**Docs** — `compilation_stages.md` and the example README gain "Controlling artifact output" and "Consuming external artifacts" sections.
  
## Notes
- The IRON runtime loads via `pyxrt.xclbin(...)`; a PDI is packed *inside* the xclbin and is not a standalone run input. `pdi_path`/`get_pdi_path()` are for *extracting* a PDI for external consumers, not feeding one back in.
- BYO run path has no `CompileTime[T]` recipe or tensor-arg validation (that lives in the JIT `CompilableDesign`), so the caller owns matching tensor shapes/dtypes.


programming_examples/basic/vector_scalar_add/vector_scalar_add.py
```diff
@@ -68,6 +94,69 @@ def _compile_kwargs(opts):
...
+
+    ``NPUKernel`` loads any xclbin + instruction-binary pair — it does not care
+    whether they came from this design's ``--aot-dir`` export, a Makefile, a
+    raw ``aiecc`` invocation, or another tool.  Inputs are passed as IRON
+    tensors; the output tensor is pre-allocated and written in place, so its
+    element count must match what the artifacts were compiled for
+    (``--problem-size``).
+    """
+    npu_kernel = NPUKernel(opts.from_xclbin, opts.from_insts)
+
```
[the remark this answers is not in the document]
hunhoffe: I think what we need is to fetch the expected device, and then make sure it's valid/matching in the runtime.

programming_examples/basic/vector_scalar_add/README.md
```diff
@@ -64,4 +64,74 @@ To run the runlist testbench (NPU2 only):
...
+
+Inputs are passed as IRON tensors; the output tensor is pre-allocated and
+written in place, so `--problem-size` must match the size the artifacts were
+compiled for. Note the run inputs are the **xclbin + insts** — the PDI is
+packed *inside* the xclbin and is not a standalone run input on the IRON
+runtime.
+
+Because this path bypasses JIT generation, there is no compiled recipe to
+validate the artifacts against. An xclbin built for the wrong NPU family runs
+without error but silently returns zeros, so pass `--dev` to declare the
```
andrej: That's not true in my experience, trying to run a Phoenix xclbin on Strix I typically get timeouts.

python/utils/compile/jit/compilabledesign.py
```diff
@@ -339,6 +358,22 @@ def get_artifacts(self) -> tuple[Path, Path] | None:
             return None
         return self._xclbin_path, self._inst_path
 
+    def get_pdi_path(self) -> Path | None:
+        """Return ``<kernel_dir>/main.pdi`` if it exists, else ``None``.
+
+        In default cache mode, aiecc emits a ``main.pdi`` into the cache
+        directory alongside the xclbin, so this locates that PDI without a
+        recompile.  It is *not* a general "PDI for the most recent compile"
```
andrej: As far as I know one PDI is generated for each aie.device. main happens to be the default device name, but maybe worth checking how this interacts / should interact with custom other device names or if there are multiple devices in the input.

andrej (approved):
Please double-check about the multi-device/alternate device name question, other than that looks good!

python/utils/compile/jit/compilabledesign.py
hunhoffe: I fixed multi-device support and added an end-to-end test. Nice catch!

programming_examples/basic/vector_scalar_add/README.md
hunhoffe: Fixed.

PR #3358, opened 2026-07-16
[dyn-seq] Dynamic BD allocator: runtime bd_id via a free-list pool for rolled runtime sequences

hunhoffe:
Part of discussion #3222. Stacks on #3293 (the dynamic BD-word encoder). Delivers the **runtime-bound `scf.for`** portion of milestone **item 8** (dynamic control flow) plus **item 9**'s dynamic passthrough hardware test.

> Note on naming: the `[dyn-seq P6]` commit prefix is an internal landing-sequence label, **not** discussion PR #6 (runtime `repeat_count`, which folded into #3293 as its "may fold into PR 5" note anticipated). This PR is the dynamic BD allocator.

 #3293 made a shim-NOC DMA's sizes/strides/offset/repeat_count runtime-valued, but the **bd_id** was still a compile-time attribute, so a runtime-bound `scf.for` in a runtime sequence had to be fully unrolled. This PR makes the bd_id runtime: the loop stays **rolled**, and each iteration draws a bd_id from a per-tile runtime **free-list pool** — so one compiled binary serves a runtime trip count.

## What's in it

- **Pool ops** `aiex.dma_bd_pool_pop` (→ `i32 bd_id`) / `aiex.dma_bd_pool_push`, and a header-only `BdPool` free-list in `TxnEncoding.h`.
- **Pass `aie-lower-dynamic-bd-pool`**: rewrites a runtime-bound sequence's configure/free/await to pool pop/push and carries the popped id through the `scf.for` iter-args. Runs only when a runtime `scf` survives unroll; straight-line and constant-bound sequences keep the static allocator, byte-identical.
- **Runtime bd_id through the lowering chain**: the shim BD-word encoder and the queue-push command word accept an SSA bd_id (the register address is linear in bd_id, so it lowers to arith over the pool result).
- **End-to-end EmitC path**: a rolled dynamic loop emits as `emitc.for`, and the C++ TXN target emits the pool + a runtime op-count, so the host builds the instruction stream at runtime (`generate_txn_*` returns `std::optional`, `std::nullopt` on pool exhaustion).
- **aiecc routing**: the pass is wired into the default NPU lowering pipeline.

## Scope / limits

- Shim-NOC only; **single-BD tasks** in the dynamic-pool path. Multi-BD chains under a runtime loop need runtime `next_bd` (a P5-level encoder change) and are diagnosed, not mis-lowered — a follow-up.
- `scf.if` (runtime condition) + arbitrary nesting are NOT in this PR — that's the rest of milestone item 8, tracked as https://github.com/Xilinx/mlir-aie/pull/3371. This delivers the `scf.for` (runtime-bound loop) half.
- Peak-liveness can't be checked at compile time for general runtime CF; the backstop is `pop` returning false → builder returns `std::nullopt`. It's an open TODO to improve error handling/error messages.

## Testing
  
- Unit: `BdPool` free-list exhaustion.
- Pass lit: `aie-lower-dynamic-bd-pool` (mixed routing, ping-pong, straight-line untouched, pinned-in-pool + multi-BD rejects).
- Equivalence: `static_vs_dynamic` register-replay — static path byte-identical, rolled-loop / pool-bdid dynamic path matches.
- **Hardware** (`test/npu-xrt/dynamic_pingpong_passthrough`, item 9 passthrough): runtime tile count on real npu2, n = 1, 2, 8. Verified PASS on Strix.

Milestone: https://github.com/Xilinx/mlir-aie/discussions/3222

### DIFF
--- a/include/aie/Dialect/AIEX/Utils/BdLowering.h
+++ b/include/aie/Dialect/AIEX/Utils/BdLowering.h
@@ -221,6 +221,18 @@ struct SsaStridePolicy {
 mlir::Value getAsValue(mlir::OpBuilder &builder, mlir::Location loc,
                        mlir::OpFoldResult ofr, mlir::Type intType);
 
+// Build the address-patch `arg_plus` (buffer BYTE offset) as an i32 Value.
+// `elementOffsets` (innermost-first) are the per-dim element offsets and
+// `strides` their matching strides; the byte offset is
+// sum(elementOffsets[i] * strides[i]) * elemWidthBytes + baseByteOffset.
+// Any entry may be a runtime Value; a fully-constant set folds to a single
+// arith.constant. Shared by the memcpy_nd and dma_task dynamic lowerings so a
+// runtime DMA offset flows into the patch instead of being rejected.
+mlir::Value buildArgPlusValue(mlir::OpBuilder &builder, mlir::Location loc,
+                              llvm::ArrayRef<mlir::OpFoldResult> elementOffsets,
+                              llvm::ArrayRef<mlir::OpFoldResult> strides,
+                              int64_t elemWidthBytes, int64_t baseByteOffset);
+
 // Pack a set of (value, mask, shift) fields into a single i32 BD word via
 // arith and/shl/or. mask == 0xFFFFFFFF skips the AND; shift == 0 skips the SHL.
 mlir::Value
--- a/include/aie/Dialect/AIEX/Utils/BdLowering.h
+++ b/include/aie/Dialect/AIEX/Utils/BdLowering.h
@@ -221,13 +221,9 @@ struct SsaStridePolicy {
 mlir::Value getAsValue(mlir::OpBuilder &builder, mlir::Location loc,
                        mlir::OpFoldResult ofr, mlir::Type intType);
 
-// Build the address-patch `arg_plus` (buffer BYTE offset) as an i32 Value.
-// `elementOffsets` (innermost-first) are the per-dim element offsets and
-// `strides` their matching strides; the byte offset is
-// sum(elementOffsets[i] * strides[i]) * elemWidthBytes + baseByteOffset.
-// Any entry may be a runtime Value; a fully-constant set folds to a single
-// arith.constant. Shared by the memcpy_nd and dma_task dynamic lowerings so a
-// runtime DMA offset flows into the patch instead of being rejected.
+// Build the address-patch `arg_plus` (buffer BYTE offset) as an i32 Value:
+// sum(elementOffsets[i] * strides[i]) * elemWidthBytes + baseByteOffset, where
+// any entry may be runtime (a fully-constant set folds to one arith.constant).
 mlir::Value buildArgPlusValue(mlir::OpBuilder &builder, mlir::Location loc,
                               llvm::ArrayRef<mlir::OpFoldResult> elementOffsets,
                               llvm::ArrayRef<mlir::OpFoldResult> strides,
--- a/include/aie/Dialect/AIEX/Utils/BdLowering.h
+++ b/include/aie/Dialect/AIEX/Utils/BdLowering.h
@@ -239,18 +239,36 @@ mlir::Value
 buildBdWord(mlir::OpBuilder &builder, mlir::Location loc,
             llvm::ArrayRef<std::tuple<mlir::Value, uint32_t, uint32_t>> fields);
 
-// Emit the per-word `npu.write32` overrides carrying a shim-NOC BD's runtime
-// sizes/strides on top of a zero-template blockwrite, plus assert_bd_field /
-// assert_bd_divisible guards for runtime values in narrow or sub-granule
-// fields. Shared by the dma_memcpy_nd and dma_task paths.
-// `mixedSizes`/`mixedStrides` are outermost-first (d3..d0). `bufLenOverride`,
-// if non-null, sets buffer_length (dma_task's runtime len); else it is the
-// d0*d1*d2 size-product. `repeatCountOut` receives the biased hardware
-// outer-dim (d3) value for the caller's queue push.
+// Emit the per-word `npu.write32` overrides that carry a shim-NOC BD's runtime
+// sizes/strides, on top of a zero-template blockwrite whose size/stride words
+// were left at 0. Shared by the dma_memcpy_nd and dma_task dynamic lowering
+// paths, which converge on the identical shim BD-word layout (words 0/3/4/5/6);
+// only the descriptor template (locks, next_bd, packet) and the queue push
+// differ, and those stay with each caller.
+//
+// `mixedSizes`/`mixedStrides` are outermost-first (d3..d0), matching
+// NpuDmaMemcpyNdOp::getMixedSizes and AIE::DMABDOp::getMixedSizes.
+// `bufLenOverride`, if non-null, is written verbatim into buffer_length
+// (word 0) -- dma_task passes its runtime `len`; dma_memcpy_nd passes null, so
+// buffer_length is computed as the d0*d1*d2 hardware-unit size-product. Emits
+// `npu.assert_bd_field` guards for runtime values landing in narrow fields
+// (d0/d1 wrap 10-bit in ND mode, iteration wrap 6-bit always). The op verifier
+// is expected to have enforced the supported scope (shim NOC, innermost stride
+// == 1) already.
+//
+// On success `repeatCountOut` receives the hardware iteration/repeat value for
+// the outer (d3) dimension, which the caller feeds to its queue push (this is
+// the biased hw value, matching the static path's `repeat_count = sizes[3]`).
+//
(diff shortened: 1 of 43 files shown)


lib/Dialect/AIEX/Transforms/AIELowerDynamicBDPool.cpp
```diff
@@ -0,0 +1,269 @@
...
+      seq.walk([&](Operation *op) {
+        if (isa<scf::ForOp, scf::IfOp, scf::WhileOp>(op))
+          hasRuntimeCF = true;
+      });
+      if (!hasRuntimeCF)
+        return WalkResult::advance();
+
+      pairedId.clear();
+      tileForTask.clear();
+
```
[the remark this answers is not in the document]
hunhoffe: Fixed in 037c20599da — originConfigure is now cleared per-sequence alongside pairedId/tileForTask. Agreed the Value-keyed stale-entry risk (and the unbounded growth) warranted it. Tests still green.

include/aie/Dialect/AIEX/Utils/BdLowering.h
```diff
@@ -221,24 +221,65 @@ struct SsaStridePolicy {
 mlir::Value getAsValue(mlir::OpBuilder &builder, mlir::Location loc,
                        mlir::OpFoldResult ofr, mlir::Type intType);
 
+// Build the address-patch `arg_plus` (buffer BYTE offset) as an i32 Value.
```
hunhoffe: Comments are too long in this file.

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -183,28 +183,39 @@ struct PushQueuetoWrite32Pattern : OpConversionPattern<NpuPushQueueOp> {
     // the offset of the task queue register in the tile
     uint32_t queue_offset = ctrl_offset + 0x4;
 
-    // Command word: bd_id [3:0], repeat_count [23:16], issue-token bit [31].
-    // bd_id is always compile-time; repeat_count may be a runtime SSA value, in
-    // which case the word is built with arith over a constant base.
+    // The command word packs bd_id [3:0], repeat_count [23:16], and the
```
hunhoffe: Comment too long!

lib/Dialect/AIEX/Transforms/AIELowerDynamicBDPool.cpp
```diff
@@ -0,0 +1,256 @@
...
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// The dynamic counterpart to AIEAssignRuntimeSequenceBDIDs: where that pass
+// rejects a runtime-bound scf.for, this keeps the loop rolled and draws bd_ids
+// from a per-tile runtime free-list pool. Each configure gets a dma_bd_pool_pop
+// (its SSA id feeding bd_id_val); each free/await gets a dma_bd_pool_push.
```
hunhoffe, line 11: Does this handle when something is awaited and freed?

tools/aiecc/IRTransforms.h
```diff
@@ -711,6 +711,14 @@ getNpuLoweringPipeline(mlir::MLIRContext *ctx, bool materialize = true) {
   dpm.addPass(X::createAIESubstituteShimDMAAllocationsPass());
   dpm.addPass(X::createAIEUnrollRuntimeSequenceLoopsPass());
   dpm.addPass(mlir::createCanonicalizerPass());
+  // Any runtime-bound scf.for still present here could not be unrolled, so it
```
hunhoffe: Comment too long.

hunhoffe (commented):
For Claude

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -86,103 +89,131 @@ struct DeviceResolved {
...
+      op->erase();
+    SmallVector<Operation *> deadSweep;
+    funcOp.walk<WalkOrder::PostOrder>([&](Operation *op) {
+      if (isa<memref::MemRefDialect, arith::ArithDialect>(op->getDialect()) &&
+          op->use_empty())
+        deadSweep.push_back(op);
+    });
+    for (Operation *op : deadSweep)
+      op->erase();
+    return count;
```
andrej: I think just adding a standard `mlir::createCanonicalizerPass()` would also do this?

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -86,103 +89,131 @@ struct DeviceResolved {
...
+      // Structural emitc ops and arith (constant / runtime-value math) are left
+      // for the later arith-to-emitc step. Recurse into scf ops (and any other
+      // region-carrying control flow) to convert their bodies; the control-flow
+      // op itself is handled by convert-scf-to-emitc afterwards.
       if (isa<emitc::EmitCDialect, arith::ArithDialect>(op->getDialect()))
         continue;
+      if (isa<scf::SCFDialect>(op->getDialect())) {
+        for (Region &r : op->getRegions())
+          for (Block &b : r)
+            convertBlockRecursive(b, count, consumed);
```
andrej, line 141: Looks like this works for any `scf` ops, but the PR description singles out `scf.for` and mentions `scf.if` is not supported. Where are the limitations?

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -210,6 +241,28 @@ class AIEXToEmitCConverter {
...
+          // Draw a BD id from the tile's runtime pool (declared in the
+          // prologue) into a fresh C++ variable; nullopt if the pool is empty.
+          // Downstream ops consume the id, so replace the SSA result with a
+          // literal naming the variable.
+          std::string var = "bd_" + std::to_string(nextPoolVar++);
+          emitc::VerbatimOp::create(
+              b, loc,
+              "uint32_t " + var + "; if (!aie_runtime::bd_pool_pop(" +
+                  bdPoolName(pop.getColumn(), pop.getRow()) + ", " + var +
+                  ")) return std::nullopt;");
```
andrej, line 250: Is there `std::nullopt` handling where the BD ID flows into the user call site? I wasn't able to immediately find the use site so thought it's quicker to ask. Just curious what happens when the allocator runs out at runtime (C++ exception, exit(), ...?)

andrej (approved):
This approach makes sense to me. Just clarifying questions and one optional suggestion about using a canonicalizer pass rather than manual removal of dead ops.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
hunhoffe: Ah, good point. I switched to `mlir::runRegionDCE` on the `emitc.func` body (`cd0209fe70d`), which is the canonical region-scoped version of what the manual sweep was doing. A plain module-level `createCanonicalizerPass()` doesn't work here: it also drops the module-level `memref.global` that the blockwrite data is inlined from, and other sequences in the same module still reference
  it (breaks the `static_vs_dynamic` register-replay tests). `runRegionDCE` stays inside the func and leaves module symbols alone.

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
hunhoffe: The converter here intentionally recurses through any `scf` region, so a self-contained `scf.if` (task configured and freed within one branch) already lowers correctly. The limitation isn't in this converter; it's in `aie-lower-dynamic-bd-pool`, which only threads pool ids through `scf.for` iter-args. So a cross-boundary `scf.if` (configured in a branch, freed after) or arbitrary nesting is diagnosed, not mis-lowered. Full `scf.if` support (`carryIdsThroughIf`, `scf.while` rejection, nesting) lands in WIP: https://github.com/Xilinx/mlir-aie/pull/3371

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
hunhoffe: No exception or exit() — it's handled at the call site before any hardware dispatch. `generate_txn_main_sequence` returns `std::optional<std::vector<uint32_t>>;` on pool exhaustion the pop
  returns `false` and the builder does `return std::nullopt`. The host checks it, e.g. in `dynamic_pingpong_passthrough/test.cpp`:
```cpp
      auto instr_opt = generate_txn_main_sequence(n);
      if (!instr_opt) { /* "exceeds BD pool" */ return 1; }
```

So a failed allocation means the instruction stream is never built and never sent to the NPU — it's a runtime build failure the caller inspects, not a device-side fault. (Improving the message/ergonomics is a tracked TODO.)

PR #3362, opened 2026-07-17
Parallel ExecutionEngine for `aiecc`

andrej:
The current implementation parallelizes items _within_ one edge (e.g. all the `object_{0}.o` per-core files can compile in parallel). This generalizes this and allows execution of entire (edge, item) tuples in parallel if they are independent of each other. For example, this allows `elfs_0.elf` already starting to compile while `object_99.o` is still compiling.

### DIFF
--- a/tools/aiecc/ExecutionEngine.h
+++ b/tools/aiecc/ExecutionEngine.h
@@ -205,13 +205,14 @@ struct Engine {
 
   // Fine-grained (edge, key) task scheduler for the reachable sub-graph.
   //
-  // A fan-out edge (Map/BundleForEach) is `open()`ed to create its per-key
+  // A fan-out edge (Map/BundleForEach) is `prepare()`d to create its per-key
   // output slots and then runs one task per key ("item"); a structural edge
   // (split / join / filter / rekey / ...) runs as a single whole-edge
   // `execute()`. A task becomes runnable as soon as the specific inputs it
-  // consumes are ready -- per key for fan-out edges -- so a core's downstream
-  // work (e.g. linking its ELF) can begin while other cores are still
-  // compiling, rather than waiting on a whole-edge barrier.
+  // consumes are ready -- per key for fan-out edges -- so an item's downstream
+  // work (e.g. a later stage of that item's pipeline) can begin while other
+  // items are still being produced, rather than waiting on a whole-edge
+  // barrier.
   //
   // A task is "exclusive" when its edge is not marked threadSafe: such edges
   // may touch shared, non-thread-safe state (today, the shared MLIRContext), so
@@ -348,7 +349,7 @@ struct Engine {
     // fan-outs finalize immediately. Called under `mtx`.
     void openEdge(unsigned u) {
       EdgeState &s = st[u];
-      if (mlir::failed(s.edge->open())) {
+      if (mlir::failed(s.edge->prepare())) {
         failed = true;
         failedEdge = s.edge;
         return;
@@ -367,9 +368,10 @@ struct Engine {
       }
     }
 
-    // Single-line progress: "(done/total) [i/n] edge_1, [j/m] edge_2, ...".
-    // The list is the edges with an in-flight task; i/n is that edge's per-item
-    // fan-out progress (0/1 for whole-edge tasks). Called under `mtx`.
+    // Single-line progress: "(done/total) [i/n] edge_1, edge_2, ...". The list
+    // is the edges with an in-flight task; fan-out edges carry an [i/n]
+    // per-item progress prefix, whole-edge edges just their name. Called under
+    // `mtx`.
     void renderProgress() {
       if (!opts.progress)
         return;
@@ -379,12 +381,12 @@ struct Engine {
       for (unsigned u = 0; u < numEdges; ++u) {
         if (st[u].running == 0)
           continue;
-        size_t total = st[u].fan ? st[u].itemsTotal : 1;
-        size_t done = st[u].fan ? st[u].itemsDone : 0;
         line += (first ? " " : ", ");
         first = false;
-        line += "[" + std::to_string(done) + "/" + std::to_string(total) +
-                "] " + displayName(st[u].edge).str();
+        if (st[u].fan)
+          line += "[" + std::to_string(st[u].itemsDone) + "/" +
+                  std::to_string(st[u].itemsTotal) + "] ";
+        line += displayName(st[u].edge).str();
       }
       size_t pad =
           progressPrevLen > line.size() ? progressPrevLen - line.size() : 0;
@@ -527,7 +529,7 @@ struct Engine {
         // Run the task without the scheduler lock.
         mlir::LogicalResult r = task.item < 0
                                     ? s.edge->execute()
-                                    : s.edge->runItem((size_t)task.item);
+                                    : s.edge->executeForItem((size_t)task.item);
 
         lock.lock();
         --s.running;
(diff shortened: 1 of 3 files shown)


tools/aiecc/Graph.h
```diff
@@ -533,6 +483,19 @@ struct EdgeBase {
   virtual NodeBase *outputNode() = 0;
   virtual mlir::LogicalResult execute() = 0;
 
+  // Fan-out interface. Fan-out edges (Map/BundleForEach) produce one output
+  // item per key and are executed by the scheduler as independent (edge, key)
+  // tasks: `open()` creates all output slots (making the key set final) without
+  // running any per-key work, and `runItem(i)` runs the action for slot `i`.
+  // Structural edges leave isFanOut() false and run as a single `execute()`.
+  // `execute()` remains valid on fan-out edges (open + sequential runItem) for
+  // non-scheduled callers.
```
andrej: What do you mean by `non-scheduled callers`? Is `execute` still used at all? If not, purge it.

tools/aiecc/Graph.h
```diff
@@ -533,6 +483,19 @@ struct EdgeBase {
   virtual NodeBase *outputNode() = 0;
   virtual mlir::LogicalResult execute() = 0;
 
+  // Fan-out interface. Fan-out edges (Map/BundleForEach) produce one output
+  // item per key and are executed by the scheduler as independent (edge, key)
+  // tasks: `open()` creates all output slots (making the key set final) without
+  // running any per-key work, and `runItem(i)` runs the action for slot `i`.
```
andrej: To be symmetric with `execute`, maybe call this `executeForItem`

tools/aiecc/Graph.h
```diff
@@ -775,24 +738,40 @@ struct MapEdge : Edge<In, Out> {
...
-  mlir::LogicalResult execute() override {
-    // Prepare all output slots sequentially (cheap: renders paths and creates
-    // parent dirs), then apply `fn` per item. When this edge is thread-safe
-    // (a context-free external tool), the per-item applications run in
-    // parallel.
+  bool isFanOut() const override { return true; }
+
+  // Prepare all output slots (renders paths, creates parent dirs) without
+  // running any per-item work; the scheduler runs the items as separate tasks.
+  mlir::LogicalResult open() override {
```
andrej: I think I'd prefer the name `prepare`

tools/aiecc/Graph.h
```diff
@@ -775,24 +738,40 @@ struct MapEdge : Edge<In, Out> {
...
+    return mlir::success();
+  }
+
+  mlir::LogicalResult execute() override {
+    if (mlir::failed(open()))
+      return mlir::failure();
+    for (size_t i = 0, n = this->out.items.size(); i < n; ++i)
+      if (mlir::failed(runItem(i)))
+        return mlir::failure();
+    return mlir::success();
```
andrej: Do we really need this specialized `execute` method, or could this be implemented generically using `numItems()` and `runItem()`? I think by doing that we can share code with BundleForEachEdge. That could just be the default implementation in the base class that other implementations then override.

tools/aiecc/ExecutionEngine.h
```diff
@@ -267,114 +713,22 @@ struct Engine {
...
+        reachableEdges.push_back(e.get());
+
+    Scheduler scheduler(opts, std::move(reachableEdges), satisfied, deserCtx);
+    mlir::LogicalResult r = scheduler.run();
+    failedEdge = scheduler.failedEdge;
+    if (mlir::failed(r))
+      return mlir::failure();
+    auto &edgeTimings = scheduler.edgeTimings;
+
+    // Force declared outputs (and, under --keep-intermediates, every reachable
```
andrej: I think `--keep-intermediates` is actually called `--dump-intermediates`?

tools/aiecc/ExecutionEngine.h
```diff
@@ -201,6 +203,450 @@ struct Engine {
...
 
+  // Fine-grained (edge, key) task scheduler for the reachable sub-graph.
+  //
+  // A fan-out edge (Map/BundleForEach) is `open()`ed to create its per-key
+  // output slots and then runs one task per key ("item"); a structural edge
+  // (split / join / filter / rekey / ...) runs as a single whole-edge
+  // `execute()`. A task becomes runnable as soon as the specific inputs it
+  // consumes are ready -- per key for fan-out edges -- so a core's downstream
+  // work (e.g. linking its ELF) can begin while other cores are still
+  // compiling, rather than waiting on a whole-edge barrier.
```
andrej: Reword to not be specific to cores. This is for an item.

tools/aiecc/ExecutionEngine.h
```diff
@@ -201,6 +203,450 @@ struct Engine {
...
+      bool first = true;
+      for (unsigned u = 0; u < numEdges; ++u) {
+        if (st[u].running == 0)
+          continue;
+        size_t total = st[u].fan ? st[u].itemsTotal : 1;
+        size_t done = st[u].fan ? st[u].itemsDone : 0;
+        line += (first ? " " : ", ");
+        first = false;
+        line += "[" + std::to_string(done) + "/" + std::to_string(total) +
+                "] " + displayName(st[u].edge).str();
```
andrej: Edges that don't fan out should not print a `[0/1]`. Just leave out the square bracket part for them.

tools/aiecc/Graph.h
andrej: (Claude) Purged the fan-out `execute()` overrides. `execute()` is now a single generic default in `EdgeBase` (`prepare()` + a sequential `executeForItem()` loop) that fan-out edges use directly, and structural edges still override for their whole-edge action. Removed the confusing "non-scheduled callers" wording.

tools/aiecc/Graph.h
andrej: (Claude) Renamed `runItem` to `executeForItem` throughout (base declaration, MapEdge, BundleForEachEdge, scheduler dispatch).

tools/aiecc/Graph.h
andrej: (Claude) Renamed `open()` to `prepare()` throughout (base declaration, MapEdge, BundleForEachEdge overrides, and the scheduler call site).

tools/aiecc/Graph.h
andrej: (Claude) Done. Removed the specialized `execute()` from both MapEdge and BundleForEachEdge; `execute()` is now a single generic default in `EdgeBase` implemented via `prepare()` + a `numItems()`/`executeForItem()` loop, shared by both fan-out edges.

tools/aiecc/ExecutionEngine.h
andrej: (Claude) Correct, the flag is `--dump-intermediates`. Fixed both comments that referenced `--keep-intermediates` (the internal field stays `keepIntermediates`).

tools/aiecc/ExecutionEngine.h
andrej: (Claude) Reworded to be item-generic: a runnable item's downstream work (a later stage of that item's pipeline) can begin while other items are still being produced, with no mention of cores/ELFs.

tools/aiecc/ExecutionEngine.h
andrej: (Claude) Fixed. `renderProgress()` now only emits the `[i/n]` prefix for fan-out edges; whole-edge (non-fan-out) edges print just their name. Updated the accompanying comment too.

hunhoffe (approved):
Do we need to test for this functionality explicitly? e.g., a regression test? Otherwise, looks good!

andrej:
I think the existing tests should cover this. It's just a different way of iterating over the graph, so if the produced artifacts are correct (which existing tests cover), I think that means it did it right. If you think a test for the internals of this would be better, let's add it as a follow-on if that's okay.

PR #3363, opened 2026-07-17
`aiecc`: Port `aiesim` flow to declarative style

andrej:
The `aiesim` flow was left as a `TODO` in #3271. This does that item and reimplements the `aiesim` as a proper member of the declarative compilation graph, rather than using the legacy code and treating it as a black box.

Here is what the full graph for aiesim + host compilation looks like, created using the `aiecc --emit-dot` option:
image


hunhoffe (approved):
Looks good! Do we need any additional tests to make sure the aiesim flow plays well with the new options?

jgmelber (approved):

jgmelber:
The original idea of breaking these out into a separate file was to keep the size of `aiecc.cpp` manageable. But this change only adds ~150 LOC to the file.

PR #3368, opened 2026-07-20
[dyn-seq] Dynamic whole-array GEMM: runtime M/K/N, multi-column

hunhoffe:
Milestone #3222 **item 9's dynamic GEMM** (the passthrough half is in #3358). Stacks on #3358 (dynamic BD allocator). A whole-array i16 GEMM whose host runtime sequence uses runtime M/K/N, so one xclbin serves many shapes.

- The row-block loop is a runtime-bound `scf.for` that survives to the EmitC C++ TXN builder (via #3358's dynamic BD pool); the host calls `generate_txn_main_sequence(M,K,N)` per shape.
- **Multi-column** (`n_aie_cols ∈ {1,2,4}`): the columns band the output N dimension. Each column has its own shim FIFOs and issues its own DMAs + await (N separate 1×1 syncs — functionally correct, uncollapsed). The column loop is a compile-time unroll; only M/K/N are runtime. col 0's N-offset folds to a literal 0, so single-column BDs stay byte-identical.
- One `seq_body` serves both lowerings: static (`dynamic=False`, M/K/N fold to constants → unrolled binary path, runnable via `@iron.jit`) and dynamic.

No compiler change: the BD pool and await→sync are already per-tile.

## Testing

- **npu2 / Strix** (`run.lit`): sweeps cols=1/2/4, verified on hardware, 0 errors (512³, 256×512×1024, 768×512×768).
- **npu1 / Phoenix** (`run_npu1.lit`): aie2-kernel sibling, authored + build-verified; awaits a Phoenix runner for HW confirmation.
- `test.cpp` is device/column-agnostic; each config's MLIR is generated at test time via `--emit-dynamic-mlir`.

Milestone: https://github.com/Xilinx/mlir-aie/discussions/3222


atassis:
I reviewed this against my own engine work. I run an engine on XDNA2 that already serves many
shapes from one pre-built kernel by patching runtime values into the lowered control code at
dispatch: an RTP the kernel reads, an instruction-stream word patch for the active length, and a
runtime buffer-descriptor offset via aiex.scratchpad_parameter. Runtime buffer-descriptor sizes
and strides are the field I had not made runtime, and that is exactly what this PR makes dynamic
in the runtime sequence itself, on top of the #3358 pool. So the shapes I serve come from
patching a lowered stream, and yours from keeping the runtime sequence dynamic and rebuilding the
TXN per call. I built it and ran it to compare. Two things I can add: an independent hardware
confirmation, and the throughput number the PR notes it lacks.

**Independent hardware run (cols 1, 2, 4).** Every config passes with 0 errors, and one
compiled xclbin per column count serves multiple runtime shapes:

- cols=1: 512x512x512 (494 insts), 256x512x1024 (249 insts)
- cols=2: 512x512x512 (984 insts), 256x512x1024 (494 insts)
- cols=4: 512x512x512 (1964 insts), 768x512x768 (2944 insts)

Runtime M varies the instruction stream on the one compiled xclbin (494 -> 249 as M goes
512 -> 256), exactly as intended.

**Throughput, and the cost of building the stream at runtime.** My own path bakes an instruction
stream per known shape and swaps it as a BO on a resident hardware context, so it pays no
per-dispatch rebuild but needs an artifact per shape and cannot serve an unseen one. This PR's
path builds the stream at runtime from one xclbin. Its extra cost over a prebuilt stream is that
host rebuild, which I measured directly, prebuilt-once versus rebuilt-per-call on the same
xclbin: generate_txn_main_sequence is 0.24 to 2.6 us median (it scales with stream length, so it
is real work), against a 689 to 745 us on-device dispatch at 512x512x512. That is 0.03 to 0.26
percent, negligible. To be precise about scope: this isolates the host rebuild on the same
xclbin; I did not separately build the static (dynamic=False, unrolled) lowering, so I make no
claim about a device-structure difference between the two, only that the runtime build itself is
close to free. Effective rate holds around 360 to 385 GFLOP/s (i16) across the shapes I ran.

**The uncollapsed per-column syncs show up as latency.** At a fixed 512x512x512, dispatch
latency rises with column count (700, 715, 745 us for cols 1, 2, 4) even though more columns
add compute parallelism, and the instruction count roughly doubles per column doubling (494,
984, 1964). At this size the path is bound by the per-column 1x1 syncs rather than compute, so
the wide-sync collapse you called out as an optional follow-up would pay off directly. I am
glad to take that follow-up if it is useful.

Nice to see item 9 landing on top of the #3358 pool.

test/npu-xrt/matmul_whole_array_dynamic/test.cpp
```diff
@@ -0,0 +1,130 @@
...
+  xrt::device device = xrt::device(device_index);
+  xrt::xclbin xclbin = xrt::xclbin(XCLBIN);
+
+  std::vector<xrt::xclbin::kernel> xkernels = xclbin.get_kernels();
+  xrt::xclbin::kernel xkernel = *std::find_if(
+      xkernels.begin(), xkernels.end(), [](xrt::xclbin::kernel &k) {
+        return k.get_name().rfind(KERNEL_NAME, 0) == 0;
+      });
+  std::string kernel_name = xkernel.get_name();
+
```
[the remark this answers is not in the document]
hunhoffe: addressed.

test/npu-xrt/matmul_whole_array_dynamic/run.lit
```diff
@@ -0,0 +1,109 @@
...
+// RUN: aie-translate --aie-npu-to-cpp %t.4.lowered.mlir > %t.4.gen.h
+// RUN: %host_clang %S/test.cpp -o %t.4.exe -std=c++23 -Wall \
+// RUN:   -DGEN_HDR='"%t.4.gen.h"' -DXCLBIN='std::string("%t.4.xclbin")' \
+// RUN:   -I%S/../../../include \
+// RUN:   -I%S/../../../programming_examples/basic/matrix_multiplication \
+// RUN:   %xrt_flags %host_link_flags %test_utils_flags -lm
+// RUN: %run_on_npu2% %t.4.exe 512 512 512 | FileCheck %s
+// RUN: %run_on_npu2% %t.4.exe 768 512 768 | FileCheck %s
+//
+// CHECK: PASS!
```
jgmelber: nit: Check PASS or use the exit condition of the program?

test/npu-xrt/matmul_whole_array_dynamic/run_npu1.lit
```diff
@@ -0,0 +1,100 @@
...
+// RUN: aie-translate --aie-npu-to-cpp %t.4.lowered.mlir > %t.4.gen.h
+// RUN: %host_clang %S/test.cpp -o %t.4.exe -std=c++23 -Wall \
+// RUN:   -DGEN_HDR='"%t.4.gen.h"' -DXCLBIN='std::string("%t.4.xclbin")' \
+// RUN:   -I%S/../../../include \
+// RUN:   -I%S/../../../programming_examples/basic/matrix_multiplication \
+// RUN:   %xrt_flags %host_link_flags %test_utils_flags -lm
+// RUN: %run_on_npu1% %t.4.exe 512 512 512 | FileCheck %s
+// RUN: %run_on_npu1% %t.4.exe 768 512 768 | FileCheck %s
+//
+// CHECK: PASS!
```
jgmelber: nit: Check PASS or use the exit condition of the program?

jgmelber (approved):
This has come a long way!

jgmelber:
> At this size the path is bound by the per-column 1x1 syncs rather than compute, so
> the wide-sync collapse you called out as an optional follow-up would pay off directly. I am
> glad to take that follow-up if it is useful.

I think this would be useful if you are willing!

test/npu-xrt/matmul_whole_array_dynamic/run.lit
hunhoffe: Addressed

test/npu-xrt/matmul_whole_array_dynamic/run_npu1.lit
hunhoffe: addressed; changed to exit condition

hunhoffe:
@jgmelber Thanks for the review!

@atassis Thank you very much for looking at this and sharing your results so far! I agree with Joe -- if you can look into the sync collapse, that would be great!

atassis:
I took a run at the wide-sync collapse and measured it on my Strix (aie2p). Starting all the columns' drains first and then waiting on a single npu.sync over the whole column block (column_num = n_aie_cols), instead of a 1x1 sync per column inside the loop, lets the columns overlap. At 512x512x512 the dispatch latency goes from rising with column count to falling with it: cols 1/2/4 were 701.7/725.8/759.5 us and are now 702.6/447.6/321.0 us (cols=1 unchanged, as expected), all shapes still passing with 0 errors. I kept it as a follow-up on top of this PR so this one can land first, and I will open it (as a draft to start) and rebase once you merge.

hunhoffe:
@atassis That sounds like a great plan! Nice results

PR #3370, opened 2026-07-20
[AIEX] Add aiex.dma_channel_reset runtime-sequence op

atassis:
## What this adds

A runtime-sequence op that resets a DMA channel, lowering to a masked pulse of the channel control
register reset bit (`npu.maskwrite32`). It mirrors `aiex.set_lock`: a non-blocking control op with no
synchronization guarantees, meant to sit inside an `aie.runtime_sequence`.

## Why it is needed

The runtime sequence is the per-dispatch control program. Today it can configure and start DMA transfers
(`aiex.dma_configure_task_for`, `dma_start_task`, `dma_await_task`, `dma_free_task`) and it can set a lock
(`aiex.set_lock`), but it cannot reset a DMA channel. That gap is the whole reason for this op.

I ran into it re-arming a resident objectFIFO across dispatches on aie2p. Some background so the motivation
is concrete. An objectFIFO's locks are initialized once, in the CDO config phase, and nothing re-initializes
them per dispatch. A FIFO that is filled and drained within each dispatch does not care, because it ends each
dispatch in its initial state on its own. But a resident FIFO, one that is acquired once and held on-chip so
its data can be reused across dispatches without re-fetching, does not return to its initial state. Its lock
values and, crucially, its consumer DMA channel queue state carry across dispatches on a live hw_context (no
context switch, no PDI reload). So the second and later dispatches inherit stale control state.

The observable effect on device: the reuse is bit-identical to a fetch baseline for the first two dispatches,
then drifts into a stable but wrong state from the third dispatch on. Re-arming the lock with `aiex.set_lock`
alone is not enough, it holds for exactly two dispatches and then drifts, because the residual is the DMA
channel's self-looping queue state, which no lock write can clear. Resetting that channel each dispatch is
what clears it. With the lock re-arm and this channel reset together, the reuse stays bit-identical across
5000 dispatches with no drift.

## The framing

The way I think about it: a dispatch behaves like a transaction over the array, and correctness across
dispatches needs isolation, so dispatch N+1 does not observe dispatch N's transient control state. Keep the
data resident, but reset the control state at each dispatch boundary. `aiex.set_lock` covers the lock half of
that, and this op covers the DMA channel half. In the resident-objectFIFO case above, re-arming the lock and
resetting the channel together were enough to keep the reuse bit-identical across dispatches, with no firmware
and no PDI reload. I am not claiming that pair is a complete data-plane re-arm in general: which control state
has to be re-armed at a dispatch boundary is a broader lifecycle question. This PR is just the missing
channel-reset mechanism.

## Why an op rather than a raw write32, and why not load_pdi

A raw `npu.write32` to the control register works but is opaque and hardcodes the address. This op computes
the address through `getDmaControlAddress`, so it is expressed by intent (tile, direction, channel) and stays
correct across target models. And it is the cheap, scoped alternative to a full `load_pdi` reload of the whole
device image, which is also correct but re-initializes everything every dispatch and defeats the point of
keeping the FIFO resident.

## Notes for review

1. The reset pulse is two `maskwrite32` writes masked to the reset bit (assert, then clear), so it preserves
   the other control fields (out-of-order, compression, controller id, FoT mode) rather than clobbering them.
   This mirrors aie-rt's own `XAie_DmaChannelReset`, which drives the reset bit with a `MaskWrite32`.
2. Only core and mem tiles have a per-channel DMA reset bit. The shim NOC DMA control register has no reset
   field (aie-rt's `Aie2PShimDmaChProp` sets `Reset.Mask = 0`; bit 1 there is `PAUSE_MEM`, not `RESET`), so
   the verifier rejects shim NOC/PL tiles. This also matches `XAie_DmaChannelReset`, which errors on
   SHIMNOC/SHIMPL.
3. I am deliberately keeping the higher-level question out of this PR, whether a resident objectFIFO should
   carry an attribute that emits this reset automatically. That is a lifecycle design question and it relates
   to the dynamic-runtime-sequence discussion (#3222). This PR is just the missing mechanism, so it can be
   reviewed on its own terms.


thomthehound:
Nice. But I *do* think the lowering should use `npu.maskwrite32` (as you point out). This is supposed to pulse the reset bit, not overwrite the entire DMA control register, after all. Otherwise, it can clobber unrelated control fields in the same register.

I think there might also be a tile-class problem here. Bit 1 is the DMA reset bit for the core- and memory-tile channels, but the shim-NOC DMA channel definition has no reset field. Since the verifier accepts shim-NOC tiles, the current lowering does not actually implement the stated operation for every tile it accepts. I think shim-NOC should be rejected unless there is a separate valid reset mechanism for it.

Otherwise, cool. Thanks!

atassis:
Thanks, both fixed in the latest push.

- The lowering now uses `npu.maskwrite32` masked to the reset bit (0x2), so the pulse preserves the other control fields (out-of-order, compression, controller id, FoT mode) instead of clobbering them. That matches aie-rt's `XAie_DmaChannelReset`, which drives the bit with `XAie_MaskWrite32`.
- The verifier now rejects shim NOC/PL tiles. You were right that there is no reset there: on the shim NOC DMA control register bit 1 is `PAUSE_MEM`, not `RESET`, and aie-rt's `Aie2PShimDmaChProp` sets `Reset.Mask = 0`. Only core and mem tiles carry the reset bit, which also lines up with `XAie_DmaChannelReset` erroring on SHIMNOC/SHIMPL. I dropped the shim case from the lowering test and added a shim-rejection case to the invalid test.

For a default-config channel the masked pulse writes the same register state as the two write32, so the reuse result I measured earlier carries over unchanged; the reset-bit-only mask is what keeps it correct once a channel does carry non-default control fields. Leaving it as a draft to defer to your RFC, still happy for this to fold in or drop out.

thomthehound:
No worries about submitting this for review and merge. You aren't stealing my thunder on anything, and I see no reason it should wait for my RFC. This is useful on its own. My work has evolved into more of a lifecycle-model cleanup and verifier layer, and I was only going to use this problem to illustrate why that is needed. That purpose is still served, and your work looks good to me.

I only ask that you frame it slightly differently: please avoid describing the lock write + channel reset as a generally "complete data-plane re-arm," or saying that the operation by itself establishes a clean dispatch baseline. It does that in this specific case, but not generally -- which is what the RFC will hopefully address.

One small thing I just noticed: could you add an NPU1 lowering test case as well, since the op is meant to cover both AIE2 and AIE2P?

atassis:
Sounds good, I'll mark it ready for review.

On the framing: I scoped the wording so it no longer reads as a general claim. The op description now says it resets the DMA channel state only and does not by itself establish a clean dispatch baseline, and the PR description limits the "lock re-arm plus channel reset is enough" statement to the specific resident-objectFIFO case measured here, not the data plane in general. What else has to be re-armed at a dispatch boundary I've left to the lifecycle discussion.

I added an npu1 (AIE2) lowering case to `test/Passes/lower-dma-channel-reset/dma_channel_reset.mlir` alongside the npu2 one. The control-register address is target-model driven, so both lower to the same local addresses, which is what the test locks in.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1303,6 +1225,40 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
+    ```
+      aiex.dma_channel_reset(2, 3, S2MM, 0)
+    ```
+  }];
+  let arguments = (
+    ins I32Attr:$column,
+        I32Attr:$row,
+        DMAChannelDir:$direction,
+        I32Attr:$channel
+  );
```
andrej, line 1424: Why row, col as I32Attrs rather than an SSA `%tile`?

andrej (approved):
Another good addition to have rather than raw register writes.

Two questions:

1. What's the plan to make this interface with ObjectFifo nicely, if any? I think something like `dma_channel_reset_for @objfifo` would be great, analogous to `dma_configure_task_for @shim_dma_alloc`.

2. Have you experimented with this and #3375 if you were able to fully reset the device? I've tried this before but have not been successful. If we could emit these ops for each core and each DMA channel in the correct sequence, I'd love to be able to replace the "empty load PDI for resetting" hack with a proper reset sequence. However, I've not been able to get that sequence right in the pass and just got deadlocks. But the use case for re-configuration across `aiex.configure`s would be there.

(To be clear, these do not need to be addressed as part of this PR, just curious about your takes.)

atassis:
@andrej thanks for review! Good catch - I switched dma_channel_reset to an SSA %tile (matching dma_configure_task), col/row derived from the tile, lowered pulse unchanged.

ObjectFifo: dma_channel_reset_for @objfifo is the right shape. A fifo has several channels though (producer + consumer ends), so _for has to pick which - the case I hit only needed the consumer end (the queue state that carried across dispatches), so I'd default to that with an option to name a side.

Full reset / deadlocks: I've run each op on aie2p in its own small resident-objectFIFO design, gated on a bit-exact output:
- #3375 core reset: pulse RESET on the two looping cores mid-dispatch. Output holds only if the pulse is masked to the RESET bit and preserves ENABLE (run condition is ENABLE && !RESET); an unmasked write clearing ENABLE breaks the run (qds_device::wait() bad command state). That's thomthehound's mask bug-class, and my guess for your deadlock (mask clearing ENABLE, or assert/clear order).
- #3370 (this op): reset the consumer channel's self-looping queue state each dispatch, which keeps a re-armed objectFIFO bit-identical across dispatches.

I haven't chained them into the full all-cores/all-channels sequence to replace the empty-load-PDI reset, but both halves hold, so it looks reachable - I can test this case if needed

hunhoffe:
@atassis I think that would be a really great addition to your work so far, to test out full all-cores/all-channels sequence to replace the empty-load-PDI reset. Also feel free to create a discussion to track the trajectory of your plans.

hunhoffe:
@atassis re-enqueued -- same failure (which I assume is a transient CI runner failure)

PR #3371, opened 2026-07-20
[dyn-seq] Phase 2 Item 8 dynamic `scf.if`

hunhoffe:
Milestone #3222 **item 8** — runtime `scf.if` (runtime condition) in a runtime sequence, lowered through the dynamic BD pool + AIEX→EmitC into a real C++ `if`. Stacks on #3368 (P7). The prior PRs delivered runtime `scf.for`; this adds conditional dispatch and arbitrary nesting/interleaving of `scf.for`/`scf.if`, so one xclbin can take a data-dependent branch at runtime.                                                                           
                                                                                                                                       
The AIEX→EmitC converter already recursed through any `scf` region, so the work here is in the **dynamic BD pool pass** (`aie-lower-dynamic-bd-pool`), which previously only threaded pool ids through `scf.for`.                                                                                 
                                                                                                                                       
## Carrying pool ids through control flow                                                                                            
                                                                                                                                       
A DMA task is an `index` SSA value the user threads through `scf.for` (iter_args) and `scf.if` (results). Each `dma_configure_task` pops a BD id from the per-tile runtime free-list; that id must be in scope at every point the task is released. Since the task value itself already threads correctly (it is valid IR), the pass shadows **every** SSA position the task occupies with a parallel `i32` id.                                                           
                                                                                                                                       
This is done in an **order-independent** way, so any nesting or interleaving works — including cases a single-direction walk cannot handle (a later sibling consuming an earlier sibling's task; a child loop reusing a parent's carried task):                                                                                                                               
                                                                                                                                       
1. **Task closure** — forward-propagate task-ness from every configure through `scf.for`/`scf.if`, giving the set of SSA values that carry a task.                                                               
2. **Grow structure** (outermost-first) — append a parallel `i32` iter_arg (`scf.for`) or result (`scf.if`) at each carried position; yields start as placeholders.                                                                                                                     
3. **Wire yields** — once every position is paired, set each appended yield to the real id. This resolves both an identity carry (a loop reusing one BD: `scf.yield %iter_arg`) and an id defined by a not-yet-grown inner op.                                                             
                                                                                                                                       
## await = sync, free = release                                                                                                      
                                                                                                                                       
`dma_await_task` lowers to an `npu.sync` (a task-completion-token wait); it does **not** return the BD id to the pool. Only `dma_free_task` returns the id (a `dma_bd_pool_push`). An `await` emits a push only when the id it carries is never reached by a `free`, so an awaited-but-never-freed task still frees its BD exactly once. This keeps the runtime free-list balanced when a BD is reused across iterations, or awaited on one SSA value and freed on another (the same physical id via the carry) — a case the earlier await-is-release model double-freed.                                                                                                                        
                                                                    
## Awaiting a task configured under control flow                  
                                                                    
An `npu.sync` needs only the physical channel (col, row, direction, channel), not the task SSA value. When a configure dominates the await (e.g. a loop's pre-loop init), the await operand is redirected to it. When none does (the task was configured inside an `scf.if` branch), the await instead records the channel via optional `sync_col`/`sync_row`/`sync_direction`/`sync_channel` attributes and its `$task` operand is dropped — BD reuse is already serialized by queue backpressure, so no ordering edge is lost. `dma_await_task`'s `task` operand is now `Optional` to allow this.


lib/Dialect/AIEX/Transforms/AIELowerDynamicBDPool.cpp
```diff
@@ -161,6 +162,74 @@ struct AIELowerDynamicBDPoolPass
     return newFor;
   }
 
+  // Carry popped ids out of an scf.if via its results, so a task configured in
```
hunhoffe: Comment too long.

test/bd-chains-and-dma-tasks/lower-dynamic-bd-pool/bad-if-divergent-yield.mlir
```diff
@@ -0,0 +1,32 @@
...
+    %r = scf.if %cond -> (index) {
+      %t = aiex.dma_configure_task(%tile_0_0, MM2S, 0) {
+        aie.dma_bd(%arg0 : memref<1024xi32> offset = 0 len = 256)
+        aie.end
+      } {issue_token = true}
+      aiex.dma_start_task(%t)
+      aiex.dma_await_task(%t)
+      scf.yield %t : index
+    } else {
+      scf.yield %other : index
```
hunhoffe, line 33: What if there's a second if/else statement that later yields the task, based on the same conditional? How fragile is this?

hunhoffe (commented):
For Claude

test/npu-xrt/dynamic_conditional_passthrough/run.lit
```diff
@@ -0,0 +1,41 @@
...
+// RUN: %host_clang %S/test.cpp -o %t.exe -std=c++17 -Wall \
+// RUN:   -DGEN_HDR='"%t.gen.h"' -DXCLBIN='std::string("%t.xclbin")' \
+// RUN:   -I%S/../../../include %xrt_flags %host_link_flags %test_utils_flags
+//
+// argv: [n] [do_copy]. Exercise the taken branch (do_copy=1) at a couple of tile
+// counts and the not-taken branch (do_copy=0, output untouched).
+// RUN: %run_on_npu2% %t.exe 2 1 | FileCheck %s
+// RUN: %run_on_npu2% %t.exe 8 1 | FileCheck %s
+// RUN: %run_on_npu2% %t.exe 8 0 | FileCheck %s
+// CHECK: PASS!
```
jgmelber: Generally we have gotten away from this and used the return of the program

hunhoffe: Fixed!

lib/Dialect/AIEX/Transforms/AIELowerDynamicBDPool.cpp
```diff
@@ -203,27 +438,88 @@ struct AIELowerDynamicBDPoolPass
...
             return WalkResult::interrupt();
           toErase.push_back(freeOp);
         } else if (auto await = dyn_cast<DMAAwaitTaskOp>(op)) {
-          if (failed(lowerRelease(await.getTask(), await, released)))
+          // Push only if no free balances this id; the await keeps its sync.
+          if (!freedByCarry.contains(await.getTask()) &&
+              failed(lowerRelease(await.getTask(), await)))
             return WalkResult::interrupt();
           if (!await.getTask().getDefiningOp<DMAConfigureTaskOp>()) {
             DMAConfigureTaskOp origin = originConfigure.lookup(await.getTask());
```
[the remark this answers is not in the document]
hunhoffe: (Claude) Good catch — fixed in bc0465f22a7. A task-less await already carries its physical channel in the sync_* attrs and owns no pool id, so the lowering now skips it before touching getTask(), rather than dereferencing the absent operand. Added a lowering test (good-taskless-await-untouched.mlir) covering the case where such an await sits in a runtime-CF sequence the idempotency guard doesn't skip.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -702,6 +702,26 @@ LogicalResult AIEX::NpuDmaWaitOp::verify() {
...
+  // full sync_* attribute set (used when no configure dominates the await). The
+  // sync attributes come as a group: all four or none.
+  unsigned nSync = getSyncCol().has_value() + getSyncRow().has_value() +
+                   getSyncDirection().has_value() +
+                   getSyncChannel().has_value();
+  if (nSync != 0 && nSync != 4)
+    return emitOpError("sync_col, sync_row, sync_direction and sync_channel "
+                       "must all be set together or all omitted");
+  if (!getTask() && nSync == 0)
+    return emitOpError("must have either a task operand or the sync_* channel "
```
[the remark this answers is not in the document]
hunhoffe: (Claude) Agreed, the "both" case was ambiguous. Fixed in bc0465f22a7: the verifier now rejects setting the sync_* attributes when a task operand is also present (the channel comes from one or the other). I went with rejecting rather than cross-validating, since the NPU lowering already prefers the attrs and there's no legitimate producer of both. Added negative verifier cases in test/dialect/AIEX/invalid.mlir.

andrej:
> dma_await_task lowers to an npu.sync (a task-completion-token wait); it does not return the BD id to the pool. Only dma_free_task returns the id (a dma_bd_pool_push). An await emits a push only when the id it carries is never reached by a free, so an awaited-but-never-freed task still frees its BD exactly once. This keeps the runtime free-list balanced when a BD is reused across iterations, or awaited on one SSA value and freed on another (the same physical id via the carry) — a case the earlier await-is-release model double-freed.

This is different from the semantics of dma_await_task up to this point. It's also a bit confusing to me to have it free sometimes and sometimes not, depending on if there is a free afterwards. Why is this change necessary in this context?

andrej:
> An npu.sync needs only the physical channel (col, row, direction, channel), not the task SSA value. When a configure dominates the await (e.g. a loop's pre-loop init), the await operand is redirected to it. When none does (the task was configured inside an scf.if branch), the await instead records the channel via optional sync_col/sync_row/sync_direction/sync_channel attributes and its $task operand is dropped — BD reuse is already serialized by queue backpressure, so no ordering edge is lost. dma_await_task's task operand is now Optional to allow this.

This smells fishy to me. The core problem seems to be that if you conditionally start a task, you should also conditionally await it, or it might deadlock (if you await something that wasn't started). The fact that there is no dominating SSA value in that case highlights that problem by design, and circumventing it seems to me to circumvent a useful compiler check. Could we do something with phi nodes here?

I also don't want to overcomplicate things and maybe am misunderstanding, but at a glance this seems to me like it works around an actual challenge rather than solving it. (workaround = break the dependence of sync on the task it is syncing on; actual challenge = that dependence still exists IRL in hardware, it's just no longer expressed in the IR)

andrej (approved):

PR #3373, opened 2026-07-21
[AIEX] Tighten DMA task op trait from HasAncestor to HasParent for RuntimeSequenceOp

bong-water-water-bong:
## Summary

Tightens the MLIR trait constraint on five DMA-related ops in the AIEX dialect from `HasAncestor<"AIE::RuntimeSequenceOp">` to `HasParent<"AIE::RuntimeSequenceOp">`.

## Motivation

`HasAncestor` matches the op anywhere in the IR subtree under a `RuntimeSequenceOp`, allowing these ops to be nested inside other constructs (e.g., `scf.for`, `func.func`) within the runtime sequence. This is almost certainly unintended — these ops should always be immediate children of `RuntimeSequenceOp`.

`HasParent` enforces the correct constraint, matching the pattern used by other runtime-sequence-only ops in the dialect.

## Affected ops

| Op | Old trait | New trait |
|---|---|---|
| `AIE_DMAConfigureTaskOp` | `HasAncestor` | `HasParent` |
| `AIE_DMAConfigureTaskForOp` | `HasAncestor` | `HasParent` |
| `AIE_DMAFreeTaskOp` | `HasAncestor` | `HasParent` |
| `AIE_DMAStartTaskOp` | `HasAncestor` | `HasParent` |
| `AIE_DMAAwaitTaskOp` | `HasAncestor` | `HasParent` |

## Testing

- All existing `check-air-mlir` LIT tests pass locally
- Verified against downstream `mlir-air` LLM inference examples (Llama-3.2, SmolLM2, Qwen2.5, Qwen3) on AMD XDNA 2 NPU hardware (Ryzen AI MAX+ 395)

/cc @samuelbayliss @erwei-wang

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -1075,7 +1075,7 @@ def AIE_NpuLoadPdiOp: AIEX_Op<"npu.load_pdi", []> {
   let hasCanonicalizeMethod = 1;
 }
 
-def AIE_DMAConfigureTaskOp : AIEX_Op<"dma_configure_task", [HasAncestor<"AIE::RuntimeSequenceOp">, TileElement]>, Results<(outs Index:$result)> {
+def AIE_DMAConfigureTaskOp : AIEX_Op<"dma_configure_task", [HasParent<"AIE::RuntimeSequenceOp">, TileElement]>, Results<(outs Index:$result)> {
   let summary = "Concrete Instantiation of a Buffer Descriptor Chain as a Task on a Channel and Direction on a Tile";
   let description = [{
     Encapsulates the DMA configuration of one task, that is the (chain of) buffer descriptors to be executed on a given channel and direction on a tile.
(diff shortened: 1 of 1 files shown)


jgmelber (approved):

andrej (changes requested):
Thanks for your contribution. I'm trying to understand what problem this solves for you?

Couple conflicts:

- The `aiex.configure` syntax requires DMA tasks to be nested inside `aiex.configure` rather than directly under the runtime sequence.
- @hunhoffe's dynamic runtime sequence work enables DMA tasks being dispatched from inside `scf.for` and other operations, so also nested, not directly in the runtime sequence.

Maybe if you can explain your use case / problem we can find another solution.

hunhoffe (changes requested):
Hello! Thank you for submitting a contribution.

> `HasAncestor` matches the op anywhere in the IR subtree under a `RuntimeSequenceOp`, allowing these ops to be nested inside other constructs (e.g., `scf.for`, `func.func`) within the runtime sequence. This is almost certainly unintended — these ops should always be immediate children of `RuntimeSequenceOp`. 

This was (somewhat) previously true, but due to recent work towards https://github.com/Xilinx/mlir-aie/discussions/3222 there may be `scf.for` (and eventually) `scf.if` control flow in the runtime sequence. Furthermore, I think this conflicts with some of the work with `aiex.configure`, as there are some notions of nesting runtime sequences.

What is the overall benefit/need for this type of change? If there is a problem that needs fixing, it's worth exploring other ways to achieving a solution.

hunhoffe:
@bong-water-water-bong If you are looking for improvements to compile code organization/quality, I think you could find some good first issues from the comments on this PR: https://github.com/Xilinx/mlir-aie/pull/3344 or on this issue: https://github.com/Xilinx/mlir-aie/issues/721

PR #3374, opened 2026-07-21
[docs] Add a runtime-sequence control-op reference

atassis:
The programming guide teaches the runtime sequence by example (section 2d) and a full design in section 3, but there is no single page that lists the control ops you can place in an `aie.runtime_sequence` and says what each one does. Today, to find out what commands exist, you have to read through the tutorials and then `AIEX.td`.

This PR adds `programming_guide/runtime_sequence_reference.md`: a catalog of the `aiex` runtime-sequence control ops, grouped by the resource they act on and the verb they perform.

1. Registers and tile memory (`write32`, `maskwrite32`, `blockwrite`)
2. Buffer descriptors and DMA queues (`dma_memcpy_nd`, `writebd`, `push_queue`, `address_patch`, the `assert_bd_*` guards)
3. DMA tasks (`dma_configure_task[_for]`, `dma_start_task`, `dma_await_task`, `dma_free_task`, `dma_start_bd_chain[_for]`, `dma_bd_pool_pop`/`push`)
4. Synchronization (`sync`, `dma_wait`)
5. Locks (`set_lock`)
6. Runtime parameters and scratchpad (`rtp_write`, `create_scratchpad`, `update_from_scratchpad`, `scratchpad_parameter`, `read_scratchpad_parameter`, `sync_scratchpad_parameters_from_host`)
7. Device image and execution (`load_pdi`, `configure`, `run`, `preempt`)
8. Control packets (`control_packet`)

Each op gets a one-line description condensed from its `let description` field in `AIEX.td`, the tile classes and DMA directions it applies to, and a one-line example. Ops the compiler emits on the dynamic buffer-descriptor path (the `assert_bd_*` and `dma_bd_pool_*` ops) are marked `(compiler-emitted)`. The page opens with a short "how to read this" section, a "which interface should I use" note that points readers at the two high-level data-movement interfaces, and states the membership rule (`HasParent`/`HasAncestor<RuntimeSequenceOp>`, the transitive `configure`/`run` case, and the `npu.*` instruction family).

I cross-linked the page from the programming-guide index and from section 2d, and registered it in the mkdocs nav. `mkdocs build --strict` passes locally.

Descriptions are drawn from the existing `let description` fields in `AIEX.td`, so the reference tracks the dialect as authored, and it stays a curated companion to the auto-generated AIEX dialect reference (which, with `AIEX.td`, is authoritative).

**Placement (for review):** I put the page under `programming_guide/` next to the other companion references and the section-2d tutorial, since that is where a reader meets the runtime sequence. If you would rather it live under `docs/` with the dialect documentation, I am happy to move it.

**Nav position (for review):** I added it to the "Extra References" group in `mkdocs.yml`. Since every design has a runtime sequence, I thought it might belong higher in the navigation, but I would rather ask your preference first than reorder the nav myself.

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -245,6 +245,7 @@ nav:
       - Compilation Stages: programming_guide/compilation_stages.md
       - Implicit MLIR Context: programming_guide/implicit_mlir_context.md
       - Kernel Library: programming_guide/kernels_library.md
+      - Runtime Sequence Reference: programming_guide/runtime_sequence_reference.md
   - API & Internals:
     - Overview: api/index.md
     - Python API:
--- a/programming_guide/README.md
+++ b/programming_guide/README.md
@@ -74,6 +74,7 @@ my_design(a, b)              # compile + run + sync back
 | Setting-by-setting configuration (cache dir, tensor backend, log level) | [Configuration options](./iron_configuration.md) |
 | What happens between `@iron.jit` and the NPU running | [Compilation stages](./compilation_stages.md) |
 | Ready-made compute kernels (matmul, conv, eltwise, vision) | [Kernel library](./kernels_library.md) |
+| Every control op you can put in a runtime sequence | [Runtime sequence op reference](./runtime_sequence_reference.md) |
 
 ## Sections
 
--- a/programming_guide/runtime_sequence_reference.md
+++ b/programming_guide/runtime_sequence_reference.md
@@ -1,0 +1,195 @@
+<!---//===- runtime_sequence_reference.md ------------*- Markdown -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//-->
+
+# Runtime Sequence Op Reference
+
+A runtime sequence is the host-side program that drives an NPU design: it fills
+tensors into the AIE-array, launches DMAs, waits on completion tokens, and
+reconfigures state between dispatches. In IRON you write it as the body of
+`Runtime.sequence()`; in the AIE dialect it is the region of an
+`aie.runtime_sequence` operation.
+
+[Section 2d](./section-2/section-2d/) teaches the runtime sequence by example
+and [Section 3](./section-3/) shows a full end-to-end program. This page is the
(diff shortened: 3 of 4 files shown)


jgmelber (approved):

hunhoffe:
This is a really excellent contribution! I do have one structural suggestion (placement) one small accuracy note.

---
Placement: consider moving to docs/AIEXDialect.md

The programming guide is IRON/Python-focused, and IRON users reach data movement through rt.fill() / rt.drain() — they don't hand-write these aiex ops. The catalog's natural audience is someone reading lowered MLIR, debugging a generated instruction stream, or writing a pass — i.e. the API & Internals → AIEX Dialect audience.

Suggested split:
1. Move the eight-category tables into docs/AIEXDialect.md as a hand-written preamble above the auto-generated per-op content (same pattern as docs/AIEDialect.md). This also lets the tables sit right next to the authoritative operand/attribute listings.
2. In the programming guide (section-2d), leave a short bridge that shows the lowering connection and links to the dialect reference (draft below).
3. Drop the new mkdocs.yml nav entry — the AIEX dialect page already has its slot under API & Internals.

Bridge paragraph for section-2d:
> Under the hood, `rt.fill()` and `rt.drain()` lower through the DMA-task API: each becomes an `aiex.dma_configure_task_for` (configuring a single-BD chain on the shim channel named by the ObjectFifo's `aie.shim_dma_allocation`) followed by `aiex.dma_start_task`, with `aiex.dma_await_task` (when `wait=True`) or a deferred `aiex.dma_free_task` on task-group finish. The full catalog of runtime-sequence ops is in the AIEX dialect reference (../../../docs/AIEXDialect.md).

 ---
Small accuracy note: `aiex.npu.dma_memcpy_nd` — note that `S2MM` always issues a token. The interface guidance reads as though `issue_token` is always opt-in. Per `AIEX.td` and `AIEDmaToNpu.cpp` (`if (!isMM2S) issue_token = true`), `S2MM` channels always issue a token; only `MM2S` needs `issue_token = true`. Worth one clause so readers don't add a redundant flag or expect they can suppress it.

I'm also ok with merging as-is and following up on this later.

atassis:
I think we might wanna merge this then. It's a bit easier for me to  create a new PR with requested changes and discuss any other requests there.

PR #3375, opened 2026-07-21
[AIEX] Add aiex.core_reset runtime-sequence op

atassis:
## What this adds

A runtime-sequence op that resets a compute core, lowering to a masked pulse of the CORE_CONTROL register
reset bit (`npu.maskwrite32`). It mirrors `aiex.set_lock`: a non-blocking control op with no synchronization
guarantees, meant to sit inside an `aie.runtime_sequence`.

## Why it is needed

The runtime sequence is the per-dispatch control program. Today it can configure and start DMA transfers,
and it can set a lock (`aiex.set_lock`), but it has no way to reset the compute core. That gap is the reason
for this op.

The motivating case is re-arming a resident kernel across dispatches on aie2p. A core that is started once
and left resident, so its program can run again on the next dispatch without a fresh PDI load, carries its
processor state from the previous dispatch. The heavy way to return it to a known state is a full `load_pdi`
reload of the device image, which re-initializes everything and defeats the point of keeping the core
resident. This op is the scoped alternative: reset just the core, in place, from the control program.

## What the reset does and does not touch

CORE_CONTROL.RESET resets the processor: program counter and registers. It does not clear the tile's data
memory (that is the whole point, the resident data stays put), and it does not touch locks or the tile DMA.
So this op is the core-state half of a dispatch-boundary re-arm; a caller that needs a full clean handshake
also resets the relevant locks (`aiex.set_lock`) and DMA, the same way aie-rt re-initializes those
separately around a core reset. It also assumes the resident core is still enabled: the pulse preserves the
ENABLE bit but does not set it, so a core that has already cleared ENABLE needs a separate re-enable. I have
kept those as separate concerns rather than folding enable/lock/DMA into this one op, so it stays a single
well-defined primitive.

## Why an op rather than a raw write32, and why the pulse is masked

A raw `npu.write32` to CORE_CONTROL works but is opaque, and a plain write of the reset bit would clobber the
rest of the register. CORE_CONTROL packs ENABLE (bit 0) and RESET (bit 1) in one word. So the lowering emits
the reset as two `npu.maskwrite32` writes masked to bit 1 (assert, then deassert), which preserves ENABLE
instead of overwriting it. This mirrors aie-rt's `XAie_CoreReset` and `XAie_CoreUnreset`, which drive the
reset bit with a `MaskWrite32`.

One deliberate difference from the sibling `dma_channel_reset`: that op computes its register address through
the target model (`getDmaControlAddress`), which varies by tile class, channel, and direction. CORE_CONTROL
does not vary, it sits at tile-local offset `0x32000` on every generation (`XAIE2PGBL_CORE_MODULE_CORE_CONTROL`
on aie2p, `XAIEMLGBL_CORE_MODULE_CORE_CONTROL` on aie2, both `0x32000`), and there is no existing
`getCoreControlAddress` on the target model. Rather than add a virtual that returns the same constant for
every override, the lowering uses the constant directly with a comment tying it to the aie-rt register name.
If maintainers would prefer the symmetry with `getDmaControlAddress`, I am happy to add
`getCoreControlAddress(col, row)` to `AIETargetModel` instead.

## Notes for review

1. The verifier accepts only core tiles. It rejects mem and shim tiles (no compute core, matching
   `XAie_CoreReset`, which errors on any non core tile), coordinates outside the device array (the column and
   row are raw attributes, so this is the only thing that bounds them), and AIE1 (the NPU runtime sequence has
   no meaning there, matching `set_lock`).
2. The op takes just (column, row). It resets the whole core, so there is no channel or direction parameter.
3. Verified by lit: a lowering test that checks the masked pulse (value/mask at the CORE_CONTROL address, on
   aie2p and aie2, with `--implicit-check-not` guarding against a stray unmasked write) and an invalid test
   that checks the verifier rejects mem, shim, out-of-range, and AIE1.
4. Validated on device (aie2p / Ryzen AI NPU). Injecting the exact masked pulse this op lowers to onto two
   resident objectFIFO cores resets and restarts them without disturbing the result (a two-stage streaming
   design keeps producing correct output, reproduced). A control that instead clears ENABLE on one of those
   cores breaks the run, which confirms the writes reach the core and that masking the pulse to the reset bit
   (preserving ENABLE) is what keeps it correct. The attr-form `maskwrite32` also survives the full lowering to
   an xclbin, so the op needs no absolute-address form. The one case I did not exercise is a run-once core that
   has cleared ENABLE after reaching Done; re-arming that also needs a re-enable, as the description notes.
5. I am deliberately keeping the higher-level question out of this PR, whether a resident core should carry an
   attribute that emits this reset automatically. That is a lifecycle design question. This PR is the missing
   mechanism.

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -1441,6 +1441,40 @@ def AIEX_DmaChannelResetOp: AIEX_Op<"dma_channel_reset", [HasParent<"AIE::Runtim
   let hasVerifier = 1;
 }
 
+def AIEX_CoreResetOp: AIEX_Op<"core_reset", [HasParent<"AIE::RuntimeSequenceOp">, SkipAccessibilityCheckTrait]> {
+  let summary = "Reset a compute core from a runtime sequence";
+  let description = [{
+    Pulses the reset bit of the CORE_CONTROL register on the tile at (`column`,
+    `row`). The pulse is two `npu.maskwrite32` writes masked to the reset bit
+    (assert, then deassert), so it preserves the ENABLE field packed in the same
+    CORE_CONTROL word instead of clobbering it. This mirrors aie-rt's
+    XAie_CoreReset followed by XAie_CoreUnreset. The operation is non blocking and
+    offers no synchronization guarantees.
+
+    Resetting clears the core's processor state (program counter and registers);
+    it does not touch the tile's data memory, locks, or DMA. The intended use is
+    re-arming a resident kernel across dispatches without a full PDI reload:
+    discard stale core state while keeping resident data in place. This assumes
+    the resident core is still enabled; re-arming a core that has cleared its
+    ENABLE bit also needs a re-enable, which this op does not emit.
+
+    Only core tiles have a CORE_CONTROL register; mem and shim tiles are rejected
+    by the verifier because they have no core to reset. This matches aie-rt's
+    XAie_CoreReset, which errors on non core tiles.
+
+    Example:
+    ```
+      aiex.core_reset(2, 3)
+    ```
+  }];
+  let arguments = (
+    ins I32Attr:$column,
+        I32Attr:$row
+  );
+  let assemblyFormat = [{ `(` $column `,` $row `)` attr-dict }];
+  let hasVerifier = 1;
+}
+
 //===----------------------------------------------------------------------===//
 // ScratchpadParameter ops
 //===----------------------------------------------------------------------===//
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -1444,12 +1444,12 @@ def AIEX_DmaChannelResetOp: AIEX_Op<"dma_channel_reset", [HasParent<"AIE::Runtim
 def AIEX_CoreResetOp: AIEX_Op<"core_reset", [HasParent<"AIE::RuntimeSequenceOp">, SkipAccessibilityCheckTrait]> {
   let summary = "Reset a compute core from a runtime sequence";
   let description = [{
-    Pulses the reset bit of the CORE_CONTROL register on the tile at (`column`,
-    `row`). The pulse is two `npu.maskwrite32` writes masked to the reset bit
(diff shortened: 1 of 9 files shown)


jgmelber (approved):

atassis:
Hi, @jgmelber! Thanks for reviewing my work, again! I have a PR and idea, on which you might be a natural reviewer ([this pr](https://github.com/Xilinx/llvm-aie/pull/1172)). I have added a `SubtargetFeature` capability set, mirroring amdgpu's model, but mostly this is a subtask of my more broader look at what whole architecture and contract could look like.
Idea is the link in PR (an article).

hunhoffe:
@atassis I'd be interested in hearing more about your proposal but I have some trouble with the link. Can you please summarize here (or in a discussion)?

atassis:
https://atassis.github.io/writing/amd-contract-first-e99255/ does this help? Temporarily deployed it onto github pages.
Long story short-  #1172 adds a `SubtargetFeature` capability set to the AIE backend (capability is a set, not the `__AIE_ARCH__` level), and the broader idea is that freezing a few interface contracts would let the stack stop moving in lockstep.

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1389,6 +1389,40 @@ def AIEX_SetLockOp: AIEX_Op<"set_lock", [HasParent<"AIE::RuntimeSequenceOp">, Sk
...
+    XAie_CoreReset, which errors on non core tiles.
+
+    Example:
+    ```
+      aiex.core_reset(2, 3)
+    ```
+  }];
+  let arguments = (
+    ins I32Attr:$column,
+        I32Attr:$row
```
andrej: Why not a reference to an SSA `%tile`?

atassis:
https://github.com/Xilinx/mlir-aie/discussions/3390
I have landed this into discussion, also so we can separate PR review from my discussion

andrej (commented):
It's useful to have an explicit instruction for core resets, thanks for this contribution.

But could we use an SSA tile, like in the channel reset op, rather than hard-coded column/row values?

include/aie/Dialect/AIEX/IR/AIEX.td
atassis: Good eye, I have missed that, thanks! Already fixed

PR #3377, opened 2026-07-21
Remove Lock ID Allocation out of ObjectFIFO pass

andrej:
ObjectFIFO needlessly had its own lock ID allocation logic, which clashed with other locks and forced awkward pass ordering. This removes lock ID allocation out of the ObjectFIFO stateful transform pass and moves the AssignLockIDs pass after the ObjectFIFO pass.


hunhoffe (approved):
😀 Excited about this one!

PR #3386, opened 2026-07-21
[dyn-seq] Collapse per-column output syncs into one wide TCT sync

atassis:
Follow-up to #3368. At a fixed 512x512x512, dispatch latency rose with column count even though more columns add compute parallelism, because each column's C drain was awaited with its own 1x1 TCT sync inside the per-column loop, which serialized the columns.

This starts every column's drain (and its A/B feeds) first, then waits on all of them with a single `npu.sync` over the contiguous column block (`column_num = n_aie_cols`, `row_num = 1`). The columns share shim row 0, the S2MM direction, and channel 0, and sit on columns `0..n_aie_cols-1`, so one TCT wait covers the whole block. `n_aie_cols == 1` keeps the original 1x1 sync so col 0 stays byte-identical to the single-column design. The change is parameterized only by `n_aie_cols`, so it holds for any runtime M/K/N.

## Testing

Measured on Strix (XDNA2, aie2p) at 512x512x512, dispatch latency (median of 200 iterations), every shape in the example matrix passing with 0 errors:

```
cols=1   701.7 -> 702.6 us   (unchanged; nothing to collapse)
cols=2   725.8 -> 447.6 us
cols=4   759.5 -> 321.0 us
```

Before the collapse latency rose with column count; after it falls, since the columns now overlap. Instruction count drops by `n_aie_cols - 1` four-word syncs per row-block (cols=4 at 512x512x512: 1964 -> 1940). Runtime M still varies the instruction stream on one xclbin (494 -> 249 as M goes 512 -> 256), so the dynamic one-xclbin-many-shapes property is unchanged.

### DIFF
--- a/test/npu-xrt/matmul_whole_array_dynamic/whole_array_dynamic.py
+++ b/test/npu-xrt/matmul_whole_array_dynamic/whole_array_dynamic.py
@@ -37,6 +37,7 @@ from aie.dialects.aiex import (
     dma_start_task,
     dma_await_task,
     dma_free_task,
+    npu_sync,
 )
 from aie.extras.dialects import arith
 from aie.helpers.util import np_dtype_to_mlir_type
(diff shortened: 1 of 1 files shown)


atassis:
Looks like the broken test here and [this](https://github.com/Xilinx/mlir-aie/pull/3385#issuecomment-5040479744) are related somehow. I am not sure but maybe there is a host-side issue on the machine (or machines), which ran these CI runs. Worth noting

hunhoffe:
@atassis yes, this has affected a few of my PRs as well. Let me rerun the failure, as a "for now" solution

hunhoffe (approved):

PR #3387, opened 2026-07-22
[dyn-seq] Rework IRON Runtime as an eager callback body

hunhoffe:
**NOTE:** Updated description after review changes.

## Motivation
  
The replayed-sequence Runtime can't express the control flow introduced in https://github.com/Xilinx/mlir-aie/discussions/3222 without abusing inline_ops. An eager callback body handles control flow for both the static and dynamic sequence-generation paths without changing the Python design, and gives Runtime the same shape as Worker.

## Description

Replaces the `rt.sequence(...)` context-manager API with a single eager callback body, mirroring `Worker(core_fn, fn_args)`: `Runtime(seq_fn, fn_args)` calls `seq_fn(*fn_args)` once at resolve time inside `@runtime_sequence`, and `Program(device, rt, workers=...)` starts Workers instead of `rt.start(...)`. `fn_args` is a single ordered list mixing runtime-input declarations (a type/dtype for a buffer, a concrete int for a folded constant) with shared objects (`ObjectFifoHandle`s, `Buffer`s, `Kernel`s, ...) passed straight through — the same body serves both the static and dynamic (EmitC) lowering depending on whether an int or a type is given. Because the body runs eagerly with live MLIR values in scope, it can use native `range_`/`if_` directly, so `inline_ops` is removed.

```python
  # before
  rt = Runtime()
  with rt.sequence(line_ty, line_ty) as (a, b):
      rt.start(worker)
      rt.fill(of_in.prod(), a)
      rt.drain(of_out.cons(), b, wait=True)
  return Program(iron.get_current_device(), rt).resolve_program()
  # after
  def sequence(a, b, in_h, out_h):
      in_h.fill(a)
      out_h.drain(b, wait=True)
      
  rt = Runtime(sequence, [line_ty, line_ty, of_in.prod(), of_out.cons()])
  return Program(iron.get_current_device(), rt, workers=[worker]).resolve_program()
```

#### Related API changes:
- fill/drain live on the ObjectFifoHandle itself (in_h.fill(a) / out_h.drain(b)) and return a Task handle to the transfer. A Task wraps the transfer's !index SSA value — pass it as a range_ iter_args entry to carry an in-flight transfer across loop iterations, and call .free()/.await_() on it directly. managed=False opts a transfer out of automatic group cleanup so the caller owns its lifetime — what a hand-rolled ping-pong needs.
- TaskGroup batches related fill/drain calls (via group=) so their waits are ordered before their frees on .finish(); it does not coalesce multiple awaits into a single hardware sync (each awaited task still lowers to its own 1x1 npu_sync). Runtime(..., strict_task_groups=True) (the default) rejects mixing the implicit default group with explicit ones.
- enable_trace(...) moves from Runtime to Program, since it configures both the traced Workers' tiles and the Runtime's trace-buffer sequencing.
- Scoped entirely to the Python iron layer — no dialect, pass, or C++ TXN-encoding changes.
- All in-tree callers (tests, programming_examples, docs) are migrated to the new API.

## Checklist


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -716,6 +716,26 @@ LogicalResult AIEX::NpuDmaWaitOp::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
```
hunhoffe: This appears out of scope. Should be removed from this PR.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -115,6 +115,21 @@ struct DMAAwaitTaskOpPattern : OpConversionPattern<DMAAwaitTaskOp> {
   LogicalResult
   matchAndRewrite(DMAAwaitTaskOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
```
hunhoffe: This seems out of scope. Should be removed from this PR.

lib/Dialect/AIEX/Transforms/AIELowerDynamicBDPool.cpp
```diff
@@ -469,29 +468,93 @@ struct AIELowerDynamicBDPoolPass
...
+              AIE::TileOp t = origin.getTileOp();
+              await.setSyncCol(t.getCol());
+              await.setSyncRow(t.getRow());
+              await.setSyncDirection(origin.getDirection());
+              await.setSyncChannel(origin.getChannel());
+              // Drop the operand: the sync only needs the channel, and keeping
+              // a use of the branch-carried task would block the configure from
+              // lowering (it requires use_empty). BD reuse stays serialized by
+              // queue backpressure, so the dropped ordering edge is not needed.
+              await.getTaskMutable().clear();
```
hunhoffe: Why did we need to make changes here? What did this fix?

programming_examples/basic/event_trace/README.md
```diff
@@ -42,21 +42,29 @@ def aie_trace(A: In, F: In, C: Out, *, tensor_size: CompileTime[int] = 4096, ...
...
+    def sequence(a_in, f_in, c_out, in_h, factor_h, out_h):
+        in_h.fill(a_in)
+        factor_h.fill(f_in)
+        out_h.drain(c_out, wait=True)
+
+    rt = Runtime(
+        sequence,
+        [tensor_ty, scalar_ty, tensor_ty],
+        fn_args=[of_in.prod(), of_factor.prod(), of_out.cons()],
+    )
```
hunhoffe, line 53: I'm wondering if enable_trace should be on program instead of runtime, since it involved orchestration on both the runtime and the workers?

python/iron/dataflow/objectfifo.py
```diff
@@ -38,6 +38,17 @@
 from ..device import Tile, AnyMemTile
 
 
+def _same_tile(a: "Tile | None", b: "Tile | None") -> bool:
+    """Whether two shim-tile pins refer to the same placement.
+
+    Tile.__eq__ is identity-based, so compare by (col, row); two unpinned
```
hunhoffe, line 44: This seems only valid for shims, which doesn't match the name of this function?

python/iron/dataflow/objectfifo.py
```diff
@@ -38,6 +38,17 @@
 from ..device import Tile, AnyMemTile
 
 
+def _same_tile(a: "Tile | None", b: "Tile | None") -> bool:
```
hunhoffe: Shouldn't we just, like, override the wquality function in that test? Seems weird that this is randomly in the objectfifo tile...

python/iron/runtime/_context.py
```diff
@@ -0,0 +1,52 @@
+# _context.py -*- Python -*-
+#
```
hunhoffe, line 2: This seems messy. Is there a way we can set in Program rather than as a general API function in runtime?

python/iron/runtime/data.py
```diff
@@ -45,6 +45,17 @@ def arr_type(self) -> type[np.ndarray]:
         """The tensor type of the buffer."""
         return self._arr_type
 
+    @property
```
hunhoffe, line 48: I feel like there's got to be a more proper way to test this.

python/iron/runtime/dmatask.py
```diff
@@ -75,12 +91,31 @@ def resolve(
...
-            issue_token=self._wait,
-            offset_parameter=self._offset_parameter,
-            packet=self._packet,  # pyright: ignore[reportArgumentType]
-        )
+        # Reference the ObjectFifo by symbol name, not its resolved op: the DMA
+        # only needs the metadata symbol (shim_dma_single_bd_task reduces an op
+        # to its sym_name anyway), and a name is a legal MLIR forward reference.
+        # This lets the runtime sequence body emit before fifos are resolved, so
+        # the body runs exactly once.
+        if self._tap is not None:
```
hunhoffe, line 106: Here and elsewhere, shouldn't there be errors if you specify both tap and any of sizes/strides/offsets? Collisions should result in hard errors, not ambiguity.

python/iron/runtime/runtime.py
```diff
@@ -20,88 +29,216 @@
...
+    def note_fifo(self, handle: ObjectFifoHandle) -> None:
+        """Record that ``handle`` is driven from the runtime (its shim endpoint)."""
+        self._runtime._fifos.add(handle)
+
+    def register_task_group(self, tg: TaskGroup) -> None:
+        self._open_task_groups.append(tg)
 
-class IronRuntimeError(Exception):
-    """Raised by the IRON Runtime when resolution encounters an unrecoverable state."""
+    def finish_task_group(self, tg: TaskGroup) -> None:
```
hunhoffe, line 89: Can't you use the task gorup object itself, rather than tracking on the sequence, to do this type of thing?? I generally don't like contexts, especially if we consider one day runtime sequences could be nested...

hunhoffe (commented):
For AI

jskimko:
Nice work on this large PR! Some comments:

- This change looks great from an architectural and modularity perspective. Decoupling the logic makes unit testing easier and aligns the API well with standard concurrency libraries.
- There seems to be an implicit positional mapping between `sequence(a, b, in_h, out_h)` and `Runtime(..., [line_ty, line_ty], fn_args=[of_in.prod(), of_out.cons()])`, i.e., `fn_args` only has two arguments for a 4-arg `sequence`. Is there a way to make this binding more explicit?
- What is the main motivation behind this change? It would be helpful to clarify the context in the description.

hunhoffe:
@jskimko I added a motivation section to the description, and I think that's a good point about the arguments! What do you think of:
`def sequence(a, b, in_h, out_h): ...` ==> `Runtime(sequence, [line_ty, line_ty, of_in.prod(), of_out.cons()])`?

In retrospect, this seems obviously better but happy to brainstorm other solutions that might work as well!

docs/api/iron.md
```diff
@@ -42,8 +42,9 @@ The objects most designs are built from.
 
 ### Runtime
 
-The host-side orchestration entry point that declares `fill` / `drain` /
-`start` operations via `rt.sequence(...)`.
+The host-side orchestration entry point. Its `fill` / `drain` operations
+are declared in the sequence body passed to `Runtime(seq, inputs, fn_args)`;
+Workers are passed to `Program(workers=...)` rather than started from the body.
```
andrej, line 47: Was there previously a meaning/use for starting a worker mid-way through the runtime sequence? Seems like the new approach would no longer allow this.

andrej (approved):
This change makes sense to me, I think this will make the runtime sequence more intuitive to use as well -- especially when trying to mix Python-binding-MLIR with IRON by dropping the inline_ops. 

Agreed on Jason's comments about the arguments. I did kind of like the separation of run-time-buffer arguments vs. compile-time components/parametrizations in `Runtime`, but as soon as they become a linear sequence of arguments in `def sequence`, it becomes a bit awkward.

jskimko:
@hunhoffe I figured it was an enabler for something else you were doing; thanks for adding the context!

Yes, that API looks better to me, but I'd go one step further to mimic [Executor.submit](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.submit) (appears to be the closest match). The call site would then look like:

```
# unpacking a list
args = [line_ty, line_ty, of_in.prod(), of_out.cons()]
rt = Runtime(sequence, *args)

# or inline
rt = Runtime(sequence, line_ty, line_ty, of_in.prod(), of_out.cons(), workers=[worker])
```

So, the difference is that `args` is unpacked at the call site rather than passed as a list.

docs/api/iron.md
hunhoffe: Uh no, it was purely cosmetic for.... historical reasons?

@jgmelber -- thoughts?

hunhoffe:
@jskimko I'll have to think about that. Symmetry within the API is important too, including matching to outside APIs. I think if I made that update to the `Runtime`, then I'd need to plan a follow-up PR for the `Worker` too.

docs/api/iron.md
andrej: Well if that's the case then the new way is better in every way because it does not imply you could do that

PR #3391, opened 2026-07-22
[AIEX] Add an experimental verifier for un-re-armed dma_channel_reset

atassis:
## What

Adds `--aie-verify-runtime-rearm`, an experimental, opt-in verification pass that
rejects an `aiex.dma_channel_reset` whose objectFIFO locks are never re-armed by
an `aiex.set_lock` in the module. It turns a common runtime deadlock into a
compile-time error with a source location and a note pointing at each frozen lock.

It is not in the default pipeline. It is a conservative, best-effort check (see
Scope and Known limitations): it catches the common missing-re-arm mistake but is
neither necessary nor sufficient, so I would like it validated on hardware and its
limitations closed before it is wired in by default. Run it explicitly with
`--aie-verify-runtime-rearm` for now.

## Why

`aiex.dma_channel_reset` (#3370) pulses a core or mem tile DMA channel's reset
bit. On AIE-ML the channel has no enable bit, so after the reset drains its task
queue the channel moves data again only once its start queue is re-pushed and
the objectFIFO's semaphore locks are re-armed to their init values. If the locks
are left at their frozen mid-transfer counts, the channel's peer blocks on an
acquire that never releases and the host sees `qds_device::wait() unexpected
command state`. That failure is invisible until you run on hardware.

A bound lock with no runtime re-arm is the common form of this mistake, so the
pass reports it as an error rather than letting it reach silicon. It does not
prove the reset is safe, nor that an un-re-armed lock always deadlocks (see Known
limitations).

## Scope

The check only fires on a channel bound to objectFIFO locks, resolved through the
tile's `aie.mem` / `aie.memtile_dma` `dma_start` and the `use_lock` ops in its BD
chain. A raw DMA with no locks has nothing to verify. A lock counts as re-armed
by an `aiex.set_lock`; the pass runs before `set_lock` is lowered, so it does not
need to recognise the lowered `npu.write32` to the lock register (which would
duplicate the `set_lock` encoding here). Re-arms are collected module-wide, so a
re-arm in another dispatch's sequence is honoured.

## Known limitations

Tracked in #3418. The check is neither necessary nor sufficient:

- False positive: a lock can be re-armed from inside the `aie.core` (balanced
  `use_lock` over the dispatch), not only by a runtime `aiex.set_lock`. The pass
  scans only runtime-sequence re-arms, so a design that re-arms from the core is
  flagged even though the reset is fine. This is why the pass is opt-in.
- False negative: it checks only the lock re-arm, not the start-queue re-push, so
  a design that re-arms locks but not the queue still deadlocks and still passes.
  The re-push primitive is `aiex.dma_channel_reset_for` (#3393); I plan to extend
  this pass to the full invariant once that lands.

## Placement and default

The pass expects the objectFIFO stateful transform to have run (channels and
locks materialised); the natural home is `getNpuDmaLoweringPipeline`, between
`aie-assign-runtime-sequence-bd-ids` and `aie-dma-tasks-to-npu`, where
`dma_channel_reset`, `set_lock`, and the channel-to-lock binding all coexist and
the IR is straight-line. It mirrors the token-balance deadlock check already in
`aie-assign-runtime-sequence-bd-ids`.

I have left it out of the default pipeline for now. Because it is conservative
(see Known limitations), I would rather close those and prove it on hardware
before enabling it by default; run it explicitly with `--aie-verify-runtime-rearm`
in the meantime.

## Tests

`test/Passes/verify-runtime-rearm/dma_channel_reset_rearm.mlir`, seven cases under
`-split-input-file -verify-diagnostics`:

- reset with both bound locks re-armed by `set_lock`: accepted;
- reset with no re-arm: rejected, with a note on each frozen lock;
- reset with a partial re-arm (one lock re-armed, one not): rejected, with a note
  on the frozen lock only, since a partial re-arm still deadlocks;
- reset on a channel that carries no objectFIFO locks (a raw DMA): accepted, so
  the pass does not false-positive;
- the same design on npu1 (AIE2): accepted, showing the check is target-model
  independent;
- the binding resolved through a mem tile (`aie.memtile_dma`): the frozen-lock
  case is rejected;
- reset and re-arm split across two runtime sequences: accepted, showing re-arms
  are collected module-wide.

### DIFF
--- a/include/aie/Dialect/AIEX/Transforms/AIEXPasses.h
+++ b/include/aie/Dialect/AIEX/Transforms/AIEXPasses.h
@@ -57,6 +57,8 @@ createAIELowerDmaChannelResetPass();
 std::unique_ptr<mlir::OperationPass<AIE::DeviceOp>>
 createAIELowerCoreResetPass();
 std::unique_ptr<mlir::OperationPass<AIE::DeviceOp>>
+createAIEVerifyRuntimeRearmPass();
+std::unique_ptr<mlir::OperationPass<AIE::DeviceOp>>
 createAIETransformBfpTypesPass();
 std::unique_ptr<mlir::OperationPass<AIE::DeviceOp>>
 createAIETxnToControlPacketPass();
--- a/include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
+++ b/include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
@@ -405,6 +405,17 @@ def AIELowerCoreReset : Pass<"aie-lower-core-reset", "AIE::DeviceOp"> {
   ];
 }
 
+def AIEVerifyRuntimeRearm : Pass<"aie-verify-runtime-rearm", "AIE::DeviceOp"> {
+  let summary = "Experimental, opt-in: reject an aiex.dma_channel_reset whose "
+                "objectFIFO lock is never re-armed (a necessary, not sufficient, "
+                "condition for a safe reset)";
+  let constructor = "xilinx::AIEX::createAIEVerifyRuntimeRearmPass()";
+  let dependentDialects = [
+    "xilinx::AIE::AIEDialect",
+    "xilinx::AIEX::AIEXDialect",
+  ];
+}
+
 def AIELowerScratchpadParameters : Pass<"aie-lower-scratchpad-parameters", "mlir::ModuleOp"> {
   let summary = "Lower scratchpad parameter ops and emit parameter-sync preambles";
   let description = [{
--- a/lib/Dialect/AIEX/Transforms/AIEVerifyRuntimeRearm.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEVerifyRuntimeRearm.cpp
@@ -33,8 +33,13 @@
 #include "mlir/IR/Matchers.h"
 #include "mlir/Pass/Pass.h"
 
+#include "llvm/ADT/STLExtras.h"
+#include "llvm/ADT/SmallPtrSet.h"
+#include "llvm/ADT/SmallVector.h"
+
 #include <array>
 #include <map>
+#include <optional>
 #include <set>
 #include <tuple>
 
@@ -158,8 +163,9 @@ struct AIEVerifyRuntimeRearmPass
           "resets a DMA channel whose objectFIFO lock is never re-armed; the "
           "semaphore counter stays frozen and the channel's peer will block "
           "forever on acquire (a runtime deadlock). Re-arm each bound lock "
-          "with "
-          "an aiex.set_lock");
+          "with an aiex.set_lock, or a write to its lock register "
+          "(aiex.npu.write32 / aiex.npu.maskwrite32 at the lock's local "
+          "address)");
       for (LockOp l : frozen)
         err.attachNote(l.getLoc())
             << "this lock is bound to the reset channel and is not re-armed";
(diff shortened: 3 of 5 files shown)


atassis:
@hunhoffe following up on testing the full reset sequence to replace the empty-load-PDI reset.

**State.** I composed and ran `core_reset` + `dma_channel_reset` + `set_lock` on aie2p. `core_reset` over all cores is bit-identical to the baseline and never hangs as a per-dispatch reset. The channel reset does not behave the same way: on its own, and even paired with `set_lock`, it deadlocks (`qds_device::wait() unexpected command state`).

**Cause.** On aie2p a DMA channel has no enable bit, so the reset drains the task queue and nothing restarts the channel, and the semaphore counters are left frozen. A channel reset only re-arms correctly when it is followed by a start-queue push (a write to `START_QUEUE` with the start BD id and repeat count) and a `set_lock` back to the objectFIFO init values. That holds inside the `aiex.configure` path, where the following configure re-streams the BDs, routing, and locks; it does not hold as a standalone per-dispatch reset.

**This PR** is a byproduct: an experimental, opt-in verifier that turns the missing-lock-re-arm case from a device hang into a compile error. It is a necessary, not sufficient, check (it does not verify the start-queue push), so I kept it out of the default pipeline.

**Next.** The full re-arm as an objectFIFO-scoped op (reset + start-queue push + `set_lock`, resolved from the fifo symbol), and the empty-load-PDI replacement in the `aiex.configure` path.

atassis:
@hunhoffe confirmed on device (aie2p). The full re-arm is `dma_channel_reset` + a `START_QUEUE` write (the channel's head BD id, plus the token bit for a channel that issues one) + `set_lock` back to the objectFIFO init values.

On a resident 3-core design, resetting the on-tile channels and re-arming them this way as a per-dispatch preamble is bit-identical to the baseline across 200 dispatches, with no hang. Reset alone, and reset with only `set_lock`, both deadlock; the `START_QUEUE` write is what restarts the channel. That matches the hardware: on aie2p a queue push is the only way to start a channel, so the reset draining the task queue is what strands it.

The start-queue push is a plain `write32`, so no new op is needed. Next is folding the three ops behind an objectFIFO-scoped re-arm, so the fifo supplies the BD id and lock values, and the empty-load-PDI replacement in the `aiex.configure` path.

hunhoffe:
@atassis This is great data! I'll review in a bit more detail -- this and your other PR -- when I have a bit more extra time, but I'll try to remember to restart the CI as needed for you.

atassis:
No trouble
I kept it drafted exactly for these reasons, wanted to hear AMD team's opinion 1st

andrej (commented):
Is this specific to ObjectFifo usage, or does this look at all locks used in BD chains and error if they are not reset from the runtime sequence?

I'm a little worried about duplicating all the write32s encodings being duplicated. Could we assume the pass runs before lowering `set_lock` and only check those?

It might be useful to have a pass that lifts write32s/blockwrites to their semantic MLIR ops so we don't have to duplicate enocdings everywhere.

atassis:
Scope: it triggers on an `aiex.dma_channel_reset` and checks the objectFIFO locks bound to that channel (resolved through the channel's `dma_start` + `use_lock` chain), not all BD-chain locks. No reset op, nothing checked.

On the write32 duplication: agreed. The pass runs before `set_lock` is lowered, so I've dropped the `write32`/`maskwrite32` recognition and now check only `aiex.set_lock`, which removes the lock-address decode I'd rather not carry here.

A pass that lifts `write32`/`blockwrite` back to their semantic ops would be the clean general fix, and it's the same reason I moved #3393's re-push from a hand-rolled `write32` to `aiex.npu.push_queue`. Worth its own issue if there's appetite.

andrej (approved):
I think this is a good addition to make the re-arming more user-friendly and catch potential errors.

However, I want to point out that I think that, as implemented, the check does not provide a _necessary_ condition for a successful reset, counter to what the description currently says. Locks might also be reset from the `aie.core`, rather than with a `set_lock`.

So the check might be too conservative. However, since it's currently opt-in and not on by default, I think that's fine.

But, before merging, can you please open an issue to document this shortcoming of this pass?

PR #3393, opened 2026-07-22
[AIE][AIEX] Add aiex.dma_channel_reset_for for resident objectFIFO re-arm

atassis:
## What this adds

`aiex.dma_channel_reset_for @objfifo` resets and re-arms every resident core/mem
DMA channel of an objectFIFO, so a resident (acquire-once, held on-chip)
objectFIFO can be reused across dispatches without carrying stale control state
into the next one. It names the fifo, not a raw channel, so the un-paired form
(resetting a channel without re-establishing its start queue and locks, which
deadlocks) is unrepresentable.

It builds on the merged `aiex.dma_channel_reset` (#3370) and lowers to the three
merged low-level ops, with no new register pokes: `dma_channel_reset` + a
`START_QUEUE` re-push (`npu.write32`) + `set_lock`, per non-shim endpoint.

## Why the re-push

I hit this re-arming a resident objectFIFO across dispatches on aie2p. On aie2p a
DMA channel has no enable bit, so once `dma_channel_reset` drains the channel's
task queue the only way to restart it is a `START_QUEUE` write. Re-arming the
locks alone leaves the channel idle; re-pushing alone runs a BD that stalls on a
stale lock. The op emits both, and reads the head BD id + repeat count off the
resident `aie.dma_start`, so the re-push reproduces the channel's load-time
arming exactly (same bd_id/repeat/token rule as `AIERT::pushToBdQueueAndEnable`,
with START_BD_ID masked to the tile class: 4 bits on a core tile, 6 on a mem
tile).

## The binding

`aie-objectFifo-stateful-transform` erases the objectfifo op and rewrites `@fifo`
uses to `@fifo_shim_alloc` (the shim endpoint only; a core-to-core fifo has
none), which drops the core/mem channels and locks the re-arm needs. So the
transform emits a new `aie.objectfifo_rearm_binding`, generalizing
`aie.shim_dma_allocation` to core/mem endpoints, and retargets the op at it. This
is on demand: a design with no `dma_channel_reset_for` emits nothing new, so its
transform output is unchanged.

## Scope

DMA control state only. Like `dma_channel_reset` it does not re-enable cores or
restream BDs, so it is not a general dispatch reset: a resident pipeline of
several fifos needs one op per fifo, and resident cores need a separate
`aiex.core_reset`. Shim endpoints and shim locks are skipped; the host owns
those.

## Tests

Round-trip and invalid tests for both ops; a transform test covering the binding
contents, the shim endpoint/lock skip, the adjacent (locks-only) fifo, and the
no-op when untargeted; the lowering on npu1 and npu2, core and mem tiles, with
exact register addresses and command words; and an end-to-end test through
`stateful-transform -> assign-bd-ids -> materialize-runtime-sequences -> lower`,
so the op is exercised through materialization, not the lowering pass alone.

## Notes for review

- I lower to the merged low-level ops rather than re-poking registers, so the
  reset pulse inherits their `maskwrite32` masking and stays correct across
  target models.
- I validated the reset + `START_QUEUE` re-push + `set_lock` sequence this emits
  on aie2p out of band: a resident conveyor reused across dispatches stayed
  bit-identical to the fetch baseline. I have not wired an in-tree npu-xrt e2e
  for the op itself; happy to add one if you would like it in this PR.


lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
```diff
@@ -0,0 +1,232 @@
...
+        uint32_t issueBit =
+            (ep.row == 0 && ep.dir == DMAChannelDir::S2MM) ? 0x80000000u : 0u;
+        uint32_t cmd = (ep.headBdId & bdIdMask) |
+                       ((static_cast<uint32_t>(ep.repeatCount) & 0xFF) << 16) |
+                       issueBit;
+        Value addr = createConstantI32(builder, loc, queueLocal);
+        Value val = createConstantI32(builder, loc, cmd);
+        NpuWrite32Op::create(builder, loc, addr, val, nullptr,
+                             builder.getI32IntegerAttr(ep.col),
+                             builder.getI32IntegerAttr(ep.row));
```
andrej: Don't we have a PushQueueOp? Do we need to duplicate the encoding here as a low-level write32?

lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
```diff
@@ -0,0 +1,232 @@
...
+    return ResolveResult::NoChannel;
+  };
+  for (MemOp mem : device.getOps<MemOp>())
+    if (mem.getTile() == tile.getResult())
+      return scanRegion(mem->getRegion(0));
+  for (MemTileDMAOp mem : device.getOps<MemTileDMAOp>())
+    if (mem.getTile() == tile.getResult())
+      return scanRegion(mem->getRegion(0));
+  return ResolveResult::NoChannel;
+}
```
andrej: Will this work in cases with mixed ObjectFifo and manual flows/bd programming usage?

Is re-scanning lowered ObjectFifo code really the best approach, or could we instead (a) either annotate the lowered ops with an attribute to show that these ops were associated with ObjectFifo X, or (b) generalize ShimDMAAllocationOp to carry this information?

andrej (changes requested):
Do we need a separate pass to lower this op? Or could we include that in the other pass you recently added?

lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
atassis: Following up on push_queue. Two snags switching to it:

1. `PushQueuetoWrite32Pattern` masks the head bd_id to 4 bits (`bd_id & 0xF`), but a mem tile's START_BD_ID field is 6 bits (register mask `0x3F`; `getNumBDs(MemTile) == 48`). So a mem-tile bd_id >= 16 truncates on the way to write32, a latent bug on the normal DMA path, not just here.
2. This pass runs after `aie-dma-to-npu`, so a `push_queue` it emits wouldn't be lowered without moving it ahead of that pass.

So switching to the op means (a) masking START_BD_ID per tile class in `aie-dma-to-npu` and (b) reordering this pass before it. I've folded both in here since they're one correctness unit; the re-arm stays bit-identical to the previous write32 (same command word and resolved queue address), which I'll also re-check on device before merge. Say if you'd rather see (a) as its own PR.

lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
atassis: The binding already does your (b): the stateful transform emits `objectfifo_rearm_binding` carrying each endpoint's tile/dir/channel and locks. The only rescan left is the head bd_id + repeat off the resident `dma_start` (not known until `aie-assign-bd-ids`, after the binding); it's keyed on the binding's own channel, so a manual flow elsewhere isn't touched. I can fold bd_id + repeat into the binding after assignment to drop the rescan if you'd rather.

atassis:
> Do we need a separate pass to lower this op? Or could we include that in the other pass you recently added?

It's an expansion, not a terminal lowering like `aie-lower-core-reset`: it emits `dma_channel_reset` + `set_lock` + `push_queue` for later passes to lower, so it has to precede all of them and can't fold into any one. Glad to group the re-arm expansions under one pass if you'd prefer.

lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
hunhoffe: I think this question also reveals a testing gap, which I've created an issue to track: https://github.com/Xilinx/mlir-aie/issues/3411

lib/Dialect/AIEX/Transforms/AIEDmaToNpu.cpp
```diff
@@ -183,11 +183,16 @@ struct PushQueuetoWrite32Pattern : OpConversionPattern<NpuPushQueueOp> {
...
+    // Command word: bd_id in the START_BD_ID field, repeat_count [23:16],
+    // issue-token bit [31]. START_BD_ID is 6 bits on a mem tile (48 BDs) and 4
+    // bits on core/shim tiles, so mask to the tile class instead of a flat 0xF
+    // (a flat 0xF silently truncates a mem-tile head bd_id >= 16).
     // bd_id and repeat_count may be runtime SSA; all-constant folds to one
     // constant (byte-identical to the static path), else built with arith.
     Location loc = op->getLoc();
     auto i32ty = rewriter.getIntegerType(32);
+    uint32_t bdIdMask =
+        tm.isMemTile(op.getColumn(), op.getRow()) ? 0x3Fu : 0xFu;
```
andrej, line 195: Thanks for fixing this!

lib/Dialect/AIEX/Transforms/AIELowerDmaChannelResetFor.cpp
andrej: Oh, thank you for clarifying this. Capturing this information in a dedicated op is good and I overlooked that. I think it would be best to keep all pertinent information with that op, rather than re-scanning, so including the BD and its repeat count. That way if (a) the ObjectFifo lowering changes its emitted BD structure down the line or (b) some other user wants to use the ream_for operato together with the ream_binding without actual ObjectFIFO code, all pertinent information is in the referenced op.

I don't know if the DMABDOp returns an SSA value. If it does, it might be simplest to just have that SSA value referenced in the rearm_binding (which would also help with the BD assignment, you can write the rearm_binding before the ID is assigned, or if the ID changes later on, the user of the rearm_binding can still pick it up).

andrej (changes requested):
Thank you @atassis for addressing my comments. The PushQueue fix is appreciated and will benefit uses beyond this re-arm functionality, if users want to program mem-tile BDs from the runtime sequence.

I'm sorry to keep going back and forth but would like to get this in the best possible shape before merging, so I'm making two more requests:
 
1. Please consider my comment about including all pertinent information about the re-arm inside the objectfifo_rearm_binding instead of re-scanning the emitted BD chain. If this turns out to be more difficult to implement, my mind can be changed about this, but if it's possible, I think I'd prefer that "cleaner" form that does not assume a certain shape of DMA emitted by ObjectFifos and leaves the door open for potential other uses of the objectfifo_rearm_binding op.
2. Please do merge all the rearm-lowering ops into one pass to keep consistent with our other lowering passes that group similar concerns in one pass. This will help reduce code maintenance burden.

Thanks again for the great thoughtful contributions!

atassis:
Hope the fix is sound

atassis:
@andrej Hi there! Both fixes are addressed already, can you review again? (DMABDOp has no result, so the ids fold in at assignment instead.)

andrej (approved):
Thank you for addressing my comments and for the useful contribution!

PR #3397, opened 2026-07-22
[test] Add npu-xrt local_reset on-board tests (core/dma)

jskimko:
## Description

Adds a `test/npu-xrt/local_reset/` suite of on-board tests for **local reset** -- returning a single tile-local block to a clean state from the runtime sequence, without rebuilding the design or reloading the array. Each family resets one block and checks that a resident design survives it.

**Direct register-write families** issue the reset from `aie.runtime_sequence`, using the same register-write type the `aie-rt` driver uses:

- `core` -- `Core_Control` (`0x32000`): masked reset -> unreset -> enable (`XAie_CoreReset` / `XAie_CoreUnreset` / `XAie_CoreEnable`) re-runs a halted core.
- `dma` -- `DMA_MM2S_0_Ctrl` (`0x1DE10`): masked reset pulse (`XAie_DmaChannelReset`), re-push BD, re-arm lock. **Updated:** the MM2S channel now holds a "bad" BD queued ahead of a "good" BD (both gated on a lock); the reset flushes the queue so only the good BD runs once the lock is armed.
- ~~`switch` -- `Stream_Switch_Slave_DMA_0_Config` (`0x3F104`): tear down the slave port at each dispatch end and re-enable it at the start of the next, re-arm lock (`XAie_StrmConnCctEnable` / `XAie_StrmConnCctDisable`; slave port only -- the master route is resident, noted in the README).~~ **Removed:** the stream switch is not a hardware reset domain (aie-rt has no stream-switch reset routine), so this was runtime connection reconfiguration rather than a reset.

**Op-based families** drive the merged runtime-sequence ops instead of writing the registers directly -- the on-board counterparts of `core` and `dma`, adding on-silicon coverage for ops that otherwise have only `aie-opt` FileCheck tests:

- `core_reset_op` -- `aiex.core_reset` (#3375). The op is reset-only, so for a core halted at `aie.end` the test composes it with a masked re-enable, giving the full `XAie_CoreReset -> XAie_CoreUnreset -> XAie_CoreEnable` driver sequence.
- `dma_channel_reset_op` -- `aiex.dma_channel_reset` (#3370), then re-push BD + re-arm lock.

Each family drives its protocol across repeated dispatches and self-checks the host-visible result; the op-based families reuse their raw sibling's `test.cpp` directly. All run on npu1 and npu2 (`REQUIRES: ryzen_ai`); every register offset is identical on both. Every offset, register-write type, and reset protocol is cross-referenced in the READMEs to the vendored aie-rt driver (`third_party/aie-rt/`).

## Checklist


jskimko:
I'll wait on #3375 to be merged, then add tests using the new API for both `core` and `dma`.

hunhoffe:
https://github.com/Xilinx/mlir-aie/pull/3375 in merge queue :)

test/npu-xrt/local_reset/core/test.cpp
```diff
@@ -0,0 +1,119 @@
...
+  auto device = xrt::device(device_index);
+  auto xclbin = xrt::xclbin(vm["xclbin"].as<std::string>());
+  std::string Node = vm["kernel"].as<std::string>();
+
+  auto xkernels = xclbin.get_kernels();
+  auto xkernel = *std::find_if(xkernels.begin(), xkernels.end(),
+                               [Node](xrt::xclbin::kernel &k) {
+                                 return k.get_name().rfind(Node, 0) == 0;
+                               });
+  auto kernelName = xkernel.get_name();
```
andrej:
```suggestion
  std::string kernelName = vm["kernel"].as<std::string>();
```

(A lot of our test.cpps have this, but I'd like to not keep propagating it.)

test/npu-xrt/local_reset/core_reset_op/aie.mlir
```diff
@@ -0,0 +1,83 @@
...
+      // run to aie.end and is no longer enabled, so the op alone does not restart
+      // it. Re-enable it with a masked write of the ENABLE bit (CORE_CONTROL bit 0,
+      // mask 0x1), mirroring aie-rt's XAie_CoreEnable (a MaskWrite32 of the enable
+      // field). op + this write is the full XAie_CoreReset -> XAie_CoreUnreset ->
+      // XAie_CoreEnable driver sequence, all masked so no other CORE_CONTROL field
+      // is clobbered.
+      %cc = arith.constant 204800 : i32
+      %enable = arith.constant 1 : i32
+      %en_mask = arith.constant 1 : i32
+      aiex.npu.maskwrite32(%cc, %enable, %en_mask) {column = 0 : i32, row = 2 : i32} : i32, i32, i32
```
andrej, line 76: Should `aiex.core_reset` be updated to include this "enable" step? Or should we add another op `aiex.core_enable`? (Not here, just asking for future PR.)

test/npu-xrt/local_reset/dma/test.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto xclbin = xrt::xclbin(vm["xclbin"].as<std::string>());
+  std::string Node = vm["kernel"].as<std::string>();
+  auto xkernels = xclbin.get_kernels();
+  auto xkernel = *std::find_if(xkernels.begin(), xkernels.end(),
+                               [Node](xrt::xclbin::kernel &k) {
+                                 return k.get_name().rfind(Node, 0) == 0;
+                               });
+  device.register_xclbin(xclbin);
+  xrt::hw_context context(device, xclbin.get_uuid());
+  auto kernel = xrt::kernel(context, xkernel.get_name());
```
andrej:
```suggestion
  std::string kernelName = vm["kernel"].as<std::string>();
  device.register_xclbin(xclbin);
  xrt::hw_context context(device, xclbin.get_uuid());
  auto kernel = xrt::kernel(context, kernelName);
```

test/npu-xrt/local_reset/switch/test.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto xclbin = xrt::xclbin(vm["xclbin"].as<std::string>());
+  std::string Node = vm["kernel"].as<std::string>();
+  auto xkernels = xclbin.get_kernels();
+  auto xkernel = *std::find_if(xkernels.begin(), xkernels.end(),
+                               [Node](xrt::xclbin::kernel &k) {
+                                 return k.get_name().rfind(Node, 0) == 0;
+                               });
+  device.register_xclbin(xclbin);
+  xrt::hw_context context(device, xclbin.get_uuid());
+  auto kernel = xrt::kernel(context, xkernel.get_name());
```
andrej:
```suggestion
  std::string kernelName = vm["kernel"].as<std::string>();
  device.register_xclbin(xclbin);
  xrt::hw_context context(device, xclbin.get_uuid());
  auto kernel = xrt::kernel(context, kernelName);
```

test/npu-xrt/local_reset/switch/aie.mlir
```diff
@@ -0,0 +1,74 @@
...
+      // isolates any failure to that one port.
+      %slave = arith.constant 258308 : i32
+      %enable = arith.constant -2147483648 : i32
+      aiex.npu.write32(%slave, %enable) {column = 0 : i32, row = 2 : i32} : i32, i32
+
+      // Re-arm the cons lock (LOCK0_VALUE, 0x1F000 = 126976; aie-rt
+      // XAie_LockSetValue, a full-word write32) so the send proceeds.
+      %la = arith.constant 126976 : i32
+      %one = arith.constant 1 : i32
+      aiex.npu.write32(%la, %one) {column = 0 : i32, row = 2 : i32} : i32, i32
```
andrej: Would `aiex.set_lock` work here? I think higher-level ops would be preferrable where available for understandability.

andrej (approved):
Looks good, my comments are mostly nitpicks, but should be easy to address if you could.

Follow-ups:
- Getting native MLIR operations for those raw-register writes would be useful. This helps with portability for future architectures, as well as understandability and debuggability. Lowerings for these additional ops could probably be tacked onto @atassis' existing core reset lowering pass.
- For the switchbox and DMA channel reset, I think it would be especially interesting if it is possible to change the switchbox or DMA channel after the reset. It looks to me that these tests currently re-run the same configuration after resetting and unresetting as before. I'd love to see this working for a reconfiguration, e.g. reset the switchbox and route it somewhere else between iterations of the test.

test/npu-xrt/local_reset/core_reset_op/aie.mlir
jskimko: This discussion was also in #3375; we should mirror the driver's protocol and add a `core_enable` op instead.

jskimko:
- Replaced the `dma` test to avoid flaky sync issues by observing a queue reset instead.
- Dropped the `switch` test since this turned out to be more of a reconfiguration test rather than a true reset.

andrej (approved):
Discussed with Jason: the deleted reconfiguration tests might be added in a future PR.

Issue #3396, opened 2026-07-22
Inlining external functions


hunhoffe:
### What are you trying to do?

External function calls happen a lot in IRON, since we typically keep the control loops in Python, while the kernels themselves are in C++. The aggregate cost of many of these calls within a loop can cause measurable performance degradation. 

This also affects mlir-air (@erwei-xilinx ) so bonus points for a portable solution for both projects.

### Proposed solution

We need to optimize external function calls on `Worker()` cores -- perhaps by doing something different with linking or with automatic inlining or something else. Microbenchmarks on function call latency would help motivate/prove any potential solution.

hunhoffe:
I believe this was adequately addressed by https://github.com/Xilinx/mlir-aie/pull/3399 and https://github.com/Xilinx/mlir-aie/pull/3493

PR #3399, opened 2026-07-23
feat(iron): inline external kernels into the core (no func.call, no separate .o)

fifield:
External function calls dominate tight IRON loops: control stays in Python while kernels are C++, so a core issues a func.call per tile (issue #3396). Add an opt-in inline path that folds the kernel body directly into the aie.core.

ExternalFunction(inline=True) / compile_cxx_core_function(inline=True): compile the C++ kernel to alwaysinline + linkonce_odr LLVM IR (.ll) via clang -S -emit-llvm instead of a .o. The func.func declaration carries a .ll link_with.

aiecc: link_files entries ending in .ll/.bc are routed to an IR-merge list, llvm-link'd into the core LLVM module before opt/llc, and excluded from object linking (ld-script INPUT() / BCF _include) so each symbol is merged once. The always-inliner folds the alwaysinline body in; linkonce_odr lets globaldce drop the now-dead definition. Per-core and unified paths; Peano only (Chess cannot llvm-link).

inline + symbol_prefix is rejected: an inline kernel is emitted as LLVM IR and cannot be symbol-renamed via llvm-objcopy.

Tests/bench:
- test/python/external_func_inline_link_with.py: IR-only check that inline=True declares a .ll link_with (no hardware).
- test/python/npu-xrt/test_jit_extern_functions_inline.py: end-to-end on NPU; the inline result matches the object-linked path.
- programming_examples/basic/inline_kernel: object-vs-inline latency microbench.

Measured Performance (best of 3, iters=200):

  | calls/iter |   object |   inline | speedup |
  |------------|----------|----------|---------|
  |        256 | 0.244 ms | 0.209 ms |   1.17x |
  |       1024 | 0.340 ms | 0.261 ms |   1.30x |
  |       4096 | 0.970 ms | 0.452 ms |   2.15x |
  |      16384 | 2.894 ms | 1.488 ms |   1.95x |


erwei-xilinx (approved):

programming_examples/basic/inline_kernel/inline_kernel.py
```diff
@@ -0,0 +1,155 @@
...
+        arg_types=[
+            np.ndarray[(16,), np.dtype[np.int32]],
+            np.ndarray[(16,), np.dtype[np.int32]],
+            np.int32,
+        ],
+        inline=inline,
+    )
+
+
+def _bench(label: str, inline: bool, num_elements: int, iters: int) -> float:
```
hunhoffe: There's some utility functions for benchmarking, example usage here: https://github.com/Xilinx/mlir-aie/blob/cd9d998ffcd6de3788a63eff1fdfeecfdee3b617/programming_examples/getting_started/00_memcpy/memcpy.py#L38

hunhoffe (approved):
Thanks for getting this ready to go!!

Might be nice to use some of the newer Python utils in the example, but that's just a nice-to-have.

jgmelber (approved):

PR #3401, opened 2026-07-23
Add Windows CI workflow for Ryzen AI self-hosted runner

jgmelber:
## Summary
- Adds `buildAndTestRyzenAIWindows.yml`, a new CI workflow that builds mlir-aie from source against the published MLIR wheel and runs the lit suite on a self-hosted Windows Ryzen AI (npu2) runner.
- Mirrors `buildAndTestRyzenAI.yml`'s from-source Linux job, but:
  - **Peano-only**: Vitis/xchesscc isn't part of the supported native-Windows toolchain (`docs/buildHostWinNative.md`), and `AIE_BUILD_CHESS_CLANG` is force-disabled under `WIN32` in the top-level `CMakeLists.txt` regardless.
  - **JIT examples only, for now**: Windows has no `make` on PATH by default, so Makefile-based `programming_examples`/`programming_guide` tests are automatically skipped by lit's existing `makefile_examples` feature gate. `@iron.jit`-style Python examples still run and exercise the NPU end to end. Broader Makefile coverage is a follow-up, not part of this PR.
- Also fixes `.gitignore`'s `build*` pattern silently matching new files under `.github/workflows/`.

## Windows-portability fixes discovered on real hardware
Getting the suite green on the NPU surfaced several native-Windows issues, fixed here:
- **`actions/setup-python` removed** — its installer runs a `setup.ps1` blocked by the runner's script execution policy; Python 3.13 is a host prerequisite instead (mirrors the Linux self-hosted job).
- **`bash.exe` by full path** — Git for Windows only puts `Git\cmd` on PATH, not `Git\bin`, so `utils/clone-llvm.sh` is invoked via its full install path.
- **Forward slashes in CMake `-D` args** — CMake parses `\W`, `\a`, etc. in `-D` string values as C escapes; all path args are converted to `/`.
- **OpenSSL** — `tools/bootgen` unconditionally requires OpenSSL; the workflow passes `-DOPENSSL_ROOT_DIR` at the documented install path.
- **`diaguids.lib` path fixup** — the published wheel's `LLVMExports.cmake` bakes in an absolute DIA SDK path from the VS Enterprise build machine; the workflow reuses the repo's own `_fixup_llvm_diaguids()` to rewrite it for the runner's VS edition.
- **git `safe.directory`** — `iron_setup.py --dev` installs pre-commit hooks that shell out to git; the runner-owned work tree is marked safe so git doesn't reject it for dubious ownership.
- **`%/t` in XCLBIN test literals** — the xclbin path was injected into a C++ string literal with native backslashes (escape-sequence corruption); switched to lit's forward-slash `%/t`.
- **`SIZE` constant renamed** — XRT's Windows headers transitively pull in `windef.h`, whose `tagSIZE` struct collides with the `constexpr int SIZE` in a few host tests (`-DNOGDI` does not suppress it, since the struct is in `windef.h`, not `wingdi.h`); renamed the local constant to `BUF_SIZE`.
- **`MPLBACKEND=Agg` for headless matplotlib** — `exercise_5b` calls taplib's `visualize()`, which unconditionally calls `plt.show()`; on the headless runner the interactive backend blocks forever (600s lit timeout). The job sets `MPLBACKEND=Agg`, and lit's env passthrough allowlist is extended to forward it into the test (Linux CI has no `DISPLAY`, so it already falls back to Agg).

## Host prerequisites
The runner-setup steps (LLVM for `llvm-objcopy`, full Win64 OpenSSL, DIA SDK via VS, antivirus exclusion for the work tree, disabling sleep) are documented separately in the Windows runner setup guide.

## Pre-existing Windows-only bugs (XFAIL'd, tracked for follow-up)
Three tests that pass on Linux fail on the Windows NPU job from platform bugs
unrelated to the tests themselves. They are `XFAIL: system-windows` so this
job can land and protect `main`; each is tracked under #3439 (with sub-issues):
- **`matmul_whole_array_dynamic`** (#3437) — Windows LLP64 infers a Python int beside the runtime `i32` M/K/N args as `index`-typed (`python/helpers/util.py` maps `np.longlong`/`np.uintp` → `T.index`), so the vendored `eudsl-python-extras` coerce path widens with `arith.extsi` (illegal to `index`) instead of `arith.index_cast`. Bug is in the Python IR-construction layer, not the test.
- **`test_jit_many_args.py` / `many_buffers`** (#3438) — both compile cleanly but the kernel aborts on the NPU (`ERT_CMD_STATE_ABORT` / `status 6`). Both exercise the >5 host-buffer DDR-address-patch path (firmware only pre-translates the first 5 buffer addresses); a Windows-only runtime failure downstream of codegen.

## Status
The workflow builds from source and runs the full lit suite on real NPU hardware, **green end to end**: `check-aie` **1007 passed / 184 unsupported (no `make`/chess, expected) / 8 expectedly-failed (5 pre-existing xfail + 3 Windows XFAILs above) / 0 failed**, plus `check-aie-concurrency`, `check-reference-designs`, and `check-programming-guide` all passing. The 3 XFAIL'd tests are pre-existing platform bugs tracked in #3439 for follow-up.

## Test plan
- [x] Self-hosted Windows runner (labels `self-hosted, windows, ryzenai, npu2`) registered and online
- [x] Workflow triggers and completes the from-source build on real hardware
- [x] `check-aie` runs on the NPU
- [x] Portability failures fixed (SIZE rename, `%/t` xclbin literals, headless matplotlib)
- [x] 3 pre-existing Windows-only bugs XFAIL'd with tracking issues (#3439)
- [x] Fully green run on real NPU hardware (with XFAILs)

### DIFF
--- a/.github/workflows/buildAndTestRyzenAIWindows.yml
+++ b/.github/workflows/buildAndTestRyzenAIWindows.yml
@@ -1,0 +1,246 @@
+# Copyright (C) 2023-2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+name: Build and Test with AIE tools on Ryzen AI (Windows)
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+      - main
+      - test-ryzen-ai
+  pull_request:
+    types: [opened, synchronize, reopened]
+  merge_group:
+  schedule:
+    # Runs at midnight (00:00) every day
+    - cron: '0 0 * * *'
+
+concurrency:
+  # A PR number if a pull request and otherwise the commit hash. This cancels
+  # queued and in-progress runs for the same PR (presubmit) or commit
+  # (postsubmit).
+  group: ci-build-test-ryzenai-windows-${{ github.event.number || github.sha }}-${{ github.event_name }}
+  cancel-in-progress: true
+
+env:
+  # Peano only: Vitis/xchesscc is not part of the supported native-Windows
+  # toolchain (see docs/buildHostWinNative.md); AIE_BUILD_CHESS_CLANG is
+  # force-disabled on WIN32 in the top-level CMakeLists.txt regardless.
+  LIT_OPTS: -sv --time-tests --timeout 600 --show-unsupported --show-excluded
+
+jobs:
+
+  build-and-test-from-source:
+    name: Run Tests and Examples on Ryzen AI (Built from Source, Windows)
+    # Mirrors buildAndTestRyzenAI.yml's Linux job: build mlir-aie from source
+    # against a published MLIR wheel and run the lit suite on real NPU
+    # hardware. Windows has no Makefile (`make`) on PATH by default, so
+    # Makefile-based programming_examples/programming_guide tests are
+    # automatically skipped by lit (see the "makefile_examples" feature gate
+    # in python/aie_lit_utils/lit_config_helpers.py); JIT-style Python
+    # examples (@iron.jit) still run and cover the NPU end to end.
+    runs-on: [self-hosted, windows, ryzenai, npu2]
+    env:
+      NPU_CACHE_HOME: ${{ github.workspace }}\iron-cache-${{ github.run_id }}
+      PIP_CACHE_DIR: ${{ github.workspace }}\.pip-cache-${{ github.run_id }}
+      # Force matplotlib's non-interactive backend: taplib visualize() calls
+      # plt.show(), which blocks forever on this headless runner (Linux CI has
+      # no DISPLAY, so it already falls back to Agg).
+      MPLBACKEND: Agg
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          submodules: "true"
+
+      - uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756  # v1
+
+      - name: Verify Python
+        # actions/setup-python isn't used here: its installer invokes a
+        # setup.ps1 under Windows PowerShell, which is blocked by this
+        # runner's default script execution policy. Python 3.13 is already
+        # provisioned on the runner as a host prerequisite (see
+        # Windows_runner_setup_mlir-aie.md), matching how the Linux
+        # self-hosted job also relies on a pre-provisioned interpreter
+        # instead of actions/setup-python.
+        shell: cmd
+        run: |
+          python --version
+          if errorlevel 1 exit /b 1
+
+      - name: Setup IRON environment
+        shell: cmd
+        run: |
+          REM Avoid git "dubious ownership" rejection when --dev's pre-commit
+          REM hook install shells out to git in the runner-owned work tree.
+          git config --global --add safe.directory "%CD%"
+          if errorlevel 1 exit /b 1
+
+          python utils\iron_setup.py --dev
+          if errorlevel 1 exit /b 1
+
+      - name: Download MLIR wheel and build mlir-aie
+        shell: cmd
+        run: |
+          call .\iron_env.cmd
+          if errorlevel 1 exit /b 1
+
+          REM utils\clone-llvm.sh is a bash script; Git for Windows ships its
+          REM own bash.exe (installed as part of "Git for Windows" in the
+          REM runner setup guide) that can run it directly. Its bin directory
+          REM isn't on PATH by default (only Git\cmd, for git.exe itself is),
+          REM so call it by its default winget install path.
+          if not exist "C:\Program Files\Git\bin\bash.exe" (
+            echo ##[error]C:\Program Files\Git\bin\bash.exe not found - install Git for Windows per the runner setup guide
+            exit /b 1
+          )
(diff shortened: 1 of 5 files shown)


.github/workflows/buildAndTestRyzenAIWindows.yml
```diff
@@ -0,0 +1,215 @@
...
+          call .\iron_env.cmd
+          if errorlevel 1 exit /b 1
+
+          REM utils\clone-llvm.sh is a bash script; Git for Windows ships its
+          REM own bash.exe (installed as part of "Git for Windows" in the
+          REM runner setup guide) that can run it directly. Its bin directory
+          REM isn't on PATH by default (only Git\cmd, for git.exe itself is),
+          REM so call it by its default winget install path.
+          for /f "usebackq delims=" %%v in (`"C:\Program Files\Git\bin\bash.exe" utils/clone-llvm.sh --get-wheel-version`) do set MLIR_VERSION=%%v
+          if not defined MLIR_VERSION exit /b 1
```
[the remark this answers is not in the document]
jgmelber: @copilot Fix the code for all comments in [this review comment](https://github.com/Xilinx/mlir-aie/pull/3401#discussion_r3661674034).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

.github/workflows/buildAndTestRyzenAIWindows.yml
```diff
@@ -0,0 +1,215 @@
...
+
+          REM CMake treats backslashes in -D string args as C-style escapes
+          REM (e.g. "\W" isn't valid, and "\a"/"\t"/etc. would be silently
+          REM misinterpreted), so pass every path with forward slashes,
+          REM which CMake accepts natively on Windows.
+          set "CD_FWD=%CD:\=/%"
+          set "PEANO_INSTALL_DIR_FWD=%PEANO_INSTALL_DIR:\=/%"
+          set "XRT_ROOT_FWD=%XRT_ROOT:\=/%"
+          set "PYTHON_EXE_FWD=%PYTHON_EXE:\=/%"
+          set "LIT_EXE_FWD=%LIT_EXE:\=/%"
```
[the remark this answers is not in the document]
jgmelber: @copilot Fix the code for all comments in [this review comment](https://github.com/Xilinx/mlir-aie/pull/3401#discussion_r3661674047).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review comment.

.github/workflows/buildAndTestRyzenAIWindows.yml
```diff
@@ -0,0 +1,240 @@
...
+  build-and-test-from-source:
+    name: Run Tests and Examples on Ryzen AI (Built from Source, Windows)
+    # Mirrors buildAndTestRyzenAI.yml's Linux job: build mlir-aie from source
+    # against a published MLIR wheel and run the lit suite on real NPU
+    # hardware. Windows has no Makefile (`make`) on PATH by default, so
+    # Makefile-based programming_examples/programming_guide tests are
+    # automatically skipped by lit (see the "makefile_examples" feature gate
+    # in python/aie_lit_utils/lit_config_helpers.py); JIT-style Python
+    # examples (@iron.jit) still run and cover the NPU end to end.
+    if: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
```
jgmelber:
```suggestion
```

Copilot suggests to guard against fork PRs but we typically allow this.

.github/workflows/buildAndTestRyzenAIWindows.yml
```diff
@@ -0,0 +1,240 @@
...
+  push:
+    branches:
+      - main
+      - test-ryzen-ai
+  pull_request:
+    types: [opened, synchronize, reopened]
+  merge_group:
+  schedule:
+    # Runs at midnight (00:00) every day
+    - cron: '0 0 * * *'
```
jgmelber, line 19: Tighten this up to match similar actions.

.github/workflows/buildAndTestRyzenAIWindows.yml
andrej: I believe our current policy is that external contributors PRs have to be approved from the web interface to run on CI, so I agree that this should be fine.

test/python/npu-xrt/test_jit_many_args.py
```diff
@@ -6,6 +6,10 @@
 # RUN: %run_on_npu1% %pytest %s
 # RUN: %run_on_npu2% %pytest %s
 # REQUIRES: xrt_python_bindings
+# XFAIL: system-windows
+# Compiles, but the kernel aborts on the NPU (ERT_CMD_STATE_ABORT) on Windows.
+# Exercises the >5 host-buffer DDR-address-patch path; pre-existing Windows-only
+# runtime failure, tracked separately.
```
andrej, line 12: We discovered that the firmware shipped with the Linux driver adds an offset to the first five DDR buffers, which is needed to access DDR from the command processor. It only does this for the first five, and that number is hard-coded. Erika's fix was to add the same offset manually for buffers 6 and beyond to work around this.

Most likely, the firmware shipped with the Windows driver does _not_ do the same address patching, which would be the root of this failure.

andrej (approved):
Awesome! Thanks for setting this up.

Maybe as future direction, we could also
- add a "smoke-test wheels" job for Windows
- make the `./utils/build-mlir-aie-from-wheels.sh` script work on Windows as well? (probably very low prio)

PR #3403, opened 2026-07-23
ci: opt-in compile/execute split for npu-xrt tests (pilot/RFC)

atassis:
> [!WARNING]
> **DO NOT MERGE AS-IS.** This draft carries a *temporary* validation harness
> (`.github/workflows/zz-experiment-split-validate.yml` + `.github/workflows/scripts/{split,cache}_experiment.sh`)
> whose only job is to gather Chess correctness/timing on a Chess-capable runner. It triggers on
> `pull_request` and runs on the self-hosted device runners, so if this PR were merged unremoved it would
> become a **permanent second per-PR device build on the capacity-1 runners** -- the exact opposite of this
> PR's aim. It must be deleted before the split is finalized. Only the lit split below is proposed for review;
> the harness is evidence-gathering scaffolding.

## Problem

The on-device `check-aie` phase is the biggest CI sink, and it is almost entirely compilation. On a
serialized `npu-xrt` test I measured device-side work at ~20 ms (device open dominates; pure NPU dispatch
0.39 ms) versus ~2.1 s of Peano compile for a one-kernel test and 26-77 s for Chess tests. So the
capacity-1 `npu-xrt` lane spends ~99% of its wall time compiling, which never touches the device.

## Change (opt-in, off by default)

Let a converted `npu-xrt` test run its compilation in a parallel, no-device phase and keep only the device
execution in the capacity-1 `npu-xrt` group. `AIE_NPU_SPLIT` selects the phase:

- unset (default): both substitutions expand to nothing, so behaviour is unchanged.
- `compile`: run the build lines (`%npu_build%`), skip the device run (`%npu_run%` becomes `:`); the test
  is ungrouped so builds run at `-jN`.
- `execute`: skip the build, run only the device lines against the compile phase's persisted `%t`
  artifacts; the test stays in the capacity-1 `npu-xrt` group.

CI would run the split subset as two passes: `AIE_NPU_SPLIT=compile lit ...` then
`AIE_NPU_SPLIT=execute lit ...`. Converting a test is a per-line prefix (`%npu_build%` on build lines,
`%npu_run%` on the `./test.exe` line).

This preserves #2737's invariant: the capacity-1 group serializes the global per-device XRT `hw_context`
pool, which is consumed only at device-execution time; compilation never opens the device. Per-test
parallelism gating keeps every unconverted device test in the capacity-1 group during the compile phase,
so running the whole directory in compile mode never parallelises unserialized device work.

## Coverage

57 `npu-xrt` tests converted by a classifier with a `--validate` pass (0 invariant violations). 14 are
left whole-test on purpose: JIT (`@iron.jit`, no separable compile/execute boundary), multi-line
continuations, and redirect/pipe/FileCheck lines that the `:` skip token cannot neutralise safely. The
dangerous classifier error is tagging a device run as `%npu_build%`; `--validate` guards exactly that.

## Evidence (Ryzen AI 9 / AIE2P)

- Behaviour-identical across the full runnable set: on a commit-matched from-source build, all 39 runnable
  Peano `npu-xrt` tests give the same verdict whole vs split. A fresh-root two-arm control is the
  load-bearing proof: execute-only from a wiped root is 0/39 PASS (no stale artifact can ride),
  compile+execute from the same wiped root is 39/39 PASS (genuine split).
- No masking, no stale ride (fault injection): for representative tests I mutate the kernel so its output
  is wrong. whole and split both FAIL identically, and split FAILs even when a prior good split has left
  good artifacts in the exec root, so the compile phase overwrites them and the execute phase never rides a
  stale-good xclbin. Reverting the mutation restores PASS.
- ~4x on this Peano subset: whole 54.2 s serial vs split 13.4 s (9.66 s parallel compile at `-j10` plus
  3.75 s serial device execute).

## What I have not measured (honest caveats)

- The Chess tests are converted and codemod-validated, but their whole-vs-split behaviour and timing are
  projected, not measured: my box is npu2/Peano and skips Chess. The compile/execute seam
  (`%run_on_npuN% ./test.exe`) is backend-agnostic, so Chess correctness is expected to hold, but I have
  not run it. The `check-aie` win is largest exactly on the Chess compiles (26-77 s), so that is the
  measurement I most want from a Chess-capable runner.
- Timing is n=1 per phase, and whole runs first, warming the aiecc on-disk JIT cache the compile phase then
  reuses, so the ratio is biased favourable; a cold or repeated-median run would tighten it.

## Scope / not in this PR

- No workflow YAML change: how and whether to wire the two passes into CI is the maintainers' call. Related
  to the CI overhaul in #3198.
- No cross-run xclbin cache: that needs a content-addressed input + toolchain key and is separate, higher
  risk.

Opening as a draft / RFC to show the mechanism and the measured win concretely. Happy to adjust the scope,
the marker names, or the phase model.

## Tracking / next

Discussion and maintainer thread: #3405. Per hunhoffe's note there that Chess compiles are memory-intensive
(they were OOM-killing runners without a Chess-specific concurrency limit), the compile phase would carry
its own memory-sized parallelism group, so Chess compiles stay as bounded as today while leaving the
capacity-1 device lane. I have a temporary, self-guarding validation job (whole==split assert plus a
fresh-root canary and no-masking fault injection, all encoded as CI assertions) ready to run on a
Chess-capable runner, to turn the projected Chess correctness/timing and the compile memory footprint into
measured numbers.

### DIFF
--- a/.github/workflows/scripts/cache_experiment.sh
+++ b/.github/workflows/scripts/cache_experiment.sh
@@ -1,0 +1,243 @@
+#!/usr/bin/env bash
+# EXPERIMENT (TEMPORARY -- removed before any cache PR is finalized).
+#
+# SEPARATE ARM from split_experiment.sh (untouched, independently readable). Proves a
+# per-test, content-addressed xclbin cache on real Ryzen AI hardware.
+#
+# THIS ARM IS MEASUREMENT-ONLY (INTRA-RUN). It demonstrates reuse WITHIN one invocation
+# (cold->warm) and proves the cache is SAFE. Genuine cross-*job* persistence (actions/cache)
+# is deliberately deferred to a follow-up so the stripped, self-contained version is validated
+# on AMD's CI first -- see the commented-out cache steps in zz-experiment-split-validate.yml.
+#
+#   [A] cold vs warm (INTRA-RUN reuse) : cold = compile every converted npu-xrt test and
+#        POPULATE the cache; warm = wipe build artifacts, RESTORE from cache, compile only the
+#        misses (~0). NOTE: warm reuses what COLD just wrote in THIS run -- it quantifies the
+#        compile time a cache WOULD avoid; it is NOT cross-job persistence (see prehit below).
+#   [B] hit is CORRECT : after the warm restore, EXECUTE every test and require the SAME verdict
+#        set as cold. A cache hit must be a correct run, not merely a fast one.
+#   [C] invalidation self-test (row 4, the load-bearing guard): cache a good kernel -> mutate its
+#        source -> assert (1) key CHANGES, (2) cache MISSES, (3) the mutated build COMPILES cleanly
+#        then EXECUTES to FAIL (a broken kernel cannot ride a stale hit / mask a regression).
+#
+# Correctness is read from lit's `-o` JSON report, never `-s`-suppressed stdout. Guards FAIL LOUD
+# not only on TOTAL vacuity but on (a) PARTIAL coverage loss (an enumerated test silently missing
+# from the report -- lit only WARNS on an unresolved input) and (b) a missing positive PASS baseline
+# (comparing two uniformly-broken sides proves nothing).
+set -uo pipefail
+
+TREE="${TREE:-$PWD}"
+BUILD="${BUILD:-$TREE/build}"
+NPU_XRT_SRC="$TREE/test/npu-xrt"
+OUT="${OUT:-$TREE/cache-experiment-out}"
+# CACHE_DIR MUST live outside $OUT so the start-of-run `rm -rf $OUT` cannot wipe it.
+CACHE_DIR="${CACHE_DIR:-$TREE/xclbin-cache-exp}"
+JOBS="${JOBS_COMPILE:-10}"
+LIT="${LIT:-lit}"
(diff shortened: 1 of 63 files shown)


atassis:
@hunhoffe Hi! This PR might be even more useful, can I ask you to run CI workflow when NPU is kinda freeish? It might substantially reduce the duration of blocking test, which takes a lot of time to run. Also, I have landed a discussion per Ci topic #3405 . Just letting you know, no hurry.

hunhoffe:
@atassis I think you have a good handle on the risks here, what I was most concerned about was that we've had issues with lit tests being run from the same directory which led to collisions between different tests in build directories. I'd really want to be certain we are matching the right artifacts to the right test.

atassis:
@hunhoffe good point, I'll try to derisk it in this PR, doesn't look like a big change.
CI-fail wise- it's a great start, I'll think through it later, this is exactly what I need!

atassis:
I think I might bank this PR for a while. It might be easier to migrate tests entirely onto Peano, then make them simpler, so they can run separately from the build step in CI.
For now a lot of moving parts, which can break.

Still, I want to ask you to start this workflow again when merge queue if free, I want to have a look at my experimental workflow again

atassis:
Came to a conclusion that this won't improve anything in current state. Solving chess migration will be far better improvement than splitting onto build -> test steps

PR #3407, opened 2026-07-24
Audit standalone Python/JIT programming examples

thomthehound:
This should close the audit of standalone, self-verifying Python/JIT paths in `programming_examples` and the *straightforward* portions of `programming_guide` as started with #3280.

Despite the large number of files touched, this pass has a narrow focus. For the examples and guides touched, the Python/JIT entry points are now the source of truth for direct lit coverage, while a Makefile remains *only* when it serves as more than a simple wrapper for the underlying JIT implementation. This avoids running a lit test on a wrapper and then again on the CLI invocation that very wrapper was already using.

## Summary of changes

- Add direct NPU1/NPU2 lit coverage for the remaining mechanically safe standalone JIT examples and guide scripts.
- Preserve existing native-host and Makefile tests where they exercise a genuinely different path; rename the ML tests to `run_makefile.lit` where needed so the distinction is explicit.
- Remove 13 Makefile wrappers that only aliased Python CLI modes, trivial output redirection, and cleanup, together with their duplicate Makefile-gated lit tests.
- Update example READMEs, the published programming guide, the generic getting-started/build documentation, and the Ryzen AI wheel smoke job to use the canonical Python interfaces, as this is the intended path for new users to learn and the only one that can be guaranteed to work on both Linux and Windows.
- Very light documentation updates to keep hardware guidance simple: since runnable JIT commands can auto-detect the attached NPU, `--dev` is typically omitted. An added guide-wide note explains when offline `--emit-mlir` commands should use `npu` or `npu2` to get the exact artifacts discussed. (I technically didn't need to do this, but I was already editing the READMEs for the programming guide and realized `npu2` users might get the wrong impression from them as written).

The workflow edit is an incidental consequence of these changes: the wheel smoke runs `passthrough_kernel`, just as before, but now invokes it through the canonical Python CLI because the wrapper Makefile is removed. This is functionally identical.

As promised in #3280, this audit also fixes concrete verifier and input-contract defects found while comparing Python and native-host paths:

- `row_wise_bias_add/test.cpp` now returns failure after a mismatch instead of printing `FAIL.` and exiting successfully.
- `memcpy`, `passthrough_pykernel`, and `vector_vector_modulo` use the same deterministic input contracts as their applicable native hosts.
- `vector_reduce_add` and `vector_reduce_min` use one deterministic rotated signed ramp in Python and C++, replacing unrelated random inputs.
- modulo diagnostics describe modulo rather than addition.

## Future work

This pass *does not* attempt to normalize examples whose tests are inherently bespoke: event tracing, packet/chaining topologies, compile-only native-host designs, matrix-multiplication matrices, vision/OpenCV applications, block datatypes, known-failing numerical examples, or model-scale demonstrations. Those are a whole other ball of wax that will require separate engineering work. The present patch is largely just mechanical.

## Validation

- header, RUN-line, and referenced-script checks for all added lit files
- Windows/NPU2 hardware validation completes successfully for all new lit tests, including the 35 added NPU2 lit files.

For `check-programming-guide`, we now get:

```bash
Total Discovered Tests: 33
  Unsupported: 14 (42.42%)
  Passed     : 19 (57.58%)
```

and for `check-reference-designs` this pass now yields:

```bash
Total Discovered Tests: 193
  Unsupported: 158 (81.87%)
  Passed     :  35 (18.13%)
```

This better than doubles (actually, more like triples) the number of tests that were previously exercised on Windows/NPU2 for the pedagogical materials.


programming_examples/basic/vector_vector_modulo/vector_vector_modulo.py
```diff
@@ -63,8 +63,9 @@ def _compile_kwargs(opts):
 
 
 def _run_and_verify(opts):
-    input0 = iron.randint(1, 100, (opts.num_elements,), dtype=np.int32, device="npu")
-    input1 = iron.randint(1, 100, (opts.num_elements,), dtype=np.int32, device="npu")
+    input0 = iron.arange(1, opts.num_elements + 1, dtype=np.int32, device="npu")
+    input1_np = np.full(opts.num_elements, 7, dtype=np.int32)
```
hunhoffe: I think you can use `iron.full` directly: https://github.com/Xilinx/mlir-aie/blob/1e6b13c2d422a8a60808f47fef48c848d58eaf8e/python/utils/hostruntime/tensor_class.py#L459

hunhoffe (approved):
Looks good to me! Thanks for the care in going through these examples to help bring them up to date and available for Windows users. I'll want to do a final sanity check on the CI outputs to make sure all the tests we are expecting to run do actually run, but otherwise this looks great!

programming_examples/basic/vector_vector_modulo/vector_vector_modulo.py
thomthehound: Nice. Done!

thomthehound:
I'm really glad this happened in the last run:
```plaintext
===== basic/passthrough_kernel =====
Traceback (most recent call last):
  File "/home/github/actions-runner/_work/mlir-aie/mlir-aie/programming_examples/basic/passthrough_kernel/passthrough_kernel.py", line 155, in <module>
    main()
  File "/home/github/actions-runner/_work/mlir-aie/mlir-aie/programming_examples/basic/passthrough_kernel/passthrough_kernel.py", line 106, in main
    in_tensor = iron.arange(n_elems, dtype=in1_dtype, device="npu")
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/github/actions-runner/_work/mlir-aie/mlir-aie/ironenv/lib/python3.12/site-packages/mlir_aie/python/aie/utils/__init__.py", line 247, in arange
    return DEFAULT_TENSOR_CLASS.arange(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/github/actions-runner/_work/mlir-aie/mlir-aie/ironenv/lib/python3.12/site-packages/mlir_aie/python/aie/utils/hostruntime/tensor_class.py", line 637, in arange
    t = cls(shape, dtype=dtype, device=device, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/github/actions-runner/_work/mlir-aie/mlir-aie/ironenv/lib/python3.12/site-packages/mlir_aie/python/aie/utils/hostruntime/xrtruntime/tensor.py", line 47, in __init__
    self.xrt_device = xrt_device if xrt_device is not None else xrt.device(0)
                                                                ^^^^^^^^^^^^^
RuntimeError:  No such device with index '0'
Error: Process completed with exit code 1.
```
Before the smoke test was introduced, it never even *occurred* to me that we would invoke an NPU run *directly* during CI! We aren't getting our usual retry protections there because it isn't a straight `lit` run!

This isn't my fault (aside from the fact that, well, you know, CI and I have a history together). But I'll patch it here in a little bit while putting it in my notes as something to think about later.

thomthehound:
Alright, I gave the smoke test an explicit NPU type for the matrix entries. Now it should be possible to safely route the CI hardware runs through `run_on_npu.py` and receive the same protections that `lit` gets. Although I distantly remember leaving a note in that file about something to fix for a future Windows CI, so I guess I have more homework now.

hunhoffe:
The CI weather forecast for today seems to be indicating turbulence (not just in this PR). A good portion of the failures so far seem to be transient network failures, but we'll see what else the day holds.

thomthehound:
> The CI weather forecast for today seems to be indicating turbulence (not just in this PR). A good portion of the failures so far seem to be transient network failures, but we'll see what else the day holds.

I was just looking at that...

thomthehound:
Welp, the patch works at least. So I guess that's good. But this is starting to look like a legitimate failure of the self-hosted device itself. Unfortunately, I don't think there is much I can do about that from this end.

Although, I do think this points out another problem concerning missing NPUs... it *looks* like a runner that *never* had an NPU (or has a dead one) will *always* pass every **lit** test given to it because it moves everything into `unsupported`. Direct runs have no such hidden failure mode. That is... interesting.

I think *potentially* many more tests are actually failing right now than are being reported as failures. That's bad.

hunhoffe:
Yeah, idk what's up with the runner. I'll try a reboot.

Ah, you might be correct. It does haunt me a little to think that we never verify the tests we want to run actually DO show up as successful in the output.

PR #3410, opened 2026-07-24
aie2p layer_norm: f32-accumulate the sum, two-pass centered variance

atassis:
## What

Make the aie2p bf16 `layer_norm` mean accurate by accumulating the row sum in
an f32 accumulator instead of a bf16 vector.

## Why

In the baseline `sum_acc` is a bf16 vector while `sum_sq_acc` is already f32, so
`E[x^2]` is accurate and only the mean is lossy: a bf16 running sum drops
low-order bits as the reduction length grows, and `embedding_dim` is typically
thousands. The centering and the variance both inherit that error, which is why
`op=layer` needed a `0.1` tolerance where `rms` passes at `0.05`.

An earlier revision of this PR also switched to a two-pass centered variance. I
dropped it. The `E[x^2] - mean^2` cancellation it guarded is unreachable under a
bf16 input contract: to make the f32 subtraction cancel you need mean/std on the
order of 100+, but a bf16 value near such a mean has an ulp larger than the std,
so the deviations are already gone. The single accumulator change is the whole
measured improvement.

## How

Accumulate the sum in an `accfloat` and reduce in f32. Nothing else changes:
same single reduction pass, same `E[x^2] - mean^2`, same bf16 in/out contract.

## Test

`programming_examples/ml/norm op=layer` on NPU2, input `uniform[-1,1]`, per-row
rel-L2 vs the f32 reference:

| cols | before (bf16 sum) | after (f32 sum) |
|------|-------------------|-----------------|
| 2048 | 1.14%             | 0.61%           |
| 4096 | 1.94%             | 0.63%           |

The trend is the point: the baseline error grows with reduction length as the
bf16 sum drops more bits; f32 accumulation flattens it (~0.6%, now bounded by
bf16 interface rounding).

Tightened the `op=layer` tolerance `0.1 -> 0.05` (matching rms) in norm.py and
test.cpp. A tighter elementwise tolerance alone can't catch a regression here:
the worst-element error is bounded by bf16 output quantization (~0.023) and is
nearly identical before and after. Added a mean per-row rel-L2 guard (ceiling
1%), which does separate them. Verified on device: the old kernel passes the
elementwise check but trips the guard at 0.0194.

## Refs

#2466, #2490, #2897.


atassis:
Github is being Github again, it has another major incident, that's we can't finish tests 👍🏼 
https://www.githubstatus.com/incidents/yjysg0xrl67m

atassis:
If current workflow won't succeed- I'll rebase the PR, so it would create new Run ID, maybe that's an issue (hail great Github)

hunhoffe (commented):
Thanks for digging into this!

A few thoughts before I approve:
1. Extra pass over input (2 loads → 3). Any sense of the perf impact? Not a blocker, just want to make sure we're making that tradeoff with eyes open.
2. mean_v still gets cast to bf16 before centering in pass 2, so that subtraction happens at bf16 precision rather than f32. Sounds like this might be intentional — just want to confirm!
3. Could we tighten norm.py's atol=0.1 for op=layer? Your numbers show it's now ~0.6%, so a tighter bound (closer to rms's 0.05) would catch a regression of this fix.

atassis:
Reworked it.

**1. Extra pass.** Dropped it. The only real defect in the baseline is the running sum: `sum_acc` was bf16 while `sum_sq_acc` was already f32, so `E[x^2]` was fine and only the mean was lossy. Accumulating the sum in an `accfloat` is the whole fix, and it keeps the original loop structure (one reduction pass, one write pass), so no third pass and no perf change. It also matches the measured trend: the error grew with reduction length because the bf16 sum kept dropping low-order bits, and f32 flattens it. The centering I'd added doesn't buy anything under a bf16 input contract (see 3), so it's gone.

**2. mean cast to bf16.** Intentional. The output centering `x - mean` stays bf16, which is fine for these inputs: the mean's bf16 rounding sits at or below the inputs' own bf16 quantization, and the output is bf16 regardless, so widening to center in f32 costs a conversion for no measurable gain.

**3. atol, plus a real guard.** Tightened to 0.05 for `op=layer` (matching rms) in norm.py and test.cpp. But a tighter atol alone won't catch a regression of this fix: the worst-element error is bounded by bf16 output quantization (~0.023) and is nearly identical before and after, so no elementwise atol separates them at e=2048. So I also added a mean per-row rel-L2 guard (ceiling 1%), which does separate them (fixed 0.61-0.63%, bf16-sum 1.14-1.94%). Verified on device: the old kernel passes the elementwise check but trips the rel-L2 guard (0.0194 > 0.01).

Separately, while measuring this I noticed the centering does matter on f32 input, where the mean/std regime is representable. The only f32 layernorm here, `layer_norm_welford`, reduces over the rows axis (per-column stats) rather than the feature axis, and isn't linked by this example. I'll send a separate PR with a correct f32 per-row layernorm.

hunhoffe (approved):
Thanks for your contribution!

PR #3415, opened 2026-07-24
[python] Fix resident-tensor stale-write bug in HostRuntime.run()

hunhoffe:
## Description
  
`Tensor.data` doesn't mark the tensor dirty, so writing a resident arg in place (`t.data[:] = x`, `np.copyto`) between dispatches is missed — the next `.to("npu")` no-ops and the kernel runs on stale bytes. Reported against #3354; same bug already fixed downstream in `amd/IRON#138`. 

  ## Fix

- `Tensor.data` now marks host-dirty on every access (read or write — numpy has no write-only hook).
- `HostRuntime._repin_outputs` reasserts `"npu"` residency on `Out`/`InOut` args after a successful dispatch, using the existing `In`/`Out`/`InOut` annotations, so reading a result between dispatches doesn't cause the *next* dispatch to push stale host data back over fresh device output.
- HRX backend needs no changes — it already syncs unconditionally both ways, sidestepping this class of bug entirely.

## Checklist

### DIFF
--- a/python/utils/callabledesign.py
+++ b/python/utils/callabledesign.py
@@ -221,6 +221,7 @@ class CallableDesign:
             kernel_name="MLIR_AIE",
             trace_config=trace_config,
             num_host_bos=num_host_bos,
+            output_flags=compilable.tensor_output_flags(),
         )
         if compilable.use_cache:
             self._kernel_cache[cache_key] = kernel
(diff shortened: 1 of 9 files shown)


hunhoffe:
Closing this — on reflection, the supported write paths — `__setitem__`, `fill_()`, `torch_view()` — already resync correctly; only bypassing all of them with a raw `.data` write is unsafe, and that pattern isn't used anywhere in mlir-aie today.

PR #3419, opened 2026-07-24
aie2p: add a correct f32 per-row layer_norm; drop wrong-axis welford

atassis:
## What

Add a correct f32 per-row `layer_norm` to `aie_kernels/aie2p/layer_norm.cc`, replacing the wrong-axis, unlinked `layer_norm_welford`, and wire it into the `ml/norm` example as `op=layer_f32`.

## Why

`layer_norm_welford` was the file's only f32 LayerNorm, but its Welford reduction runs down the **rows** axis, accumulating per-**column** statistics. LayerNorm normalizes each row over its feature axis (cols), so it computed the wrong thing, and nothing linked it (the example only builds the bf16 `layer_norm`). There is currently no working f32 per-row LayerNorm here.

## How

`layer_norm_f32`: f32 in/out, reduces over `cols` per row, with a numerically stable centered two-pass variance (`sum((x-mean)^2)/cols`) instead of `E[x^2] - mean^2`.

The dtype is the point. The bf16 `layer_norm` uses `E[x^2] - mean^2`, which is safe under a bf16 input contract: a bf16 value near a large mean has an ulp wider than the std, so the large-mean / small-std regime where that form cancels is unrepresentable. f32 **can** represent that regime, so the f32 path centers first.

Wired into `ml/norm` as `op=layer_f32`: f32 tensors, an f64 gold reference, and a non-zero-mean input (row mean ~100) that exercises the centering. Verified through the standalone Python path (`run_strix.lit`, the same `%python %S/norm.py -o ...` form the rms/layer JIT checks already use), not the bf16 `test.cpp`. Two guards fall out of the dtype switch:

- The verify pins `rtol=0` so the f32 kernel's tight `atol=1e-3` actually governs (`assert_pass` otherwise defaults `rtol=0.128`, which would swamp an O(1) LayerNorm output and mask the error).
- `embedding_dim` is required to be a multiple of the 16-wide vector, since the kernels process full vector chunks with no scalar tail.

## Test

`programming_examples/ml/norm op=layer_f32` on NPU2 at `embedding_dim=2048` (an f32 row is twice the bytes of bf16, and a 4096-wide f32 row does not fit the tile's local memory double-buffered):

| embedding_dim | max-abs vs f64 reference |
|---------------|--------------------------|
| 1024          | 1.5e-5                   |
| 2048          | 2.4e-5                   |

Passes at `atol=1e-3, rtol=0`; a reintroduced `E[x^2] - mean^2` kernel fails that gate on device. The bf16 `rms` and `layer` paths are unchanged and still pass.

## Notes

Touches the same file as #3410, but a disjoint region of the kernel (that PR changes the bf16 `layer_norm`; this changes the f32 kernel), so those auto-merge. The two do overlap in `norm.py`'s `_VERIFY_CFG` dispatch table; whichever lands first, the other is a one-line rebase.

### DIFF
--- a/aie_kernels/aie2p/layer_norm.cc
+++ b/aie_kernels/aie2p/layer_norm.cc
@@ -55,48 +55,52 @@ void layer_norm(const T *restrict input, T *restrict output, int32_t cols) {
   event1();
 }
 
+// f32 per-row LayerNorm. The bf16 layer_norm above centers with a single
+// E[x^2] - mean^2 reduction, which is fine because the bf16 input contract
+// keeps the mean/std ratio small (a bf16 value near a large mean has an ulp
+// wider than the std, so that regime is unrepresentable). On f32 input the mean
+// can be large relative to the std and E[x^2] - mean^2 then catastrophically
+// cancels, so this variant reduces over the feature axis (cols) per row with a
+// numerically stable two-pass centered variance: center first, then square.
 template <typename T, int N>
-void layer_norm_welford(const T *restrict input, T *restrict output,
-                        int32_t rows, int32_t cols) {
+void layer_norm_f32(const T *restrict input, T *restrict output, int32_t cols) {
   event0();
   constexpr float epsilon = 1e-5f;
   const float gamma = 1.0f;
   const float beta = 0.0f;
-  ::aie::vector<T, N> reg_a, delta, delta2, mean_v, m2_v, variance_v, just_div,
-      just_prod;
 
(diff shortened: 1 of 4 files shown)


hunhoffe (approved):
A nice improvement, thanks for the clear description & merge order details! I've queued https://github.com/Xilinx/mlir-aie/pull/3410 -- once that has landed and the conflict is resolved, I'll queue this to merge too!

atassis:
I have fixed conflicts and rebased, CI is reset

PR #3422, opened 2026-07-25
CI: fail NPU jobs when their hardware is missing

thomthehound:
This follows from the side discussion in #3407 after a hardware runner repeatedly failed to expose its NPU during direct invocation of a test.

The lit setup, as it exists now, treats a missing NPU as an unsupported configuration, skipping all tests that were intended to run on it. There can be circumstances where that makes sense locally, but it is **not** acceptable for a CI runner whose job is *specifically* to test on hardware. 

For instance, if a machine suddenly loses access to its NPU more than just transiently (perhaps due to a persistent deadlock or hardware failure), it will inadvertently **pass** hardware tests that might otherwise have failed *for every run or GitHub action that takes place using that physical host after the loss* unless it is manually reset. The smoke test run was previously the only reliable way to discover that problem.

This patch makes each hardware job declare its expected NPU. At the start of a test run, lit retries device discovery three times, then fails configuration if the expected device is missing or inaccessible. Runs without an explicit hardware expectation keep the existing behavior. 

It also:

* finishes the native-Windows diagnostic path in `run_on_npu.py`;
* fixes the pure-HRX tests so they use HRX-specific NPU gating instead of the XRT-backed `%run_on_npu2%` substitution.
> I believe the HRX tests weren't actually running during CI before, because they were not exposing themselves as actual hardware tests. This should fulfill that intent, although it might expose other problems I haven't considered.

As a result of this, an NPU CI job (regardless of backend) cannot pass merely because its hardware disappeared and its tests became `UNSUPPORTED`.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -45,7 +45,7 @@ env:
 jobs:
 
   build-and-test-from-source:
-    name: Run Tests and Examples on Ryzen AI (Built from Source)
+    name: Run Tests and Examples on Ryzen AI (Built from Source) (${{ matrix.runner_type }})
     # The goal for this job is to run the full test suite, including programming guide and examples, on a freshly built-from-source MLIR-AIE.
     # This also exercises the build-mlir-aie-from-wheels.sh script, which is the most common developer-facing way to rebuild MLIR-AIE from source.
     runs-on: ${{ matrix.runner_type }}
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -52,8 +52,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner_type: [ aie2-4col, aie2p-8col ]
+        include:
+          - runner_type: aie2-4col
+            expected_npu: npu1
+          - runner_type: aie2p-8col
+            expected_npu: npu2
     env:
+      AIE_EXPECTED_NPU: ${{ matrix.expected_npu }}
       NPU_CACHE_HOME: ${{ github.workspace }}/iron-cache-${{ matrix.runner_type }}-${{ github.run_id }}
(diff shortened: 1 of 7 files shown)


thomthehound:
yup, that was pretty much confirmation: the HRX tests were never actually running because they were gated behind `run_on_npu`, which is obviously not going to run anything when XRT is missing. Now they are at least trying. Honestly, though, I'm going to be firing blind on this one. I'm hoping the CI driver is up to the task.

thomthehound:
Additionally linking #3198 in this thread, because it should be catalogued.

hunhoffe:
@thomthehound good find -- nice confirmation the work in this PR is doing what it should! If you want to mark the HRX tests as unsupported and create an issue, we can get this PR in now if you don't want to be debugging blind.

If you don't mind debugging blind, then fixes are of course appreciated!

thomthehound:
> @thomthehound good find -- nice confirmation the work in this PR is doing what it should! If you want to mark the HRX tests as unsupported and create an issue, we can get this PR in now if you don't want to be debugging blind.
> 
> If you don't mind debugging blind, then fixes are of course appreciated!

Looks like it's going to work! Hooray!

hunhoffe (approved):
Thanks for another helpful contribution!

hunhoffe:
@thomthehound would you mind bumping this to retrigger CI actions? It seems that a few actions have been lost into the ether...

thomthehound:
I think the slight changes to the matrix might have altered some sort of launch name or something. I'm looking into this right now.

Edit: no, they launched and passed, I think there is just a handoff problem. Still looking.

Edit2: ah, I see. Branch protection still wants the old job names, even though the exact same jobs were automatically superseded with an appended name (i.e. "aie2-4col" became "aie2-4col, npu1"). No problem, I can make sure they have the old, expected names. Just a mintue.

thomthehound:
@hunhoffe Sorry about that. I hadn't noticed that the jobs didn't have a stable name. Now they do.

thomthehound:
green to go!

PR #3423, opened 2026-07-26
fix: make pathfinder flow routing order-independent

### LINKED ISSUE #3360: create-pathfinder-flows is sensitive to aie.flow ordering
### What happened?

`--aie-create-pathfinder-flows` produces different switchbox routings for two modules that contain the **identical set** of `aie.flow` ops but list them in a **different order**. The routing should be a function of the flow set (and the device), not of the textual order the flows happen to appear in the IR.

This is a determinism/canonicalization gap: two semantically-identical programs compile to different routings. Both routings are individually valid, but because the physical switchbox ports differ, a host/runtime that targets fixed physical ports (e.g. a host-driven design whose shim DMA descriptors are pinned) will work with one ordering and dead-lock with the other — with no diagnostic.

### Reproducer

Two modules, **identical flow set**, only the order of the three `aie.flow` ops differs.

`repro_order_a.mlir` — the `shim(0,0)->core(0,3)` flow declared **first**:
```mlir
module {
  aie.device(npu2) {
    %s0 = aie.tile(0, 0)
    %c02 = aie.tile(0, 2)
    %c03 = aie.tile(0, 3)
    %s1 = aie.tile(1, 0)
    %c12 = aie.tile(1, 2)
    aie.flow(%s0, DMA : 0, %c03, DMA : 0)
    aie.flow(%s0, DMA : 1, %c02, DMA : 0)
    aie.flow(%s1, DMA : 0, %c12, DMA : 0)
  }
}
```

`repro_order_b.mlir` — the **same three flows**, that flow declared **last**:
```mlir
module {
  aie.device(npu2) {
    %s0 = aie.tile(0, 0)
    %c02 = aie.tile(0, 2)
    %c03 = aie.tile(0, 3)
    %s1 = aie.tile(1, 0)
    %c12 = aie.tile(1, 2)
    aie.flow(%s0, DMA : 1, %c02, DMA : 0)
    aie.flow(%s1, DMA : 0, %c12, DMA : 0)
    aie.flow(%s0, DMA : 0, %c03, DMA : 0)
  }
}
```

### Commands

```
aie-opt --aie-create-pathfinder-flows repro_order_a.mlir
aie-opt --aie-create-pathfinder-flows repro_order_b.mlir
```

### Output (differs) — `switchbox(%tile_0_2)`

A:
```mlir
aie.connect<South : 1, North : 0>
aie.connect<South : 2, DMA : 0>
```
B:
```mlir
aie.connect<South : 1, DMA : 0>
aie.connect<South : 2, North : 1>
```

The `shim(0,0) DMA:0 -> core(0,3)` flow is routed through different physical ports depending only on declaration order (e.g. at `switchbox(%shim_noc_tile_0_0)` it takes `South:3->North:1` in A vs `South:3->North:2` in B; at `core(0,3)` it arrives on `South:0` in A vs `South:1` in B). Neither pass run emits any warning or error.

### Expected

Routing should be invariant under reordering of an identical `aie.flow` set (e.g. by canonicalizing/sorting flows before pathfinding, or otherwise making the allocation order-independent). At minimum, the order-sensitivity should be documented, since it can silently change physical port assignment.

### Impact

In a larger host-driven design this manifested as a hardware hang (`ERT_CMD_STATE_TIMEOUT`): converting one DMA leg to an `ObjectFifo` did not change the flow set, but the objectfifo-stateful-transform emits its `aie.flow` at a different point in the module than the equivalent hand-written flow, which reordered the routing... and then the design hung. I'm still looking into why the hang happened.

### Environment

- mlir-aie commit `6e5796c`
- Device: `npu2`.

### Related

Possibly related in spirit to #437 (routing correctness not verified against intent), though that issue concerns `AIEFindFlows` validation of overlapping **packet** rules; this is `create-pathfinder-flows` order-sensitivity for **circuit** flows.

### Hardware / device

npu2

w3lld1:
## Description

This canonicalizes pathfinder inputs after both packet and circuit flows have been collected. Priority packet flows still route first, followed by normal packet flows and then circuit flows; sources and destinations are sorted within those classes so equivalent flow sets no longer depend on textual declaration order.

The regression test routes two modules with the same three circuit flows in different orders and requires byte-identical output.

Fixes #3360

## Validation

- `clang-format 17.0.1 --dry-run --Werror` on the changed C++ files
- `git diff --check`
- Standalone ordering probe covering all six circuit-flow permutations, priority/normal-packet/circuit precedence, and fanout destination ordering
- Reproduced the order-dependent output with the `4976487` wheel; canonicalizing the same inputs before routing produced byte-identical `aie-opt` output

The full source build and the new lit test are deferred to CI because the local environment did not have enough quota to extract the pinned MLIR build toolchain.

## Checklist

- [x] Docs / docstrings updated if the change is user-facing (not user-facing)

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
@@ -188,7 +188,7 @@ public:
   virtual void addFlow(TileID srcCoords, Port srcPort, TileID dstCoords,
                        Port dstPort, bool isPacketFlow,
                        bool isPriorityFlow) = 0;
-  virtual void sortFlows(const int maxCol, const int maxRow) = 0;
+  virtual void sortFlows() = 0;
   virtual bool addFixedConnection(SwitchboxOp switchboxOp) = 0;
   virtual std::optional<std::map<PathEndPoint, SwitchSettings>>
   findPaths(int maxIterations) = 0;
@@ -201,7 +201,7 @@ public:
                   const AIETargetModel &targetModel) override;
   void addFlow(TileID srcCoords, Port srcPort, TileID dstCoords, Port dstPort,
                bool isPacketFlow, bool isPriorityFlow) override;
-  void sortFlows(const int maxCol, const int maxRow) override;
+  void sortFlows() override;
   bool addFixedConnection(SwitchboxOp switchboxOp) override;
   std::optional<std::map<PathEndPoint, SwitchSettings>>
   findPaths(int maxIterations) override;
--- a/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
+++ b/lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
@@ -71,10 +71,6 @@ LogicalResult DynamicTileAnalysis::runAnalysis(DeviceOp &device) {
     }
   }
 
-  // Sort ctrlPktFlows into a deterministic order; concat ctrlPktFlows to flows
-  pathfinder->sortFlows(device.getTargetModel().columns(),
-                        device.getTargetModel().rows());
-
   // Add circuit flows.
   for (FlowOp flowOp : device.getOps<FlowOp>()) {
     TileOp srcTile = cast<TileOp>(flowOp.getSource().getDefiningOp());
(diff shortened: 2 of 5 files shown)


w3lld1:
I fixed the failing order-independence regression in `98334d99`.

The generated routing was semantically identical, but the test compared printer order: identical switchbox/connect operations appeared in a different textual order. The test now associates each connection with its containing switchbox and sorts those pairs before comparison, so it remains sensitive to routing changes while ignoring operation ordering.

Validation:
- normalization probe for switchbox/connection association
- `git diff --check`

CI has not started on the pushed commit yet.

hunhoffe (approved):
Thank you for your contribution!

PR #3424, opened 2026-07-26
[python] JIT cache: wait for the lock instead of spinning on it

atassis:
## Problem

`file_lock` is documented to wait up to `timeout_seconds` for the cache lock and
then raise. It does neither. On a held lock it spins.

`_try_acquire_lock` signals contention by **returning False** -- it catches the
`OSError` from `flock` / `msvcrt.locking` itself:

```python
try:
    fcntl.flock(lock_file.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
    return True
except OSError:
    return False
```

but the wait loop puts the sleep *and* the deadline check inside an `except
OSError` that nothing raises into:

```python
while True:
    try:
        if _try_acquire_lock(lock_file):
            break
    except OSError:                      # unreachable
        if time.time() - start_time > timeout_seconds:
            raise TimeoutError(...)      # unreachable
        time.sleep(0.1)                  # unreachable
```

So a contended `file_lock` retries with no delay and no deadline. Measured
against a held lock, the waiter makes **15,329,952 acquisition attempts per
second** and never returns until the holder releases. `timeout_seconds` has no
effect at any value.

## Concretely

The waiter's own latency is not the problem -- it gets the lock as soon as it is
free either way. The cost lands on everything else running on the machine.

Ten processes queued on a held lock, with ten unrelated CPU-bound jobs running
alongside them, on a 20-core host:

| | wall for the unrelated jobs |
|---|---|
| no waiters at all (control) | 1.01 s / 0.99 s |
| ten waiters, spinning (today) | 1.51 s / 1.73 s |
| ten waiters, with this change | 1.07 s / 0.99 s |

Two repeats of each. A queued build currently taxes every concurrently running
job by 50-75%; with this change the tax is gone and unrelated work runs as fast
as if nothing were waiting at all.

Measured from the waiters' side instead: 8 waiters over a 10 s hold burn
**81.3 s of CPU today and 6.3 s after**, for identical wall time (10.06 s vs
10.07 s). The waiting was never slow -- it was expensive.

That is worst exactly where it is least affordable: behind a long compile, on
shared runners.

## Fix

Wait on the kernel instead of repairing the retry loop.

`flock` without `LOCK_NB` parks the caller until the holder releases, which is
the event the loop was approximating. Running it on a helper thread lets
`timeout_seconds` apply to the handoff rather than to a poll interval. The
uncontended path is unchanged: one non-blocking attempt, no thread.

Windows has no indefinite blocking lock -- `msvcrt.locking`'s `LK_LOCK` gives up
after about ten seconds -- so that platform keeps a retry, now with a reachable
deadline.

Handoff, with 16 waiters over a cache-hit-sized critical section:

| wait strategy | handoff | polls during a 1 s wait |
|---|---|---|
| spin (today) | 11.3 ms | 15,329,952 |
| fixed 100 ms poll | 531 ms | 17 |
| 1 ms -> 5 ms backoff | 35 ms | 200 |
| blocking (this change) | 7.5 ms | 0 |

Faster than the spin it replaces, with no polling at all. Single-waiter handoff
after a 300 ms hold is 0.2 ms.

**A waiter that gives up must not go on to win the lock and hold it.** The helper
thread can acquire in the same instant the caller times out; if it hands the lock
to nobody, it stays held until the process exits. The thread's handoff and the
caller's timeout are therefore serialised, and whichever loses releases.

## The timeout this switches on

Repairing the loop makes `timeout_seconds` load-bearing for the first time, and
both call sites in `compilabledesign` took the 60 s default. A waiter there is
waiting out another process' entire **compile**, not an acquisition handshake,
and AIE compiles run long enough that CI budgets individual tests 600-1200 s.
Left alone, making the timeout work would convert "slow but eventually correct"
into a spurious `TimeoutError` whenever a concurrent build ran over a minute.
Both sites now pass a bound sized for a build.

## Test

Adds `test/python/test_cache_file_lock.py`:

- `test_contended_lock_times_out` -- a lock held past the deadline must raise,
  not block indefinitely.
- `test_contended_lock_does_not_busy_wait` -- the wait must park rather than
  retry, asserted as CPU consumed against time spent waiting, measured in-process
  so it is attributable.
- `test_abandoned_waiter_does_not_keep_the_lock` -- sweeps the caller's deadline
  across the holder's release, including the instant the lock changes hands, and
  asserts the lock is free afterwards in every case.
- the uncontended path, and release-on-exception.

The first two fail on `main`; all five pass with this change.

## Validation

Host-side only; no device involved. `test/python/` excluding the `npu-xrt` and
`npu-hrx` device directories, same tree, before and after:

| | passed | failed |
|---|---|---|
| `main` | 578 | 7 |
| with this change | 580 | 5 |

The delta is exactly the two new contention tests flipping to pass; no other test
changes state. The failures common to both runs are unrelated to this change and
are artifacts of the local harness, not of `main`.

A full `@iron.jit` design was compiled end to end before and after to confirm the
ordinary path is unaffected: 1.4 s cold, cache hit at 0.00 s, identical xclbin
size.

`black` (26.5.1, the pinned pre-commit revision) reports all three files
unchanged.

## Two things left alone

A permanent, non-contention failure -- `ENOLCK`, `EBADF`, a mount without working
advisory locks -- is swallowed by `_try_acquire_lock` and surfaces as a
`TimeoutError` after the full wait rather than as the underlying errno. Still
better than spinning on it forever, but it is a diagnostic the caller does not
get. Narrowing `_try_acquire_lock` to treat only `EACCES`/`EAGAIN` as contention
and re-raise the rest would fix it; that is left out here because the Windows
branch cannot be exercised from this change's test job, which is also why these
tests skip on Windows.

`TimeoutError` now propagates out of `compile()` where the code previously
blocked indefinitely, and nothing in the tree catches it. A timed-out waiter
raises before `file_lock` yields, so it never enters the body and cannot disturb
the holder's in-progress build or its cache directory.

## Why it has been quiet

Both call sites lock a per-design directory under `$NPU_CACHE_HOME`, so
contention needs two processes building the *same* design concurrently. The
on-device lit group runs at capacity 1 and each CI run gets its own cache
directory, which keeps that rare today. It stops being rare as soon as
compilation is fanned out in parallel against a shared cache root.

### DIFF
--- a/python/utils/compile/cache/utils.py
+++ b/python/utils/compile/cache/utils.py
@@ -9,6 +9,8 @@ import contextlib
 import hashlib
 import os
 import pickle
+import queue
+import threading
 import time
 
 # Cross-platform file locking:
@@ -20,6 +22,10 @@ else:
     import fcntl
 
 from aie.utils.hostruntime.tensor_class import Tensor
+
+# Windows has no indefinite blocking lock -- msvcrt.locking's LK_LOCK gives up
+# after ~10s -- so that platform waits by retrying.  See _wait_for_lock.
+_LOCK_POLL_SECONDS = 0.005
 
 
 def _cell_val_to_key(val):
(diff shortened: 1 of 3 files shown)


hunhoffe (approved):
Thank you for the contribution! It's great to see someone exercising the JIT cache enough to notice these types of bugs/shortcomings.

I think it's in good shape to merge if we get the CI green check. One note --  `test/python/test_cache_file_lock.py` makes timing-sensitive assertions (burned < waited / 10, wall < 8s for a 1s timeout, the abandoned-waiter sweep). test/python/ runs under check-aie's default lit scheduling, which is parallel by default (explicit -j4/-j12 on the Ryzen AI job, unset-but-still-parallel elsewhere) — so this test can end up sharing CPU with a dozen unrelated tests at once, which risks flaking the tighter assertions under load.

There's already precedent for isolating this class of test: test/python-concurrency/'s lit.local.cfg sets parallelism_group for exactly this reason. But that directory is wired to a separate check-aie-concurrency target that's only invoked from the NPU-hardware workflow (check-aie itself does --filter-out python-concurrency) — moving the file there would mean it stops running in the plain wheel-build CI, which seems like the wrong trade for a test that's explicitly host-side-only.

Since parallelism_group is set per-directory (via lit.local.cfg), not per-file, this test could get its own small subdirectory (e.g. test/python/cache_lock/) with a lit.local.cfg setting parallelism_group, so it's isolated from CPU contention without losing coverage on the non-hardware CI path.

But, let's consider this if we experience any flakiness. For now it's ok, I think.

PR #3428, opened 2026-07-26
[python] JIT cache: give code objects a stable identity in the recipe hash

atassis:
## Problem

`_compute_recipe_hash` folds `repr(code.co_consts)` into the key. Every design defines its worker
body inline, so `co_consts` holds a code object, and `repr()` of one embeds its address:

```
(<code object core at 0x7fd184538570, file "...", line 4>,)
```

Three runs of the same unchanged design:

```
412a7f8e248e20cb254ee1ba0891edf0fc7b7dc4
2619e562f66c815e2a764ec37a02a83a5bd21692
6fbe6b036d9d32aa4a967e6e4a6260efbf196236
```

`_compute_hash` names the directory under `$NPU_CACHE_HOME`, so the cache writes a fresh entry on
every run and reads none. `use_cache=True` has no effect across processes, and the cache directory
grows by one per invocation of the same design. This has gone unnoticed because a cache that never
hits never serves anything, so it has never served anything wrong.

## Fix

Hash `marshal.dumps()` of the code object. That is what CPython writes into a `.pyc`: bytecode,
names, varnames, flags and constants, recursing into nested code objects, with no addresses and a
frozenset order that does not depend on the hash seed.

File and line information is stripped first. It is not part of the design, and keying on it would
give two checkouts of one repo separate cache entries, and would invalidate a design because a line
was added above it. `CodeType.replace()` does not reach nested code objects, hence the walk in
`_without_location`.

The callable branch of `_kwarg_repr` had the identical defect for a callable passed as a
compile-time value, and takes the same substitution.

## Two blind spots that have to close with it

A key that never hits cannot serve a wrong entry. Making it hit turns both of these from latent into
reachable, so they belong in the same change.

**`co_names`.** `matmul_bf16(a, b, c)` and `matmul_i8(a, b, c)` compile to byte-identical
instructions, because the symbol is an index into `co_names`. A key built from `co_code` and
`co_consts` cannot tell those two designs apart. marshal covers names, varnames and flags.

**Generator defaults.** A `CompileTime[T]` left at its Python default never reaches `compile_kwargs`,
and the default is stored on the function rather than in the code object, so editing
`TILE: CompileTime[int] = 512` to `1024` did not move the key. `__defaults__` and `__kwdefaults__`
are now hashed, on the generator and in `_kwarg_repr`.

## Measured

One single-core design built three times in three separate processes, against a fresh
`$NPU_CACHE_HOME`:

| | keys | cache entries | wall |
|---|---|---|---|
| `main` | 3 distinct | 3 | 0.17s, 0.16s, 0.16s |
| with this change | 1 | 1 | 0.15s, then 0.01s, 0.01s |

The cold run is unchanged; what changes is that the second one can hit. The key is unchanged under
`PYTHONHASHSEED` of 0, 1 and random.

## Tests

Eight in `test/python/test_compilabledesign.py`. Two fail on `main`:

- `test_hash_is_stable_for_a_generator_with_a_nested_function`, one key across three separate
  interpreters.
- `test_hash_survives_a_move_of_the_design_file`, unchanged by a file move, by a line inserted above
  the design, and by a reflowed body. That last case is the one that moves `co_linetable`; a uniform
  line shift does not.

The other six pin the key against blindness: nested body, nesting past any fixed depth, symbol name,
compile-time default, and the callable-kwarg path. Each pair is built from source so that it differs
in exactly one way and shares a name and module, since a differing qualname would discriminate on
its own and the test would then pass without proving anything. I checked that by mutation: disabling
the recursion into nested code, any part of the location strip, the defaults, or the `_kwarg_repr`
substitution each fails at least one test.

## Validation

Host side only, no device. `test/python/` excluding the `npu-xrt` and `npu-hrx` device directories,
same tree, before and after:

| | passed | failed |
|---|---|---|
| `main` | 598 | 0 |
| with this change | 606 | 0 |

The delta is exactly the eight new tests. `black` 26.5.1, the pinned pre-commit revision, reports
both files unchanged.

Every existing cache entry is orphaned once, since the keys move. They were already being orphaned
on every run.

## What this does not do

It does not make the key see closures. Two designs from one factory, differing only in a captured
value, still hash identically:

```python
def make(tile):
    def design(...):
        ...uses tile...
    return design

make(64)   # and
make(128)  # -> same key
```

The same holds for a decorator that does not use `functools.wraps`, where the wrapper's code and
name are what get hashed. Neither is introduced here, and closure values cannot simply be folded in:
the in-process key handles them separately via `_cell_val_to_key`, which starts at `hash()`,
randomized per process for `str`, and ends at `repr()`. Folding that into the on-disk key would
trade this bug for a worse one.


hunhoffe (approved):

andrej (commented):
Seems like a reasonable change, but I have to wonder if there isn't an existing serializer library (pickle?) that would do the same? Seems a bit like reinventing the wheel for something that is potentially more involved than it looks like on the surface.

atassis:
I have cleaned this up, also I have introduced a `marshal` dep. You were right on suggesting a serializer lib, it helped to fix few blind spots

hunhoffe (approved):
Good call on serialization/marshalling the data!

PR #3429, opened 2026-07-26
Minimize packet-flow slave rules with exact subcube cover

jskimko:
## Description

The packet-flow router (`AIECreatePathFindFlows`) groups slave rules by identical destination-set and encodes each group with a single mask = the common bits of the group's packet ids. That mask is coarse whenever the group is a suffix/subset rather than an aligned power-of-2 block: it over-claims ids outside the group and the pass fails with `'aie.rule' op can lead to false packet id match for id N`. Distinct-data distributes and cross-column scatters therefore **fail to route** at modest density and force users into hand-authored switchbox MLIR.

This replaces the single common-bits mask with the exact-minimum set of subcube rules that cover the group's ids while avoiding the other ids present on the same slave port (ids absent from the port are don't-cares, free to over-claim -- a rule only acts on ids that arrive). It is exact two-level logic minimization of a 5-bit function (based on Quine-McCluskey): enumerate prime implicants, pick a minimum-cardinality cover by branch-and-bound, then tighten each chosen cube to the bounding cube of the ids it serves. Branch-and-bound is used for the cover step (Petrick's method is the classical alternative) -- it reaches the same optimum without Petrick's product-of-sums expansion, and its incumbent bound doubles as the 4-slot cutoff. It is tractable because the packet-id space is only 5 bits -- the 3^5-cube enumeration plus cover runs in about **1.5 us per rule group**.

The old greedy packing is the degenerate case: when one cube suffices it is optimal and the tightened cube is exactly the common-bits mask, so output is byte-identical wherever the old single-rule packing was valid; **it only emits more (still minimum) rules where the coarse mask used to be rejected**. A slave port's packet-rule budget (`AIETargetModel::getNumSlaveSlots`, currently 4) is enforced: a fan-out whose minimum cover exceeds it is reported as an error rather than silently over-programmed (this also catches the pre-existing case where the greedy pass could emit more than 4 rules).

New lit tests in `test/create-packet-flows`:
- `subcube_cover.mlir` - {2,4} routes while avoiding sibling id 0 on the same slave port.
- `subcube_cover_optimal.mlir` - {0,3,4} avoiding 1 -> the masked rule `(27,0)`, the minimum 2 rules (not 3).
- `subcube_cover_multihop.mlir` - the id-{26,28,29} multi-hop scheme (previously a false match) now routes.
- `subcube_cover_overbudget.mlir` - five destinations on one slave port trip the slot budget.

## Checklist

- [x] Code is formatted (`clang-format` for C++, `black` for Python - see [CONTRIBUTING](../CONTRIBUTING.md))


jskimko:
@andrej This might resolve some of the packet routing issues we both were running into; at least for `false packet id match`.

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -404,6 +404,10 @@ class AIETargetModel {
   /// the origins of connect operations in the switchbox.
   virtual uint32_t getNumSourceShimMuxConnections(int col, int row,
                                                   WireBundle bundle) const = 0;
+  /// Return the number of packet-rule slots per stream-switch slave port.
+  virtual uint32_t getNumSlaveSlots() const { return 4; }
+  /// Return the largest packet id the stream switch can route (5-bit field).
+  virtual uint32_t getMaxPacketId() const { return 31; }
```
andrej: Are these verified across architectures? If not, it might be better to leave this pure-virtual here and put this in the AIE2P-specific target model class.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -265,6 +268,140 @@ bool AIEPathfinderPass::findPathToDest(SwitchSettings settings, TileID currTile,
...
+  int mask;
+  int value;
+  uint32_t cov; // bitmask of the onset ids this cube matches
+};
+
+// 5-bit packet-id space; kept local (mirrors getMaxPacketId()) since these
+// helpers take no target model.
+constexpr int ID_BITS = 5;
+constexpr int NUM_IDS = 1 << ID_BITS;
+constexpr int ID_MASK = NUM_IDS - 1;
```
andrej: Is this duplication from the target model required for performance? If not, can we avoid it?

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -265,6 +268,140 @@ bool AIEPathfinderPass::findPathToDest(SwitchSettings settings, TileID currTile,
...
+
+// Cover `onset` (a group's packet ids) with the fewest (mask, value) rules that
+// match every onset id and no `offset` id (other ids on the same slave port).
+// Ids on neither set are don't-cares, free to over-claim.
+// Based on Quine-McCluskey: enumerate prime implicants -> minimum cover
+// (branch-and-bound) -> tighten.
+// Note: the old common-bits packing is the one-cube degenerate case.
+static SmallVector<std::pair<int, int>>
+computeSubcubeCover(const SmallVector<int, 4> &onset,
+                    const llvm::SmallSet<int, 8> &offset) {
```
andrej: nit: `onset` and `offset` here are named confusingly because they're English words. `match_set` and `avoid_set` might be clearer?

jgmelber (approved):
LGTM. Good to see 4 new tests covering the feature.

test/create-packet-flows/badpacket_flow.mlir
andrej, line 1: Was this test removed or just moved? If it now passes and you removed it for that reason, can we instead re-enable it and change it to expect a pass?

test/create-packet-flows/subcube_cover_multihop.mlir
```diff
@@ -0,0 +1,37 @@
...
+
+// RUN: aie-opt --aie-create-pathfinder-flows %s | FileCheck %s
+
+// ids 26,28,29 from one source: 28 drops at tile(0,2) while {26,29} forward via
+// rules that avoid 28 (the common-bits mask (24,24) would false-match it).
+
+// CHECK-LABEL: aie.device(npu1_1col) {
+// CHECK:         aie.switchbox({{.*}}) {
+// CHECK:           aie.packet_rules(South : 1) {
+// CHECK-DAG:         aie.rule(31, 28, %[[DROP:.*]])
```
andrej: nit: I find `DROP` in this test confusing -- maybe this is established terminology and I'm just unfamiliar -- but to me the packet isn't dropped -- it is routed out to `TileControl : 0`.

andrej (approved):
This is a really nice improvement!

I'm not going to pretend I tried to understand the actual algorithm implemented, but it appears to be a well-established method and the added tests give me confidence it works.

Only left a couple nitpick comments.

Open-ended question: Would a packet-ID-allocator pass with a global view of all routes be a tractable problem? If, rather than using hard-coded numeric IDs, we used SSA "symbolic" IDs, the allocator could pick IDs that are distinct with simpler masks on the most congested nodes -- but I could picture that this explodes the search space.

include/aie/Dialect/AIE/IR/AIETargetModel.h
jskimko: Moving to derived class.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
jskimko: I also debated this. I'll try threading it in from the model and measure performance.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
jskimko: Renamed.

test/create-packet-flows/badpacket_flow.mlir
jskimko: This negative test doesn't trigger anymore; new behavior is in `subcube_cover_multihop.mlir`.

test/create-packet-flows/subcube_cover_multihop.mlir
jskimko: Renamed to `DEST` (against `FWD`).

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
jskimko: Negligible difference. Added https://github.com/Xilinx/mlir-aie/pull/3429/commits/7cf2aaa5addd9107b98bd719617606bb07947d5a.

Issue #3420, opened 2026-07-25
Add a sub-region view to the host-runtime Tensor so amd/IRON can drop its XRTSubBuffer fork


atassis:
## Summary
The host-runtime `Tensor` (`python/utils/hostruntime`) has no sub-region view, so amd/IRON forked one (`XRTSubBuffer`). I would like to add the view upstream as a general primitive, let IRON delete its fork, and make the base a single canonical host-tensor class both repos share. This is a dedup between mlir-aie and amd/IRON, not a new feature for mlir-aie alone.

## The demand
amd/IRON carries `XRTSubBuffer` only because `XRTTensor` has no sub-region view. Its own source asks for this upstream:

```
# TODO: replace with XRTTensor.__getitem__ slice support when available upstream
(amd/IRON: iron/common/utils.py)
```

It carves per-head and per-KV-group windows out of one contiguous buffer, in the fused-sequence arena runner and the llama application. In #3412 you noted the mlir-aie / amd/IRON dedup is wanted but has not reached the top of the list. This is one concrete instance of it.

## Why the fork is awkward
With no sub-region view on `XRTTensor`, `XRTSubBuffer` subclasses it and bypasses `__init__` to avoid re-allocating, re-implements the host-dirty residency handling on its own leaf (the same rule that already lives on `Tensor`/`XRTTensor`), and carries a FIXME because it is unsure whether `bo.sync()` on a sub-buffer moves the slice or the whole parent, so it conservatively syncs the whole parent and can disturb sibling views. That last point was the one real unknown, so I resolved it.

## Sub-buffer sync is slice-scoped (the FIXME, resolved)
`xrt::bo(parent, size, offset).sync(dir)` transfers only its own slice, never the whole parent. `buffer_sub::sync(dir, sz, off)` computes `off += m_offset` and calls `m_parent->sync(dir, sz, off)` (`core/common/api/xrt_bo.cpp`), and the no-arg `sync(dir)` passes the sub-buffer's own `size()`. The amdxdna shim honors `(offset, size)` in every path: the driver-sync ioctl carries them, the CLFLUSH path flushes exactly that byte range, and on cache-coherent parts sync is a no-op. So a sub-view owns its own coherence and the whole-parent workaround is unnecessary.

## Proposal
[...]

PR #3430, opened 2026-07-26
[python] hostruntime: NpuTensor, sub-region views, and Storage

atassis:
Addresses #3420: the sub-region view and the `NpuTensor` rename. The remaining piece is
on the IRON side, deleting `XRTSubBuffer` once a wheel carries this.

Fifteen commits. The first six, up to 610d9ec, add the sub-region view itself. Commits
29a7485 through c198d0d are the consequence of adding views to a residency model that
assumed one tensor per allocation. The last four are performance, each measured. They are
ordered so the later ones can be dropped without disturbing the earlier.

The host-runtime `Tensor` base has no sub-region view, so amd/IRON forked one
(`XRTSubBuffer`) with an in-code TODO asking for it upstream. The fork subclasses the
backend to reach the storage and re-implements the host-dirty residency handling on its
own leaf. The same rule was then worked out independently on both sides of the repo
boundary: IRON#138 fixed it there, and #3415 proposed the same fix here before being
closed as unnecessary for this tree. An invariant a subclass cannot inherit is one each
consumer has to rediscover. This adds the view as a primitive that shares storage rather
than subclassing, so it cannot re-fork the invariant, and IRON can delete its copy.

`subview(offset_bytes, shape, dtype=None)` returns a tensor of the same backend sharing
the parent's storage, with no allocation and no copy. It bounds-checks the region, keeps the
storage alive for as long as any view of it, and reports where it came from through
`base`, `is_view` and `storage_offset`, following numpy and torch.

### Alignment

Views must not share a cache line, and the primitive enforces it. Host and device are
reconciled a line at a time rather than a byte at a time, so two views sharing a line are
not separable: synchronizing one acts on the other's bytes, and a neighbour's stale host
copy can be written back over what the device just produced.

Measured on npu4, which the shim reports as not cache-coherent
(`pdev_kmq::is_cache_coherent()` returns false), so a sync really does walk cache lines
rather than returning early. At a 4-byte-aligned but not 64-byte-aligned offset a view loses 47 to 49
of the 64 bytes it shares with its neighbours, once those neighbours are dirtied. Every
damaged byte lies inside the head or tail line it shares, and every one reads back its
pre-run contents. The remaining shared bytes are the ones where the correct output
happened to equal the stale value, so the damage is there but invisible. At
64-byte-aligned offsets, nothing is damaged.
Separately, an offset that is not a multiple of 4 returns garbage regardless of what the
host does, so element alignment (2 bytes for bf16) is not sufficient either. Both rules
are met by requiring the coherence granule. Numbers and method are in #3420.

The shape of the failure is why this is a check and not a test: it only appears when the
shared line is still dirty at the moment of the sync, so unrelated host activity that
evicts it hides the damage. Silent and timing dependent is not something a suite catches
reliably, so the layout is rejected instead.

Alignment bounds the damage but does not make a sync exactly slice-scoped. In
`amdxdna_flush_bo` only the vmap branch honours `(offset, size)`; `drm_clflush_sg` and
`drm_clflush_pages` maintain the whole buffer. The docstring says so, and you invited a driver
patch for it on #3420, which I will send separately.

### Nested views

A sub-buffer of a sub-buffer does not compose the way it looks like it does. XRT takes
the host pointer from the parent, so it picks up the parent's offset, but the device
address from the underlying allocation, so it picks up only the innermost one. A
two-level view therefore has the host reading one region while the device writes another.
Measured on npu4: a nested view that should have landed at element 16 was written at
element 0, while the equivalent single-level view landed correctly. Every handle is now
derived from the allocation at an offset accumulated from the root, which composes to any
depth.

I read this as an XRT defect rather than something to work around here: `buffer_sub`
composes the host pointer and the sync range through its parent, but takes the device
address from the underlying allocation, so two of its three paths compose and one does
not. Deriving from the root makes this branch correct either way. I have a reproducer if
it is worth filing there.

Worth flagging for review: the CPU backend composed correctly all along, so the test
backend disagreed with the real one. That divergence is what a shared base class exists to
prevent, and it is why the alignment rule is enforced on every backend rather than only
where the hazard lives.

### Coherence is a property of the memory

Adding views broke an assumption the residency model was entitled to make. With one tensor
per allocation, one `device` flag per tensor says exactly what is true. With views, one
allocation holds regions in different states, and a per-object flag cannot express that:
two tensors over the same bytes each keep their own answer and nothing reconciles them.

Four properties a caller is entitled to, written as tests in
`test/python/npu-xrt/test_coherence_is_per_region.py`. Both columns are this branch:
the left column is 610d9ec, the last commit before this change, where the view exists but
residency is still stored on each tensor; the right is the branch tip, where it is stored
per region of the allocation instead. Upstream is not in the table, because without
views none of these questions can be asked there.

| property | per-tensor (610d9ec) | per-region (tip) |
|---|---|---|
| two tensors over the same bytes report the same residency | fails | passes |
| reconciling a view leaves its parent agreeing with it | fails | passes |
| a host write through a view is visible to the buffer containing it | fails | passes |
| views of different regions still hold their own state | passes | passes |

The first three are what a per-tensor flag cannot express once two tensors can name the
same bytes. The fourth is the thing not to break while fixing them: regions that really
are independent must stay that way.

So the state moved to where the fact lives. `Storage` owns an allocation and the coherence
of the memory in it, and delegates the transfers that reconcile it to a `Transport`; it has
no shape and no dtype, because those are interpretations and interpretations are what
`NpuTensor` is for. It is not subclassed: what varies between backends is where the host's
bytes come from and what reconciling a range does, so those are the transport
(`HostOnlyTransport`, `XrtTransport`, `WholeExtentTransport`) and not a class hierarchy. `device`
became a query against the range a tensor covers rather than a field it stores, so two
names for the same bytes read one answer because there is only one to read. This is the
split torch draws between `UntypedStorage` and `Tensor`, and is where `storage_offset` comes
from.

It also gives the alignment rule its reason, and a home. No granule may hold regions in
two different states, because reconciling one range acts on whole cache lines and two
states sharing a line cannot be maintained independently. That is a property of the map,
so the map enforces it: it carries the granule its backend reconciles at and refuses a
write that would leave a state boundary off one. `subview` keeps its own check as the
early error, but it is no longer the only defence, which matters for any future path that
records state without going through it.

A buffer whose regions are all written by the same agent is safe whatever its layout;
alignment only matters when a buffer mixes host-written and device-written regions.
Splitting an allocation by direction, where that is possible, avoids the question
entirely and is usually the better answer.

Ranges stay coalesced, so a reconcile covers what was written rather than everything that
might have been. Measured on npu4 with a 4 MiB arena, 8 of its 1024 windows written by the
host and then handed to the device, again comparing 610d9ec with the tip. Pinned to one
core, minimum of many runs, since unpinned this machine varies by half:

| | per-tensor (610d9ec) | per-region (tip) |
|---|---|---|
| bytes maintained per reconcile | 4096 KiB | 32 KiB |
| reconcile wall time | 34.8 us | 6.8 us |
| residency read | 28 ns | 34-48 ns |
| `to("npu")`, already resident | 39 ns | 39 ns |

Those figures are XRT. HRX does not get them: it is on `WholeExtentTransport`, which
reconciles the whole allocation whatever range it is handed, so per-region state is tracked
there but not yet acted on. Over-maintaining is safe, which is why this is a caveat and not
a bug, and the transport is named for it rather than taking a range and ignoring it. It is
also the reason that backend wants a transport of its own: `hrx_buffer_flush_range` already
takes an offset and a size, and `HRXTensor._sync_to_device` discards them before the
transport could use them.

Worth being precise about what that first row counts: the memory is
shared, so a reconcile moves no bytes anywhere. It is cache maintenance on the host, and
the row is the span it covers, 65536 cache lines against 512. That is why the time follows
it.

Moving a residency question from an attribute read to a range query does cost something, a
few nanoseconds, and the row says so. It stays that small because almost every query is
against an allocation wholly in one state, and that answer does not depend on the range
being asked about, so it is kept on the map and checked before anything is computed. A
buffer whose regions disagree walks its spans, which is when the question needs a real
answer.

Construction is unchanged. `XRTTensor(shape, dtype=...)` still allocates and still returns
a tensor; it gets its bytes from a buffer now. No call site moves.

Nor does anything that already worked get slower. Against this branch's merge base, same
machine, pinned, minimum of many runs:

| | upstream | this branch |
|---|---|---|
| allocate | 13.5 us | 14.9 us |
| `fill_` | 44.5 us | 44.6 us |
| read one element | 8.7 us | 8.9 us |
| write one element | 17.3 us | 17.6 us |
| `to("npu")`, already resident | 0.02 us | 0.04 us |

Allocation gains about a microsecond, which is the buffer object. Everything else is within
a percent or two, and the residency fast path is tens of nanoseconds either way. Those
figures are large in absolute terms because each of them reconciles: reading one element
synchronizes the whole buffer, which is worth its own look one day and is not this PR.

For a sense of what the bookkeeping costs in the abstract, the same operations against
plain numpy arrays on the same machine. numpy is not an alternative here, since it does
not reconcile anything between two agents; it is a floor, to say how much of the time is
this class rather than the memory:

| | numpy | NpuTensor |
|---|---|---|
| fill a whole buffer | 33.1 us | 33.0 us |
| write a 4 KiB region | 0.17 us | 1.65 us |
| create a view | 0.07 us | 2.16 us |
| ask what it views | 0.02 us | 0.04 us |

Where there is real work the difference disappears: filling a buffer is memory traffic and
none of it is this class. The gaps are on operations whose numpy baseline is tens of
nanoseconds, where any bookkeeping at all is most of the cost, and they are what pays for
a write being recorded and a region knowing where it came from. Creating a view is the one
I would still like smaller; it is already down from 3.2 us and the remainder is validation
and object construction rather than anything about the device.

### Writing

`mutate()` is a scope: it reconciles on entry so partial updates see current contents,
records the write on exit, and retires the borrowed array so a reference kept past the
block fails loudly rather than writing bytes nobody will flush.

`overwrite()` is the same scope without the reconcile on entry, for a caller replacing
every byte, which nothing can observe reading back. Filling a buffer that way costs one
transfer rather than two. The caller is promising to write the whole region, and that is
the one promise here nothing checks. It is still narrower than reaching for `data`, which
makes the same promise and records nothing either way.

Marking on `.data` instead was tried and does not work. The class reads through `.data`
too, in `__repr__`, `__array__`, `__getitem__` and `numpy()`, so every read marked the
whole tensor written and collapsed the map to a single span. `.data` keeps its current
behaviour and is documented as the unmediated handle.

### Rename

The base is renamed `NpuTensor`, with `Tensor` kept as an alias so existing callers and
out-of-tree subclasses keep working; a test pins it to the same class object, since an
alias that was a separate class would silently break `isinstance`. `XRTTensor` and
`HRXTensor` keep their names, since a backend name should say which mechanism backs the
buffer.

### HRX

Not touched. A backend that still owns its storage gets a `WholeExtentTransport`, which
delegates reconciliation to the transfer methods it already has, so `HRXTensor` keeps
working unchanged and gains per-region coherence without being restructured by someone
with no hardware to verify it on.

It is also the backend this interface fits best: libhrx already flushes and invalidates by
range, which is exactly `to_device(offset, nbytes)`, where XRT needs sub-buffer derivation
to say the same thing. Adopting it properly is an `HrxTransport` of about twenty lines,
after which the tensor class has nothing backend-specific left in it. I have not written
that, because I cannot run it, and an untested range-accurate flush is worse than a tested
whole-extent one.

### Tests

652 tests, including the existing `test_tensor.py` and `test_torch_comparison.py` suites,
which are unchanged. The span map is also checked against a byte-wise reference model over
4000 randomized trials, the splice being fiddly enough that worked examples are not enough
on their own.

The new ones are built around what they fail to catch rather than what they confirm: 30
plausible regressions of this code, each a mutation of a single line, are each caught by
at least one test. Dropping either half of the alignment check, freezing the granule at
import, taking the offset in the wrong dtype's units, returning a copy, losing the parent
reference, ignoring the accumulated offset, downcasting a subclass, transposing a view's
shape. The XRT backend's derivation is covered without hardware, so a lane with no NPU
still catches a backend that ignores the offset or nests from the wrong allocation.

### Not in scope

The runtime still discards the `In`/`Out`/`InOut` direction before dispatch, so it cannot
record that a kernel wrote a buffer, and consumers re-supply that by hand. That is the
remaining reason `device` is assigned from outside the class. It wants fixing where the
information is dropped rather than here.


programming_guide/iron_configuration.md
```diff
@@ -11,7 +11,7 @@ There are several options that exist to configure the IRON Python programming en
 
 ## Default IRON Tensor Class
 
-This is a variable that controls the types of [```aie.utils.Tensor```](../python/utils/hostruntime/tensor_class.py)s that are produced by the utility functions ```tensor```, ```ones```, etc. Right now there are two tensor implementations: [```CPUOnlyTensor```](../python/utils/hostruntime/tensor_class.py) and [```XRTTensor```](../python/utils/hostruntime/xrtruntime/tensor.py).
+This is a variable that controls the types of [```aie.utils.NpuTensor```](../python/utils/hostruntime/tensor_class.py)s that are produced by the utility functions ```tensor```, ```ones```, etc. Right now there are two tensor implementations: [```CPUOnlyTensor```](../python/utils/hostruntime/tensor_class.py) and [```XRTTensor```](../python/utils/hostruntime/xrtruntime/tensor.py).
```
hunhoffe: Side note (not caused by this PR, but noting so I don't forget), the `HRXTensor` needs to be listed here too.

test/python/npu-xrt/test_xrt_tensor_subview.py
```diff
@@ -0,0 +1,173 @@
...
+buffer a view is carved from and at what offset -- on any machine, so a lane
+without an NPU still catches a backend that ignores the offset or nests from the
+wrong buffer. Only the XRT allocation call is substituted; the code under test is
+the real ``XRTTensor._subview``.
+"""
+
+import numpy as np
+import pytest
+
+pytest.importorskip("pyxrt")
```
hunhoffe, line 23: Let's keep the tests that need xrt in `test/python/npu-xrt`

test/python/npu-xrt/test_xrt_tensor_subview.py
```diff
@@ -0,0 +1,173 @@
...
+
+from aie.utils.hostruntime.tensor_class import NpuTensor  # noqa: E402
+from aie.utils.hostruntime.xrtruntime import tensor as xrt_tensor_module  # noqa: E402
+from aie.utils.hostruntime.xrtruntime.tensor import XRTTensor  # noqa: E402
+
+GRANULE = 64
+
+
+class FakeBo:
```
hunhoffe, line 33: I think it's worth doing these tests on device w/ real allocations.

I see you have both kinds of tests... what are your thoughts about consolidating to all on-device tests?

python/utils/hostruntime/tensor_class.py
```diff
@@ -77,20 +169,424 @@ def _array_to_torch(array: np.ndarray):
...
+
+    def binding_handle(self, offset, nbytes):
+        return getattr(self._tensor, "_bo", None)
+
+    @property
+    def host_bytes(self):
+        return self._tensor._data.reshape(-1).view(np.uint8)
+
+
+class CPUBuffer(NpuBuffer):
```
hunhoffe: Can you remind of the the reason for the rename from `Tensor` -> `NPUTensor`, and why `NPUBuffer` instead of `Buffer`? It seems a little odd to have CPU implementations of thing explicitly marked NPU.

python/utils/hostruntime/tensor_class.py
```diff
@@ -4,10 +4,102 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 from abc import ABC, abstractmethod
+import bisect
```
hunhoffe: This module got really big -- can we separate out into:
* One class per module, unless classes are VERY trivial?
* Helper utilities, either here or in various util files?

hunhoffe (commented):
This is a well-reasoned redesign — the per-region coherence map, the `NpuBuffer`/`NpuTensor` split, and the cache-line-alignment guard. Definitely a more sophisticated improvement to what we had so far!

 A few comments, because this is both a) hard to get fully correct, and b) hard to present to the user in a way that matches expectations from torch and/or numpy. I think making sure we really land this improvement is worth a bit of careful review.

## Correctness 
I think this might be a correctness issue, but not fully sure. The whole-tensor write methods weren't migrated, and silently lose data on a fragmented buffer. `NpuTensor`'s docstring says writes through `__setitem__`/`fill_`/factories "resynchronize." But those methods (plus `numpy()`, `to_torch()`, `torch_view()`, `__array__`, `__getitem__`, `__repr__`) are unchanged and still gate on `if self.device == "npu"`. That was fine when device was a per-instance flag; it's now a property that collapses a mixed extent to one answer, so once a subview fragments the buffer the guard fires on the wrong bytes.

Extracting `tensor_class.py` (stdlib + numpy only) with a fake backend whose host and device bytes are separate, so a missed flush is observable:
```
start:               device=npu  map=[[0,128,'npu']]
after subview write:  device=cpu  map=[[0,64,'cpu'],[64,128,'npu']]   <- extent now mixed
after t.fill_(0x22):  host[64:]=0x22  device[64:]=0xAA   <- fill_ skipped its sync
after t.to('npu'):    device[64:]=0xAA                   <- second window never pushed

driven by:

t = T((2 * G,))                      # G = COHERENCE_GRANULE; device holds 0xAA everywhere 
with t.subview(0, (G,)).mutate() as w:
     w[:] = 0x11                      # dirty one window -> map is now mixed, t.device == "cpu"
t.fill_(0x22)                        # writes 0x22 to ALL host bytes; device=="cpu" so no sync,
                                      # and the map is never told [64,128) is host-dirty
t.to("npu")                          # pushes only the pre-existing [0,64); [64,128) is lost
assert (t.buffer._device[G:] == 0x22).all()   # fails: device kept 0xAA
```

Reachable from ordinary use — write a window via `mutate()`, then touch the enclosing tensor before reconciling. The suite misses it because every `fill_`/`numpy()` in the new tests runs on a fresh (uniform) tensor or after a `.to()` that re-collapses the state first. Fix: make these methods range-aware like `to()`, or raise on a genuinely mixed extent instead of guessing. Maybe add a few more test cases too.



## API boundaries / matching user expectations

The ownership layer is a faithful match to both numpy and torch — `.base` collapsing to the root owner (I verified numpy does exactly this, even across an intermediate dtype-view), `.is_view`, zero-copy aliasing, nested-view lifetime. A user from either library will find those correct.

The subview operation matches neither, in a few places worth documenting explicitly:
- Offset units are split three ways: offset is in the parent's dtype, shape is in the view's dtype, storage_offset is reported in bytes. Neither numpy (offsets always in the array's own dtype) nor torch (storage_offset in elements) works this way. This is the single biggest surprise surface — it's tested, so it's intentional, but if the goal is matching expectations I'd reconsider it: making offset count bytes (raw-buffer model) or view-dtype elements (matching shape) collapses the footgun. As written, the closest mental model is a C typed-pointer cast, not numpy/torch.
- Reject-don't-clamp is the opposite of numpy slicing (which wraps negatives and clamps out-of-bounds). This is the correct choice for a DMA region, and the method-with-args API surface is different enough from [start:stop] that it doesn't wrongly invite numpy intuition — just worth a doc note.
- The alignment constraint has no numpy/torch analogue. Correct for coherence; novel to callers.

Suggestion: the docstring cites `torch.Storage` once as an aside — I'd make that the headline framing, since the buffer/tensor split is a torch-storage idea with no numpy equivalent (numpy users think one ndarray + `.base`, and won't expect the two-level split). Framed as "torch-storage semantics with a restricted, alignment-constrained narrow," the boundaries match expectations; framed as "like numpy," they won't.

## Smaller items
- `_CoherenceMap._coalesce` (staticmethod) appears unused — `.set()` has its own inline merge. Remove or wire in?
- `_subview` builds the view via `type(self).__new__` + manual field-setting, replaying only `NpuTensor.__init__`. A backend subclass that adds state in its own `__init__` would get an incomplete view. Nothing in-tree does this, but it's an undocumented constraint on backend authors.

python/utils/hostruntime/tensor_class.py
atassis: The rename question has a bigger answer than the name, and your `CPUBuffer` observation is what surfaces it.

`NpuBuffer` is not a DMA buffer. It is one allocation plus which agent owns which of its bytes, and DMA is one implementation of one method on it. That is why `CPUBuffer(NpuBuffer)` is legal: the class survives a CPU subclass only because it is not about the NPU. The name describes a different object than the class does, which is what made you look twice.

Two things the name is hiding.

`host_bytes` is load-bearing, since `_subview`, `CPUOnlyTensor.__init__` and `XRTTensor.__init__` all reach their data through it, and it is not on the ABC. All three concrete buffers define it and the base never requires it, so a backend that omitted it would construct fine and fail later at first use. That one is just a miss and I am fixing it.

`_TensorBackedBuffer.sync_to_device(offset, nbytes)` ignores both arguments and reconciles the whole allocation, and `binding_handle` ignores offset and returns the root BO. So on HRX the per-region state is tracked but never acted on. Over-maintaining is safe, so this is not a correctness bug, but it does mean the 32 KiB against 4096 KiB row in the description is XRT only, and I have caveated it. The root BO for a view is closer to wrong than slow. HRX is also the backend that least needs the shim: `hrx_buffer_flush_range` already takes an offset and a size, and `HRXTensor._sync_to_device` throws them away before the buffer could use them.

Across the three, exactly two things vary:

| | host bytes from | reconciling a range |
|---|---|---|
| `CPUBuffer` | `np.zeros` | nothing |
| `XRTBuffer` | mmap of a BO | `bo.sync` on a derived sub-handle |
| `_TensorBackedBuffer` | borrow the tensor array | whole extent, range discarded |

Two axes is not a hierarchy. It is one class with a transport: one concrete buffer holding nbytes, host_bytes and coherence, delegating to a `NullTransport`, `XrtTransport` or `HrxTransport`. With no subclass, nothing can re-fork the range contract the way `_TensorBackedBuffer` did, and a transport that cannot do ranges has to say so in its name instead of silently ignoring them.

On the name itself I would rather not use `Buffer`, which is taken by `aie.iron.Buffer`, a compile-time `Resolvable` that lowers to `aie.buffer` on a Tile and owns no host bytes and syncs nothing. `Storage` is free, and it makes `storage_offset` explain itself. It is also your point about framing: if the split is a torch-storage idea then the class may as well say so.

Which shape do you want before I build it? I can do it here or as a follow-up, since this PR already has the write and read path fix to land.

test/python/npu-xrt/test_xrt_tensor_subview.py
atassis: There are two axes here and this file sits on only one of them: it needs the pyxrt module, not a device. The `importorskip` is for the import, and `FakeBo` substitutes only the allocation call, so what runs is the real `XRTTensor._subview`.

That is deliberate. `npu-xrt/` is the needs-a-device tier, and the device tests there already prove the derivation is right on hardware. What this file adds is that a lane with XRT installed and no NPU, which is most of CI, still catches a backend that ignores the offset or nests from the wrong allocation. Consolidating to on-device would not move that coverage, it would drop it, and the failure it catches is the silent kind: a nested view whose host pointer and device address disagree reads back plausible data from the wrong place.

So I would rather keep both, but I do not want to invent a directory. Is there an existing home for needs-xrt-but-not-a-device, or would you like me to make one? If the answer is that `npu-xrt/` is that home and it is not gated on hardware being present, I will move it there and drop the distinction.

hunhoffe: I don't know if you can install XRT without a device? I have not tried it. Right now, we assume XRT is only installed when a device is present... so I think npu-xrt is the right location. And I know we have separate NPU-identification and lit XRT features for gating, but I'm honestly not sure how well they mesh within npu-xrt.

Since the current convention is XRT => NPU, I think we should put the tests in npu-xrt; restructuring of tests/expectations I think should be a separate PR.

python/utils/hostruntime/tensor_class.py
hunhoffe: That's a much better shape than a rename — collapsing the three subclasses into one Storage class with a `NullTransport`/`XrtTransport`/`HrxTransport` removes the thing that was bugging me, since right now the hierarchy implies "NPU" is the concept when ownership-of-bytes is. And `Storage` is a good name for it — it's the same split `PyTorch` already draws between `Tensor` and `UntypedStorage`, not just an analogy to fit our case.
  
I don't have a strong preference between the division between this and future PRs -- do whatever is easier for you to put together.

test/python/npu-xrt/test_xrt_tensor_subview.py
atassis: That settles it, thanks. Moved to `npu-xrt/`, with the directory's RUN lines and REQUIRES.

I have rewritten the docstring rather than carry it across, since "runs where hardware does not" is no longer the claim. What the file still buys is that a wrong offset, or a nest from the wrong allocation, shows up as a wrong argument rather than as wrong bytes, which is a faster thing to read off a failure than a mismatched array. The on-device tests remain the proof that the result is right.

hunhoffe:
@atassis this is a tricky, but very nice contribution! Glad to see you continue on in this direction; I'll keep an eye on this until it's merged.

programming_guide/iron_configuration.md
atassis: Listed, along with a note that HRX is never selected automatically, so a reader does not go looking for the condition that picks it.

test/python/npu-xrt/test_xrt_tensor_subview.py
atassis: Moved to `npu-xrt/`, with the directory RUN lines and REQUIRES. Discussion in the thread below.

python/utils/hostruntime/tensor_class.py
atassis: Split, into `coherence.py` (granule detection and the region map), `buffer.py` (Storage and the transports) and `torch_interop.py` (the numpy/torch dtype bridge). `tensor_class.py` keeps the tensor and goes from 1452 lines to about 1100. Every name the old module exported is re-exported, so no import in or out of tree changes.

Worth flagging what the move exposed: two tests patched `_read_sysfs_line_size` on `tensor_class`, and rebinding a re-exported name never reached the caller, which resolves against its own module globals. One of them was passing only because the real detector returned the value it asserted. They patch the module that calls the function now.

python/utils/hostruntime/tensor_class.py
```diff
@@ -3,94 +3,272 @@
...
+        self._tensor = tensor
+        self._reconcile = reconcile
+
+    def __enter__(self):
+        tensor = self._tensor
+        if self._reconcile:
+            tensor.to("cpu")
+        array = tensor.data[...]
+        self._array = array
+        return array
```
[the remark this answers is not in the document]
atassis: Fixed, and you are right that it was inconsistent rather than just wasteful. Entry now pulls without claiming, so the two failure paths agree: a body that raises leaves both the device's copy and the residency alone, whichever scope it was.

Tests for both halves, since the reconcile on entry is what a partial update depends on and dropping it would have been the easy wrong fix.

python/utils/hostruntime/tensor_class.py
atassis: Built, in this PR. `NpuBuffer` and its three subclasses are one concrete `Storage` now, delegating to a `Transport`: `HostOnlyTransport`, `XrtTransport`, `WholeExtentTransport`. Storage is not subclassed, so a backend cannot re-fork what a range means by inheriting.

It is in this PR rather than a follow-up for one reason: `NpuBuffer` is unreleased, so the rename is free today and a breaking change to a public class the moment this merges.

Two things fell out of it. `_TensorBackedBuffer` is `WholeExtentTransport`, which says in its name what it was doing silently, and the description now caveats the 32 KiB against 4096 KiB row as XRT-only for that reason. And `_storage` meant the parent tensor while `storage_offset` meant an offset into the allocation; the parent is `_parent` now.

HRX keeps whole-extent reconciliation. Its own transport is about twenty lines and `hrx_buffer_flush_range` already takes the range, but I cannot run HRX, and an untested range-accurate flush is worse than a tested whole-extent one.

`Storage` and the transports are on the host runtime API page now, since `storage_offset` previously pointed at a class the docs did not mention.

python/utils/hostruntime/buffer.py
```diff
@@ -0,0 +1,106 @@
+# tensor.py -*- Python -*-
```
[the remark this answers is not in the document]
atassis: Fixed. The split copied the header along with the license block.

python/utils/hostruntime/coherence.py
```diff
@@ -0,0 +1,199 @@
+# tensor.py -*- Python -*-
```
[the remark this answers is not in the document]
atassis: Fixed. The split copied the header along with the license block.

python/utils/hostruntime/torch_interop.py
```diff
@@ -0,0 +1,76 @@
+# tensor.py -*- Python -*-
```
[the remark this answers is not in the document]
atassis: Fixed. The split copied the header along with the license block.

atassis:
Sorry for multiple commits, had to address all the issues.
I have also landed a documentation for Storage class as well.

hunhoffe (approved):
Nice work!

hunhoffe:
looks like a few snags?

```
# | ..\..\..\test\python\test_tensor_subview.py:53: AttributeError
# | _______________ test_granule_falls_back_when_no_source_reports ________________
# | 
# | monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x000002728962F570>
# | 
# |     def test_granule_falls_back_when_no_source_reports(monkeypatch):
# |         """Windows has no os.sysconf at all; the floor is the answer there."""
# |     
# |         def no_sysconf(_):
# |             raise AttributeError("no sysconf on this platform")
# |     
# | >       monkeypatch.setattr(coherence.os, "sysconf", no_sysconf)
# | E       AttributeError: <module 'os' (frozen)> has no attribute 'sysconf'
# | 
# | ..\..\..\test\python\test_tensor_subview.py:63: AttributeError
# | ________ test_granule_falls_back_to_sysfs_when_sysconf_is_unavailable _________
# | 
# | monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x000002728962E9C0>
# | 
# |     def test_granule_falls_back_to_sysfs_when_sysconf_is_unavailable(monkeypatch):
# |         def no_sysconf(_):
# |             raise ValueError("unrecognized configuration name")
# |     
# | >       monkeypatch.setattr(coherence.os, "sysconf", no_sysconf)
# | E       AttributeError: <module 'os' (frozen)> has no attribute 'sysconf'
# | 
# | ..\..\..\test\python\test_tensor_subview.py:72: AttributeError
```

atassis:
@hunhoffe that error should be fixed now

atassis:
It should be just a test flake, you might wanna rerun only a single test instead of whole fleet of tests

PR #3432, opened 2026-07-27
Unify HRX + XRT test infra by having a runtime-agnostic npu test suite

Abhishek-Varma:
## Summary
This PR unifies the HRX and XRT on-device Python test infrastructure so a single
shared suite runs under both host runtimes, and decouples the DDR-patch
instruction ABI so HRX and XRT no longer have to share a folded `insts.bin`.
It also bumps the pinned HRX (amdxdna) release to v2026.07.30.

## Motivation
The JIT / tensor / objectfifo / e2e tests under `test/python/npu-xrt` only needed
*a* host runtime, not XRT specifically, yet they were gated on XRT and never
exercised the HRX backend. 
(Incorporates the HRX test-invocation direction from #3422. - Thank you for this @thomthehound !)
## What changed
**1. Runtime-agnostic `npu` test suite**
- Renamed the shared `test/python/npu-xrt` tests that only need a host runtime to
  `test/python/npu`, and run them under **both** runtimes; genuinely XRT-internal
  tests (cached-XRT-runtime, `full_elf_*`, `scratchpad_*`) stay in `npu-xrt`.
- Added per-RUN-line gating substitutions in `test/lit.cfg.py`:
  - `%run_on_npu{1,2}_xrt%` → device wrapper only when `pyxrt` is importable, else a no-op `echo`;
  - `%run_on_npu2_hrx%` → device-decoupled `env NPU_RUNTIME=hrx`, gated only on `hrx_python_bindings` (no longer inherits the XRT probe).
- Shared tests carry both backends' RUN lines gated by
  `xrt_python_bindings || hrx_python_bindings`; the pure-HRX CI job now also runs
  `test/python/npu`.
- HRX coverage brought to parity with XRT: tensor/torch-comparison tests build
  their `TENSOR_CLASSES` from the host's available backends, and HRX gains the
  cached-runtime tests mirroring the XRT suite.

**2. Producer-independent DDR-patch ABI**
- New `aiecc` option `--fold-ddr-addr-offset` (default `true`) controlling whether
  `appendAddressPatch` folds the AIE DDR aperture offset into `arg_plus` for
  firmware-translated args (index ≥ 5).
- The JIT selects it per active runtime (`npu_runtime_folds_ddr_addr_offset`):
  XRT/CPU keep the folded firmware ABI; `NPU_RUNTIME=hrx` compiles `insts.bin`
  unfolded so `libhrx` adds exactly one aperture offset per arg.

**3. HRX release bump**
- Bumped the pinned amdxdna release `v2026.07.20 (fcaecde)` → `v2026.07.30 (eb0b39f)`.


thomthehound:
Would you mind taking a look at #3422? There are a few changes there involving how HRX is invoked for testing.

Abhishek-Varma:
Thank you @thomthehound for catching the no-op issue. I've tried adapting the current PR accordingly. Can you please take a look at it?

python/utils/tensor_factory.py
```diff
@@ -110,6 +110,25 @@ def _probe_hrx() -> bool:
 if _NPU_RUNTIME == "auto":
     _NPU_RUNTIME = "xrt" if _probe_xrt() else "cpu"
 
+
+def npu_runtime_kind() -> str:
```
hunhoffe: Can't you just get the class name and get this information in an object-oriented way by querying the type/instance of the active Runtime object?

python/utils/tensor_factory.py
```diff
@@ -110,6 +110,25 @@ def _probe_hrx() -> bool:
...
+    """DDR-patch ABI: XRT (and CPU) consume the folded firmware ABI; HRX consumes
+    the producer-independent (unfolded) insts.bin and adds the AIE DDR aperture
+    offset for every arg itself. cl::opt defaults to true, so only pass the
+    flag when unfolding is requested.
+    """
+    return _NPU_RUNTIME
+
+
+def npu_runtime_folds_ddr_addr_offset() -> bool:
```
hunhoffe: I think this should also be a class attribute; default can be True, HRX can override.

test/python/npu-hrx/test_hrx_runtime.py
```diff
@@ -22,6 +22,7 @@
 
 import os
 import time
+from pathlib import Path
```
hunhoffe, line 26: Should we be using lit for requires? I see several different mechanisms for discerning between XRT and HRX tests -- would be nice to standardize (but not necessary for this PR, just a nice to have).

python/utils/tensor_factory.py
hunhoffe: Actually, on second thought -- is this a windows vs linux thing, or a runtime thing? Or both? @jgmelber @thomthehound did you get a feel for this?

hunhoffe (commented):
Looks good! I did have one question about the fold DDR Addr Offset control.

python/utils/tensor_factory.py
thomthehound: > Actually, on second thought -- is this a windows vs linux thing, or a runtime thing? Or both? @jgmelber @thomthehound did you get a feel for this?

Yeah, I'm thinking this is a runtime thing and not a platform thing. Not that Windows is a capable platform for this at the moment, anyway. Someday soon...

I like the class-attribute idea. Defaulting it to `True` and then overriding it for HRX seems cleaner to me.

I'm not sure about querying an actual runtime object from the compiler/cache path, though. That feels a little backwards to me.

thomthehound:
Overall: I think this is great. I'm really excited about the work you've been doing here. Super cool!

The only thing that sticks out to me, personally, is the cache handling. I don't love silently assuming the folded ABI if that capability can't be resolved. I don't think it's a huge problem, but it seems like something that should just fail clearly. I also think it might be cleaner and simpler to resolve that once and use it for both the cache key and the compilation. These are quibbles, though.

python/utils/tensor_factory.py
Abhishek-Varma: I have removed it now.

python/utils/tensor_factory.py
Abhishek-Varma: That makes sense. I've added `FOLDS_DDR_ADDR_OFFSET` as the class attribute now (overriding it for HRX as you mentioned). Can you please take a look?

test/python/npu-hrx/test_hrx_runtime.py
Abhishek-Varma: Yeah, good to standardize. Leaving it out of this PR's scope.

Abhishek-Varma:
Hi @hunhoffe @thomthehound - thank you for your review comments! I've addressed them all. Can you please take a look?

thomthehound (approved):
Looks good to me! Great work!

We are experiencing a slight CI issue right now. Hopefully we'll have that resolved shortly.

hunhoffe (approved):
Looks good to me! Excited to get this merged (once we get our CI snags ironed out, at least).

andrej:
@Abhishek-Varma the Windows CI failures are random, due to an access control policy issue. I think I figured out the issue but will need to physically access the box tomorrow to fix it. The Linux failures might be transient as well -- let me restart to see.

andrej:
For the transient Linux failures, unfortunately the best we've come up with so far is to add [this retry logic](https://github.com/Xilinx/mlir-aie/blob/main/.github/workflows/buildAndTestRyzenAI.yml#L103-L128). This only happens on Phoenix/Hawk Point, and we've not been able to nail down why, but with retrying it's reliable enough. You might need to add the same type of thing to your workflow job (re-run the job once if it fails with the set of errors listed in the linked code).

hunhoffe:
@andrej and @Abhishek-Varma would it make sense to merge this as-is, and then (if you are amenable) @Abhishek-Varma could do a follow-on PR for the retry logic?

andrej:
@hunhoffe I'll leave it up to you, no retry logic might turn CI flaky again but maybe not.

Abhishek-Varma:
Hi @andrej @hunhoffe - I've addressed the retry logic in the CI flows being added via this PR. Please take a look and let me know if anything needs tweaking. Since I don't have merge rights, I'd need help merging this in (if all looks okay). Thanks! :)

hunhoffe:
@Abhishek-Varma thank you! If we ever miss a response on a PR please do feel free to ping myself, Joe, or André as well -- sometimes we miss the github emails :)

PR #3417, opened 2026-07-24
Add ruff linter

hunhoffe:
## Description

I don't enjoy enforcing Python style constraints in PRs, so I'd like the CI to do it for me, Part 1.

## Checklist

- [ ] Tests pass locally, or the change is covered by CI
- [ ] Docs / docstrings updated if the change is user-facing
- [ ] Code is formatted (`clang-format` for C++, `black` for Python — see [CONTRIBUTING](../CONTRIBUTING.md))

PR #3436, opened 2026-07-27
Extend pyright/ruff coverage to programming_examples/ and enforce docstring style

hunhoffe:
## Description

Part 2 of https://github.com/Xilinx/mlir-aie/pull/3417

- Redefined `CompileTime[T]` as `Annotated[T, ...]` (was `Generic[T]`), fixing the root cause of most pyright errors across examples using `@iron.jit` with compile-time params.
- Extended pyright + ruff coverage from `python/{iron,utils,helpers,compiler}` to `programming_examples/algorithms`, `basic/`, `getting_started/`, `vision/`, `lit.cfg.py`, and most of `ml/` (block_datatypes, mobilenet, and 12 smaller dirs).
- Two more library-level typing fixes to eliminate whole classes of per-call-site suppressions: `Runtime.sequence()` now has `@overload`s so tuple-unpacking type-checks without a cast; `jit()` has a proper return annotation so `@iron.jit(...)` resolves to `CallableDesign` everywhere.
- Replaced ~30 bare `# pyright: ignore` suppressions on `Program(get_current_device(), ...)` with `assert device is not None`, matching the pattern already used elsewhere — `get_current_device()` can genuinely return `None`, so the ignores were masking a real (if narrow) robustness gap.
- Widened `Worker`/`ObjectFifo.{join,split,forward}`'s `tile` params to `Tile | None` (they already handled `None` at runtime; only the annotation was wrong) instead of the dishonest `cast(Tile, ...)` used in `ml/mobilenet`.
- Enforced docstring style (ruff `D`, Google convention) in `python/{iron,utils,helpers,compiler}`; fixed real content gaps (missing `Args:` entries) and reflow bugs along the way.
- A handful of real bugs found and fixed incidentally: a never-imported bare `set_current_device()` call, a variable-shadowing bug in two `tiling_exploration` examples, several genuinely-dead code blocks, and a branch-timing regression in `test/python/test_markers.py`.
- Updated `CONTRIBUTING.md`, the PR template, and dev-setup docs to reflect the new ruff/pyright scope and docstring convention.

## Checklist


jskimko:
FYI, I ran into some noise on non-ruff-covered files on pushes. Do we need to mirror the config in this hook? I.e.:

``` 
# .pre-commit-config.yaml
...
  hooks:
    - id: ruff-check
      files: ^python/(iron|utils|helpers|compiler)/ # mirrors pre-PR ruff.toml
      stages: [pre-push]
...
```

andrej (approved):
(skimmed only)

hunhoffe:
@jskimko good catch! Should be fixed now.

Issue #3438, opened 2026-07-27
Windows: >5 host-buffer designs abort on NPU (ERT_CMD_STATE_ABORT / status 6)


jgmelber:
## Summary
Two tests compile cleanly on Windows but the kernel aborts at NPU execution:
- `test/python/npu-xrt/test_jit_many_args.py` → `HostRuntimeError: Kernel returned ert_cmd_state.ERT_CMD_STATE_ABORT`
- `test/npu-xrt/many_buffers` → `Kernel did not complete. Returned status: 6`

Both pass on Linux. Currently `XFAIL: system-windows`.

## Common factor
Both exercise the **>5 host-buffer** path. The NPU firmware only pre-translates the first 5 host buffer addresses into the AIE address space; buffers beyond that have the translation offset folded into their DDR address patch by the compiler (`aie-npu-fold-ddr-addr-offset`, `lib/Targets/AIETargetNPU.cpp` `appendAddressPatch`, `kNumFirmwareTranslatedArgs = 5`, `kDDRAIEAddrOffset = 0x80000000`).

The failure is at runtime (post-codegen), so this is likely a Windows-specific issue in the address-patch values, the instruction stream, or a host/XRT interaction on the >5-buffer path — not a compile error.

## Investigation starting points
- Diff the emitted `insts.bin` / address patches for these designs between Linux and Windows.
- Confirm whether the abort is buffer-count-triggered (does a ≤5-buffer variant pass on Windows?).
- Check XRT runlist/BO handling differences on Windows.

## Done when
- Both tests pass on the Windows NPU job with `XFAIL: system-windows` removed.

thomthehound:
After *extensive* diagnostics, I can confidently say that the current Windows driver or runtime aborts any commands that bind more than seven host BOs in addition to the instruction BO. 

We should leave these XFAIL'd for now, but I'll see if there is something I can do upstream about this.

jgmelber:
> We discovered that the firmware shipped with the Linux driver adds an offset to the first five DDR buffers, which is needed to access DDR from the command processor. It only does this for the first five, and that number is hard-coded. Erika's fix was to add the same offset manually for buffers 6 and beyond to work around this.
> 
> Most likely, the firmware shipped with the Windows driver does _not_ do the same address patching, which would be the root of this failure.

_Originally posted by @andrej in https://github.com/Xilinx/mlir-aie/pull/3401#discussion_r3666921190_

jgmelber:
[...]

PR #3441, opened 2026-07-28
Fix remaining Windows NPU lit failures

### LINKED ISSUE #3437: Windows: dynamic runtime-sequence args emit invalid arith.extsi(i32)->index
## Summary
On the native-Windows NPU CI job, `test/npu-xrt/matmul_whole_array_dynamic` fails MLIR verification:

```
error: 'arith.extsi' op result #0 must be signless-fixed-width-integer-like, but got 'index'
```

The same Python passes on Linux. Currently `XFAIL: system-windows`.

## Root cause
Platform integer width (LLP64 vs LP64). On Windows, a Python `int` placed beside a runtime `i32` scalar arg is inferred as `index`-typed because `python/helpers/util.py` maps `np.longlong`/`np.uintp` → `T.index` (NumPy's default int is 32-bit on Windows, 64-bit on Linux). The vendored `eudsl-python-extras` coerce/`_binary_op` path then widens the `i32` runtime arg to `index` via `arith.extsi` — which is illegal — instead of `arith.index_cast`.

## Candidate fixes
1. **Bump the `eudsl-python-extras` pin** (`python/requirements.txt`, currently `0.1.0.20260520.827+af15e49`). A newer eudsl clone already uses `index_cast` in coerce — verify and bump.
2. Patch eudsl coerce so an `IndexType` target always uses `arith.index_cast`, never `arith.extsi`/`arith.trunci`.
3. Stop routing runtime-sequence scalar args through `T.index` in `python/helpers/util.py`.

Both (2) and (3) touch core cross-platform IR construction — needs Linux regression coverage.

## Done when
- `matmul_whole_array_dynamic` passes on the Windows NPU job with the `XFAIL: system-windows` removed.

thomthehound:
## Description

Designed to be merge compatible with #3401.

However, please note that this replaces the Windows XFAIL for `matmul_whole_array_dynamic` with a source-level fix.

This addresses the remaining Windows host-side failures not covered by #3401:

- preserve the `np.int64` to `i64` mapping when `np.longlong` aliases it on Windows, without changing the existing `np.uintp` index mapping;
- replace GNU-only `<bits/stdc++.h>` dependencies in the shared matrix multiplication host code with the standard headers and types it actually uses;
- include `libm` in the non-Windows host-link substitution instead of asking Windows clang to resolve `m.lib`; and
- keep `stale-io-bos` focused on BO destruction order by creating the kernel and I/O BOs through the same `xrt::device` handle.

Fixes #3437.

Only the >5-buffer `many_buffers` and `test_jit_many_args.py` failures (#3438) remain outside this change.

## Checklist

- [X] Tests pass locally, or the change is covered by CI
- [x] Code is formatted (`clang-format` for C++, `black` for Python — see `CONTRIBUTING.md`)

### DIFF
--- a/python/aie_lit_utils/lit_config_helpers.py
+++ b/python/aie_lit_utils/lit_config_helpers.py
@@ -595,9 +595,10 @@ class LitConfigHelper:
     def setup_host_link_substitution(config_obj) -> None:
         """Add host linker flags for tests that build XRT host executables.
 
-        Linux-hosted tests need librt/libstdc++. Windows-hosted tests link
-        against CMake-built dynamic MSVC libraries, matching CMake's default
-        /MD runtime selection.
+        Linux-hosted tests link librt, libstdc++, and libm explicitly because
+        the host compiler substitution resolves to clang rather than clang++.
+        Windows-hosted tests link against CMake-built dynamic MSVC libraries,
+        matching CMake's default /MD runtime selection.
         """
         if os.name == "nt":
             host_link_flags = " ".join(
@@ -610,7 +611,7 @@ class LitConfigHelper:
                 ]
             )
         else:
-            host_link_flags = "-lrt -lstdc++"
+            host_link_flags = "-lrt -lstdc++ -lm"
         config_obj.substitutions.append(("%host_link_flags", host_link_flags))
 
     @staticmethod
--- a/python/helpers/util.py
+++ b/python/helpers/util.py
@@ -72,10 +72,13 @@ _np_dtype_to_mlir_type_ctor = defaultdict(
         # Index Types
         # Not strictly correct, but numpy casts Python scalars to these types by
         # default, so we map them to index type to support passing lists of ints.
-        np.longlong: T.index,
         np.uintp: T.index,
     },
 )
+
+# np.longlong aliases np.int64 on Windows. Keep the explicit i64 mapping
+# authoritative there while preserving the distinct index mapping elsewhere.
+_np_dtype_to_mlir_type_ctor.setdefault(np.longlong, T.index)
 
 NpuDType = (
     np.int8
--- a/programming_examples/basic/matrix_multiplication/common.h
+++ b/programming_examples/basic/matrix_multiplication/common.h
@@ -12,14 +12,25 @@
 #define MATRIX_MULTIPLICATION_H
 
 #include <algorithm>
-#include <bits/stdc++.h>
-#include <cassert> // for assert(); not guaranteed via other headers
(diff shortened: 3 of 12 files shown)


python/helpers/util.py
```diff
@@ -72,11 +72,14 @@ def get():
...
         # default, so we map them to index type to support passing lists of ints.
-        np.longlong: T.index,
         np.uintp: T.index,
     },
 )
 
+# np.longlong aliases np.int64 on Windows. Keep the explicit i64 mapping
+# authoritative there while preserving the distinct index mapping elsewhere.
+_np_dtype_to_mlir_type_ctor.setdefault(np.longlong, T.index)
+
```
andrej, line 82: I don't follow what this buys us over having that same mapping in the dictionary above? Is this overwritten elsewhere?

python/aie_lit_utils/lit_config_helpers.py
```diff
@@ -595,9 +595,10 @@ def setup_host_compiler_substitutions(config_obj) -> None:
     def setup_host_link_substitution(config_obj) -> None:
         """Add host linker flags for tests that build XRT host executables.
 
-        Linux-hosted tests need librt/libstdc++. Windows-hosted tests link
-        against CMake-built dynamic MSVC libraries, matching CMake's default
-        /MD runtime selection.
+        Linux-hosted tests link librt, libstdc++, and libm explicitly because
+        the host compiler substitution resolves to clang rather than clang++.
```
andrej, line 599: We should really fix that ... (not as part of this PR but I think clang++ would be the correct solution)

andrej (approved):
Thank you!

python/helpers/util.py
thomthehound: Great question! That's actually the *first* way that I tried this! In short, the answer is "yes, it gets overwritten".

thomthehound: `np.longlong` aliases `np.int64` on Windows, so putting both of them in the dictionary causes the later `np.longlong: T.index` entry to overwrite `np.int64: T.i64`.

python/aie_lit_utils/lit_config_helpers.py
thomthehound: Agreed. It's something I've had in the back of my mind for months now. But there is always a more important fire to put out.

hunhoffe: Would either of you like to create an issue to record this? It sounds like something that should be recorded officially, even if it's not priority right now.

PR #3444, opened 2026-07-28
Windows CI: smoke-test wheels + build via shared build-from-wheels script

jgmelber:
## Motivation

Follow-up to the native-Windows Ryzen AI CI work in #3401. In his approval,
@andrej suggested two future directions:

> Maybe as future direction, we could also
> - add a "smoke-test wheels" job for Windows
> - make the `./utils/build-mlir-aie-from-wheels.sh` script work on Windows as well? (probably very low prio)

This PR implements both. Now that #3401 has merged, this branch is rebased
directly on `main` as 4 focused commits.

## Task 1 — smoke-test-wheels-windows job (done)

Adds a `smoke-test-wheels-windows` job to `buildRyzenWheels.yml` that
`needs: build-windows`. For each `{python_version, ENABLE_RTTI}` matrix
element it:
- runs on a clean GitHub-hosted `windows-2022` runner (no MSVC / vcpkg /
  build toolchain),
- downloads the same-run wheel artifact
  (`mlir_aie_windows_rtti_<RTTI>-<py>`, matching `build-windows`'s upload name),
- creates a clean venv, `pip install`s the wheel, and runs
  `python -c 'import aie.ir; import aie.extras; import aie.helpers'`.

This is the Windows analogue of the existing in-job Linux venv smoke test in
`build-repo` and mirrors `mlirDistro.yml`'s `smoke_test_wheels` job. It is
import-only by design; NPU execution requires the self-hosted runners and stays
on the existing `smoke-test-npu` job. Uses the same pinned action SHAs already
present in the file (`actions/setup-python` v6, `actions/download-artifact`
v8.0.1).

## Task 2 — build-mlir-aie-from-wheels.sh on Windows (done, wired up)

Andrej's second point was to make `utils/build-mlir-aie-from-wheels.sh` work on
Windows. Before this PR, `buildAndTestRyzenAIWindows.yml` hand-inlined its own
copy of the cmake/ninja/install logic while the Linux Ryzen AI job called the
script — two divergent copies of the same build. This PR makes the script
portable **and** switches the Windows workflow to call it, so the build logic
lives in one place.

**1. Make the script portable** — minimal, OS-guarded changes so it runs under
Git-for-Windows bash without regressing Linux (all Windows branches gated on
`case "$(uname -s)" in MINGW*|MSYS*|CYGWIN*)`):
- `winpath()` converts MSYS `/c/...` paths to CMake's mixed `C:/...` form for
  every path-valued `-D` arg (`CMAKE_PREFIX_PATH`, `CMAKE_MODULE_PATH`,
  `LLVM_EXTERNAL_LIT`, `PEANO_INSTALL_DIR`, `VITIS_VPP`). On Linux it is a pure
  pass-through, so Linux behavior is byte-for-byte unchanged.
- `stamp_mlir_past()` replaces the per-file `find -exec touch` (which blows
  MSYS bash's env-block limit on the Windows runner) with a single
  `python os.utime` pass; Linux keeps the original `find` path.
- Skip `-DVITIS_VPP` when `v++` is not on PATH (Vitis is Linux-only).

**2. Wire the Windows workflow to call it** — `buildAndTestRyzenAIWindows.yml`
now invokes `build-mlir-aie-from-wheels.sh` (via Git-for-Windows bash) in place
of its inlined cmake/ninja block (net −22 lines). The Windows-specific wheel
prep (python `zipfile` extract, the `llvm-debuginfod.exe` presence check, and
the `_fixup_llvm_diaguids` DIA-SDK path rewrite) stays inline, then the
already-extracted `mlir` dir is passed as arg 1 so the script skips its own
download. Windows-only `-D` options (`XRT_ROOT`, `Python3_EXECUTABLE`,
`OPENSSL_ROOT_DIR`+static, XRT python bindings, python-passes off) are passed
through `EXTRA_CMAKE_ARGS` — forward-slashed and space-free (OpenSSL via its 8.3
short path) so the script's unquoted arg expansion doesn't split them; the
provisioned Peano is passed as arg 4.

The script is therefore now **exercised end-to-end on real hardware** by the
self-hosted Windows Ryzen AI job, not just groundwork.

### Notes
- `lld`/`ccache` branches in the script only fire when those tools are on PATH;
  they are absent on the current Windows runner, so they stay no-ops there.

## Test plan
- [x] `smoke-test-wheels-windows` passes on CI for all py/RTTI matrix elements
- [x] Downloaded artifact name matches `build-windows`'s upload name
- [x] Windows Ryzen AI job (self-hosted) builds green via the script and the
      full lit suite still passes on the NPU (equivalent to the pre-rewire
      inline build)
- [x] Linux Ryzen AI job still builds via the script unchanged (no regression
      from the `uname`-gated edits)

Refs #3401

### DIFF
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -637,8 +637,8 @@ jobs:
     name: Smoke-Test Wheel on Ryzen AI (Windows)
     # Windows analogue of smoke-test-npu: install the same-commit 3.13/RTTI-ON
     # build-windows wheel via iron_setup.py (runtime mode, so the installed
-    # wheel is exercised, not a from-source build) and run a make-free @iron.jit
-    # example on the self-hosted Windows NPU. Only npu2 Windows runners exist.
+    # wheel is exercised, not a from-source build) and run a few @iron.jit
+    # designs on the self-hosted Windows NPU. Only npu2 Windows runners exist.
     # 3.13 matches the runner's pre-provisioned interpreter (docs/buildHostWinNative.md).
     needs: build-windows
     runs-on: [self-hosted, windows, ryzenai, npu2]
@@ -671,11 +671,28 @@ jobs:
           mkdir "%NPU_CACHE_HOME%"
           set XRT_CONTEXT_CACHE_SIZE=2
 
-          REM Make-free @iron.jit example (Windows has no make on PATH).
+          REM A few representative @iron.jit designs: an elementwise kernel, a
+          REM vector-vector op, a reduction, and a 2D data-movement design. Each
+          REM resolves the device via iron.get_current_device(), so they target
+          REM this npu2 runner.
           REM Use the venv interpreter explicitly: bare "python" picked up the
           REM system 3.13, which has no numpy.
           set "IRON_PY=%CD%\ironenv\Scripts\python.exe"
+
+          echo ===== basic/passthrough_kernel =====
           "%IRON_PY%" utils\run_on_npu.py npu2 "%IRON_PY%" programming_examples\basic\passthrough_kernel\passthrough_kernel.py --in1_size 4096
+          if errorlevel 1 exit /b 1
+
+          echo ===== basic/vector_vector_add =====
+          "%IRON_PY%" utils\run_on_npu.py npu2 "%IRON_PY%" programming_examples\basic\vector_vector_add\vector_vector_add.py
+          if errorlevel 1 exit /b 1
+
+          echo ===== basic/vector_reduce_add =====
+          "%IRON_PY%" utils\run_on_npu.py npu2 "%IRON_PY%" programming_examples\basic\vector_reduce_add\vector_reduce_add.py
+          if errorlevel 1 exit /b 1
+
+          echo ===== basic/matrix_scalar_add =====
+          "%IRON_PY%" utils\run_on_npu.py npu2 "%IRON_PY%" programming_examples\basic\matrix_scalar_add\matrix_scalar_add.py
           if errorlevel 1 exit /b 1
 
       - name: Cleanup NPU_CACHE_HOME
--- a/.github/workflows/buildRyzenWheels.yml
+++ b/.github/workflows/buildRyzenWheels.yml
@@ -671,8 +671,21 @@ jobs:
           mkdir "%NPU_CACHE_HOME%"
           set XRT_CONTEXT_CACHE_SIZE=2
 
+          REM TEMP diagnostics: confirm iron_env.cmd activated the venv.
+          echo ----- iron_env.cmd -----
+          type iron_env.cmd
+          echo ------------------------
+          echo VIRTUAL_ENV=%VIRTUAL_ENV%
+          echo PEANO_INSTALL_DIR=%PEANO_INSTALL_DIR%
+          echo XRT_ROOT=%XRT_ROOT%
+          echo NPU2=%NPU2%
+          where python
+
           REM Make-free @iron.jit example (Windows has no make on PATH).
-          python utils\run_on_npu.py npu2 python programming_examples\basic\passthrough_kernel\passthrough_kernel.py --in1_size 4096
+          REM Use the venv interpreter explicitly: bare "python" picked up the
+          REM system 3.13, which has no numpy.
+          set "IRON_PY=%CD%\ironenv\Scripts\python.exe"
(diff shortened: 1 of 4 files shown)


.github/workflows/buildRyzenWheels.yml
```diff
@@ -585,6 +585,54 @@ jobs:
...
+          path: wheelhouse
+
+      - name: Smoke test wheel in clean venv
+        shell: bash
+        run: |
+          python -m venv /tmp/smoke-venv
+          /tmp/smoke-venv/Scripts/python -m pip install --upgrade pip
+          /tmp/smoke-venv/Scripts/python -m pip install wheelhouse/mlir_aie*whl
+          /tmp/smoke-venv/Scripts/python -c 'import aie.ir; import aie.extras; import aie.helpers'
+
```
andrej, line 635: Do we additionally also want to test if the wheel can compile/run some simple designs like [here](https://github.com/Xilinx/mlir-aie/blob/94537df8ecc2f1a0068b3a8744052d2b0a428ad3/.github/workflows/buildRyzenWheels.yml#L325-L338)? This would require `runs-on:` one of our local runners with an NPU, of course.

andrej (approved):
Great! Like to see that the build script is also tested on Windows

.github/workflows/buildRyzenWheels.yml
jgmelber: Good idea — added in 3d49da3 as a new `smoke-test-wheels-npu-windows` job. It installs the same-commit `build-windows` wheel via `iron_setup.py --wheelhouse` (runtime mode, so the installed wheel is exercised rather than a from-source build) on a self-hosted `[self-hosted, windows, ryzenai, npu2]` runner, then runs a make-free `@iron.jit` example (`passthrough_kernel.py --in1_size 4096`) through `run_on_npu.py`.

A couple of Windows-specific differences from the Linux `smoke-test-npu` you linked:
- Uses `iron_setup.py --wheelhouse` + `iron_env.cmd` instead of `env_install.sh`/`env_setup.sh` (those are bash/Linux; the Windows job already standardizes on `iron_setup.py`).
- Runs a `@iron.jit` Python example directly rather than the `make -C programming_examples/... run` targets, since Windows has no `make` on PATH.
- npu2 only — there's no Windows npu1 runner.

.github/workflows/buildRyzenWheels.yml
```diff
@@ -585,6 +585,106 @@ jobs:
...
+          if errorlevel 1 exit /b 1
+
+          if exist "%NPU_CACHE_HOME%" rmdir /s /q "%NPU_CACHE_HOME%"
+          mkdir "%NPU_CACHE_HOME%"
+          set XRT_CONTEXT_CACHE_SIZE=2
+
+          REM Make-free @iron.jit example (Windows has no make on PATH).
+          REM Use the venv interpreter explicitly: bare "python" picked up the
+          REM system 3.13, which has no numpy.
+          set "IRON_PY=%CD%\ironenv\Scripts\python.exe"
```
jgmelber: @thomthehound any advice on this?

thomthehound: Yes; just run a few representative Python JIT examples. As long as they activate the NPU, that should be enough for the smoke test. If we are already moving away from `make`, there is no need to comment on it. In the example @andrej sited, I had only converted the `passthrough_kernel` at that point, which is why it looks disjointed. At some point in the future, it may be wise to implement a CMake based example as one of those instead.

jgmelber: I was thinking more about needing to hardcode `python.exe` from `ironenv` rather than the workflow picking up the right python from the start.

jgmelber: Done in 27ff53f — the smoke test now runs four representative `@iron.jit` designs instead of just `passthrough_kernel`: an elementwise kernel (`passthrough_kernel`), a vector-vector op (`vector_vector_add`), a reduction (`vector_reduce_add`), and a 2D data-movement design (`matrix_scalar_add`). Each resolves its device through `iron.get_current_device()`, so they target the npu2 runner without needing a `-d` flag. Also dropped the `make` commentary per your note.

Agreed on a CMake-based example being a good addition here later — #3454 is landing the CMake build/run helpers now, and #3459 tracks converting the rest, so once a converted example is available on npu2 it'd be easy to swap one in.

thomthehound: Great! Having tests directly invoked from the CLI has already paid dividends in the past.

PR #3446, opened 2026-07-28
test: drop Chess test variants that duplicate an existing Peano test

atassis:
Following up on the Chess to Peano scoping discussion (#3414), this is the
uncontroversial first slice: tests whose only purpose is to run a design through
Chess when a Peano test of the same design already exists.

Nothing here loses coverage of a design or a configuration. Each deleted test has
a Peano counterpart in the tree that runs the same Makefile parameters on the same
device, so the only thing dropped is a second compile of the same thing through
the backend we are moving away from. No production code changes, no new tests, no
extra device runs.

### programming_examples/basic (12 lit files)

| deleted | Peano test that already covers it |
| --- | --- |
| `matrix_vector/tests/run_makefile_chess.lit` | `run_makefile.lit` |
| `matrix_vector/tests/run_strix_makefile_chess.lit` | `run_strix_makefile.lit` |
| `single_core/tests/run_makefile_chess.lit` | `run_makefile.lit` (both all defaults) |
| `single_core/tests/run_makefile_bf16_chess.lit` | `run_makefile_bf16.lit` |
| `single_core/tests/run_strix_makefile_bf16_chess.lit` | `run_strix_makefile_bf16.lit` |
| `single_core/tests/run_strix_makefile_bf16_emulated_chess.lit` | `run_strix_makefile_bf16_emulated.lit` |
| `whole_array/tests/run_makefile_chess.lit` | `run_makefile_4_col.lit` (`n_aie_cols?=4`) |
| `whole_array/tests/run_makefile_bf16_chess.lit` | `run_makefile_bf16.lit` |
| `whole_array/tests/run_strix_makefile_bf16_chess.lit` | `run_strix_makefile_bf16.lit` |
| `whole_array/tests/run_strix_makefile_bf16_emulated_chess.lit` | `run_strix_makefile_bf16_emulated.lit` |
| `vector_scalar_mul/run_makefile_chess.lit` | `run_makefile.lit` (superset, adds `int_bit_width=32`) |
| `vector_scalar_mul/run_strix_makefile_chess.lit` | `run_strix_makefile.lit` (superset) |

I compared the resolved Makefile variables rather than the file names, so the
mapping accounts for defaults: `whole_array/run_makefile_chess.lit` passes no
`n_aie_cols`, which resolves to 4, matching `run_makefile_4_col.lit`.

### test/npu-xrt (3 directories)

`add_one_cpp_aiecc_xchesscc`, `add_one_func_link_with_chess` and
`add_one_scale_func_link_with_chess`. Their counterparts `add_one_cpp_aiecc`,
`add_one_func_link_with_peano` and `add_one_scale_func_link_with_peano` test the
same feature on the same devices through Peano/lld, as the file headers say
("End-to-end test for func-level `link_with` (Chess/xbridge backend)" against
"(Peano/lld backend)"). The only substantive difference I found between the pairs
is an unused `%c8 = arith.constant 8 : i64` in the Peano copy.

### Evidence

From the CI run on the base commit `a079d8b0a87` (run 30278176623, all checks green),
on both device runners:

* Each of the 15 Peano tests listed above passed on its matching runner and was
  correctly Unsupported on the other. Neither device suite reported a single
  FAIL or TIMEOUT, and none of the 15 carries an `XFAIL` marker, so "a Peano test
  already covers it" is a green test rather than a skipped one.
* The tests I am deleting were live, not dead weight: they ran and passed too.
  That is the point, they were running the same designs a second time through
  Chess.

Their measured cost in that run was about 1535s of device-runner time, 911s on
`aie2-4col` and 624s on `aie2p-8col`. That is a lower bound, since the log only
times the 64 slowest tests per job and `add_one_cpp_aiecc_xchesscc` falls outside
that.

### What I deliberately left alone

`single_core/tests/run_strix_makefile_chess.lit` and
`whole_array/tests/run_strix_makefile_chess.lit` look like the same pattern but
are not. Each runs its example with all defaults on npu2, and no Peano test covers
that configuration: the npu2 Peano tests for single_core all pin different
parameters, and for whole_array they cover 1, 2 and 8 columns plus 4 columns at
i8, but not 4 columns at i16/i32. Deleting those would drop a configuration rather
than a backend, so they want converting to Peano instead. I can send that
separately with the device run attached.

The `block_datatypes` examples still need Chess and are untouched, and I left the
`use_chess` plumbing in `makefile-common` and the designs alone so this change
stays purely subtractive and easy to revert.

### One question while I am here

The discussion says Peano is the default, but `aiecc` still has
`--xchesscc`/`--xbridge` at `cl::init(true)` and CMake defaults `AIE_COMPILER` and
`AIE_LINKER` to `XCHESSCC`. `%backend_flags` only steers tests onto Peano when
Chess is absent, so on a Vitis-equipped runner the 62 tests that invoke a bare
`%aiecc` still go through Chess. Is flipping those defaults something you would
take, or is the Chess-when-present default deliberate for now? Happy to prepare it
if it is wanted, though it needs your CI to validate.


hunhoffe:
@atassis I polled a few folks here and we're all in agreement that peano as default in aiecc would be good!

atassis:
Some great news, nice! I would like to, per se 'tear out' the chess, since the single clearly valuable part is ISS, it seems. Most other stuff is stale or can be rewritten and tested. 
Per ISS- it might take a while to implement a decent design to create its alternative. And also I have too much info now to consume, since, still, I am kinda new to the discipline. I want to become fluent in the ecosystem, this 'hobby' definitely helps=)

hunhoffe:
Sounds good! I think this PR is a good start! Your work in this area is appreciated, and we're glad that you picked this ecosystem as your 'hobby' 😁

atassis:
I'm working towards making it more than a hobby over time:)
And also on the aiecc defaults- should I prepare the flip?

hunhoffe:
@atassis Yep! I'm happy to review the flip whenever it's submitted.

jgmelber (approved):

PR #3451, opened 2026-07-29
test: state each test's backend, and move off Chess where it is not needed

atassis:
A test's backend should be readable in the test. For a good part of the suite it currently is not.

`aiecc` defaults to Chess (`xchesscc` and `xbridge` are both `cl::init(true)`), and `%backend_flags`
expanded to `--no-xchesscc --no-xbridge` only when Chess was *absent*. On a machine with Vitis it expanded
to nothing, so the test compiled with Chess; on a machine without, the same test compiled with Peano. That
is visible in CI today: those tests build with Chess on the Linux Ryzen AI runners and with Peano on the
Windows one, from the same source.

This does two things.

**Make the choice explicit.** 31 Chess-gated tests relied on aiecc's default; they now pass
`--xchesscc --xbridge`. That is a no-op, since `resolveOptions()` already resolves the unflagged case to
`(xchesscc=true, xbridge=true)`, and these tests are UNSUPPORTED where Chess is absent. Then
`%backend_flags` is replaced by the literal flags at each of its 50 call sites and the substitution and its
three call sites are removed, so no test's backend depends on the machine any more.

**Move the tests that never needed Chess.** 38 tests built an ordinary `aie_api` or scalar kernel with
`xchesscc_wrapper`, or simply asked aiecc for Chess, and only ran where Vitis was installed. They now
compile with Peano and gate on `peano`. Chess-gated tests go from 111 to 73.

The Peano kernel invocation follows the one already in
`test/npu-xrt/matmul_whole_array_dynamic/run_npu1.lit`, including `-D__AIE_API_AIE_ADF_HPP__` so stock
`aie_api` does not reach for `<adf.h>`, which ships only with Vitis. Per-test `-D` flags and architectures
are preserved.

Commits are separated by class, so a failure maps to one: declaration, substitution removal, then the
conversions grouped by shape.

### Verified

For every converted test that builds a kernel I parsed the RUN lines back out, joined the continuations,
expanded `%S` and `%PEANO_INSTALL_DIR`, and ran them. All produce an object with the pinned Peano. The
remaining conversions are flag and gate changes with nothing to compile locally, so CI is their first
check.

I also swept for the inverse mistake: no test gated on `peano` reaches `aiecc` without `--no-xchesscc`.
One did, `vector_scalar_using_dma`, which passed a bare `--xbridge`; since either flag promotes to both,
it would have compiled its core with Chess while requiring only `peano`, and failed outright without
Vitis. That is fixed in the same commit that converts it.

### What stays on Chess

73 tests, deliberately:

- 53 drive the Chess instruction-set simulator. `--get-aiesim` requires `--xbridge`, and there is no open
  equivalent, so these cannot move.
- 11 are the `block_datatypes` bfp16 examples. Peano's AIE2P backend still aborts the legalizer on
  `<8 x s8> G_BUILD_VECTOR` in `matmul_vectorized_bfp16`; llvm-aie#1184 fixes it and is open.
- 9 exist to test the Chess path itself, including the `simple*.mlir` cases that FileCheck both backends
  side by side. Converting those would delete coverage rather than move it.

### Follow-on

With this in, flipping aiecc's default to Peano becomes a no-op for every test in the tree: Chess-gated
tests pass explicit Chess flags, `--get-aiesim` still promotes `xbridge` back on, and everything else
declares Peano. @hunhoffe indicated on #3446 that the default flip would be welcome; I would rather land it
against a tree where nothing silently changes backend.


thomthehound:
The reason that `%backend_flags` was there is because it seems silly to keep saying `--no-xchesscc --no-xbridge` when, in the future, Chess is intended to be the exception to the rule instead of the rule itself. I had been hoping that, at some point, the ceremony could be dropped and the default would be "no flags", and "`--xchesscc --xbridge`" only when Chess or Vitis are truly needed. It was intended as a transition forward, so this seems kind of reversed to me. But I realize the repo is fighting you on that right now.

Otherwise, this tests green for me!

atassis:
I have removed that so it can be easy to grep search tests, that's why I am aiming explicitness now. Also, `%backend_flags` introduces another mind game about how flags work, it should be either this or explicit statement, I have chosen last to be executed. So yes, you are right on both "no flags" and that repo is fighting not me, but everyone=).
I'll just also admit that all this leads us for easier removal of all flags related to chess/peano branching. I mean that once we move all onto single type of rails (migrate all chess related stuff onto Peano)- it would be an easy move to flip flags and remove redundant flags from tests. 
Later I'll try to make tests be grouped by where they should be run. What I mean is to removed REQUIRE statement, because each hardware will receive it's own directory with tests, which let's us to still do compilation first and tests later as #3403 suggests. But these ideas are not polished enough, still. But iterating slowly through the fixes will help to get a better idea of what to change next

atassis:
@hunhoffe I have opened a lot of PRs in the last few days, so to save you the triage: this is the one I would take first.
It should both allow easier refactoring of the CI in future and shorten longest workflow duration.

hunhoffe:
@atassis whoops! Somehow I missed the one I was supposed to see first.... I see it now!

atassis:
Fixed!

thomthehound:
> I have removed that so it can be easy to grep search tests, that's why I am aiming explicitness now. Also, `%backend_flags` introduces another mind game about how flags work, it should be either this or explicit statement, I have chosen last to be executed. So yes, you are right on both "no flags" and that repo is fighting not me, but everyone=). I'll just also admit that all this leads us for easier removal of all flags related to chess/peano branching. I mean that once we move all onto single type of rails (migrate all chess related stuff onto Peano)- it would be an easy move to flip flags and remove redundant flags from tests. Later I'll try to make tests be grouped by where they should be run. What I mean is to removed REQUIRE statement, because each hardware will receive it's own directory with tests, which let's us to still do compilation first and tests later as #3403 suggests. But these ideas are not polished enough, still. But iterating slowly through the fixes will help to get a better idea of what to change next

Works for me! Glad you've thought about it. I can't wait to get rid of those flags.

atassis:
@thomthehound These shenanigans are my stronger side:)

atassis:
@hunhoffe pushed another commit with a fix of tests

atassis:
@thomthehound I see that, I could produce more work, but everything is stalled with CI. I have some work going on on CI topic, will try to land a PR in next few hours. I think, few more iterations and we can reduce `buildAndTestRyzenAI.yaml` to less than 10 minutes (this is my goal).

hunhoffe (approved):
Would be good to record the tests/reasons why things couldn't be migrated in the discussion and/or an issue. Maybe you can remember, but I probably can't!

PR #3452, opened 2026-07-29
HSA host runtime

ypapadop-amd:
## Description

Adds an opt-in **HSA/ROCR host runtime** for IRON, so `@iron.jit` designs can be
dispatched on the XDNA NPU through `libhsa-runtime64.so` instead of XRT. It uses
ROCR's native AIE support (`HSA_DEVICE_TYPE_AIE`,
`hsa_amd_aie_kernel_dispatch_packet_t`) and needs no XRT userspace at runtime.

Selected with `NPU_RUNTIME=hsa`, alongside the existing `xrt` and `hrx`
backends. It is **strictly opt-in** — `auto` never selects it, so an XRT-less
host still degrades to CPU-only tensors.

```bash
NPU_RUNTIME=hsa python programming_examples/basic/vector_scalar_add/vector_scalar_add.py
# PASS!
```

The backend is pure Python (ctypes bindings to ROCR). There is no C++ host stack
and no build-system component.

### Status

Runs end-to-end on hardware (Phoenix / npu1). `vector_scalar_add` passes, as
does a 9-argument design and a 100-dispatch chain that wraps the 64-slot queue.

## What's here

New package `python/utils/hostruntime/hsaruntime/`:

| Module | Responsibility |
|--------|----------------|
| `discovery.py` | Locate a ROCm install. No `dlopen`, so it doubles as the cheap `has_hsa` probe. |
| `_bindings.py` | Raw C ABI: constants, ctypes structs, lazily-bound `lib` handle. Importing the package performs no `dlopen` and no device init. |
| `context.py` | `HSAContext` — process-wide singleton owning the agents, memory pools, queue, kernarg pool and completion signal; fills and enqueues dispatch packets. |
| `tensor.py` | `HSATensor` — a vmem allocation mapped for CPU + AIE, exposed as a zero-copy numpy view. |
| `hostruntime.py` | `HSAHostRuntime` / `CachedHSAHostRuntime` - the IRON `HostRuntime` interface. |

Plus `NPU_RUNTIME=hsa` wiring in `tensor_factory.py` / `utils/__init__.py`, the
package files in `python/CMakeLists.txt`, and documentation.

## Design notes

A few things are non-obvious and were established against hardware rather than
docs:

- **The PDI must come from the device-heap pool.** The driver's `aie2_config_cu`
  accepts only `AMDXDNA_BO_DEV` buffer objects for a CU's PDI, and ROCR produces
  those only from the DEVICE_SVM pool. Anything else fails the submit ioctl with
  `EIO`. ROCR distinguishes that pool by reporting a recommended allocation
  granule of 0.
- **Device generation comes from the HSA agent's ISA name** - `aie2` is
  Phoenix/npu1, `aie2p` is Strix/npu2. An unrecognized name raises rather than
  defaulting: guessing wrong compiles for the wrong architecture and wedges the
  NPU until the driver's timeout fires.
- **Kernarg slots are indexed by queue ring slot.** Slot *i* belongs to ring slot
  *i*, so reuse needs no bookkeeping - a ring slot is only rewritten once the
  device has consumed its previous packet, which is exactly when it has finished
  reading those kernargs.

## Limitations

- **No trace capture.** A design carrying a `trace_config` is rejected up front;
  use `NPU_RUNTIME=xrt` for trace-enabled designs.
- **Python flow only.** There is no C++ host stack, so `make`-driven examples
  ignore `NPU_RUNTIME=hsa`.

## Documentation

- `programming_guide/hsa_runtime.md` - architecture overview (added to the
  mkdocs nav).
- `python/utils/hostruntime/hsaruntime/README.md` - prerequisites, ROCm
  selection, troubleshooting.
- `programming_guide/iron_configuration.md` - `hsa` added to the `NPU_RUNTIME`
  table, plus a configuration section for the four environment variables
  (`ROCM_PATH`, `IRON_HSA_DEVICE`, `HSA_EXE_CACHE_SIZE`, `IRON_HSA_TIMEOUT`).

ROCm is located by installation *root*, in order: `ROCM_PATH`, a pip-installed
ROCm from [TheRock](https://github.com/ROCm/TheRock/blob/main/RELEASES.md), then
`/opt/rocm`. Note that both `libhsa-runtime64.so` and the bare SONAME
`libhsa-runtime64.so.1` are accepted - TheRock's runtime wheels contain no
symlinks, so a pip-installed ROCm provides only the versioned name.

## Testing

41 host-side tests (no NPU required) covering discovery precedence, backend
selection, the dispatch/chain/timeout logic against a fake ROCR, and the vmem
teardown ordering. On hardware: `vector_scalar_add`, a 9-argument design, and
chain wrap-around past the queue capacity, with the XRT backend re-run
throughout to confirm no regression.

## Checklist


python/utils/hostruntime/hsaruntime/_bindings.py
andrej, line 1: Is this auto-generated from `hsa.h` or manual?

andrej (approved):
Congratulations on getting this to work with some examples.

I have scanned it and the code looks high-quality. I can't claim to understand (or have looked at) all the intricacies but from 1,000 ft seems reasonable.

I do have two future direction questions:

1. A lot of low-level logic (queue submission, allocation into pools, ...) is now implemented in Python here. Is support for other languages an explicit non-goal? Or is the plan to prototype this logic in Python and then later write a library in C++ (or other language) that can export bindings into other languages for all users? As it stands, it seems getting the C++-based examples running would required duplicating a lot of this code in C++.
2. XRT/aiebu has some advanced "patching" functionality, namely for DDR buffers (patch address_patch on the host rather than executing it in firmware), for load_pdi instructions (XRT will append PDIs into the right memory region, and patch the load_pdi address, enabling multiple "designs/reconfigurations" in a single dispatch), and for the scratchpad parameter functionality. Is the plan to replicate any of this in your runtime (with any additional metadata that may be needed for this emitted by MLIR-AIE, or parsing the ELF in your runtime)?

I know both of these are wildly expanding the scope of this PR. Don't mistake it for me asking you why you haven't done this. I'm just interested in future direction.

python/utils/hostruntime/hsaruntime/discovery.py
```diff
@@ -0,0 +1,122 @@
...
+from collections.abc import Iterator
+from pathlib import Path
+
+__all__ = ["find_hsa_include_dir", "find_libhsa", "hsa_available"]
+
+# Explicit user-chosen root. ROCM_PATH is the ROCm-wide convention.
+_ENV_ROOT_VARS = ("ROCM_PATH",)
+
+# Conventional system install location.
+_SYSTEM_ROOTS = (Path("/opt/rocm"),)
```
hunhoffe: Does this need to be windows safe? Overall, it would be neat to make sure this is as platform independent as possible, as a matter of best practice even if HSA does not support windows now.

python/utils/hostruntime/hsaruntime/hostruntime.py
```diff
@@ -0,0 +1,338 @@
...
+from .context import HSAContext
+from .tensor import HSATensor
+
+if TYPE_CHECKING:
+    from aie.iron.device import Device
+
+_TRACE_UNSUPPORTED_MSG = (
+    "Trace capture is not supported on the HSA backend. Re-run without a "
+    "trace_config, or use the XRT backend (NPU_RUNTIME=xrt) for trace-enabled "
+    "designs."
```
hunhoffe, line 41: Is this a future feature that should be tracked in an issue? What is the main blocker here?

python/utils/hostruntime/hsaruntime/README.md
```diff
@@ -0,0 +1,116 @@
+<!---//===- README.md ---------------------------*- Markdown -*-===//
+//
```
hunhoffe, line 2: Is this file included in the webpage in a reasonable spot?

hunhoffe (approved):
Looks like good work that fits nicely within the existing structure, and nice documentation too! I left a few minor comments/questions.

@ypapadop-amd do you want to add yourself as the code maintainer for the HSA backend? And if you do intent to maintain it longer-term, let's discuss setting up the runners for testing.

python/utils/hostruntime/hsaruntime/_bindings.py
ypapadop-amd: It's agent-generated, minimum API needed to support it in Python.

ypapadop-amd:
> 1. A lot of low-level logic (queue submission, allocation into pools, ...) is now implemented in Python here. Is support for other languages an explicit non-goal? Or is the plan to prototype this logic in Python and then later write a library in C++ (or other language) that can export bindings into other languages for all users? As it stands, it seems getting the C++-based examples running would required duplicating a lot of this code in C++.

HSA as a spec defines only the bare minimum (queues, pools, etc.), leaving the rest to the user of the runtime. ROCR does not offer these capabilities either; the rationale is that you get to pick how you want to do allocation that fits best your use-case. We could abstract it in a different library, but we'll be adding another layer that would potentially be IRON specific. Other libraries based on HSA follow a slightly different pattern when it comes to allocation, e.g., if you want to allocate kernargs that can be used from either GPU or NPU and decide at dispatch, then you need additional logic that is not needed here.

> 2. XRT/aiebu has some advanced "patching" functionality, namely for DDR buffers (patch address_patch on the host rather than executing it in firmware), for load_pdi instructions (XRT will append PDIs into the right memory region, and patch the load_pdi address, enabling multiple "designs/reconfigurations" in a single dispatch), and for the scratchpad parameter functionality. Is the plan to replicate any of this in your runtime (with any additional metadata that may be needed for this emitted by MLIR-AIE, or parsing the ELF in your runtime)?

Yes, this probably belongs in ROCR; I am planning to remove PDI/insts buffers and switch to hsaco loading for transparent loading between archs. I can add some metadata in the ELF if this requires extra work from the runtime and do it kernel load time.

ypapadop-amd:
> @ypapadop-amd do you want to add yourself as the code maintainer for the HSA backend? And if you do intent to maintain it longer-term, let's discuss setting up the runners for testing.

Yes. I'll add myself to the maintainers. Runners should be set-up too, I'll need some guidance there.

python/utils/hostruntime/hsaruntime/discovery.py
ypapadop-amd: AIE support in ROCR is Linux only for now unfortunately; and work to support ROCR is still ongoing AFAIK. I can guard it better if needed.

python/utils/hostruntime/hsaruntime/hostruntime.py
ypapadop-amd: I have never used the tracer. Does XRT do anything special for the tracer?

python/utils/hostruntime/hsaruntime/README.md
ypapadop-amd: It's included in a top-level file. Do I need to move it?

python/utils/hostruntime/hsaruntime/hostruntime.py
hunhoffe: I don't think so? But I don't think we should merge something that is missing features without documenting it somewhere.

python/utils/hostruntime/hsaruntime/README.md
hunhoffe: I don't believe that README's are consumed from code directories. Most of the documentation is written directly into the code in docstrings; most other forms of documentation live in programming_examples or docs. So you would need to include this explicitly, I think.

python/utils/hostruntime/hsaruntime/hostruntime.py
ypapadop-amd: https://github.com/Xilinx/mlir-aie/issues/3552

andrej:
@ypapadop-amd What requirements do you have for runners beyond the current setup? If there's a couple extra dependencies you need installed, happy to do so.

python/utils/hostruntime/hsaruntime/README.md
ypapadop-amd: Fixed.

python/utils/hostruntime/hsaruntime/discovery.py
ypapadop-amd: Guarded it to make sure that is Linux only.

ypapadop-amd:
> @ypapadop-amd What requirements do you have for runners beyond the current setup? If there's a couple extra dependencies you need installed, happy to do so.

It works with the latest TheROCK wheels. It shouldn't need anything past that. Let's try to set it this week.

PR #3455, opened 2026-07-29
[AIE] getTargetModel: reject an unknown AIEDevice instead of answering VC1902

atassis:
## Root cause

`getTargetModel(AIEDevice)` switches over all 15 enumerators and then ends with:

```c++
  case AIEDevice::npu2_7col:
    return NPU2model7col;
  }
  return VC1902model;
```

The switch is exhaustive, so `-Wswitch` stays quiet and the trailing `return` is dead for every
valid enumerator. It is live for anything that is not an enumerator at all, and it answers with a
different device family -- Versal VC1902, 50 columns x 9 rows -- rather than reporting an error.

## Why it is reachable

The enum is not the only entry point. `aieGetTargetModel` casts a raw `uint32_t`:

```c++
AieTargetModel aieGetTargetModel(uint32_t device) {
  return wrap(
      xilinx::AIE::getTargetModel(static_cast<xilinx::AIE::AIEDevice>(device)));
}
```

`python/AIEMLIRModule.cpp` binds that as `aie.dialects.aie.get_target_model`, and
`python/iron/device/device.py` passes its `device` parameter straight through. So an out-of-range
device integer yields VC1902 column and row counts instead of raising.

`AIEDevice` values run 1..15 contiguously, so both `0` and anything above `15` land here. Zero is
the value a default-initialized device carries.

## Fix

Replace the silent fallback with `llvm::report_fatal_error`, naming the offending value.
`report_fatal_error` is already the idiom in this file (four existing call sites).

No `default:` label is added. Keeping the switch exhaustive is what makes `-Wswitch` report a newly
added device, which is a separate and useful signal; the fallthrough exists only for values that are
not enumerators, and those cannot be covered by a case label anyway.

## Evidence

New `test/CppTests/target_model_unknown_device.cpp`, alongside the existing `target_model.cpp` /
`target_model_rtti.cpp`. It installs a fatal-error handler and checks the message in-process, so it
stays a plain exit-status test like its siblings.

A first version asserted the abort with CTest's `PASS_REGULAR_EXPRESSION` instead. That does not
work: CTest reports a signal-terminated child as `Subprocess aborted` and fails the test regardless
of the property. The handler approach also keeps this portable to MSVC.

```
$ ctest --test-dir build/test/CppTests
    Start 1: TargetModel                    Passed
    Start 2: TargetModelRtti                Passed
    Start 3: TargetModelUnknownDevice       Passed
    Start 4: RegisterDatabase               Passed
    Start 5: Bfloat16Fallback               Passed
100% tests passed out of 5
```

Reverting only the `AIEDialect.cpp` hunk and rebuilding makes the new test exit 1, so it fails on
the unfixed compiler rather than passing either way.

## What this does not change

`getTargetModel(Operation *)` keeps its documented VCK190 fallback for an op with no enclosing
`AIETarget` -- that one is deliberate and marked "for backward compatibility". Only the `AIEDevice`
overload changes.

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -239,7 +239,13 @@ const AIETargetModel &xilinx::AIE::getTargetModel(AIEDevice device) {
   case AIEDevice::npu2_7col:
     return NPU2model7col;
   }
-  return VC1902model;
+  // The switch above is exhaustive over AIEDevice, so no default: -- that keeps
+  // -Wswitch reporting a newly added device. A value outside the enum can still
+  // arrive through the C API, which casts an unchecked uint32_t
+  // (aieGetTargetModel), so fail loudly rather than silently answering with an
+  // unrelated device family.
+  llvm::report_fatal_error("getTargetModel: unknown AIEDevice value " +
+                           llvm::Twine(static_cast<uint32_t>(device)));
 }
 
 // Walk the operation hierarchy until we find a containing TileElement.
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -239,11 +239,9 @@ const AIETargetModel &xilinx::AIE::getTargetModel(AIEDevice device) {
   case AIEDevice::npu2_7col:
     return NPU2model7col;
   }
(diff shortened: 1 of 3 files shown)


fifield (approved):

atassis:
Hi there!
Applied copilot changes, trimmed comments  a bit, were too verbose, and fixed CI lint issues too.

PR #3456, opened 2026-07-29
[AIEX] Resolve the reset registers through the register database

atassis:
Both reset lowerings restated hardware layout as local constants: the CORE_CONTROL offset and its reset
bit in `AIELowerCoreReset`, and the DMA channel CTRL reset bit in `AIELowerDmaChannelReset`. That layout
is already in the register database, so the constants were a second, hand-transcribed copy of it.

They now resolve through `lookupRegister` + `getField` + `encodeFieldValue` + `getFieldMask`. Two files,
and it deletes the constants rather than adding another way to spell them.

## What changed since the first version

This PR originally added a `RegField` / `encodeRegisterField` / `createMaskWriteField` helper under
`lib/Dialect/AIEX/Utils/`. That is withdrawn — the database does the same job better. It is generated
from the documentation instead of transcribed by hand, it resolves the offset per tile module instead of
hardcoding `0x32000`, and describing a field as a bit range makes the malformed mask my helper had to
validate at runtime unrepresentable. The helper, its unit test and their build entries are gone.

## The DMA address

Still from `getDmaControlAddress`. The database does carry a per-channel, per-direction register whose
offset matches that function for every channel on both core and mem tiles, so keeping it is consistency
with the other caller in `AIEDmaToNpu.cpp` rather than a gap in the database.

## One behaviour change

`CoreResetOp`'s verifier rejects AIE1; `DmaChannelResetOp`'s does not. AIE1 has no register database, so
`aiex.dma_channel_reset` on AIE1 now fails with a diagnostic where it previously emitted a
`maskwrite32` built from an AIE2 mask against an AIE1 address.

Erroring is the better answer, but it is a change rather than a no-op, and the verifier asymmetry is
worth closing on its own — happy to send that separately.

## Evidence

`test/Passes/lower-core-reset` and the five files under `test/Passes/lower-dma-channel-reset` pass
**unmodified**, including the six-pass `dma_channel_reset_for_pipeline` case and both
`-verify-diagnostics` files. Same addresses, values and masks, which is the equivalence check for a
change of this shape.

The new lookup-failure branches have no lit coverage: through a legal device they are unreachable except
on the AIE1 path above, so there is nothing to write a positive case against.


fifield:
At first glance some of this overlaps with the regdb functionality or could land there? Take a look at `AIETargetModel::lookupRegister/encodeFieldValue/getFieldMask` and see if they work or could be improved for your use case. Another solution I've considered is to generalize (rename) the `aie.trace.reg` op and its lowering to apply to all register writes. The trace implementation is a bit of a trial-run of that more general functionality. Then fields can be accessed by name and the encoding is handled by the lowering.

atassis:
They work, thanks. Dropping the helper and converting `AIELowerCoreReset` and `AIELowerDmaChannelReset` onto `lookupRegister` + `getFieldMask`, which deletes the hardcoded address and mask. Existing lit tests should be unchanged.

One small thing while I was in there: `encodeFieldValue` truncates on overflow with a warning on `errs()`, where a pass would rather have a diagnostic. Does not affect these callers. Happy to send separately if useful.

Generalizing `aie.trace.reg` sounds right to me, and the reset ops would be a natural first non-trace consumer if you want one to design against.

atassis:
Pushed. The helper is withdrawn and the two reset lowerings now go through the database; net diff against main is the two lowering files. Retitled to match.

I have not touched the `aie.trace.reg` generalization here. Between trace and these two there are now three callers doing the same lookup-and-encode sequence, so rather than answer that with a yes or no I would like to look at the register-access surface as a whole first and come back with something more considered.

fifield (approved):

PR #3458, opened 2026-07-29
[AIEX] Document the bulk-reset loop and point re-arm at dma_channel_reset_for

atassis:
Two gaps in `aiex.dma_channel_reset`'s description.

## There is no reset-all op, and no statement of why

The question recurs because aie-rt exposes `XAie_DmaChannelResetAll`, so it looks like a missing
primitive. It is not one: that function is a bare loop over `XAie_DmaChannelReset`, MM2S then S2MM,
with no tile-wide reset bit underneath it (`third_party/aie-rt/driver/src/dma/xaie_dma.c`). A caller can
already write that loop with this op. The description now says so, and gives the target-model query
that bounds it -- the same one the op's verifier uses for `channel`.

Also stated: resetting a channel that was never configured is safe. aie-rt gates the reset on the
channel index being in range for the tile and direction, not on configuration state.

## The resident-objectFIFO case pointed at the hard path

A resident FIFO needs three things, not one: the reset, a lock re-arm (the reset freezes the bound lock
counters), and a START_QUEUE re-push (an aie2p DMA channel has no enable bit, so once the reset drains
the queue nothing else restarts it). #3393 added `aiex.dma_channel_reset_for`, which resolves all three
from the fifo's `aie.objectfifo_rearm_binding`. The description now points there for the single-FIFO
case and reserves the raw loop for the genuinely tile-wide case `dma_channel_reset_for` does not cover.

It also notes that `AIEVerifyRuntimeRearmPass` checks only the lock half. Its own comment says as much:

> the complementary start-queue re-push ... is not checked, so a design that re-arms locks but not the
> queue still deadlocks and still passes.

Without that note the pass reads like a completeness check for the whole re-arm, which is the failure
mode worth warning about -- it is silent, with no diagnostic.

## Evidence

Both claims were read out of the sources they cite rather than restated from memory:
`AIELowerDmaChannelReset.cpp` enumerates the trio in the order the lowering emits it, and
`AIEVerifyRuntimeRearm.cpp`'s header comment states its own limit.

`mlir-tblgen --gen-dialect-doc -dialect=aiex` regenerates cleanly and the new paragraphs render as
prose, not as code blocks -- worth checking here specifically because #3374/#3416 moved the
runtime-sequence op catalog into this same generated page.


hunhoffe (approved):

hunhoffe:
I do enjoy some good documentation!

atassis:
I am thinking about if we can strip out running tests against documentation, if that won't break anything. This looks excessive

atassis:
The red aie2-4col job is the runner, not this change. The diff is 22 lines of op description text in `AIEX.td` and nothing else, so there is no code here that a runtime test can reach.

The failure is `DRM_IOCTL_AMDXDNA_CREATE_HWCTX IOCTL failed (err=-2)` out of `pyxrt.hw_context()`, in `basic/matrix_scalar_add`. The same ioctl fails the same way on main: https://github.com/Xilinx/mlir-aie/actions/runs/30561912778/job/90936848624, the merge-queue run for #3466, on `basic/matrix_multiplication/whole_array/tests/run_makefile_4_col_i8.lit`. In the same window Ryzen AI jobs went red on ten other branches, three of them merge-queue runs on main.

Could someone re-run the job once the runners are healthy?

hunhoffe:
@atassis updated & restarted!

PR #3463, opened 2026-07-30
[AIEX] Guard verifyTaskBDDimensions against a mismatched dma_bd sizes list

atassis:
### What

Add a guard in `verifyTaskBDDimensions` (`AIEXDialect.cpp`) so a `dma_bd` with a malformed `static_sizes`/`sizes` operand count, nested inside a `DMAConfigureTaskOp`, gets a diagnostic instead of crashing the verifier. Adds `bad-20.mlir`, a hand written generic-syntax test for this exact state, and fixes a one character typo (`sepcify` to `specify`) in `bad-2.mlir`'s comment while I was in the file.

### Why

`AIE::DMABDOp::verify()` already checks that a `dma_bd`'s `static_sizes` dynamic marker count matches its `sizes` SSA operand count, via `mlir::verifyListOfOperandsOrIntegers`, but only when the op's parent is a `MemOp`/`MemTileDMAOp`/`ShimDMAOp`/`DMAOp` (see that function's own comment). A `dma_bd` nested inside a `DMAConfigureTaskOp` has none of those as parent, so it reaches `verifyTaskBDDimensions` (called from `DMAConfigureTaskOp::verify()` to enforce the per tile ND dimension cap) completely unchecked. That function calls `bd.getMixedSizes()` on the never validated op, which routes through `mlir::getMixedValues()`. On a mismatch, `getMixedValues()` indexes past the end of the dynamic operand range instead of diagnosing it: an assertion failure when the linked MLIR core is built with assertions, an out of bounds read otherwise. A verifier's job is turning invalid IR into a diagnostic, not crashing on it.

This state is not reachable through the pretty `sizes = [...]` syntax (its custom parser always keeps the two lists consistent), so it models malformed or fuzzed IR rather than anything a real producer emits. Still, a verifier that can be made to assert or read out of bounds on adversarial input is worth closing, and this was flagged as an unresolved Copilot finding on #3351.

### How

Add the same `verifyListOfOperandsOrIntegers` check `DMABDOp::verify()` already uses, ahead of the `getMixedSizes()` call inside `verifyTaskBDDimensions`'s per BD walk. On failure it marks the containing result as failure and returns before reaching `getMixedSizes()`, so the existing dimension cap check downstream never runs on unchecked data.

### Test

`bad-20.mlir` constructs the mismatched state directly (`static_sizes` declares one dynamic, `kDynamic` sentinel entry, but `operandSegmentSizes` gives the `sizes` operand group zero entries) via the generic op syntax, since the pretty parser cannot produce it. With the fix, `aie-opt --verify-diagnostics bad-20.mlir` exits 0 (the expected `expected 1 dynamic sizes values` diagnostic matches). Without the fix (only this file's guard reverted, tests unchanged), the same invocation aborts:

```
aie-opt: mlir/lib/Dialect/Utils/StaticValueUtils.cpp:214: llvm::SmallVector<mlir::OpFoldResult> mlir::getMixedValues(...): Assertion `dynamicValues.size() == ...' failed.
```

exit 134, SIGABRT, with a stack trace through `DMABDOp::getMixedSizes()`, the `verifyTaskBDDimensions` walk lambda, and `DMAConfigureTaskOp::verify()`, confirming the exact call path the commit message describes. I checked this both ways rather than trusting that a passing test with the fix applied was sufficient. I also ran the rest of the `dma-tasks-to-npu` directory (46 files total, each via its own `RUN:` line) both ways: 46/46 pass with the fix, and without it exactly one file fails, `bad-20.mlir` itself (the other 45 are unaffected, including `bad-2.mlir`, whose own check is downstream of this guard and unrelated to it).

### Refs

Follow up to #3351 (mine, merged), closing a Copilot review finding left unresolved there: that `verifyTaskBDDimensions` calls `getMixedSizes()` purely to count dimensions, and `getMixedValues()` can assert when the `static_sizes` and `sizes` lists disagree. This is that check.

### DIFF
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -19,6 +19,7 @@
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Interfaces/DataLayoutInterfaces.h"
 #include "mlir/Interfaces/FoldInterfaces.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
 #include "mlir/Transforms/InliningUtils.h"
 
 #include "llvm/ADT/TypeSwitch.h"
@@ -1125,6 +1126,20 @@ verifyTaskBDDimensions(const AIE::AIETargetModel &targetModel, int col, int row,
     ++maxNDims; // leading dim is hoisted into the iteration/repeat register
   LogicalResult result = success();
   body.walk([&](AIE::DMABDOp bd) {
+    // AIE::DMABDOp::verify() only runs this same check when its parent is a
+    // MemOp/MemTileDMAOp/ShimDMAOp/DMAOp (see its own comment); a BD nested in
+    // a DMA task op reaches here unchecked. Validate the dynamic-operand count
+    // against the kDynamic-sentinel count in static_sizes before calling
+    // getMixedSizes(): mismatched lists make it call mlir::getMixedValues(),
+    // which indexes past the end of the operand range on a mismatch instead
+    // of diagnosing it (asserts in a debug build, OOB read otherwise).
+    llvm::ArrayRef<int64_t> staticSizes =
+        bd.getStaticSizes().value_or(llvm::ArrayRef<int64_t>{});
+    if (failed(mlir::verifyListOfOperandsOrIntegers(
+            bd, "sizes", staticSizes.size(), staticSizes, bd.getSizes()))) {
+      result = failure();
+      return;
+    }
     size_t numDims = bd.getMixedSizes().size();
     if (numDims > maxNDims) {
       bd.emitOpError() << "Cannot give more than " << std::to_string(maxNDims)
--- a/lib/Dialect/AIEX/IR/AIEXDialect.cpp
+++ b/lib/Dialect/AIEX/IR/AIEXDialect.cpp
@@ -19,7 +19,6 @@
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Interfaces/DataLayoutInterfaces.h"
 #include "mlir/Interfaces/FoldInterfaces.h"
-#include "mlir/Interfaces/ViewLikeInterface.h"
 #include "mlir/Transforms/InliningUtils.h"
 
 #include "llvm/ADT/TypeSwitch.h"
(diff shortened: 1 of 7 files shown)


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -19,6 +19,7 @@
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Interfaces/DataLayoutInterfaces.h"
 #include "mlir/Interfaces/FoldInterfaces.h"
+#include "mlir/Interfaces/ViewLikeInterface.h"
 #include "mlir/Transforms/InliningUtils.h"
```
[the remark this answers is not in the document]
atassis: `mlir::verifyListOfOperandsOrIntegers` is declared in `mlir/Interfaces/ViewLikeInterface.h` (line 228, sole declaration in the tree), so that include is the declaring header and this diff adds it for exactly that call. Nothing transitive involved.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -1125,6 +1126,20 @@ verifyTaskBDDimensions(const AIE::AIETargetModel &targetModel, int col, int row,
...
+    // getMixedSizes(): mismatched lists make it call mlir::getMixedValues(),
+    // which indexes past the end of the operand range on a mismatch instead
+    // of diagnosing it (asserts in a debug build, OOB read otherwise).
+    llvm::ArrayRef<int64_t> staticSizes =
+        bd.getStaticSizes().value_or(llvm::ArrayRef<int64_t>{});
+    if (failed(mlir::verifyListOfOperandsOrIntegers(
+            bd, "sizes", staticSizes.size(), staticSizes, bd.getSizes()))) {
+      result = failure();
+      return;
+    }
```
andrej: For completeness, shouldn't we also be checking getStaticStrides() and make sure it is equal length as getStaticSizes()?

andrej (approved):
Thanks for adding the check!

lib/Dialect/AIEX/IR/AIEXDialect.cpp
atassis: Yes, and both turn out to be crashes rather than symmetry. A mismatched `strides` list asserts in `getMixedStrides()` from `rewriteSingleBD`, the same `getMixedValues` path as `sizes`. A sizes/strides rank mismatch overruns the 4-element pad in `rewriteSingleBDDynamic` (`strides4[4 - strides.size() + i]`, loop bounded by `sizes.size()`), which asserts on a runtime-bd_id BD; on the static path it degrades into "internal error folding BD dimensions".

Pulled the three checks `DMABDOp::verify()` already runs into `DMABDOp::verifyMixedSizesAndStrides()` and call that from `verifyTaskBDDimensions`, with `bad-21.mlir` (strides) and `bad-22.mlir` (rank). The shared helper also drops the `ViewLikeInterface.h` include from AIEXDialect.cpp.

atassis:
@andrej Can you check this again, pls?

andrej (approved):

PR #3467, opened 2026-07-30
aie2p: three new kernels and one layer_norm extension [MED]

atassis:
### Problem

`aie_kernels/aie2p/` has no standalone f32 -> bf16 narrowing cast, no depthwise
conv1d and no software `2^x`, and `layer_norm_f32` computes its affine with
`gamma = 1.0f, beta = 0.0f` pinned. One commit per piece, each independently
droppable.

### Fix

**`cast_f32_bf16`** (commit 1). Row-wise cast, 8 cores splitting rows, plus
`programming_examples/ml/cast_f32_bf16`. The narrow goes through an `accfloat`
accumulator under `conv_even`; the AIE default truncates toward zero, so
without it an on-chip cast and a host pack disagree by up to 1 ULP per element.

**`layer_norm`** (commit 2). The existing core becomes
`layer_norm_f32_impl<TIn, TOut, N, kAffine>`, and a new `layer_norm_affine_cast`
entry point takes per-column gamma/beta and narrows to bf16 in the same
dispatch. `layer_norm_f32` keeps its signature and its generated code. gamma and
beta arrive as one `[2 * embedding_dim]` buffer, loaded once per core, since two
DMA inputs is the compute-tile limit. `ml/norm` gains the op as a second design,
inheriting the 0.05 elementwise and 0.01 per-row rel-L2 gates from #3410.

**`dwconv1d`** (commit 3). 'same' padding, stride 1, bf16, 8 cores splitting
channels, on `aie::sliding_mul_ops`. The caller supplies the haloed row so the
kernel does no boundary handling. The 32-lane window is a `concat` of two
256-bit loads rather than one 512-bit load, since `in_pad` is only 256-bit
aligned and a 512-bit access on XDNA 2 requires 512-bit alignment. The
weights row is `kernel_size + 1` wide with the extra column zero-filled:
`kernel_size` is odd, so a bf16 row of exactly that width has a DMA transfer
length that is not a multiple of 4 bytes, which `aie.dma_bd` rejects.

**`exp2f_vec`** (commits 4-6: kernel, IRON factory, example). Degree-5 minimax
`2^x` over `aie::vector<float, 16>`. The hardware LUT behind `bf16_exp.cc` is
bf16-output; measured on aie2p against a float64 reference, same call and input
on both paths:

| domain | hw LUT | this poly |
|---|---|---|
| [-1, 0] | 6.1% | 8.49e-5 |
| [0, 10] | 6.1% | 8.49e-5 |
| [-10, 0] | 10.1% | 8.49e-5 |
| [-100, 0] | 49.1% | 8.51e-5 |

`[-100, 0]` is softmax's range, since scores are shifted by the row max before
exponentiating. The kernel clamps both ends. `2^k` is reconstructed by writing
`k + 127` into the exponent field, so past k = 127 it overflows into the sign
bit and returns finite wrong-signed values like `-0.0` and `-2.0`. Clamping at
128 and letting the hardware saturate does not work here, because `aie::mul` on
this part returns NaN rather than `+inf` on f32 overflow; instead the overflow
predicate is computed on the true input, the arithmetic clamps to 127.999, and
`+inf` is selected by lane copy.

### Test

Every kernel built through `aiecc` to an xclbin and run on NPU2 (Strix):

- `cast_f32_bf16`: bit-exact against a host `bfloat16_from_float`, 0 of 262144
  elements mismatched at 64x4096 and 0 of 4096 at 64x64.
- `layer_norm_affine_cast`: max abs 0.0078, rel-L2 4.0e-5 at 64x1024. The three
  pre-existing norm ops re-verified after the template change: `rms`
  0.0156/0.0042, `layer` 0.0234/0.0065, `layer_f32` 2.4e-5/1.0e-5.
- `dwconv1d`: bit-exact (max abs 0, rel-L2 0), with bias at K=1,9,17 and
  without at K=1,3,5,9,17.
- `exp2f_vec`: max rel 8.53e-5 over [-100, 0], rel-L2 3.00e-5, zero non-finite.
  Boundary tests assert on values, with sign-bit checks so `-0.0` cannot pass.

`test/python/test_kernels_specs.py` covers the `exp2f_vec` factory with no
hardware. `clang-format` 17.0.1 reports no changes.

### Known limitations

No npu1 path is exercised; these kernels are aie2p-only. The `dwconv1d` and
`cast_f32_bf16` device figures were measured through their own harnesses and
those lit RUN lines executed by hand, because the generated `lit.site.cfg.py` in
my toolchain instance has its Peano and XRT paths unset.

`exp2f_vec` carries `__attribute__((noinline))`. With it removed and inlining
forced into a register-pressure-heavy caller, that caller goes from rel-L2
5.37e-3 against a 0.08 gate to rel-L2 3.35 at correlation 0.224. I have not
reduced it to a standalone case, so that is the A/B behaviour of one caller, not
a compiler-defect claim.


test/python/test_kernels_specs.py
```diff
@@ -653,6 +676,34 @@ def _flat_ids(rows, label):
...
+    generic tests are parametrized per-spec, not per-fixture.
+    """
+    if spec.requires_npu2:
+        set_current_device(NPU2Col1())
+        try:
+            yield
+        finally:
+            set_current_device(None)
+    else:
+        yield
```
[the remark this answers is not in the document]
atassis: Fixed in 81171460c7a. `_device_for` now snapshots with `get_current_device(probe_runtime=False)`, which reads the explicit binding without initializing the default runtime, and restores it in the `finally`. Left conftest's `npu2_device` fixture clearing unconditionally, since pytest scopes a fixture's teardown to one test while this helper runs inline, per parametrized spec.

programming_examples/ml/dwconv1d/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. `dwconv1d.cc`: AIE2P kernel, pulled from [`aie_kernels/aie2p/`](../../../aie_kernels/aie2p/). Vectorized 16-outputs/iteration: two aligned loads build a 32-lane window, and `aie::shuffle_down_fill` slides a 16-wide tap window across it in-register (no unaligned loads); each tap's product is `aie::mac`'d into one `accfloat` accumulator. `aie::sliding_mul_ops` is the general, dedicated primitive for this shape; this kernel does not use it, for reasons the kernel's own header comment records in full (a standalone repro later found the primitive itself correct, so this is not a workaround for a library defect). No performance comparison between the two has been made.
+
+1. `test.cpp`: C++ testbench. Loads the compiled XCLBIN + `insts.bin` via `setup_and_run_aie`, recomputes the reference directly from the already-padded input buffer (the same convolution the kernel itself performs, so no separate host-side padding step), and checks the output against it with a tolerance (bf16 accumulation is not bit-exact against a double-precision host reference, though on this target it measures bit-exact in practice for both the `bias=True` and `bias=False` paths).
```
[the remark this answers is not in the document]
atassis: Right, the README was describing an earlier version. `test.cpp` bit-compares exactly and says so in its own comment. Corrected the README to match in 81171460c7a.

programming_examples/ml/cast_f32_bf16/test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    DATATYPE_OUT got = bufOut[i];
+    if (expected != got) {
+      std::cout << "Mismatch at index " << i << ": expected "
+                << test_utils::bfloat16_to_float(expected) << ", got "
+                << test_utils::bfloat16_to_float(got) << std::endl;
+      errors++;
+    } else {
+      pass++;
+    }
+  }
```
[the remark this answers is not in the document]
atassis: Confirmed: `setup_and_run_aie` passes `(in..., out, first_input_volume, verbosity)` (xrt_test_wrapper.h:324), so that parameter was receiving `verbosity` under a name that hid it. Renamed and used it to lift the mismatch cap in 81171460c7a. `dwconv1d`'s verifier in this PR already had it right; `ml/norm` and `ml/rope` carry the same misnaming but predate this branch.

aie_kernels/aie2p/cast_f32_bf16.cc
```diff
@@ -0,0 +1,51 @@
...
+#include <aie_api/aie.hpp>
+#include <stdint.h>
+
+// Element-wise f32 -> bf16 narrowing cast, round-to-nearest-even.
+//
+// A resident on-chip dataflow that hands an intermediate activation from an
+// f32-producing kernel (e.g. a numerically-stable norm or a K-split f32
+// accumulate) to a bf16-consuming one (e.g. a bf16 matmul) needs this cast to
+// stay on-chip; otherwise the f32 buffer has to round-trip through the host to
+// be requantized. `aie::vector<float, N>` has no `to_vector<bfloat16>`
```
andrej:
```suggestion
// `aie::vector<float, N>` has no `to_vector<bfloat16>`
```

(i.m.o. somewhat obvious)

aie_kernels/aie2p/cast_f32_bf16.cc
```diff
@@ -0,0 +1,51 @@
...
+// N (16). Rounding is round-to-nearest-even (`conv_even`) rather than the AIE
+// default (truncation toward zero), matching a host f32->bf16 pack that also
+// rounds to nearest even (e.g. `_mm512_cvtne2ps_pbh` on AVX512-BF16) so an
+// on-chip cast and its host equivalent agree bit-for-bit.
+//
+// The rounding mode is a single sticky register shared by every kernel that
+// runs on this core, so this kernel saves the caller's mode and restores it
+// before returning rather than leaving conv_even set for whatever runs next,
+// the same save/restore mm.cc's AIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16
+// block uses.
```
andrej: Good for now -- opened issue #3481.

aie_kernels/aie2p/cast_f32_bf16.cc
```diff
@@ -0,0 +1,51 @@
...
+                       int32_t cols) {
+  event0();
+  ::aie::rounding_mode saved_rounding =
+      ::aie::swap_rounding(::aie::rounding_mode::conv_even);
+  for (int i = 0; i < cols; i += N) {
+    ::aie::vector<float, N> v = ::aie::load_v<N>(input + i);
+    ::aie::accum<accfloat, N> a;
+    a.from_vector(v);
+    ::aie::store_v(output + i, a.template to_vector<bfloat16>());
+  }
```
andrej, line 34: Can we have either
1. an assert that fails if this is called with `cols` not divisible by `N`.
2. a scalar "remainder" conversion section?

aie_kernels/aie2p/dwconv1d.cc
```diff
@@ -0,0 +1,136 @@
...
+// the caller supplies, this kernel does no boundary handling and needs no
+// local scratch buffer:
+//
+//   in_pad[0 .. T+15]: [P zeros | T real samples | P zeros | 16-(K-1) junk]
+//                       \_______________ T+K-1 valid, halo-padded ________/
+//   P = (K - 1) / 2
+//
+// The trailing "don't-care" slack past the halo (up to a fixed 16 elements,
+// independent of K) exists only so the vectorized kernel's aligned 16-wide
+// loads never read past the end of the buffer; its VALUES never reach the
```
andrej:
```suggestion
// loads never read past the end of the buffer; its values never reach the
```

aie_kernels/aie2p/dwconv1d.cc
```diff
@@ -0,0 +1,136 @@
...
+//
+//===----------------------------------------------------------------------===//
+//
+// Depthwise conv1d, 'same' padding, stride 1, bf16, one channel (row) per
+// call:
+//
+//   out[t] = bias + sum_{p=0..K-1} w[p] * in_pad[t + p],   t = 0 .. T-1
+//
+// This is a cross-correlation (no kernel flip), matching torch.nn.Conv1d and
+// most framework "same" depthwise convs. `in_pad` is the ALREADY-PADDED row
```
andrej:
```suggestion
// most framework "same" depthwise convs. `in_pad` is the already-padded row
```

aie_kernels/aie2p/dwconv1d.cc
```diff
@@ -0,0 +1,136 @@
...
+// An earlier version of this kernel used
+// aie::sliding_mul_ops<L, K, 1, 1, 1, bfloat16, bfloat16>::mul() and, in a
+// full multi-channel pipeline (not isolated), observed wrong bf16 output.
+// A later, standalone single-core repro of that same call, independent of
+// this kernel's dataflow, found the primitive itself correct at K in
+// {1, 4, 9, 16} (rel-L2 ~0.003, the known bf16 noise floor for this class of
+// kernel, zero non-finite outputs), against both the pinned toolchain and
+// the mlir_aie wheel's aie_api. The original wrong output is attributed to a
+// dataflow-wiring bug in the surrounding multi-channel pipeline, not to
+// sliding_mul_ops. No head-to-head performance comparison between the two
```
andrej: Am I reading this correct as follows:

"Kernel using `sliding_mul_ops` appeared broken, so we re-implemented it. But, the error turned out to be outside of the kernel, in the data flow instead. But we're still keeping our custom impl rather than `sliding_mul_ops`?"

If so, can we revert to using `sliding_mul_ops`?

Also, anti-pattern of documenting code history in comments. There's commits/git for that.

aie_kernels/aie2p/exp2f_vec.cc
```diff
@@ -0,0 +1,235 @@
...
+// Cost of the fix: the op count for the whole kernel body goes from 31 to
+// 39 aie:: calls (device-verified against the actual compiled kernel, not
+// estimated), broken down as: +1 broadcast/+1 aie::ge (the overflow test), +1
+// broadcast/+1 aie::min (the safety clamp, replacing the old single
+// lower-only clamp), +1 broadcast/+1 aie::select (the +inf override). All
+// of it is cheap scalar-broadcast-and-compare work; none of it touches the
+// 5-step Horner poly, so it costs nothing in the accuracy-critical path.
+// Device-verified bit-identical to the pre-fix kernel everywhere below
+// x = 127.999 (the [-100, 0] and [0, 127] domains both reproduce the exact
+// same max-relative-error figures as before this commit).
```
andrej: Can we tighten these extremely verbose comments and remove references to "this commit"? This will be a comment in the code, so it will be unclear what "this commit" is supposed to refer to.

aie_kernels/aie2p/layer_norm.cc
```diff
@@ -55,53 +56,114 @@ void layer_norm(const T *restrict input, T *restrict output, int32_t cols) {
...
+      ::aie::vector<TIn, N> scaled_v = ::aie::mul(norm_v, gamma_v);
+      ::aie::vector<TIn, N> out_v = ::aie::add(scaled_v, beta_v);
+      ::aie::accum<accfloat, N> a;
+      a.from_vector(out_v);
+      ::aie::store_v(output + i * N, a.template to_vector<TOut>());
+    }
+    ::aie::set_rounding(saved_rounding);
+  } else {
+    // gamma = 1, beta = 0, TOut == TIn: `layer_norm_f32`'s original body,
+    // unchanged, no gamma/beta loads, no rounding-mode calls.
```
andrej:
```suggestion
    // gamma = 1, beta = 0, TOut == TIn
```

aie_kernels/aie2p/layer_norm.cc
andrej, line 1: other reviewers: this file probably needs most scrutiny since it touches an existing kernel

aie_kernels/aie2p/layer_norm.cc
```diff
@@ -55,53 +56,114 @@ void layer_norm(const T *restrict input, T *restrict output, int32_t cols) {
...
-    ::aie::vector<T, N> out_v = ::aie::add(scaled_v, beta_v);
-    ::aie::store_v(output + i * N, out_v);
+  ::aie::vector<TIn, N> inv_std_v = ::aie::broadcast<TIn, N>((TIn)inv_std);
+
+  // The two instantiations diverge only in where gamma/beta come from and
+  // whether the write narrows, kept as two separate loop bodies (selected
+  // at compile time, no runtime branch in either) rather than one loop with
+  // per-iteration `if constexpr` dispatch, so each instantiation's codegen is
+  // exactly the sequence of ops it needs, nothing left over from the other.
+  if constexpr (kAffine) {
```
andrej, line 100:
```suggestion
  // The two instantiations diverge only in where gamma/beta come from and
  // whether the write narrows.
  if constexpr (kAffine) {
```

No need to explain what `constexpr` is.

andrej (changes requested):
Thanks for the contribution. 

I reviewed the first ~ 1/4 of this. I did not check the algorithms themselves and will trust that tests/programming examples will cover their correctness. Seems solid and not a real risk to merge overall since it only adds kernels (layer_norm changes seem minimal).

Requested changes may appear nitpicky but I'd like to continue avoiding AI slop comments as they will eventually blow up our codebase and make the actual code hard to parse. 

Overall seems like good work! :) 

Agent skill that may be useful (I welcome contributions/improvements to this skill as I think in its current form it still isn't aggressive enough in removing and tightening comments):

```
---
name: code-commenting
description: Instructions for writing good code comments. Read this every time you write or review code.
license: Apache-2.0 WITH LLVM-exception
---

# Code Comments

Comments are expensive: they draw the reader's limited attention. Spend that
attention only on what the code cannot say for itself. Default to FEWER comments.

## A comment must earn its place
Write one ONLY to capture something not evident from the code itself:
- a *why* / rationale, a constraint, an invariant, a gotcha, a non-obvious
  consequence, or genuinely complex/surprising behavior.
Do NOT write a comment that restates what the adjacent code plainly does.

## Never narrate the change or the request
Code describes what IS, never its own history. The reader has no idea (and does
not care) what the code used to be, or that an edit happened. Do NOT:
- Turn my prompt / the change request into a comment.
- Document that code was previously wrong, broken, non-idiomatic, or slow.
- Explain that something "was changed", "now does X instead of Y", or "was
  refactored to follow <abstraction>". No changelog/diff narration in code —
  that belongs in the commit message, not the source.
- Reference a "baseline", "legacy", "old", "original", "previous", or "existing"
  path/behavior — or say a change "preserves"/"keeps"/"still" does something.
  These all frame the code against a prior version the reader can't see. Describe
  the code as it stands, with no before/after framing. (If two code paths
  genuinely coexist, name them for what they ARE, not which came first.)
- Justify following the codebase's normal conventions (that's just the default).

## Specific anti-patterns to avoid
- Paraphrasing the next line(s) of code.
- Explaining what code no longer does after fixing an issue (e.g., "this 
  replaces the legacy O(n^2) lookup", or "uses a hash table rather than a
  O(n^2) lookup").
- Enumerating where code is NOT (e.g. "parsing lives here, not in the driver").
- Restating a well-named function/variable's obvious purpose.

## Default
When in doubt, delete the comment and let the code speak. 
Keep only high-signal comments.
```

atassis:
All four requested changes are in:

- `cast_f32_bf16.cc` `cols % N`: took option 1, the assert (`cast_f32_bf16.cc:25`). Callers here pass tile widths that are multiples of the vector width, so a scalar remainder arm would be untested dead code. If you'd rather the kernel be safe for arbitrary `cols`, I'll add it.
- `dwconv1d.cc`: reverted to `aie::sliding_mul_ops` in 03c035b76.
- `exp2f_vec.cc` and `dwconv1d.cc` comments: cut to the directory norm in 5c367ca30 and 772b723e4, change-history narration removed.
- python: ruff and black in 81e53836b.

On your reading of the dwconv1d history: close, but the correction is bigger than that comment made it sound. I re-tested `sliding_mul_ops` in isolation on a single core and could not reproduce a defect at all -- every configuration I had blamed came out correct at the bf16 noise floor. The original numbers came from a full pipeline rather than an isolated kernel test, so the kernel was never shown to be broken in the first place. That is why reverting was right, and why the comment claiming otherwise should not have been there.

One thing I have not measured: whether `sliding_mul_ops` or the in-register window is faster here. I only established that they agree numerically. Happy to benchmark it if that matters for merging.

`layer_norm.cc` still needs the second pair of eyes you asked for. Merging main now, it is behind.

programming_examples/basic/vector_exp2f/README.md
```diff
@@ -0,0 +1,71 @@
...
+right at the upper clamp). See
+[`aie_kernels/aie2p/exp2f_vec.cc`](../../../aie_kernels/aie2p/exp2f_vec.cc)
+for the full domain contract, the accuracy table, and the derivation
+(`x = k + f`, `2^k` built directly in the float32 exponent field, `2^f`
+from the poly). That file's header also documents two device-only bugs
+found and fixed in this kernel's history: an unclamped upper bound that
+wrapped `2^x` into a wrong-signed finite value for `x >= 129`, and a
+device-measured hardware quirk where `aie::mul` returns NaN, not +inf, on
+f32 overflow, which is why the fix is an explicit compare against the
+true `x` plus a select rather than a wider clamp.
```
andrej:
```suggestion
from the poly).
```

programming_examples/basic/vector_exp2f/vector_exp2f.py
```diff
@@ -0,0 +1,332 @@
...
+    # Upper-clamp boundary block: THE regression test for the sign-wrap
+    # defect. k = 127 must still be an accurate finite value (rel-err gate,
+    # same as above); every k >= 128 has a true 2**x that already exceeds
+    # FLT_MAX (2^128 = 3.402823669e38 > FLT_MAX = 3.402823466e38), so the
+    # kernel's contract is EXACT +inf there, not merely "close to", so assert
+    # bit-for-bit equality to +inf, not a tolerance. A pre-fix kernel (lower
+    # clamp only) fails this block outright: k=129 reads back as -0.0,
+    # k=150 as -1.23e-32, k=257 as -2.0, all finite, all wrong-signed, all
+    # invisible to an isfinite() check. np.signbit() catches -0.0 too, which
+    # a plain `< 0` comparison would not (IEEE -0.0 compares equal to 0.0).
```
andrej:
```suggestion
    # bit-for-bit equality to +inf, not a tolerance.
```

Can you please do another pass cleaning up comments? "pre-fix kernel" documents code history, is confusing.

programming_examples/basic/vector_exp2f/vector_exp2f.py
andrej, line 1: I think I understand the k < 128 upper limit -- anything larger would overflow a float 32. It would be good to document that reasoning here as well.

The lower bound -100 seems arbitrarily chosen to me. Please add at a minimum a rationale as to why we choose that number/what it means, and ideally make it parametrizable so users can generate designs with lower values if they need accuracy for these tiny inputs.

programming_examples/ml/cast_f32_bf16/cast_f32_bf16.py
```diff
@@ -0,0 +1,158 @@
...
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+"""Element-wise f32 -> bf16 narrowing cast, IRON API + ``@iron.jit``.
+
+NPU2-only: the underlying ``cast_f32_bf16.cc`` kernel lives under
+``aie_kernels/aie2p/`` and has no aie2 counterpart.
+
+Eight cores each cast ``sequence_length // 8`` rows; one row is
+``embedding_dim`` f32 values in, ``embedding_dim`` bf16 values out.
```
andrej: If this is supposed to be a generic f32->bf16 cast, `sequence_length` and `embedding_dim` are poorly chosen names for these dimensions. Other workloads might not have these. Can we rename? Maybe `vector_size` and `n_vectors`?

programming_examples/ml/cast_f32_bf16/cast_f32_bf16.py
```diff
@@ -0,0 +1,158 @@
...
+Rounding is round-to-nearest-even, matching a host f32->bf16 pack that also
+rounds to nearest even, so an on-chip and a host cast of the same input agree
+bit-for-bit. Structurally this mirrors ``ml/norm``'s row-split ``@iron.jit``
```
andrej:
```suggestion
Rounding is round-to-nearest-even.
```

That's crazy many words to say "rounding is to nearest even." The rest of that sentence just says if the host does the same as the core, the results will match, which is obvious.

aie_kernels/aie2p/exp2f_vec.cc
```diff
@@ -0,0 +1,92 @@
...
+#include <stdint.h>
+
+using namespace aie;
+
+// 512-bit vector register / 32-bit lanes.
+static constexpr int EXP2F_VEC_LEN = 16;
+
+static __attribute__((noinline)) aie::vector<float, EXP2F_VEC_LEN>
+exp2f_vec(aie::vector<float, EXP2F_VEC_LEN> x) {
+  x = aie::max(x, aie::broadcast<float, EXP2F_VEC_LEN>(-100.0f));
```
andrej: -100 rationale / parametrization

programming_examples/ml/cast_f32_bf16/Makefile
andrej, line 1: My understanding is that the JIT-compiled Python files also execute and test the code (run_strix.lit). Is this correct?

If so, unless there is strong need/motivation, can we remove the test.cpp path (including removing CMakeLists.txt, run_makefile.lit and Makefile) from each example, please?

andrej (changes requested):
Thanks! After reading through some more of this, I'd really like to see two more things addressed:

- Are the test.cpps really needed?
- I didn't see (or understand) the rationale for the -100 lower bound in the exp2f_vec kernel.

andrej:
And just to draw your attention to it if you haven't seen it already, the ml/norm programming example does currently time out on Strix in CI, so that'll need investigation as well before merging.

atassis:
All addressed.

**-100.** It was arbitrary. Two limits, not one. Hard floor -126: `2^k` is written into the f32
exponent field as `(k + 127) << 23`, so below it the biased exponent carries into the sign bit
(k = -128 gives -inf, k = -129 gives -1.7e38), the mirror of the k = 129 wrap at the top. What binds
first is accuracy, at -111: `aie::mul(p, 2^k)` loses low order bits near the bottom of the normal
range. Measured on aie2p, f in [0, 1):

| k | >= -111 | -114 | -116 | -117 | -118 | <= -119 |
| --- | --- | --- | --- | --- | --- | --- |
| max rel err | 8.9e-5 | 1.1e-4 | 4.6e-4 | 1.7e-3 | 3.2e-3 | 6.5e-3 |

Default clamp is -111, moves with `kernels.exp2f_vec(min_x=...)` or `-DEXP2F_VEC_MIN_X`, floor
checked at -126. Derivation and the k < 128 reasoning are in the header. That does push
`exp2f_vec.cc` to 34% comment lines against a ~17% directory median; happy to move the table to the
example README.

**test.cpp.** Gone from both examples, along with `CMakeLists.txt`, `Makefile` and
`run_makefile.lit`. `run_strix.lit` already builds and asserts.

**ml/norm.** I do not think it is this PR. `-o layer_f32 -e 2048` predates the branch and is green on
main's nightlies, and the compiled `layer_norm_f32` is byte identical to main's, 1584 bytes and 320
bytes of stack either way. It passes 20/20 locally. The failing run waited 125s on the device lock
before `ERT_CMD_STATE_TIMEOUT`, and the branch was 21 commits behind, so it predated #3532. Merged
main; if it reproduces I will dig further.

The rest of the comments are in. All 12 lit invocations this PR touches pass on device.

andrej (approved):
Looking good now I think!

atassis:
Pushed b2731d7 for the two device-leg failures.

`layer_norm_f32`'s frame is 1024 bytes on its own (`llvm-readelf --stack-sizes` on the built kernel object), which is exactly `getDefaultCoreStackSize()`, and `norm.py` never set `stack_size`. The linker places the stack directly below the core's objectFIFO buffers, so the kernel overran into them and the core hung: `run_strix.lit`'s `layer_f32` line timed out with `ERT_CMD_STATE_TIMEOUT` on both legs, while `rms` (0 B) and `layer` (448 B) passed. On main that f32 work sat in a called template at 832 B and fit; my rewrite moved it into the entry point, and it no longer does. Fixed by reserving 2048 on that Worker.

Verified on an npu2 box: reproduced the timeout without the reservation, and all four `run_strix.lit` lines pass with it.

The Lint failure is unrelated to this branch. clang-tidy 18, which came with the 24.04 runner image in #3534, exits non-zero on the `'aie_api/aie.hpp' file not found` error that clang-tidy 14 printed but tolerated, so any PR touching `aie_kernels/**.cc` went red in that window (#3462 has the same error in its log and passed on 22.04). #3518's explicit file list drops `aie_kernels`, so a fresh run should be clean.

The runs are sitting at `action_required` - could someone approve them?

atassis:
@andrej Hi! Can we run CI here?

atassis:
@hunhoffe can you look at this?

andrej:
@atassis Sorry for missing this, restarted the workflows.
@hunhoffe If it passes I think this can be safely merged.

andrej (approved):

atassis:
@andrej fixed linter

PR #3471, opened 2026-07-30
Allow limiting build parallelism in `build-mlir-aie-from-wheels.sh`

jskimko:
## Description

On machines with limited RAM, the build can OOM with high job parallelism. This PR adds two env vars, COMPILE_JOBS and LINK_JOBS, that limits parallelism using job pools, which are enforced by the generated build files.

It also swaps the raw `ninja` calls for `cmake --build`, which respects the standard `CMAKE_BUILD_PARALLEL_LEVEL` for portable `-j` behavior.

Minor: fixed error handling w.r.t. `pipefail`.

```
# Default: no caps — Ninja uses all cores for everything (unchanged behavior).
./utils/build-mlir-aie-from-wheels.sh

# Limit linking to 2 at a time, compiles unbounded.
LINK_JOBS=2 ./utils/build-mlir-aie-from-wheels.sh

# Cap both: at most 8 concurrent compiles, 2 concurrent links.
COMPILE_JOBS=8 LINK_JOBS=2 ./utils/build-mlir-aie-from-wheels.sh

# Set the overall -j (no per-phase pools), via cmake --build.
CMAKE_BUILD_PARALLEL_LEVEL=4 ./utils/build-mlir-aie-from-wheels.sh
```

### DIFF
--- a/utils/build-mlir-aie-from-wheels.sh
+++ b/utils/build-mlir-aie-from-wheels.sh
@@ -17,6 +17,20 @@
 # <build dir>    - optional, mlir-aie/build dir name, default is 'build'
 # <install dir>  - optional, mlir-aie/install dir name, default is 'install'
 #
+# Optional 4th positional arg / environment:
+#
+# <peano dir>              - optional 4th arg, Peano (llvm-aie) install dir;
+#                           default: located via 'pip show llvm-aie'
+#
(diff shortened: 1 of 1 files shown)


andrej (approved):
Thanks!!

PR #3472, opened 2026-07-30
[router] Fix packet flow channel sharing

andrej:
One of the FastFlowLM designs uncovered a fundamental routing bug for packet flows.

We want to enable channel sharing -- being able to multiplex a single channel is kind of the whole point of packet flows. But, when sharing channels, we need to be careful about packet IDs, so we can distinguish packets to different destinations. We don't need globally-unique packet IDs. But, as soon as two flows traverse the same channel _and_ need to later be split up again to different destinations, those two flows can't have the same ID.

The previous router had two bugs/shortcomings:
1. Routing was not packet ID-aware. 
2. The router only enabled sharing channels for flows that originated from the same source port or that ended at the same destination port. This is somewhat arbitrary (though it might make sense if you look at it from a bandwidth-bottleneck perspective).

**(Edit)** A previous iteration of this PR attempted to address both of those points, but that turned out to be a bigger challenge than anticipated. So, for now, I've decided to only fix (1) here; After this PR, the router never allows sharing a channel between two packet flows if they have the same ID.

The following is the concrete bug that the FastFlowLM design uncovered:
image

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
@@ -15,6 +15,7 @@
 #include <algorithm>
 #include <iostream>
 #include <list>
+#include <optional>
 #include <set>
 
 namespace xilinx::AIE {
@@ -49,6 +50,9 @@ using SwitchboxConnect = struct SwitchboxConnect {
   std::vector<std::vector<int>> packetFlowCount;
   // only sharing the channel with the same packet group id
   std::vector<std::vector<int>> packetGroupId;
+  // packet ids currently routed through each channel (and its crossbar
+  // row/column); a channel may be shared only among distinct ids.
+  std::vector<std::vector<std::set<int>>> packetIds;
   // flags indicating priority routings
   std::vector<std::vector<bool>> isPriority;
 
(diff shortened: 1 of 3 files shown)


andrej:
Ready for review

hunhoffe (approved):
Did you get a chance to do any perf testing with any complicated designs? Ok if no, mostly just curious.

andrej:
> Did you get a chance to do any perf testing with any complicated designs? Ok if no, mostly just curious.

Very valid question, and no I haven't. Do any come to mind for you that would use packet flows? Circuit routing should be unaffected (unless it is present in conjunction with packet flows).

Edit: Looks like it is fully failing to route the vector_reduce_max design, so I'll have to investigate that too.

andrej:
@hunhoffe I reduced the scope of this PR since the original change didn't work as generally as I would have liked it to, and attempted to solve additional problems that we really hadn't run into yet (sharing additional channels under extreme congestion).

For now, this is now just the minimal fix needed to get the FastFlowLM design working. As such, it also shouldn't affect the routing performance of other designs.

fifield:
Does this remove support for Option 2, which is a valid hardware feature?

jskimko:
@fifield No, if I understand it correctly, Option 2 (downstream merge) is still available; it just moves same-id packets onto separate ports (higher resource cost).

Update: if the diagram is referring to same-port, same-id merge, then yes, I think it's not possible anymore (even with same destination flows).

PR #3476, opened 2026-07-31
[python] JIT cache: key design inputs on content, not mtime [MED]

atassis:
## Problem

The cache key identifies its file inputs by `st_mtime`: source and object files in
`_compute_artifact_hash`, and a `Path` generator in `_compute_recipe_hash`. mtime records how
a file arrived, not what is in it, so the key breaks in both directions.

*Stale hits.* Restore an mtime and the key does not move even though the content did.
`git checkout` of another revision, `cp -p`, and archives that carry timestamps all do
this, so the second build is served the first build's artifact and a changed kernel runs
as the old binary.

*Spurious misses.* A fresh clone, a `pip install`, a `cp` or a `touch` restamps every file
without changing a byte, and invalidates every entry. Two checkouts on one machine cannot
share one.

## Fix

The same inputs, identified by content instead. Digests are streamed in 1 MiB chunks so a
large input is never materialised. An input that cannot be read contributes a
marker naming the error class rather than being skipped, so an unreadable input does not
collapse onto the same key as a readable one.

**Tool identity is deliberately left alone.** `peano_cxx_path()` and `aiecc_path()` keep
their mtime keys and that block is byte-identical to main. Content-hashing the toolchain
costs 158 ms here, re-paid per process, which is not worth putting on a lookup path, and it
would not fix the failure this PR is about: what `git checkout` restamps is sources, and a
locally rebuilt compiler moves its own mtime anyway.

Four docstrings in `compilabledesign.py` and `compileconfig.py` described the inputs as
mtime-keyed and are corrected to match.

## Evidence

Behaviour differential on `_hash.py`, main against this branch, same checks both sides:

| check | main | branch |
|---|---|---|
| touch of an unchanged source | moves the key | stable |
| touch of an unchanged object | moves the key | stable |
| touch of a `Path` generator | moves the key | stable |
| content changed under a preserved mtime | **key unchanged** | moves |
| end-to-end key, preserved mtime | **key unchanged** | moves |
| content edit moves the key | moves | moves |

The bolded rows are the stale hit. Six new tests, four of which assert behaviour main does
not have; the rest cover the chunked digest and the unreadable-input marker. One existing
test is renamed and tightened, since it no longer needs a `time.sleep` to outwait mtime
granularity.

Cost, on a 36 KB kernel plus a 64 KB object over 200 reps: 85.6 us median per key on main,
127.1 us on this branch, so about 42 us more against a compile measured in seconds. No
input is read that the key did not already `stat`.

## Scope

The key still includes each input's path string, so this enables reuse across runs on one
machine, not across machines.

Changing the key orphans entries written by an earlier version, as #3428 and #3432 also did.
There is no eviction in tree, so a store only accumulates: mine is 624 entries and 262 MB.
Not addressed here.

This changes the key only. `--xclbin-path` still bypasses the cache entirely, which is most
of the example suite, so the practical hit rate does not move much until that is addressed
separately. No workflow file is touched and no CI behaviour changes.


atassis:
I can't reach my pc, but this PR might not land now,  I suppose. It can break windows workflow in current state. I'll push the fix when I come back.

hunhoffe:
No worries! I'm taking some time to digest this. The JIT cache and management of that cache in CI has a rich (and unfortunate) history of breaking things (mostly me, but some other folks along the way have felt the pain too). So I wanted to think really carefully about this one.

Thoughts so far:
1. As noted, I'll make sure https://github.com/Xilinx/mlir-aie/pull/3457 lands first.

2. Cache key doesn't fingerprint the compiler or Python infra — and hashing the tool binaries costs more than it looks. aie-opt/aie-translate (the actual pass logic) and aie.iron.*/aie.utils.compile.* aren't hashed at all — only aiecc's own binary + the design script's bytecode. A compiler-pass fix with no kernel changes → cache hit → CI never runs the new compiler. Separately, I measured _tool_identity(aiecc) at 153 ms (219 MB, SHA-256) — and since _tool_digest_cache only memoizes within one process, every separate CI test process (lit test, pytest file) re-pays that cost from scratch.

While in general, I do think content hashing is a more correct solution, I think 1) it's too expensive for general use, and 2) it doesn't ensure test coverage of infrastructure components the CI is also designed to exercise. I'm open to more discussion in this area.

3. Eviction isn't lock-aware. rm -rf on aged entries ignores the .lock file file_lock() uses — can delete an entry a concurrent process (e.g. an orphan from cancel-in-progress) is actively writing.

4. Hard-linked publish is a silent-corruption risk. Published paths share an inode with the cache entry; nothing stops a downstream in-place write from corrupting it for every future consumer. chmod cache artifacts read-only after write.

5. No tests for the explicit-path cache-hit/miss/publish logic — only a manual on-hardware run. Worth unit tests before merge.

6. Minor: the "cache bypassed" debug log still fires when the cache is in fact being used (explicit-path + use_cache=True miss).

andrej:
On the caching discussion, I agree that in principle content-matching would be best, but for CI I think I'd never want to use cached artifacts. I think I wouldn't trust it. We've had it in the past that tests continued passing after some PR broke them because they were not recompiled. We already delete the caches between each run anyways at the start of the runs, and I think we should keep doing that.

atassis:
I understand that concern, that's why I am reproducing as much as I can locally. I understand you points, I'll try to make to them:)
My point is that making caches work properly is a requirement not only because of CI, but also for local development. For example, I also build Peano locally, that's why I need all incremental changes to be applied as fast as it can be

hunhoffe:
@atassis That's a good point re: local development. Do you think there's a better (but still cheap) way to tag binaries/versions that's more expressive/forgiving than relying on last git commit id?

atassis:
I have thought about your feedback.
Right now I still think, that caching is not smth not to do, and the code should land in one form or another
But I agree that it needs to be more thoroughly reasoned about, tested and else. 
Since it's friday- it is not a good idea to land this now, anyway, for all the sounded reasons.
Let's keep this PR as draft for now, then. I'll bring later a more thorough analysis.

atassis:
What do you think if I make a separate PR with 4th commit, part 2? And strip it out of this PR

andrej:
> What do you think if I make a separate PR with 4th commit, part 2? And strip it out of this PR

I think it's worth a try. I'd be more inclined to merge only that change while we consider caching. However, there are some concerns with that as well: Our CI runners have limited memory, and we've run into OOM kills frequently before. If that pops up again, we'd have to stick to a serialized schedule (having jobs randomly fail due to OOM and needing to restart them is a bigger time and effort cost than having every job be a bit slower). But agreed that it is worth a try.

atassis:
Rescoped down to the key itself, in response to both of you, and rebased onto current main.

Dropped: the CI commit, so no workflow file is touched at all; the tool content-hashing,
because @hunhoffe's 153 ms measurement is right and hashing the driver would not have fixed
this bug anyway; and the explicit-path commit, which needs the unit tests point 5 asked for
and belongs on its own. Points 3, 4, 5 and 6 went with the commits that carried them, and
#3457 landed, so point 1 is done.

What is left is the input key alone. `git checkout` of another revision restores mtimes, the
key does not move, and a changed kernel is served the previous revision's artifact. The tool
identity block is byte-identical to main.

@hunhoffe, on your question about a cheaper and more forgiving tag than the last git commit
id: the ELF build-id. Measured on the current Peano dist, `clang++` is 224 KB but `ldd` shows
it loading `libLLVM.so` and `libclang-cpp.so`, 217 MiB between them, and all three carry
`.note.gnu.build-id`. Reading all three notes takes 29 us; SHA-256 over the same three files
takes 158 ms. So it is about 5000x cheaper, and it covers the two libraries that hashing
`clang++` alone does not reach at all.

The tradeoff is worth stating: a build-id is an identity the linker stamped at link time, not
a check of the bytes on disk now, so it will not catch a truncated or corrupted install the
way a content hash would. For "did the toolchain change", which is what the key is asking, I
think that is the right trade. These are 20-byte sha1-style notes; a toolchain linked with
`--build-id=none` would need a fallback.

That says nothing about the Python infra in your point 2, which is a real gap and a separate
question. Not proposing either one here.

hunhoffe:
Thank you for sharing your measurements and for making the amendment!

hunhoffe (approved):
If profiling ever shows this hashing cost mattering at scale (many source/object files, hot repeated-build loops), ccache's "inode cache" pattern — memoize (inode, size, mtime) → hash so repeat lookups on unchanged files skip the read entirely — is a proven way to get there without reintroducing the mtime-only correctness gap. Not needed now given the measured cost, just noting it as a lever if it ever becomes one.

Verdict: LGTM!

PR #3478, opened 2026-07-31
update_peano_version: require wheels for every CI platform

jgmelber:
## Problem

`update_peano_version.py` picks the newest version string on the nightly index without checking which wheels that version actually shipped:

```python
WHEEL_VERSION_RE = re.compile(r"llvm_aie-([0-9][0-9.]*\+[0-9a-f]+)-")
...
return max(versions, key=Version)
```

Publishing is not atomic across platforms. A nightly can ship its Linux wheel while the Windows one is delayed — or never arrives. **`2026072101` through `2026073001` published no `win_amd64` wheel at all.**

That is how `2026072701` reached `main` and broke every Windows job at `Setup IRON environment` (fixed by #3477). The failure is easy to misdiagnose, because pip filters its `from versions:` list to the current platform, so the pinned version looks *withdrawn* rather than platform-incomplete.

This defeats the guarantee in `peano-requirements.txt`'s own header — that the workflow exists "so a bad nightly is caught in a PR instead of breaking main."

## Fix

Parse each wheel's platform tag alongside its version, and only treat a nightly as a bump candidate once it has wheels for **both** Linux and Windows. `REQUIRED_PLATFORMS` is the single place to extend if CI grows another platform.

When newer-but-incomplete nightlies exist, the script reports what it skipped instead of silently pinning something older:

```
note: skipping 21.0.0.2026073001+2d0a7d4e and newer; no wheel for windows.
      Falling back to 21.0.0.2026072001+ce8c0f8f.
```

An explicit `--peano-version` is validated the same way — hand-picking an incomplete nightly is the same failure mode:

```
error: llvm-aie 21.0.0.2026072701+bc83427c has no wheel for windows at
       https://github.com/Xilinx/llvm-aie/releases/expanded_assets/nightly.
```

## Verification

Against the live index, platform detection is correct across the broken window:

| Nightly | Detected |
|---|---|
| `2026072001` | linux, windows |
| `2026072601` … `2026072901` | linux |
| `2026073101` | linux, windows |

Replaying the index as it stood on 30 Jul (with `2026073101` withheld), the script selects **`2026072001`** — the last complete nightly — instead of the `2026072701` that broke CI. A bad explicit pin exits 1; the default run resolves to `2026073101` and leaves the file untouched under `--identify-only`.

Depends on nothing; independent of #3477, which fixes the currently-pinned version.

### DIFF
--- a/utils/update_peano_version.py
+++ b/utils/update_peano_version.py
@@ -27,8 +27,22 @@ NIGHTLY_INDEX_URL = (
 REPO_ROOT = Path(__file__).resolve().parent.parent
 REQUIREMENTS_FILE = REPO_ROOT / "utils" / "peano-requirements.txt"
 
-# Wheel filenames look like: llvm_aie-21.0.0.2026062501+c83e305a-py3-none-...whl
-WHEEL_VERSION_RE = re.compile(r"llvm_aie-([0-9][0-9.]*\+[0-9a-f]+)-")
+# Wheel filenames look like:
+#   llvm_aie-21.0.0.2026062501+c83e305a-py3-none-manylinux_2_27_x86_64...whl
+#   llvm_aie-21.0.0.2026062501+c83e305a-py3-none-win_amd64.whl
+# Capture the version and the platform tag separately: a nightly is only usable
+# once it has published a wheel for every platform CI builds on.
+WHEEL_RE = re.compile(r"llvm_aie-([0-9][0-9.]*\+[0-9a-f]+)-[^\"]*?-([^\"-]+)\.whl")
+
+# Platform tags CI needs. The Linux runners install the manylinux wheel and the
+# Windows runners the win_amd64 one, so a nightly missing either breaks half of
+# CI at IRON setup -- with a confusing error, because pip's "from versions:"
+# list is filtered to the current platform and so appears to omit the pin
+# entirely rather than reporting a platform mismatch.
+REQUIRED_PLATFORMS = {
+    "linux": lambda tag: "linux" in tag,
+    "windows": lambda tag: tag.startswith("win"),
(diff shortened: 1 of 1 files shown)


hunhoffe (approved):

jgmelber:
Good catch — addressed in 9186704.

The matchers were OS-only (`"linux" in tag` / `tag.startswith("win")`), so a nightly shipping only `win32`, `win_arm64` or `linux_aarch64` would have counted as complete while still being uninstallable on the runners. Every CI runner is x86_64 (`ubuntu-*`, `windows-2022`, and the self-hosted `[windows, ryzenai, npu2]` box), so that is now required explicitly:

```python
REQUIRED_PLATFORMS = {
    "linux-x86_64": lambda tag: "linux" in tag and tag.endswith("x86_64"),
    "windows-x86_64": lambda tag: tag == "win_amd64",
}
```

Verified against the live index — the tags llvm-aie actually publishes still classify correctly, and the ones you raised now match nothing:

| Tag | Matches |
|---|---|
| `manylinux_2_27_x86_64.manylinux_2_28_x86_64` | `linux-x86_64` |
| `win_amd64` | `windows-x86_64` |
| `win32` | — |
| `win_arm64` | — |
| `linux_aarch64` | — |

`musllinux_1_2_x86_64` still matches `linux-x86_64`, which is intended — it is x86_64 Linux. In practice llvm-aie publishes only the two real tags above.

The end-to-end behaviour is unchanged: the default run resolves to `2026073101`, a `--peano-version` of the platform-incomplete `2026072701` exits 1, and replaying the index as it stood on 30 Jul still falls back to `2026072001` instead of the pin that broke CI.

PR #3483, opened 2026-07-31
ci: retry the aie2-4col CREATE_HWCTX ENOENT transient

jgmelber:
## Problem

The retry classifier only matches `EXEC_CMD err=-5`, so a hardware-context creation failure takes an otherwise-green run down and costs a manual re-run:

```
RuntimeError: DRM_IOCTL_AMDXDNA_CREATE_HWCTX IOCTL failed (err=-2): No such file or directory
```

Seen **twice today** on `aie2-4col`, both passing on re-run:

| Test | Context |
|---|---|
| `basic/memcpy/run.lit` | ungated Python JIT test, unrelated to the PR's diff |
| `matrix_multiplication/single_core` | the `trace` step, after `run` had already passed |

The driver could not hand out a context — typically because a previous test's context had not been torn down yet.

## Why this is safe to retry

It fails during **context setup, before the design executes**, so there is no computed output to compare. It cannot mask a wrong-result regression. As with the existing signatures, a deterministic failure still fails the retry.

## Retry shape

`lit` completes the suite normally here — it prints a `Failed Tests` summary and a test index (`40 of 185`) — so the retry **narrows to the failed tests**, like the read-race path:

```
FAIL: AIE_PROGRAMMING_EXAMPLES :: basic/memcpy/run.lit (40 of 185)
Failed Tests (1):
```

The `eio` path, which can abort lit mid-suite leaving tests un-run, keeps its full-suite re-run. The narrowing condition is now explicit about which transients qualify rather than relying on `hwctx` falling through by accident.

## Verification

Replayed the classifier against real and adversarial log samples:

| Sample | Runner | Classified |
|---|---|---|
| `CREATE_HWCTX err=-2` (today's flake) | aie2-4col | **hwctx** ✅ |
| `CREATE_HWCTX err=-2` | aie2p-8col | none (not gated) |
| `EXEC_CMD err=-5` | aie2-4col | eio (unchanged) |
| `CREATE_HWCTX` with a **different errno** | aie2-4col | none |
| Wrong value (`41 != 42`) | aie2-4col | none |
| Compile error | aie2-4col | none |

Gated on `aie2-4col` to match the existing `EXEC_CMD` rule, since both observations were on that runner. Uses `grep -qF`, so no GNU-only `-P` dependency is added.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -107,6 +107,12 @@ jobs:
           #      host read an output buffer before the NPU DMA had fully landed;
           #      by the time the value is printed it has settled. A real logic
           #      error produces UNEQUAL values and is therefore never retried.
+          #   3. aie2-4col hardware-context creation failing with ENOENT
+          #      (DRM_IOCTL_AMDXDNA_CREATE_HWCTX err=-2). The driver could not
+          #      hand out a context, usually because a previous test's context
+          #      had not been torn down yet; the same test passes on a re-run.
+          #      This is a setup failure before the design executes, so it can
+          #      never mask a wrong-result regression.
           # A deterministic regression fails both runs, so the retry only masks
           # nondeterministic failures.
           #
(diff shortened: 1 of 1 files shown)


jgmelber:
Superseded by #3484, which carries this retry plus the concurrency de-duplication.

Two corrections in #3484 worth noting, since they are behavioural rather than cosmetic:

- **Not gated on `aie2-4col`.** Both observed failures were on npu1, but there is no evidence npu2 is immune, so restricting the match risked leaving the same flake unhandled on the other runner.
- **Full-suite retry rather than narrowing to failed tests.** `CREATE_HWCTX` can abort lit mid-suite, leaving tests un-run, so it belongs with the EIO path; only the read-race case is safe to narrow.

Also: the title here was misleading. This only touches `buildAndTestRyzenAI.yml` (the Linux workflow, which is where the retry harness lives) — it is not Windows-related.

PR #3486, opened 2026-07-31
ci: report diff coverage and fix the code coverage set

hunhoffe:
## Description

The `code-coverage` job had two problems that made its PR comment misleading:
  
1. **Wrong binary set.** `llvm-cov` output was attributed to `aie-opt` alone, but `check-aie` also exercises `aie-translate` and `aiecc` (now a compiled C++ driver). Code reached only through those tools showed as **0% / uncovered** even when tests ran it. Now the report is generated against all three binaries so counters merge correctly.
     
2. **Wrong question, fragile rendering.** The comment pasted whole-file HTML coverage of the changed files, mangled through six `sed` passes to survive as a GitHub comment. Reviewers actually want *diff coverage* — which newly-added/changed lines are untested. We now emit an lcov file from the report target and run [`diff-cover`](https://github.com/Bachmann1234/diff_cover) against `origin/main`. The per-file table is retained as context inside a `<details>` block under a heading that states plainly it measures whole-file (not diff) coverage of the changed files.
     
Also swaps the hardcoded `/home/runner/work/mlir-aie/mlir-aie` paths for `$GITHUB_WORKSPACE` / `${{ github.workspace }}`, and adds `diff-cover` to `requirements_dev` (lock regenerated with the pinned `uv`).
  
This stays a non-blocking advisory comment — no failing status check on coverage regressions.
  
Verified locally without triggering the full `check-aie` lit suite (using version-matched `llvm-cov-18`/`llvm-profdata-18` and a small two-binary instrumented harness).

## Checklist


andrej (approved):

hunhoffe:
@andrej Thanks for getting this merged!

PR #3489, opened 2026-08-01
[AIEVec] Preserve operand signedness through vector.contract lowering

erwei-xilinx:
## Problem

An i8 matmul lowered from `vector.contract` computes `zext(A) * sext(B)` on AIE2 — the LHS is zero-extended while the RHS is sign-extended. Any negative value in A gives a wrong result:

| A | B | expected | actual |
|---|---|---|---|
| -1 | 1 | -1 | 255 |
| -7 | 5 | -35 | 1245 |
| -3 | -3 | 9 | -759 |
| 1 | -3 | -3 | -3 ✅ |

Non-negative A is always correct, so the bug is invisible to any test using only non-negative inputs — which is why it has gone unnoticed.

## Cause

The AIE2 MAC reads operand signedness from bits 9 (`signX`) and 8 (`signY`) of its configuration word (`AIE2InstrPatterns.td`, `aiev2_compute_control` in `aiev2_vmult.h`). The intrinsic always takes `v64uint8`, so the operand *type* carries no signedness.

`MatMulOpConversion` tries to recover it by looking for an `arith.extsi`/`extui` on each operand. But by the time it runs, `LowerVectorContractionOpToAIEVecMatMulPattern` has already peeled those ops off via `getSourceOfWideningOp` in order to match the narrow AIE2 shapes. The lookup always misses and falls through to an asymmetric default — unsigned LHS, signed RHS — emitting conf `0x108` where a signed × signed matmul needs `0x308`. The existing comment in `AIEVecToLLVM.cpp` notes the information is unavailable; this PR makes it available.

## Fix

Record the signedness on the op while it is still known:

- add optional `lhsSigned`/`rhsSigned` attributes to `aievec.matmul` and `aievec.matmul_aie2p`;
- set them in the contraction pattern at the point it peels the widening op;
- prefer them in `MatMulOpConversion`.

The attributes are **optional** and the previous type-based defaults are retained, so hand-written `aievec.matmul` IR behaves exactly as before.

Detection also consults the original, pre-conversion contraction operands: another pattern in the same conversion may already have rewritten `arith.extsi` into an `aievec.ups`/`aievec.srs` pair, so `getSignednessOfWideningOp` additionally reads `aievec.srs`'s own `sign` attribute.

The uint8-activation × int8-weight case used by quantized CNNs is unchanged — it now derives conf `0x108` from the actual `extui`/`extsi` ops instead of assuming it.

## Verification status

Verified end-to-end on **NPU1 (Phoenix, AIE2)** hardware, plus NPU2 as noted in the thread. The NPU1 runs below use mlir-aie **v1.4.0 + these three commits** (they cherry-pick cleanly onto the tag), driving the real `aircc`/`aiecc` flow with mlir-air `d3c5f87`. Note `aiecc` runs its passes in-process, so the patched `aiecc` binary — not just `aie-opt` — is what is exercised here.

**Config word in the generated core ELF**, same `vmac ..., rN` instruction:

```
before:  mova r22, #0x108     (signX=0, signY=1)  -> zext(A) * sext(B)
after:   mova r18, #0x308     (signX=1, signY=1)  -> sext(A) * sext(B)
```

**mlir-air `programming_examples/matrix_multiplication/i8`, `run4x4` (512x512x512, 4x4 herd, `--direct-codegen`):**

| inputs | unpatched | patched |
|---|---|---|
| `np.arange(...) % 7 - 3` (signed) | `failed.` | **`PASS!`** |
| `np.arange(...) % 7` (as shipped, non-negative) | `PASS!` | **`PASS!`** |

The as-shipped example uses values 0..6. That is exactly the non-negative pattern that hides this bug, which is why it passes either way — the signed variant is the one that exposes it.

**Triton i8 GEMM** (`tl.dot(i8,i8) -> i32`, 256x256x256, exact compare `atol=0 rtol=0`):

| probe | unpatched | patched |
|---|---|---|
| signed single products (`A=-1,B=1`; `-7,5`; `-3,-3`; ...) | wrong, `= zext8(A)*sext8(B)` | **all exact** |
| vary along M | 28416 / 65536 wrong | **0 wrong** |
| vary along N / K | 0 wrong | **0 wrong** |
| random `[-8, 8)` | 65293 / 65536 wrong | **0 wrong** |
| full example, all shapes | fails | **exit 0** |

**No regression:** a bf16 GEMM sweep (27 kernels) passes on NPU1 with the patched toolchain in 352 s, matching the unpatched baseline exactly.

## Tests

- `test/Conversion/AIEVecToLLVM/matmul.mlir` — all four signedness combinations pinned to their exact configuration words (`0x008` / `0x108` / `0x208` / `0x308`). The pre-existing attribute-free case still checks `264`, covering the unchanged fallback.
- `test/Conversion/VectorToAIEVec/test-contract.mlir` — all four `extsi`/`extui` pairings, plus the bf16 path which correctly gets no attributes (`arith.extf` carries no signedness).
- `test/dialect/AIEVec/roundtrip.mlir` — printing/parsing with and without the attributes.

`check-aie`: 90/90 AIEVec tests pass. The 20 `python/` failures in the full suite are pre-existing in my environment — I confirmed they reproduce identically on a clean `main` build.

---

## Update: AIE2P (NPU2) now honors the attributes

The first revision added `lhsSigned`/`rhsSigned` to `aievec.matmul_aie2p` but `MatMulOpAIE2pConversion` still hardcoded the i8 MAC configuration word (`776` = `0x308`, signed×signed) and ignored them — so mixed signedness silently computed signed×signed on AIE2P. The AIE2P integer lowering now derives the conf word as `0x8 | (signX<<9) | (signY<<8)` from the attributes (defaulting to signed×signed when absent), matching the AIE2 encoding (`sgn_x` at bit 9, `sgn_y` at bit 8 per `aie2p_compute_control`).

Verified end-to-end on NPU2 (Strix / AIE2P) through the mlir-air `matrix_multiplication/i8` example with `--direct-codegen`:

| test | result |
|---|---|
| uint8 LHS × int8 RHS (`cast_unsigned` on A), 64×64×64, LHS bytes straddling 127 | **PASS** vs uint8 reference |
| same run, reference recomputed as *signed* A (negative control) | **fails** (expected −1026, hardware 8190 = the unsigned result) → confirms `signX=0` reached the MAC |
| signed × signed (regression) | **PASS** |

## Review comments addressed

- **AIE2P lowering** now consumes `lhsSigned`/`rhsSigned` instead of hardcoding `776`.
- Signedness is recorded on the **first** `aievec.matmul` create, not only on the widening-peel fallback, so it survives a later canonicalization stripping the `extsi`/`extui`.
- `getSignednessOfWideningOp` doc corrected: `aievec.srs` carries its own `sign` attribute; `aievec.ups`/`aievec.cast` do not.
- `.td` fallback description clarified to match the actual asymmetric defaults (lhs unsigned unless `extsi`, rhs signed unless explicitly unsigned).

## Added tests

- `test/Conversion/AIEVecToLLVM/matmul-aie2p.mlir` — i8 conf words `8`/`264`/`520`/`776` pinned for all four signedness combinations plus the attribute-free default.

### DIFF
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -658,9 +658,11 @@ def AIEVec_MatMulOp:
     operands to recover the narrow input types. That peel would otherwise
     discard whether each operand is signed, which the AIE2 MAC needs in order
     to set the `signX`/`signY` bits of its configuration word. The optional
-    `lhsSigned`/`rhsSigned` attributes carry that information forward. When
-    absent, the lowering falls back to inferring signedness from the operand
-    types (preserving behavior for hand-written `aievec.matmul` IR).
+    `lhsSigned`/`rhsSigned` attributes carry that information forward. When an
+    attribute is absent (e.g. hand-written `aievec.matmul` IR), the lowering
+    uses the historical asymmetric defaults: the lhs is treated as unsigned
+    unless it is fed by an `arith.extsi`, and the rhs is treated as signed
+    unless its element type is explicitly unsigned.
 
     Currently, this intrinsic supports the following type combinations:
 
@@ -700,6 +702,12 @@ def AIEVec_MatMulOp_AIE2P:
   let description = [{
     AMD AIEv2P-specific intrinsic that performs a matrix multiplication
     between `lhs` and `rhs`, and accumulates the result in `acc`.
+
+    As for `aievec.matmul`, the optional `lhsSigned`/`rhsSigned` attributes
+    carry each operand's signedness (lost when the pattern peels the
+    `arith.extsi`/`arith.extui`) so the integer MAC configuration word can set
+    its `signX`/`signY` bits. When absent, the integer lowering defaults to
+    signed x signed.
 
     Currently, this intrinsic supports the following type combinations:
 
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -644,25 +644,18 @@ def AIEVec_MatMulOp:
   ]>,
   Arguments<(ins AIE2MatMulLHS:$lhs,
                  AIE2MatMulRHS:$rhs,
-                 AIE2MatMulACC:$acc,
-                 OptionalAttr<BoolAttr>:$lhsSigned,
-                 OptionalAttr<BoolAttr>:$rhsSigned)>,
+                 AIE2MatMulACC:$acc)>,
   Results<(outs AIE2MatMulACC:$result)> {
   let summary = "AIE2 matrix-multiply and accummulate";
   let description = [{
     AMD AIEv2-specific intrinsic that performs a matrix multiplications
     between `lhs` and `rhs`, and accumulates the result in `acc`.
 
-    MLIR integers are signless, and the pattern that creates this op from a
-    `vector.contract` looks through any `arith.extsi`/`arith.extui` on the
-    operands to recover the narrow input types. That peel would otherwise
-    discard whether each operand is signed, which the AIE2 MAC needs in order
-    to set the `signX`/`signY` bits of its configuration word. The optional
-    `lhsSigned`/`rhsSigned` attributes carry that information forward. When an
-    attribute is absent (e.g. hand-written `aievec.matmul` IR), the lowering
-    uses the historical asymmetric defaults: the lhs is treated as unsigned
-    unless it is fed by an `arith.extsi`, and the rhs is treated as signed
-    unless its element type is explicitly unsigned.
+    Operand signedness is carried in the integer element types: a signed
+    element type (`si8`) sets the MAC `signX`/`signY` bit, an unsigned type
+    (`ui8`) clears it. A signless element type (`i8`) keeps the historical
+    asymmetric defaults for hand-written IR: the lhs is treated as unsigned and
+    the rhs as signed.
 
     Currently, this intrinsic supports the following type combinations:
 
@@ -694,20 +687,16 @@ def AIEVec_MatMulOp_AIE2P:
   ]>,
   Arguments<(ins AIE2PMatMulLHS:$lhs,
                  AIE2PMatMulRHS:$rhs,
-                 AIE2PMatMulACC:$acc,
-                 OptionalAttr<BoolAttr>:$lhsSigned,
-                 OptionalAttr<BoolAttr>:$rhsSigned)>,
+                 AIE2PMatMulACC:$acc)>,
   Results<(outs AIE2PMatMulACC:$result)> {
   let summary = "AIE2P matrix-multiply and accummulate";
   let description = [{
     AMD AIEv2P-specific intrinsic that performs a matrix multiplication
     between `lhs` and `rhs`, and accumulates the result in `acc`.
 
-    As for `aievec.matmul`, the optional `lhsSigned`/`rhsSigned` attributes
-    carry each operand's signedness (lost when the pattern peels the
-    `arith.extsi`/`arith.extui`) so the integer MAC configuration word can set
-    its `signX`/`signY` bits. When absent, the integer lowering defaults to
-    signed x signed.
+    As for `aievec.matmul`, operand signedness is carried in the integer
+    element types (`si8`/`ui8`) so the integer MAC configuration word can set
+    its `signX`/`signY` bits. A signless element type defaults to signed.
 
     Currently, this intrinsic supports the following type combinations:
 
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -653,9 +653,14 @@ def AIEVec_MatMulOp:
 
     Operand signedness is carried in the integer element types: a signed
     element type (`si8`) sets the MAC `signX`/`signY` bit, an unsigned type
-    (`ui8`) clears it. A signless element type (`i8`) keeps the historical
-    asymmetric defaults for hand-written IR: the lhs is treated as unsigned and
-    the rhs as signed.
+    (`ui8`) clears it. Prefer spelling the signedness out.
+
+    A signless element type (`i8`) falls back to the pre-existing asymmetric
+    behavior -- lhs treated as unsigned, rhs as signed. That fallback is kept
+    only for compatibility with IR written before signedness was expressible
+    here; it is not a recommended default, and it disagrees with
+    `aievec.matmul_aie2p`, which defaults signless operands to signed.
+    Reconciling the two is left to a follow-up.
 
     Currently, this intrinsic supports the following type combinations:
 
--- a/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
+++ b/include/aie/Dialect/AIEVec/IR/AIEVecOps.td
@@ -644,12 +644,23 @@ def AIEVec_MatMulOp:
   ]>,
   Arguments<(ins AIE2MatMulLHS:$lhs,
                  AIE2MatMulRHS:$rhs,
-                 AIE2MatMulACC:$acc)>,
+                 AIE2MatMulACC:$acc,
+                 OptionalAttr<BoolAttr>:$lhsSigned,
+                 OptionalAttr<BoolAttr>:$rhsSigned)>,
   Results<(outs AIE2MatMulACC:$result)> {
   let summary = "AIE2 matrix-multiply and accummulate";
   let description = [{
     AMD AIEv2-specific intrinsic that performs a matrix multiplications
     between `lhs` and `rhs`, and accumulates the result in `acc`.
+
+    MLIR integers are signless, and the pattern that creates this op from a
+    `vector.contract` looks through any `arith.extsi`/`arith.extui` on the
+    operands to recover the narrow input types. That peel would otherwise
+    discard whether each operand is signed, which the AIE2 MAC needs in order
+    to set the `signX`/`signY` bits of its configuration word. The optional
+    `lhsSigned`/`rhsSigned` attributes carry that information forward. When
+    absent, the lowering falls back to inferring signedness from the operand
+    types (preserving behavior for hand-written `aievec.matmul` IR).
 
     Currently, this intrinsic supports the following type combinations:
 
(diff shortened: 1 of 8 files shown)


lib/Dialect/AIEVec/Transforms/VectorToAIEVecConversions.cpp
```diff
@@ -178,6 +178,30 @@ static std::optional<Value> getSourceOfWideningOp(Value src) {
   return std::optional<Value>();
 }
 
+// Given a Value, if it is defined by an integer widening op, return whether
+// that widening is signed. Returns std::nullopt when the defining op is not a
+// recognized integer widening op, or when it carries no signedness (e.g. a
+// floating-point extension, or an accumulator round-trip through
+// aievec.ups/srs). Callers use this to record the operand signedness on the
+// created matmul op, since peeling the widening op would otherwise discard it
+// (MLIR integers are signless).
```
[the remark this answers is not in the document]
erwei-xilinx: Fixed in a328e3d0f — corrected the doc: aievec.srs carries its own `sign` attribute; aievec.ups/cast do not.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -644,13 +644,24 @@ def AIEVec_MatMulOp:
...
     between `lhs` and `rhs`, and accumulates the result in `acc`.
 
+    MLIR integers are signless, and the pattern that creates this op from a
+    `vector.contract` looks through any `arith.extsi`/`arith.extui` on the
+    operands to recover the narrow input types. That peel would otherwise
+    discard whether each operand is signed, which the AIE2 MAC needs in order
+    to set the `signX`/`signY` bits of its configuration word. The optional
+    `lhsSigned`/`rhsSigned` attributes carry that information forward. When
+    absent, the lowering falls back to inferring signedness from the operand
+    types (preserving behavior for hand-written `aievec.matmul` IR).
```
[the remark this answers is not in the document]
erwei-xilinx: Fixed in a328e3d0f — reworded to the actual asymmetric defaults (lhs unsigned unless fed by extsi, rhs signed unless the element type is explicitly unsigned).

lib/Conversion/AIEVecToLLVM/AIEVecToLLVM.cpp
```diff
@@ -4311,6 +4316,9 @@ class MatMulOpConversion
       rhs = lookThroughShapeCasts(extUIOp.getIn());
       rhsVecTy = cast<VectorType>(rhs.getType());
       rhsScaTy = cast<IntegerType>(rhsVecTy.getElementType());
+    } else if (auto rhsSignedAttr = op.getRhsSigned()) {
+      // See the lhs case above.
+      signY = *rhsSignedAttr ? 1 : 0;
     } else {
```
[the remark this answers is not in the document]
erwei-xilinx: Fixed in a328e3d0f — MatMulOpAIE2pConversion now derives the i8 conf word from lhsSigned/rhsSigned instead of hardcoding 776; verified uint8×int8 on NPU2 and added matmul-aie2p.mlir tests for all four combinations.

jsetoain:
> The AIE2 MAC reads operand signedness from bits 9 (`signX`) and 8 (`signY`) of its configuration word (`AIE2InstrPatterns.td`, `aiev2_compute_control` in `aiev2_vmult.h`). The intrinsic always takes `v64uint8`, so the operand _type_ carries no signedness.

A `v64uint8` does carry signedness, it's _unsigned._ `v64sint8` and `v64uint8` are signed and unsigned values, respectively. `v64int8` is the signless value.

> ## Fix
> 
> Record the signedness on the op while it is still known:
> 
>     * add optional `lhsSigned`/`rhsSigned` attributes to `aievec.matmul` and `aievec.matmul_aie2p`;
> 

MLIR already offers natural ways to express signedness in types (sint/uint vs int), there's not need for additional attributes.

test/Conversion/AIEVecToLLVM/matmul.mlir
```diff
@@ -111,3 +111,75 @@ func.func @matmul(%A : vector<4x2xi32>, %B : vector<2x4xi16>,
...
+// CHECK:      "xllvm.intr.aie2.I512.I512.ACC1024.acc32.mac.conf"(
+// CHECK-SAME:   %{{.*}}, %{{.*}}, %{{.*}}, %[[CONF]])
+
+// -----
+
+func.func @matmul_i8i8_signed_unsigned(%A : vector<4x8xi8>, %B : vector<8x8xi8>,
+                                       %C : vector<4x8xi32>) -> vector<4x8xi32> {
+  %0 = aievec.matmul %A, %B, %C {lhsSigned = true, rhsSigned = false} :
+                                  vector<4x8xi8>, vector<8x8xi8>
+                                  into vector<4x8xi32>
```
jsetoain: For instance, the equivalent without the need for attributes would be:

```mlir
%0 = aievec.matmul %A, %B, %C : vector<4x8xsi8>, vector<8x8xui8> into vector<4x8xsi32>
```

jsetoain (commented):
I'm not going to block it because I'm no longer actively involved in this, but this is a bad solution and thought I'd give you a heads up 🙂

erwei-xilinx:
NPU1 (Phoenix / AIE2) end-to-end verification is now in, complementing the NPU2 results above. This closes the gap I flagged earlier — I had previously only verified the reworked fix at IR level.

Built **v1.4.0 + these three commits** (they cherry-pick cleanly onto the tag) and ran the real `aircc`/`aiecc` flow against mlir-air `d3c5f87`. Worth noting for anyone reproducing: `aiecc` now runs its passes in-process, so swapping `aie-opt` alone has no effect — the patched `aiecc` binary is what matters.

Config word in the generated core ELF, same `vmac` instruction:

```
before:  mova r22, #0x108     (signX=0, signY=1)
after:   mova r18, #0x308     (signX=1, signY=1)
```

**This repo's own i8 example**, `run4x4` (512³, 4×4 herd, `--direct-codegen`):

| inputs | unpatched | patched |
|---|---|---|
| `% 7 - 3` (signed) | `failed.` | **`PASS!`** |
| `% 7` (as shipped) | `PASS!` | **`PASS!`** |

The shipped inputs are 0..6, which is precisely the non-negative pattern that masks this bug — worth considering whether that example should carry a signed case so this cannot regress silently.

**Triton i8 GEMM** (`tl.dot(i8,i8) -> i32`, exact compare): random `[-8,8)` goes from 65293/65536 elements wrong to **0**; the "vary along M" probe from 28416 wrong to **0**; vary-N and vary-K were already clean and stay clean. bf16 sweep (27 kernels) unchanged at 352 s, no regression.

Also — thanks for the correction on `v64uint8`/`v64sint8`/`v64int8`. My original description claimed the operand type carried no signedness, which was simply wrong, and it was the premise for reaching for attributes in the first place. The description has been rewritten accordingly.

erwei-xilinx:
Thanks for taking the time to review this, and for the suggestion — you're right that the element type is the natural home for the signedness. Reworked to carry it in the operand type (si8/ui8) and dropped the attributes. (2afbc727e)

Implementation note: since the vector.contract inputs are signless the whole way down (even arith.extsi/extui produce signless results) and arith.bitcast won't produce a signed/unsigned result, the contraction pattern materializes a builtin.unrealized_conversion_cast from the signless operand to si8/ui8; the AIE2/AIE2P lowerings read signX/signY from the element type and look through that cast so the intrinsic consumes the signless value, leaving the cast dead for canonicalize to DCE before LLVM translation.

Validated on NPU2 (AIE2P): `uint8 x int8` and `signed x signed` pass, and a signed-reference negative control fails — so signX=0 genuinely reaches the MAC; all four i8 combinations pin conf 8/264/520/776, and check-aie is green.

Thanks again for the steer.

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
```diff
@@ -651,6 +651,12 @@ def AIEVec_MatMulOp:
     AMD AIEv2-specific intrinsic that performs a matrix multiplications
     between `lhs` and `rhs`, and accumulates the result in `acc`.
 
+    Operand signedness is carried in the integer element types: a signed
+    element type (`si8`) sets the MAC `signX`/`signY` bit, an unsigned type
+    (`ui8`) clears it. A signless element type (`i8`) keeps the historical
+    asymmetric defaults for hand-written IR: the lhs is treated as unsigned and
+    the rhs as signed.
```
andrej: Is there really a reason to keep this historically weird implementation? Why not just make both unsigned for signless types for consistency?

andrej (approved):
Seems reasonable, one small question/comment

include/aie/Dialect/AIEVec/IR/AIEVecOps.td
erwei-xilinx: Agreed, no good reason — it's preserved for backward compatibility, not by design. I'll do it in a follow-up PR so this one stays scoped to the signedness-recovery fix.

One thing I'll dig into there: signless also reaches generated code for the native i16 shapes (no ext op to peel), and those want *signed* rather than unsigned — which would also line AIE2 up with `aievec.matmul_aie2p`, which already defaults signless to signed.

Reworded the op description in b94bae8 so it reads as compatibility-preserved rather than intentional.

jsetoain: I'm gong to leave here a good reference to help understand signedness, for those curious about this peculiarity in MLIR/LLVM: [MLIR Rationale: Integer signedness semantics](https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics)

erwei-xilinx: Thanks — that's a useful anchor, and it bears directly on the follow-up in #3503.

The rationale's point that interpretation belongs to the op rather than the type is exactly what was missing here: `aievec.matmul` had no defined reading for a signless operand, just an undocumented asymmetric guess (lhs unsigned, rhs signed) that `aievec.matmul_aie2p` didn't share. #3503 gives it one — signless means signed, same rule for both ops, resolved per operand.

Worth noting that guess wasn't only reachable from hand-written IR: when an operand's type is already a legal AIE2 narrow type nothing gets extended, so the contraction pattern leaves it signless and generated code lands on the default. The native i16 shapes all hit it (conf 280/314/282, signX=0), as do the mixed ones like `vector<4x4xi16>, vector<4x8xsi8>`.

PR #3492, opened 2026-08-01
Windows XRTKernelResult.run fix npu_time precision

astrelsky:
## Description

Switch to `time.perf_counter_ns` since `time.time_ns` is not precise …enough on Windows causing start and stop to be the same value. Previously a majority of the IRON tests fail with a divide by zero error because the `npu_time` was 0.

## Checklist

### DIFF
--- a/programming_examples/ml/magika/test_group0.py
+++ b/programming_examples/ml/magika/test_group0.py
@@ -105,9 +105,9 @@ def main(opts):
     # Main run loop
     # ------------------------------------------------------
     for i in range(num_iter):
-        start = time.time_ns()
+        start = time.perf_counter_ns()
         ret = DefaultNPURuntime.run(kernel_handle, buffers)
-        stop = time.time_ns()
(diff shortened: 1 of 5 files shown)


thomthehound:
Huh. I can't believe this snuck by for so long. Thanks for catching it!

Would you mind fixing `_run_full_elf()` in the same way for consistency? Should just be two more lines that need changing.

astrelsky:
> Huh. I can't believe this snuck by for so long. Thanks for catching it!
> 
> Would you mind fixing `_run_full_elf()` in the same way for consistency? Should just be two more lines that need changing.

Sure can. I didn't realize the problem was elsewhere too.

astrelsky:
@thomthehound I searched for all uses of `time.time_ns`, and they were all being used exactly the same way, so I applied a simple search and replace to change them to `time.perf_counter_ns()`.

andrej (approved):
Thanks, good catch!

PR #3495, opened 2026-08-02
[test_lib] Fix two tile-address bugs in the status printers

atassis:
## Problem

`mlir_aie_get_tile_addr(ctx, col, row)` is called as `(ctx, row, col)` at all 8
sites in `test_library.cpp`, so a design whose only tile is (7,3) reads (3,7).

`mlir_aie_print_dma_status` is the core-tile printer (it asserts
`XAIEGBL_TILE_TYPE_AIETILE`) but takes `write_count` from `0x1D230`, the
shim/NoC module's register. The core tile's own is `0x1DF18`, already inside the
range `installDma` claims for core-tile status.

## Fix

Correct the argument order at the 8 sites, and `0x1D230` -> `0x1DF18` in the
core-tile printer only. The shim and memtile siblings already use theirs.

## Test

Found running `test/unit_tests/aie2/08_tile_locks`; the tile-address bug shows
as a fault at the transposed coordinate.

### DIFF
--- a/runtime_lib/test_lib/test_library.cpp
+++ b/runtime_lib/test_lib/test_library.cpp
@@ -203,7 +203,7 @@ static void print_aie2_dmachannel_status(aie_libxaie_ctx_t *ctx, int col,
                                          const char *channel, int channelNum,
                                          u32 statusOffset, u32 controlOffset,
                                          int &current_bd) {
-  u64 tileAddr = mlir_aie_get_tile_addr(ctx, row, col);
+  u64 tileAddr = mlir_aie_get_tile_addr(ctx, col, row);
   u32 status, control;
   XAie_Read32(ctx->XAieDevInst, tileAddr + statusOffset, &status);
   XAie_Read32(ctx->XAieDevInst, tileAddr + controlOffset, &control);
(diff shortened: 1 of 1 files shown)


andrej (approved):
Another good catch. I actually didn't know we had this functionality. Does this allow us to read out the state of DMAs when there is e.g., a deadlock, on aie2? Does this work for you?

atassis:
Yes, on aie2 that is exactly what it gives you: per-channel IDLE/STARTING/RUNNING plus the four stall causes (Acquire, Release, Data, Completion) from status bits 2-5, the current BD from bits 24+, s2mm write_count, then every BD with its acquire/release lock id and value.

The caveat is that it is aie2 only. The gate is AieGen == XAIE_DEV_GEN_AIEML and the file has two arms, that one and an else meaning AIE1, so an aie.device(npu2/npu4) design falls to the AIE1 offsets. It does not fault: print_tile_status reads 0x30280 for PC, which on aie2p is a valid CORE_BMHL2_PART1, so you get a plausible wrong number instead.
print_memtiledma_status asserts, and that assert is compiled out under NDEBUG.

For the DMA printers this is only the gate: all 11 offsets the aie2 arm uses are identical in xaiemlgbl_params.h and xaie2pgbl_params.h. The core register window is not (0x30E00 is M0 on aie2 and PC on aie2p), so print_tile_status needs a real third arm. I can send that as a follow-up if useful.

To your second question, not on hardware yet: I hit this on aie2 (08_tile_locks, xcve2802) under simulation. My box is npu4, where these take the AIE1 arm, which is why the aie2p arm matters to me.

PR #3496, opened 2026-08-02
[AIE] Move the core stack-size default onto AIETargetModel

atassis:
## Problem

`aie.core`'s `stack_size` default of `0x400` was stated in the TableGen default
and again in the op docs, and could not vary per target although
`getLocalMemorySize()` does. The attribute was also unconstrained, so
`stack_size = -1` was read as a huge unsigned and accepted without a diagnostic.

## Fix

`AIETargetModel::getDefaultCoreStackSize()` holds the default; consumers read
`CoreOp::getEffectiveStackSize()`. The attribute is constrained positive, and
`verify()` rejects a reservation that leaves no local memory for the tile's
buffers.

## Scope

This does not check the reservation against a core's actual frames. That needs
the linked ELF's `.stack_sizes`, which the generated script already keeps and
nothing reads; `iree-amd-aie` does that check downstream. Separate change.

## Test

New `test/dialect/AIE/bad_stack_size.mlir`. Diffed `aie-opt` error text against
a binary built from this branch's own base over 1062 files in `test/`: the new
test is the only difference. BCF and ldscript output byte-identical with and
without the attribute.

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -1833,6 +1833,14 @@ LogicalResult CoreOp::verify() {
                  << "' appears in both 'link_files' and 'link_merge_files'; an "
                     "artifact must be either merged or linked, not both";
     }
+  // Checked last so it does not pre-empt the diagnostics above on an op with
+  // more than one defect.
+  if (uint32_t stackSize = getEffectiveStackSize(),
+      localMem = getTargetModel(*this).getLocalMemorySize();
+      stackSize >= localMem)
+    return emitOpError("stack_size ")
+           << stackSize << " leaves no local memory for this tile's buffers ("
+           << localMem << " bytes total)";
   return success();
 }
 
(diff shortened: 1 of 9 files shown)


lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1833,6 +1833,14 @@ LogicalResult CoreOp::verify() {
...
                     "artifact must be either merged or linked, not both";
     }
+  // Checked last so it does not pre-empt the diagnostics above on an op with
+  // more than one defect.
+  if (uint32_t stackSize = getEffectiveStackSize(),
+      localMem = getTargetModel(*this).getLocalMemorySize();
+      stackSize >= localMem)
+    return emitOpError("stack_size ")
+           << stackSize << " leaves no local memory for this tile's buffers ("
+           << localMem << " bytes total)";
```
andrej, line 1843: That's a valuable extra check for too large stack size values

andrej (approved):
Very reasonable change, thanks for these improvements. See also #2347

PR #3497, opened 2026-08-02
[aiecc] Fail on a backend directory that does not exist

atassis:
## Problem

`--peano` and `--aietools` were both dropped when the stated directory did not
exist, and the search fell through to PATH -- which answers `opt` and `llc` with
the host LLVM. A bogus `--peano` ran 30 of 42 pipeline edges and died inside
`/usr/bin/opt`, never naming the flag. The Peano arm can also succeed against the
host toolchain and emit an artifact built by the wrong compiler.

## Fix

Both flags fail at the flag when the stated directory is missing. Discovery
candidates are guesses, so their absence stays non-fatal, and an unset or empty
value still reaches discovery unchanged. This is what `--xclbinutil-path`
already does with an unusable override.

## Test

`test/aiecc/cpp_backend_dir_missing.mlir`. Measured against the built binary:
`--peano=<missing>` and `--aietools=<missing>` now exit 1 naming the flag
(previously 30/42 edges, then a failure inside the wrong tool); `--peano=` and a
real `--peano=<dir>` both still reach 42/42.

### DIFF
--- a/test/aiecc/cpp_backend_dir_missing.mlir
+++ b/test/aiecc/cpp_backend_dir_missing.mlir
@@ -1,0 +1,65 @@
+//===- cpp_backend_dir_missing.mlir ----------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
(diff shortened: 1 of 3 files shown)


andrej (approved):
Good change -- erroring explicitly if someone specified `--peano` to an invalid value is much less surprising behavior than silently ignoring it and falling through to regular discovery.

PR #3499, opened 2026-08-03
ci: lock the device, not the whole test

atassis:
## Problem

`test/npu-xrt/lit.local.cfg` puts the whole directory in the capacity-1 `npu-xrt` parallelism group.
The NPU cannot take concurrent dispatch, so something has to serialize, but the group also serializes
each test's compile, which is nearly all of a device test's wall time and needs no device.

Measured on an AIE2P box over the whole npu-xrt filter: 197.55s at `-j1`, 189.71s at `-j12`. Twelve
workers requested, same wall time, eleven idle. The device-only portion of one of these tests is
about 20ms.

## Fix

Serialize at the dispatch instead. Every device run already goes through `utils/run_on_npu.py`, so an
`flock` there covers dispatch and nothing else, and lit schedules the tests freely. It is keyed on
NPU kind, released by the kernel on process death, and `AIE_NPU_NO_DEVICE_LOCK=1` opts out. Where
`fcntl` is unavailable the lit group stays in force.

Moving eight `.lit` files to `%t.d` is the precondition: lit's working directory is the test
directory, and sibling tests there write the same artifact names (all three of
`xrt_handle_lifetime`'s write `aie.xclbin` and `test.exe`). Invisible serialized, a clobber in
parallel.

## Test

```
197.55s  -j1,  parallelism group   (before)
189.71s  -j12, parallelism group   (before)
 47-67s  -j12, device lock         (after, three runs)
```

56 passed / 3 failed in every run before and after; the three are pre-existing and unrelated
(`test_cascade_flow`, `test_compile_link`, `test_device_override`). The lock was checked at 12, 64
and 256 contenders: no overlapping critical sections.

## Memory

This does not raise the concurrency ceiling. `buildAndTestRyzenAI.yml` already picks `-j` from the
runner's actual RAM (`-j4` under 48GB, `-j12` otherwise); npu-xrt now runs under that ceiling instead
of below it. If the suite proves memory-heavy, the narrower fix is a parallelism group for the
specific tests, as `gemm_asymmetric_tile_buffering` has via `atb_chess`, rather than the whole
directory.

### DIFF
--- a/test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
+//
+// RUN: rm -rf %t.d && mkdir -p %t.d
+// RUN: cd %t.d
 // RUN: %aiecc -v --get-full-elf --load-pdi-to-ctrl-pkt --get-ctrlpkt --no-xchesscc --no-xbridge %S/aie.mlir
 // RUN: %host_clang -o test %S/test.cpp -std=c++17 %host_link_flags %xrt_flags
 // RUN: %run_on_npu2% ./test
--- a/test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit
@@ -6,9 +6,7 @@
 // Sibling tests in this directory write the same artifact names, and lit's
 // working directory is the test DIRECTORY rather than the test file, so they
 // would clobber each other if they ever ran concurrently.  Give each test its
-// own directory (%t is per test file), the way the makefile-driven examples
-// already do.  This is what lets the device lock replace the capacity-1
-// parallelism group without the tests racing on their artifacts.
+// own directory (%t is per test file).
 //
 // RUN: rm -rf %t.d && mkdir -p %t.d
 // RUN: cd %t.d
--- a/test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
@@ -6,9 +6,7 @@
 // Sibling tests in this directory write the same artifact names, and lit's
 // working directory is the test DIRECTORY rather than the test file, so they
 // would clobber each other if they ever ran concurrently.  Give each test its
-// own directory (%t is per test file), the way the makefile-driven examples
-// already do.  This is what lets the device lock replace the capacity-1
-// parallelism group without the tests racing on their artifacts.
+// own directory (%t is per test file).
 //
 // RUN: rm -rf %t.d && mkdir -p %t.d
 // RUN: cd %t.d
--- a/test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
+//
+// RUN: rm -rf %t.d && mkdir -p %t.d
+// RUN: cd %t.d
 // RUN: %aiecc -v --get-full-elf --no-xchesscc --no-xbridge %S/aie.mlir
 // RUN: %host_clang -o test %S/test.cpp -std=c++17 %host_link_flags %xrt_flags
 // RUN: %run_on_npu2% ./test
--- a/test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
+//
+// RUN: rm -rf %t.d && mkdir -p %t.d
+// RUN: cd %t.d
 // RUN: %aiecc -v --get-full-elf --expand-load-pdis --no-xchesscc --no-xbridge %S/aie.mlir
 // RUN: %host_clang -o test %S/test.cpp -std=c++17 %host_link_flags %xrt_flags
 // RUN: ./test
--- a/test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
+++ b/test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
@@ -6,9 +6,7 @@
 // Sibling tests in this directory write the same artifact names, and lit's
 // working directory is the test DIRECTORY rather than the test file, so they
 // would clobber each other if they ever ran concurrently.  Give each test its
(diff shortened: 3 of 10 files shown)


atassis:
@andrej this is a PR (commit 4) from #3476 we have agreed on to merge and have a look at it how it behaves

test/npu-xrt/reconfigure_loadpdi/run_ctrlpkt.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/reconfigure_loadpdi/run_loadpdi.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/reconfigure_loadpdi/run_write32s.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/reconfigure_loadpdi_persistent_memtile/run_loadpdi.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/reconfigure_loadpdi_persistent_memtile/run_write32s.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai_npu2, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/xrt_handle_lifetime/run_ordered.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/xrt_handle_lifetime/run_stale_instr_bo.lit
```diff
@@ -4,6 +4,15 @@
 // REQUIRES: ryzen_ai, peano
 // XFAIL: system-windows
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

test/npu-xrt/xrt_handle_lifetime/run_stale_io_bos.lit
```diff
@@ -3,6 +3,15 @@
 //
 // REQUIRES: ryzen_ai, peano
 //
+// Sibling tests in this directory write the same artifact names, and lit's
+// working directory is the test DIRECTORY rather than the test file, so they
+// would clobber each other if they ever ran concurrently.  Give each test its
+// own directory (%t is per test file), the way the makefile-driven examples
+// already do.  This is what lets the device lock replace the capacity-1
+// parallelism group without the tests racing on their artifacts.
```
andrej:
```suggestion
// Sibling tests in this directory write the same artifact names, and lit's
// working directory is the test DIRECTORY rather than the test file, so they
// would clobber each other if they ever ran concurrently.  Give each test its
// own directory (%t is per test file).
```

utils/run_on_npu.py
```diff
@@ -138,6 +145,47 @@ def run_command(command: list[str]) -> tuple[int, str]:
...
+# tests in parallel. Properties that matter:
+#   * flock is released by the kernel when the fd closes or the process dies, so
+#     a crashed or timed-out test cannot wedge the queue.
+#   * the lock file lives at a stable path and is never unlinked: flock is per
+#     INODE, so deleting and recreating the file would let two holders through.
+#   * it is machine-wide, which is stronger than a lit parallelism group; that
+#     only serializes within one lit invocation, and gives nothing if a second
+#     job on the same host also touches the device.
+#   * keyed on the NPU kind, so npu1 and npu2 dispatches do not block each other.
+# AIE_NPU_NO_DEVICE_LOCK=1 opts out, for tests that exercise concurrency itself.
```
andrej, line 157:
```suggestion
# The NPU cannot sustain concurrent dispatch, so device runs must be serialized.
# Serialize around the dispatch only, so lit is free to run the tests in parallel. 
#   * flock is released by the kernel when the fd closes or the process dies, so
#     a crashed or timed-out test cannot wedge the queue.
#   * the lock file lives at a stable path and is never unlinked: flock is per
#     INODE, so deleting and recreating the file would let two holders through.
#   * it is machine-wide, which is stronger than a lit parallelism group; that
#     only serializes within one lit invocation, and gives nothing if a second
#     job on the same host also touches the device.
# AIE_NPU_NO_DEVICE_LOCK=1 opts out, for tests that exercise concurrency itself.
```

andrej (approved):

andrej:
Cleaned up the comments for you and merged main. 

I think it's not needed to key the concurrency lock on NPU type. At least currently, there's only ever going to be one type of device per machine. But since it doesn't hurt I left it.

My approval is conditioned on that this doesn't OOM on the runners. I will re-run the job a couple times to make sure this does not introduce transient failures.

PR #3500, opened 2026-08-03
ci: run the ATB artifact suite nightly, not on every PR

atassis:
## Problem

`gemm_asymmetric_tile_buffering` runs on every PR and costs a quarter to two
fifths of the aie2p-8col device job. Measured on three aie2p-8col merge-queue
jobs, 2026-08-03:

| runner | lit workers | ATB | whole job | share |
|---|---|---|---|---|
| aie4 | 12 | 5.61 min | 17.51 min | 32% |
| aie6 | 12 | 7.72 | 19.56 | 39% |
| aie8 | 4 | 6.08 | 23.84 | 26% |

It does not shrink on a bigger host. For the memory reason the file's own comment
gives, it already has its own `-j4` invocation and 1200 s timeout, serialized by
lit's `atb_chess` group, so its share rises as the rest of the job gets faster.
It contributes nothing on aie2-4col, where the same suite reports ~0.01 s.

These three configs reproduce a published result (arXiv 2511.16041). They are an
artifact, not a regression gate on the compiler.

## Fix

Run the `atb` suite on the existing nightly `cron: '0 0 * * *'`, and on
`test-ryzen-ai` so a kernel change can still be exercised on demand. No new
trigger, no change to how the suite runs when it does run.

#3104 raised running the Chess tests nightly rather than on every run. This does
it for the one that costs the most.

## Coverage

This is a real reduction: presubmit no longer builds or runs the three ATB
configs, so a change that breaks only them surfaces at the next nightly, against
a later commit than the one that caused it. `check-reference-designs` already
excludes them via `LIT_FILTER_OUT`, so nothing else in presubmit covers them
today either.

## Test

None: a workflow file is only exercised by a run. The gate is one bash
conditional, checked standalone against every `(event, ref)` pair that can occur:

| event | ref | atb |
|---|---|---|
| `pull_request`, `merge_group` | PR/queue ref | excluded |
| `push` | `test-ryzen-ai`, its only trigger | included |
| `schedule` | `main` | included |

The retry path needs no change: it iterates `$SUITES`, and `FAILED_SUITE` is
validated against that same list.

### DIFF
--- a/.github/workflows/buildAndTestRyzenAI.yml
+++ b/.github/workflows/buildAndTestRyzenAI.yml
@@ -217,7 +217,13 @@ jobs:
             esac
           }
 
-          SUITES="check-aie concurrency refs guide atb"
+          # gemm_asymmetric_tile_buffering reproduces a paper rather than gating
+          # a regression, and is ~6 min of an 18-24 min job that does not shrink
+          # on a bigger host. Nightly, plus test-ryzen-ai for on-demand runs.
+          SUITES="check-aie concurrency refs guide"
+          if [ "$GITHUB_EVENT_NAME" = schedule ] || [ "$GITHUB_REF_NAME" = test-ryzen-ai ]; then
+            SUITES="$SUITES atb"
+          fi
 
           # Only treat FAILED_SUITE as valid if it EXACTLY names a known suite
           # (fixed-string match, not a glob) -- else force the full re-run path.


andrej:
I see the motivation for the performance gain, but this does have the risk that a regression might break the test and we wouldn't notice until the next push to main or nightly run. This is an important test for us since it's one of the most performant GEMMs currently checked in.

Let me first check how performance compares when this is compiled using peano instead of chess. That would probably already speed compilation up, and if it does without harming performance, that would probably be the best middle ground.

atassis:
Ah, so you do rely on this test, I thought it was just paper related, that's why it has some importance. I haven't thought about this in this key

andrej:
It would definitely be good not to break it. That being said, pushes to main and nightly are probably "quick enough" to find the regression and be able to follow up with a quick fix PR if we do end up breaking something.

Claude's evaluation seems to imply that switching to peano is not an option for this test due to the performance regression:

Config | chess compile | peano compile | peano result
-- | -- | -- | --
config1 | 2:13 | 0:31 | compiles, fails verification (977/1000 wrong)
config2 | 2:14 | fails (~2.4 s) | backend can't legalize G_BUILD_VECTOR <8 x s8> in matmul_vectorized_bfp16
config3 | 1:45 | 0:29 | compiles, PASS, but ~35% slower (18.2 vs 27.9 TFLOPS avg)

So I vote let's merge this. @jgmelber do you agree?

atassis:
config2 is a Peano bug I already fixed. `unable to legalize instruction: G_BUILD_VECTOR <8 x s8>` in `matmul_vectorized_bfp16` is the `<8 x s8>` block-exponent splat from Xilinx/llvm-aie#1184, merged Jul 31. The pin here predates it: `utils/peano-requirements.txt` is on `cfd8aae2` (Jul 29). #3498 moves it to `c9c5ecb7`, which is the tip of that series, so config2 should get past the legalizer there. #3498 has been green since Aug 3.

That does not change your verdict though. config1 and config3 still stand, and they are the worse two: config1 compiling and then getting 977/1000 wrong is a silent bfp16 miscompile, and 18.2 vs 27.9 TFLOPS is a 35% regression on the GEMM. I would like to take both. Can you share the exact configs and commands you ran, so I reproduce the same numbers rather than a nearby thing?

andrej:
Looks like this is the commands it ran to benchmark the timed runs:

````
# ---- config1 : 4096x4096x2048, tile 128x64x128 ----
cd programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config1
for r in 1 2 3 4 5; do
  ./n32_core.exe \
    -x build/final_4096x4096x2048_128x64x128.xclbin \
    -i build/insts_4096x4096x2048_128x64x128.bin \
    -k MLIR_AIE -M 4096 -K 4096 -N 2048 -w 128 -y 64 -z 128 \
    -v 0 --verify 0 --warmup 5 --iters 20
done
```

```bash
# ---- config2 : 3072x4096x1536, tile 192x128x96 ----
cd programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config2
for r in 1 2 3 4 5; do
  ./n32_core.exe \
    -x build/final_3072x4096x1536_192x128x96.xclbin \
    -i build/insts_3072x4096x1536_192x128x96.bin \
    -k MLIR_AIE -M 3072 -K 4096 -N 1536 -w 192 -y 128 -z 96 \
    -v 0 --verify 0 --warmup 5 --iters 20
done
```

```bash
# ---- config3 : 4096x4096x2048, tile 128x64x128 ----
cd programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config3
for r in 1 2 3 4 5; do
  ./n32_core.exe \
    -x build/final_4096x4096x2048_128x64x128.xclbin \
    -i build/insts_4096x4096x2048_128x64x128.bin \
    -k MLIR_AIE -M 4096 -K 4096 -N 2048 -w 128 -y 64 -z 128 \
    -v 0 --verify 0 --warmup 5 --iters 20
done
```

This should mirror what `make run` does roughly, except no verification, no verbosity, and 5 warmups followed by 20 timed iterations for this timed run.

hunhoffe (approved):

PR #3501, opened 2026-08-04
[aiecc] Default the core backend to Peano

atassis:
## Problem

README ("Optional: Install AIETools") says Vitis may be skipped for AIE2/AIE2P, and `env_install.sh`
installs Peano unconditionally. `aiecc` disagreed: `xchesscc` and `xbridge` were both
`cl::init(true)`, so a Vitis-free install ran 14 of 17 stages and died on a hardcoded Chess path.

```
$ aiecc --get-core-elfs mini.mlir
aiecc: '/tps/lnx64/target_aie_ml/bin/LNa64bin/chess-llvm-link' failed:
       posix_spawn failed: No such file or directory
$ aiecc --no-xchesscc --no-xbridge --get-core-elfs mini.mlir   # same input, exit 0
```

## Fix

Drop both `cl::init(true)`. The flag algebra in `resolveOptions()` is unchanged.

`--help` now names the default on all four backend flags. `--no-xchesscc` read as "use this to get
Peano", which is no longer true -- on its own it is now a no-op.

`--get-aiesim` no longer turns Chess back on for callers who named no backend. It selects a
simulator, not a compiler, so inheriting the backend means a second simulator would silently take
the flag over; it now requires `--xchesscc`/`--xbridge` stated.

Every aiecc RUN line now names its backend. 44 AIE1 and benchmark lines take `--xchesscc --xbridge`
-- most of them because AIE1 has no Peano support (#3414); a few because their directory is gated on
`AIE in vitis_components` regardless of the target device, so they stay Chess-only for now even where
Peano could in principle run them. 17 lines that never reach a core compiler take `--no-xchesscc
--no-xbridge`; none needs `REQUIRES: peano`, verified by running them with no Peano on PATH.

`python/utils/compile/utils.py`'s `compile_mlir_module` had the same implicit dependency: its
`use_chess` branch passed only `--unified` and relied on the old default. Left alone, it would have
silently started compiling every `use_chess=True` IRON caller (e.g. the `block_datatypes` bfp16
examples) with Peano instead of Chess. It now states `--xchesscc --xbridge` explicitly, matching the
`else` branch's existing explicit `--no-xchesscc --no-xbridge`.

Of 301 RUN lines, 8 remain implicit and cannot be otherwise: six `--resume` (aiecc rejects other
arguments and replays the manifest argv, whose creating runs are explicit) and two that test the
unflagged behaviour itself.

`cpp_aiesim.mlir`'s new negative-path cases (the `--get-aiesim` contradiction/omission diagnostics)
moved to their own file, `cpp_aiesim_negated.mlir`, without the parent file's `REQUIRES: chess`: they
fire in `resolveOptions()` before any toolchain is touched, so gating them on Chess would mean they
never ran in the Vitis-free lane this change exists to unblock.

## Evidence

Measured over all 11 flag combinations. Three rows move, the other eight are identical:

| flags | before | after |
|---|---|---|
| *(none)* | chess/chess | peano/peano |
| `--no-xbridge` | chess/peano | peano/peano |
| `--get-aiesim` | chess/chess | error: requires `--xbridge` |

`--no-xbridge` alone meant Chess-compile + Peano-link; it has no call site in the tree.

Full lit suite, two arms built fresh from the same base, Peano wired, no Vitis/XRT: 1228 -> 1229
discovered (new `test/aiecc/cpp_peano_default.mlir`), 27 failures both arms, failure set identical.

This run predates three follow-up fixes made during review (the two `cpp_backend_dir_missing.mlir`
lines, the `cpp_aiesim_negated.mlir` split, and the `compile_mlir_module` fix) -- each is either a
verbatim relocation of already-covered RUN/CHECK lines or an addition of flags already exercised
elsewhere in the suite, but none has been re-run through lit since. clang-format 17.0.1 clean.

## Scope

The 44 Chess-converted RUN lines are `UNSUPPORTED` without Vitis, so they were not executed here;
the table above is their evidence (`--xchesscc --xbridge` resolves identically before and after).
They need a Vitis runner. The 17 Peano-converted lines do run here.

Callers invoking bare `aiecc` and expecting Chess must now pass `--xchesscc`. A `--chesscc` typo that
left one such test permanently rejected by the argument parser is fixed in a separate one-line commit
in this PR, kept out of the 44-line mechanical sweep so it can be reviewed and reverted independently.


andrej (commented):
Yes to peano default, thanks for this! Could we drop the `--no-xchesscc` and `--no-xbridge` flags altogether then? Would need updating a lot of tests but that could be an easy grep.

atassis:
I wanted to do that in new PR. I have rewritten some tests onto peano, but also they were blocked by https://github.com/Xilinx/llvm-aie/pull/1184, and also some other smaller things, I can't remember exactly which ones. If that's ok to land this and I'll come a bit later with replaced flags in tests and else- would that be ok?

andrej:
Maybe I'm misunderstanding, but doesn't this changeset leave `--no-xchesscc` and `--no-xbridge` completely as no-ops? Or do they still do something?

If they're truly no-ops, I'd prefer removing them in this unit and running a quick search/replace to remove all occurrences of this flag across the codebase.

atassis:
Not quite no-ops -- they were no-ops standing alone, but not in combination, so I checked before sweeping.

`--no-xchesscc` or `--no-xbridge` on their own selected exactly the default, which is every use in the tree. But `--xchesscc --no-xbridge` still meant a Chess compile with a Peano link, and nothing else expressed it: `--xchesscc` alone re-coupled both through the `(xchesscc || xbridge) && !noXchesscc && !noXbridge` clause. That combination is under test in `test/unit_tests/chess_compiler_tests_aie2/00_itsalive`, so a plain search/replace would have dropped it.

`cl::opt<bool>` already takes `--xbridge=false`, so the `no-` flags were redundant syntax rather than redundant capability. Done in 3795dc0: both deleted, each Chess flag now implies the other unless that one is stated too, and the ~228 standalone uses are gone.

It also simplifies `--get-aiesim`. The occurrence counting and the contradiction-vs-omission split only existed because the flags defaulted on; with Peano the default the check is just "both resolved on", so `cpp_aiesim_negated.mlir` collapses to one diagnostic and `cpp_no_xchesscc_implies_no_xbridge.mlir` goes away entirely -- `cpp_peano_default.mlir` already covers the Peano path. Net -96 lines.

I verified the resolution table by compiling `CommandLineOptions.h` into a harness that calls `resolveOptions()` directly, rather than by reading it:

```
<none>                     -> xchesscc=0 xbridge=0
--xchesscc                 -> 1 1
--xbridge                  -> 1 1
--xchesscc --xbridge=false -> 1 0
--xchesscc=false --xbridge -> 0 1
--xbridge=false            -> 0 0
```

and every `--get-aiesim` line in the rewritten test rejects, while `--xchesscc`, `--xbridge` and both together pass.

One wart: this adds `cpp_aiesim_negated.mlir` in an earlier commit and rewrites it here. Happy to squash the series before merge if you'd rather it not read that way.

andrej:
Thanks for doing the sweep -- it's queued now. As for squashing -- no worries, it gets squashed anyways when the PR merges and other PRs (including mine) have extremely bad discipline keeping commits meaningful.

atassis:
20 minutes to merge PR! Noice

hunhoffe:
Very nice!

PR #3506, opened 2026-08-06
ci: bound device dispatch instead of serializing it

atassis:
## Problem

#3499 was mine, and it regressed the device job overall.

It moved device serialization out of a lit parallelism group and into
`run_on_npu.py`. For `test/npu-xrt` that was a loosening: its group was 1, so whole tests ran one at
a time including compile. Locking only the dispatch took `check-aie` from 285s to 224s, which was
the intent.

`programming_examples` had no such group. It ran at full lit width with concurrent dispatch, and the
same lock serialized it for the first time. Measured on the same runner (aie4), same suite set:

| suite | before #3499 | after |
|---|---|---|
| check-aie | 285s | 224s |
| reference-designs | 283s | **697s** |
| programming-guide | 13s | **41s** |
| ATB (already at group=1) | 337s | 337s |

Net **+384s (+42%)** per `aie2p-8col` job. ATB is the control: it was serialized before and after, and
did not move.

## Fix

Replace the exclusive `flock` with a counting semaphore of N lock files, probed non-blocking so a
free slot is taken immediately rather than queueing behind one chosen holder. npu2 gets 8, npu1
gets 1; `AIE_NPU_MAX_CONCURRENT_DISPATCH` overrides it, and 1 reproduces today's behaviour exactly.

The second commit adds the GET_INFO resume transient to the retry list; it is separate so it can be
reviewed or reverted on its own. The third holds npu1 at 1: it caps at 6 hardware contexts against
npu4's 16 and enables AIE2_PREEMPT only on npu4, and at 8 the aie2-4col leg lost 6 tests.

The fourth covers one test a slot cannot serve. `scratchpad_regwrite` reads back, in its second
dispatch, what the first left in a core tile's `params_buf`, and opening any other hardware context
clears that buffer -- the other context need not dispatch. A slot count bounds dispatch, so it
cannot express that. Add a gate file over the semaphore: slot holders take it shared,
`AIE_NPU_EXCLUSIVE_DEVICE=1` takes it exclusive, and that test opts in.

## Evidence

15 real device runs over 5 distinct designs, medians of 9 batches per width:

| N | wall | speedup | jobs | failures |
|---|---|---|---|---|
| 1 | 68.2s | 1.00x | 135 | 0 |
| 2 | 36.2s | 1.89x | 135 | 2 |
| 4 | 24.3s | 2.81x | 135 | 0 |
| **8** | **16.2s** | **4.21x** | 135 | 0 |
| 12 | 14.7s | 4.65x | 135 | 0 |

4 to 8 buys another 50%; 8 to 12 buys 10%. The bound itself is verified rather than assumed: peak
concurrency reconstructed from an enter/exit trace equals N at every width.

Scaling the measured N=8 penalty to reference-designs' ~121 device tests predicts it returns to
roughly 295s. The same model predicts a 432s penalty for N=1, against the 414s actually observed.

The gate, same box: `scratchpad_regwrite` driven through `run_on_npu.py` against three neighbours
churning hardware contexts through the same lock failed 14 of 15 runs as a slot holder, and 0 of 15
exclusive. The mechanism is isolated rather than inferred -- a neighbour that opens a context and
never dispatches is enough to clear the buffer, and a 3s idle gap with no neighbour is not.

## Scope

The two failures at N=2 are a driver race, not a lock problem: on runtime-PM resume the hardware
context restart looks up an already-freed CU BO, and the -EINVAL surfaces as GET_INFO. dmesg
confirms it, correlated to the second. It needs two clients plus an idle window, so it never fires
at N=1 and did not fire at N=8 or N=12 where the device stays busy. Filed separately against the
driver; the retry commit covers it here.

Two limits worth stating. All concurrency measurements are from one Strix box, not a CI runner. And
N=8 does not by itself reduce flake exposure relative to the unbounded state before #3499 -- the
retry is what covers that, not the bound.

With lit at 12 workers, 8 binds. On a runner configured with fewer workers it is an inert ceiling.

The gate inherits flock's fairness, so a steady stream of slot holders can delay an exclusive run.
One test opts in, at roughly 100ms against a 210-280s suite.


atassis:
The retry commit now has a live confirmation, from chasing the same transient down to its cause.

The GET_INFO/-22 abort is a runtime-PM resume race in amdxdna: on resume the driver restarts the hardware contexts, `aie2_config_cu()` re-resolves a userspace GEM handle that the client has already dropped, and the -EINVAL surfaces on whichever ioctl triggered the resume. Reported as amd/xdna-driver#1566, fix proposed as amd/xdna-driver#1567.

Running an unpatched build of that driver under this PR's wrapper, the signature hit twice in 300 two-client runs and **both recovered on the retry**, so the tests still passed:

```
terminate called after throwing an instance of 'xrt_core::system_error'
  what():  DRM_IOCTL_AMDXDNA_GET_INFO IOCTL failed (err=-22): Invalid argument
[run_on_npu.py] Device-enumeration failure on attempt 1/3.
[run_on_npu.py] Retrying in 3s.
...
PASS!
```

That is the intended behaviour for a transient that fails during context setup, before the design runs -- there is no output to compare, so it cannot mask a wrong result, and a deterministic failure still fails all three attempts.

Worth stating plainly: raising the dispatch bound is what makes this reachable at all, since the race needs at least two clients. At one client it cannot fire. So the two commits here belong together -- the second is the cost of the first, and it is bounded by the driver fix landing.

atassis:
https://www.githubstatus.com/incidents/qcvjkzcs7j74
😿

utils/run_on_npu.py
```diff
@@ -145,16 +160,26 @@ def run_command(command: list[str]) -> tuple[int, str]:
...
+#     same host also touches the device.
+# AIE_NPU_MAX_CONCURRENT_DISPATCH overrides the bound; 1 serializes dispatch.
+# AIE_NPU_NO_DEVICE_LOCK=1 opts out entirely, for tests that exercise concurrency.
+#
+# npu1 gets 1: the driver caps it at 6 hardware contexts against npu4's 16 and
+# enables frame-boundary preemption (AIE2_PREEMPT) only on npu4, so it cannot
+# time-share a whole-array design. At 8 the aie2-4col leg lost 6 tests -- five
+# aborted in CREATE_HWCTX/-ENOENT, and writebd_tokens returned all zeros instead
+# of failing. Anything between 1 and 8 is unmeasured there.
+DEFAULT_MAX_CONCURRENT_DISPATCH = {"npu1": 1, "npu2": 8}
```
hunhoffe, line 180: Do you think we could get npu1 up to 3 or so?

atassis: I am not quite sure how this would behave on npu1, that's why I set 1 only. I'ld rather merge it like this first and then test on a hardware (I have none locally)

atassis: Worth flagging why I'm cautious rather than just unsure: at 8 the failure wasn't clean. Five tests aborted in CREATE_HWCTX/-ENOENT, but writebd_tokens returned all zeros and passed. So a green run at 3 wouldn't tell us 3 is safe. CI is the only npu1 I can reach, so I'd want that experiment to check outputs, not exit codes.

hunhoffe: Makes sense with me to merge as-is and then maybe try bumping in a follow on!

hunhoffe (approved):

atassis:
I have made a fix for flaky CI, can we run it again?

hunhoffe (approved):

PR #3507, opened 2026-08-06
[AIE][AIEX] Do not abort on an unplaced tile in the DMA passes

atassis:
Found while following up your `getStaticStrides()` question on #3463: the same
sweep turned this up nearby, and it is unrelated to that BD sizes/strides
family, so it is its own PR.

### Problem

`TileElement::getTileOp()` reports a fatal error when the tile is an
`aie.logical_tile`. Two passes call it unguarded, so `aie-opt` aborts instead of
diagnosing, on well-formed input:

```
$ aie-opt --aie-dma-tasks-to-npu unplaced.mlir
LLVM ERROR: Calling getTileOp requires TileOp.
```

(`unplaced.mlir` is an `aiex.dma_configure_task` on an
`aie.logical_tile<ShimNOCTile>(?, ?)`; `--aie-decompose-large-dma-bd` aborts the
same way.) `aiecc` always places before either lowering pipeline, so this is
reachable through `aie-opt`'s per-pass flags rather than a full compile: the
cost is a crash instead of an error, not bad output.

The tree already avoids the call twice without naming why.
`DMAConfigureTaskForOp::verify()` hand-rolls the safe form under "Do not call
`allocOp.getTileOp()`: it hard-asserts when the allocation is still bound to an
unplaced (logical) tile", and `resolveTaskAndTile` guards with
`if (!tile) return std::nullopt;` after a call that cannot return null.

### Fix

`tryGetTileOp()` beside `getTileOp()`, null when the tile is still logical.
`getTileOp()` delegates to it, so a placed tile behaves exactly as before.

The passes then diverge by what they need: `aie-dma-tasks-to-npu` emits absolute
column/row and so reports an error; `aie-decompose-large-dma-bd` is a shape
rewrite and declines, running again after placement.

`aiex.core_reset` and `aiex.dma_channel_reset` reach `getTileOp()` the same way,
but their verifiers reject a tile operand not produced by an `aie.tile`, so the
call is unreachable there. Checked rather than assumed.

### Test

`unplaced-tile-bad.mlir` (the diagnostic) and
`decompose_large_dma_unplaced_tile.mlir` (the design left untouched); both abort
without this change. Substituting `aie.tile(0, 0)` gives byte-identical output
before and after, and `check-aie` is unchanged against `main`.

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEInterfaces.td
+++ b/include/aie/Dialect/AIE/IR/AIEInterfaces.td
@@ -134,7 +134,11 @@ def TileElement : OpInterface<"TileElement", [
   ];
 
   let extraClassDeclaration = [{
+    // Requires a placed tile: fatally errors if the tile is still logical.
     TileOp getTileOp();
+    // Null when the tile is still logical, so callers that run before
+    // placement can defer instead of aborting.
+    TileOp tryGetTileOp();
   }];
 
   let extraTraitClassDeclaration = [{
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -1297,6 +1297,9 @@ def AIE_DMAConfigureTaskOp : AIEX_Op<"dma_configure_task", [HasAncestor<"AIE::Ru
 
   let extraClassDeclaration = [{
     AIE::TileOp getTileOp();
+    // Null when the tile is still an aie.logical_tile, so passes that can run
+    // before placement defer instead of aborting inside getTileOp().
+    AIE::TileOp tryGetTileOp();
     std::optional<uint32_t> getFirstBdId();
     // The repeat count as an OpFoldResult: the runtime operand if present, else
     // the compile-time attribute. Mirrors dma_bd's offset/len duality.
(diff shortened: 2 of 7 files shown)


hunhoffe (commented):
Nice fix! 

For some history, I believe we aborted because there were too many instances of segfaults in the compiler after logical tiles were introduced. It seemed better to abort cleanly (and early) than potentially let things progress further. And I do think that there are a few other errors of this type related to logical tiles.

From my review agent ---

As a follow on, I think there's one more site in the same shape, though, in the same file this PR touches: AIEDecomposeLargeDmaBd.cpp has two more unguarded getTileOp() calls this PR doesn't touch — the DMAConfigureTaskForOp branch in resolveTaskAndTile, and the direct call in DecomposeLargeDmaBdPattern::matchAndRewrite — both on a ShimDMAAllocationOp:

```
  TileOp ShimDMAAllocationOp::getTileOp() {
    return cast<TileOp>(getTile().getDefiningOp());
  }
```

This one doesn't go through TileElement at all, so it isn't covered by the tryGetTileOp() split — it's a hard cast<>, which is UB in a release build (not even a clean abort) if the tile is still a logical_tile. And ShimDMAAllocationOp::verify() looks like it tolerates an unplaced tile the same way DMAConfigureTaskOp did before this fix — it only checks TileLike + shim-tile-ness, and the error path already accounts for tryGetCol()/tryGetRow() returning nullopt:
```
  LogicalResult ShimDMAAllocationOp::verify() {
    TileLike tileLike = llvm::dyn_cast<TileLike>(getTile().getDefiningOp());
    if (!tileLike) return emitOpError(...);
    if (!tileLike.isShimNOCorPLTile()) { ... }
    return success();
  }
```

That's different from core_reset/dma_channel_reset, whose verifiers you confirmed reject a non-aie.tile operand outright. So a hand-written aie.shim_dma_allocation(%logical_tile, ...) module looks verifier-legal, and running it through aie-opt --aie-decompose-large-dma-bd looks like it'd hit the same class of bug this PR fixes, just via DMAConfigureTaskForOp/a raw NpuDmaMemcpyNdOp rather than DMAConfigureTaskOp — neither of the two new tests exercise that shape.
  
I haven't actually reproduced this (worth doing before treating it as certain), but given how close it sits to what you already swept, could you either check it the same way you did the reset ops, or extend the guard to ShimDMAAllocationOp::getTileOp() too?

PR #3512, opened 2026-08-07
[AIE] wire pad_value to hardware and move it to the DMA channel

jskimko:
## Description

MemTile DMAs can insert constant padding at the buffer-descriptor level, and the IR already had a `pad_value` attribute on `aie.dma_bd` to describe the fill value. But that attribute was **never consumed** -- nothing in lowering or the runtime read it, so a non-zero `pad_value` was silently dropped and padded regions always came out as zeros. It was also on the wrong op: the constant value is a per-channel property in hardware (one `CONSTANT_PAD_VALUE` register per MM2S channel), not a per-BD one.

This PR wires the feature to hardware for the first time, and in doing so moves the value to where the hardware puts it and threads it through every abstraction level.

### Op / dialect                                                                                                
Drops the dead `pad_value` from `aie.dma_bd` and adds it to the channel ops `aie.dma_start` and `aie.dma` (default `0`). The pad *geometry* stays per-BD on `aie.dma_bd`; only the fill value moves. Verifiers reject a non-zero pad value outside a MemTile DMA, and `aie.objectfifo` rejects `padValue` without `padDimensions`. The objectfifo attribute is spelled `padValue` to match the op's camelCase convention.

Before (value on the BD, but a no-op):

```mlir
aie.dma_bd(%buf : memref<256xi32> offset = 0 len = 256
           pad [<const_pad_before = 1, const_pad_after = 1>] pad_value = 7)
```

After (geometry on the BD, value on the channel):

```mlir
aie.dma_start("MM2S", 0, ^bd0, ^end) {pad_value = 7 : i32}
  ...
aie.dma_bd(%buf : memref<256xi32> offset = 0 len = 256
           pad [<const_pad_before = 1, const_pad_after = 1>])
```

Region form and objectfifo:

```mlir
%0 = aie.dma(MM2S, 0) {pad_value = 5 : i32} [ ... ]

aie.objectfifo @of(%mem dimensionsToStream [...], {%shim}, 2 : i32)
  {padDimensions = #aie<bd_pad_layout_array[...]>, padValue = 7 : i32}
  : !aie.objectfifo<memref<64xi8>>
```

### Lowering / runtime

`AIEObjectFifoStatefulTransform` and `AIELowerMemcpy` carry the value onto the channel op; `AIERTControl::pushToBdQueueAndEnable` now emits `XAie_DmaSetPadValue` for MemTile MM2S channels with a non-zero value -- the step that was missing before.

### IRON API

`ObjectFifo`, `ObjectFifo.forward()` / `split()`, and the `tile_dma` path accept a per-*element* `pad_value` and pack it into the raw 32-bit stream word using the fifo/BD element width (new `pack_pad_value` helper), so callers write `pad_value=7` for an `i8` region instead of hand-packing `0x07070707`. Sub-32-bit types are replicated across the word; types wider than 32 bits are rejected. On the `tile_dma` path the value is validated against the channel's padded BDs (must have a padded region; all padded BDs must share an element size).

ObjectFifo (forward pads a memtile-staged output):

```python
of_out = of_in.cons().forward(
    placement,
    pad_dimensions=[(pad_before, pad_after)],
    pad_value=42,          # per element; packed by dtype width
)
```

tile_dma (geometry per-BD, fill value per-channel):

```python
DmaChannel(
    direction=DMAChannelDir.MM2S,
    channel=0,
    pad_value=7,           # i8 buffer -> 0x07070707 in the register
    bds=[
        Bd(
            buffer=buf,
            pad_dimensions=[(0, 1), (0, 0)],
        ),
    ],
)
```

## Checklist                                                                                                    


include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -449,6 +449,10 @@ class AIETargetModel {
   // Returns true if the target model supports the given block format.
   virtual bool isSupportedBlockFormat(std::string const &format) const;
 
+  // Returns true if the target supports a constant pad value on MemTile MM2S
+  // DMA channels (register=CONSTANT_PAD_VALUE).
+  virtual bool isMemTilePadValueSupported() const { return false; }
```
jskimko, line 454: Thoughts on this placement?

```
  AIETargetModel                         virtual … = false   ← base default
  ├── AIE1TargetModel                                        → false
  │   └── VC1902TargetModel              (VCK190)            → false
  └── AIE2TargetModel                                        → false
      ├── VE2302TargetModel              (Versal VE2302)     → false
      ├── VE2802TargetModel              (Versal VE2802)     → false
      ├── BaseNPU1TargetModel            (Phoenix / aie2)    → false
      │   └── VirtualizedNPU1TargetModel (npu1_1col…4col)    → false
      └── BaseNPU2TargetModel            (Strix / aie2p)  override = true
          ├── NPU2TargetModel            (npu2, full 8-col)  → true
          └── VirtualizedNPU2TargetModel (npu2_1col…7col)    → true
```

python/dialects/aie.py
```diff
@@ -919,6 +923,7 @@ def dma_start(
     dest: Successor | Block | ContextManagedBlock | None = None,
     chain: Successor | Block | ContextManagedBlock | None = None,
     repeat_count: int = 0,
+    pad_value: int = 0,
```
hunhoffe, line 928: should pad value be none by default here too?

python/iron/dataflow/objectfifo.py
```diff
@@ -1080,6 +1119,8 @@ def forward(
             dims_from_stream=dims_from_stream_arg,
             plio=plio,
             repeat_counts=[repeat_count] if repeat_count is not None else None,
+            pad_dimensions=[pad_dimensions] if pad_dimensions is not None else None,
+            pad_value=[pad_value] if pad_value else None,
```
hunhoffe, line 1123: is 0 not a valid pad value? because, IIRC "if pad_value" will evaluate to False if 0

hunhoffe (commented):
Nice work! Do we need to add documentation of this feature in a programming example or the programming guide?

python/dialects/aie.py
jskimko: My reasoning is that `dma_start` faces the user and converts `0` to `None` for `DMAStartOp` consumption. `int = 0` matches the register expectation (not enabled _is_ value `0`) and mirrors `repeat_count` handling; when `DMAStartOp` receives `None`, it will omit the attr from the IR.

python/iron/dataflow/objectfifo.py
jskimko: Yes, `pad_value=0` takes the else/None branch, but aie-rt only writes `CONSTANT_PAD_VALUE` when it's non-zero, so they produce identical output.

The alternative `if pad_value is not None` would break: `pad_value` defaults to `0` (from above), so `is not None` is always true and the downstream `.itemsize` runs for every fifo, raising `TypeError` on non-dtype elements like `v8bfp16ebs8`.

jskimko:
@hunhoffe Added `programming_examples/basic/dma_padding`.

hunhoffe (approved):

PR #3513, opened 2026-08-07
Pin clang-tidy, make it blocking, scope to one file

hunhoffe:
## Update

Ran the pilot's ruleset across all 102 production `.cpp` files (not just the one enabled file) to see what a full rollout would surface, then researched what newer clang-tidy versions offer. That turned up two problems with the original config and an opportunity worth taking now:

**Dropped `misc-include-cleaner`.** It accounted for ~94% of warning volume in the full-repo test run. LLVM's own `.clang-tidy` disables this exact check for the same reason we'd hit here — it doesn't handle forward-declaration-heavy, template-heavy code well, and its results vary with CMake configuration. Not tunable, only droppable.

**Narrowed `clang-analyzer-*` to four suites.** The full wildcard fires inside MLIR/LLVM's own header boilerplate (dialect registration lambdas, `APInt`/`SmallBitVector` internals) rather than on code we own. Chromium hand-restricts to `core`/`cplusplus`/`unix`/`deadcode` for the same reason ([docs/clang_static_analyzer.md](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/clang_static_analyzer.md)); we do the same instead of the full wildcard. This still catches real things — e.g. `clang-analyzer-unix.StdCLibraryFunctions` flagged a `dup2()` call in `aiecc.cpp` with a potentially-unchecked `-1` fd argument.
  
**Bumped clang-tidy 17.0.1 → 20.1.0** (and clang-format along with it, since this repo deliberately pins them to one version). This unlocks several `bugprone-*` checks that don't exist in 17.0.1 at all and are aimed at real bug classes rather than style: `bugprone-bitwise-pointer-cast`, `bugprone-incorrect-enable-shared-from-this`, `bugprone-optional-value-conversion`, `bugprone-pointer-arithmetic-on-polymorphic-object`, `bugprone-return-const-ref-from-parameter`, `bugprone-suspicious-stringview-data-usage`, `bugprone-unused-local-non-trivial-variable`. Verified clang-format 20.1.0 produces zero new diffs in `lib/`, `include/`, `tools/` (the pilot's enforcement path), so the bump doesn't force any reformatting.
  
**Re-enabled `bugprone-unchecked-optional-access`.** It was disabled because it hung 180s+ on `AIEAssignBufferDescriptorIDs.cpp` under 17.0.1 — a known long-standing LLVM bug ([llvm/llvm-project#55530](https://github.com/llvm/llvm-project/issues/55530)) also cited by PyTorch and Arch Linux for disabling it. Under 20.1.0 it completes in ~7s on that same file. Fixed the 3 findings it surfaces there.

**Fixed the file currently in scope** (`AIEAssignBufferDescriptorIDs.cpp`) for all newly-enabled findings: collapsed three `has_value()`/`.value()` pairs into a single cached optional (the repeated calls were why the analyzer couldn't correlate the guard with the access — also just better code), and changed two `setBdId(*nextId)` calls to `setBdId(nextId)` since the generated setter already accepts `std::optional<int32_t>` directly.

## Original Description

clang-tidy's CI step only diffed the PR and uploaded an artifact nobody looked at, and its version silently floated with whatever apt install clang-tidy resolved to. This pins clang-tidy to 17.0.1 (matching clang-format's pin) via requirements_dev.lock, enables clang-analyzer-*, and makes both CI and a new pre-commit hook fail on real findings — scoped for now to AIEAssignBufferDescriptorIDs.cpp, meant to grow one file at a time like ruff/pyright coverage already does.
- .clang-tidy: dedupe, add clang-analyzer-*
- Fixed all findings in the one enabled file (naming, missing includes)
- CI: pinned install, diff-scoped to enabled files, blocks on failure
- New clang-tidy pre-commit hook (utils/run_clang_tidy.sh)
- Docs + PR template updated

-----------

## Checklist


hunhoffe:
Closing in favor of https://github.com/Xilinx/mlir-aie/pull/3518

PR #3514, opened 2026-08-07
[AIE] fail on incomplete packet-flow routing

jskimko:
## Description

The congestion-aware pathfinder (`--aie-create-pathfinder-flows`) can produce a routing solution whose per-tile switch settings do not form a complete source -> dest path for every declared packet source. This shows up with high-fan-in merges when the destination sits at a column edge. In `AIEPathfinderPass::runOnPacketFlow` the "reject false broadcast" `findPathToDest()` guard then drops the source's own connection, so no source connect is emitted and the sender is silently omitted -- the design compiles cleanly but under-delivers on device with no diagnostic.

This PR tracks whether each declared packet source keeps a surviving connection to its destination and emits an op error when it does not, turning a silent miscompile into a compile-time diagnostic that names the source and destination that could not be connected.

Adds a regression test `test/create-packet-flows/incomplete_route_edge_merge.mlir` that must now fail loudly. The underlying limitation -- that some column-edge high-fan-in placements cannot be routed at all -- is separate and remains open.

## Checklist

### DIFF
--- a/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
@@ -538,6 +538,8 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
         if (analyzer.processedFlows[srcPoint])
           continue;
         SwitchSettings settings = analyzer.flowSolutions[srcPoint];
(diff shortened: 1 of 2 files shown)


hunhoffe (approved):
This makes sense to me! For the underlying problem -- that this pattern cannot/is not routed correctly -- is that something that should be tracked in an issue?

PR #3516, opened 2026-08-08
[AIE] Reject packet rules that misroute a live packet id (fixes #437)

### LINKED ISSUE #437: AIEFindFlows can report a packet flow exists even when the packet rules route the flow incorrectly
In the AIEFindFlows pass, packet flows are found by tracing from the destination back to the source of the flow. If the packets originating from the source are incorrectly routed, AIEFindFlows can still report that the flow exists. This can happen if the packet rules "overlap" such that a particular packet ID should go on the correct route, but instead is routed differently by a higher priority packet rule. For example: 

%35 = AIE.switchbox(%34) {
    %40 = AIE.amsel<0> (0)
    %41 = AIE.amsel<0> (1)
    %42 = AIE.masterset(South : 0, %41)
    %43 = AIE.masterset(West : 0, %40)
    AIE.packetrules(DMA : 0) {
      AIE.rule(26, 10, %40)
      AIE.rule(24, 8, %41)
    }
  }

In this example, packet flow ID 14 matches both rules! It is intended to be routed south by the 2nd rule, but in fact it will be routed West because the first rule has top priority. This rule overlap should be investigated and avoided. It is possible that this issue was caused by the use of Pathfinder for packetflows.

The dangerous consequence of this issue is that AIEFindFlows can report that a packetflow exists when in fact it will be routed incorrectly. Tests that rely on AIEFindFlows to determine correctness of packetflows can therefore pass even when the routing is incorrect.

hunhoffe:

## Description

Closes #437.                                                                                                                                                                                                       
                                                                                                                                                                                                                     
A stream switch matches a slave port's `aie.rule` entries **in order and routes on the first hit**. Two rules that both match a packet id therefore send it to whichever comes first, and the later rule never sees it. Nothing rejected that, and `-aie-find-flows` — which traces backwards from destination to source — evaluated each rule independently, so it reported the shadowed route as a live flow. That is the report in #437, open since 2023.                                                                                                                                                                 
                                                                                                                                                                                                                     
### It is real on hardware                                                                                                                                                                                         
                                                                                                                                                                                                                     
`test/npu-xrt/packet_flow_overlapping_rules/` pins the issue's rule pair into an otherwise router-generated single-column design. On NPU Phoenix (npu1):                                                                                                                                            

```                                                                                                                                                                                                                
buf_a (packet id 10, not checked): 200 201 202 203 204 205 206 207                                                                                                                                                 
buf_b (packet id 14): -1 -1 -1 -1 -1 -1 -1 -1                                                                                                                                                                      
```                                                                                                                                                                                                                
                                                                                                                                                                                                                     
`buf_a` holding `200..207` is id 14's payload arriving at id 10's destination. Substituting the exact masks the router would have emitted makes the same design pass, so the masks are the only variable.

### Overlap alone is not the bug                                                                                                                                                                                   
                                                                                                                                                                                                                     
The first attempt at this was an op verifier rejecting any cross-route overlap. That is wrong, and fuzzing showed it: 6 of 800 random designs were rejected, every one an overlap on an id **no flow uses**. A slave port has four rule slots, so `computeSubcubeCover` (#3429) relaxes masks to merge ids into one rule, and a relaxed rule necessarily claims ids nothing sends. For live ids `{10,22} -> North:1` and `{16,19} -> North:4` the router emits                                                                                                                                                     
                                                                                                                                                                                                                     
```mlir                                                                                                                                                                                                            
aie.rule(3, 2, %up)     // ids {2,6,10,14,18,22,26,30}                                                                                                                                                             
aie.rule(28, 16, %mem)  // ids {16,17,18,19}                                                                                                                                                                       
```                                                                                                                                                                                                                
which share only 18. Rejecting that would break designs that currently fit.                                                                                                                                        
                                                                                                                                                                                                                     
Whether an overlap misroutes depends on which ids the device actually produces, which is not knowable from one op. So the check is a **DeviceOp pass**, `-aie-verify-packet-rules`, that collects the ids the device injects and reports only an overlap on one of those. It lives in AIEX because a hand-routed design names its ids in the runtime sequence, and runs in aiecc's routing pipeline so it covers hand-written switchboxes too.                                                                     

Re-fuzzed: 0 false positives across 1300 designs, including 700 with ids deliberately reused across four source ports.   

### Also here                                                                                            
 
- **`-aie-find-flows` honours rule priority.** Packet ids reaching a point are tracked as an explicit set rather than a `(mask, value)` cube, because first-match priority leaves a rule carrying the ids no earlier rule took, which is not in general a cube.
- **A crash fix.** Of three `getNewUniqueAmselPerArbiterID` call sites, the partial-arbiter-match one did not check `INVALID_AMSEL_VALUE`; on msel exhaustion a null `AMSelOp` reached `getResult()`. Found by fuzzing.
- **First hardware coverage of the mask minimiser.** Every existing on-device packet test uses exact `mask = 31`. `test/npu-xrt/packet_flow_subcube_masks/` drives a genuine two-cube cover, each id carrying a distinguishable payload. Falsified: diverting one rule in the routed IR gives `packet id 3 slot 0: got 1400 want 1300`.

 ### Coverage                                                                                             

Measured with gcov over `find-flows`, `create-packet-flows`, `create-flows`, `objectFifo-stateful-transform`:

  | file | before | after |                                                                                
  | --- | --- | --- |                                                                                      
  | `AIECreatePathFindFlows.cpp` | 91.6% | 93.0% |                                                         
  | `AIEVerifyPacketRules.cpp` | — | 95.1% |

New negative tests pin three diagnostics that had none: packet id above the target maximum, the illegal stream-switch connection a shim fan-in can produce, and msel exhaustion.                                                                                     

### Known gaps                                                                       
                                                                                                           
- `AIECreatePathFindFlows.cpp:1158-1162` (`can lead to false packet id match`) appears unreachable: every rule in a `PacketRulesOp` comes from `computeSubcubeCover` in the same pass, and covers on one port avoid each other's ids by construction. Left in place.
- The `port assigned to multiple arbiters` diagnostic and the arbiter-reuse branches around 819-853 are still uncovered; ~4000 fuzzed designs did not reach them.                                                                                                  
- The live-id set is device-wide, which over-approximates per-port arrival. Safe in the sound direction and clean across the fuzzing above.

### Behaviour change

A rule with `mask == 0` now yields `aie.packet_flow(0)` rather than `aie.flow`. The old code keyed "circuit-switched" off a zero mask, which such a rule also has. This affects `--aie-flows-to-json` output for those designs.                                        

### Testing                                                                                              
                                                                                                           
Full `test/npu-xrt` on NPU Phoenix (npu1): 81 passed, 0 failed. Scoped lit (`find-flows`, `create-packet-flows`, `create-flows`, `objectFifo-stateful-transform`): 214 passed, 1 pre-existing XFAIL.


lib/Dialect/AIEX/Transforms/AIEVerifyPacketRules.cpp
```diff
@@ -0,0 +1,166 @@
...
+// The check is deliberately restricted to *live* ids -- those some op in the
+// device injects. Overlap on an id nothing produces is both normal and wanted:
+// a slave port has only getNumSlaveSlots() slots, so the router relaxes masks
+// to merge ids into one rule, and a relaxed rule claims ids no flow uses. Only
+// ids that can arrive make an overlap observable.
+//
+// The live set is device-wide, which over-approximates per-port arrival: an id
+// live only on some disjoint port still counts here. That direction is safe for
+// the router, which never leaves a live id ambiguous on the port it arrives at,
+// and it keeps the pass from having to re-derive routing.
```
andrej, line 22: Is this pass going to reject a design that relies on this?

lib/Dialect/AIEX/Transforms/AIEVerifyPacketRules.cpp
```diff
@@ -0,0 +1,166 @@
...
+//
+// This lives in AIEX because a hand-routed design names its ids only in the
+// runtime sequence (aiex.npu.dma_memcpy_nd / aiex.npu.writebd), which the AIE
+// dialect cannot see. Run it after routing and before those ops are lowered.
```
andrej, line 26: The "only" here isn't fully right and also contradicts the code below: DMAs can also prepend packet IDs onto their packets with `aie.dma_bd_packet`, which the code _does_ appear to also consider as a source of the live ID set.

lib/Dialect/AIEX/Transforms/AIEVerifyPacketRules.cpp
```diff
@@ -0,0 +1,166 @@
...
+    });
+    return live;
+  }
+
+  // Where an amsel actually sends a packet: the master ports of every
+  // masterset naming an amsel with the same (arbiter, msel), since that pair is
+  // all the switch matches on. Two rules whose destinations agree route alike
+  // and may overlap freely -- one masterset can list several amsels, and
+  // distinct AMSelOps can spell the same pair.
+  static SmallVector<Port, 4> destPorts(PacketRuleOp rule) {
```
andrej, line 94: Is it possible we already have reusable infrastructure for this in `AIEFindFlows.cpp`? I'd assume that pass also has to reason about exactly this.

tools/aiecc/IRTransforms.h
```diff
@@ -887,6 +887,10 @@ getRoutingPipeline(mlir::MLIRContext *ctx) {
   auto pm = std::make_unique<mlir::PassManager>(ctx);
   pm->nest<xilinx::AIE::DeviceOp>().addPass(
       xilinx::AIE::createAIEPathfinderPass());
+  // Runs on both routed-here and hand-written switchboxes, and while the
+  // runtime sequence still names its packet ids.
+  pm->nest<xilinx::AIE::DeviceOp>().addPass(
+      xilinx::AIEX::createAIEVerifyPacketRulesPass());
```
andrej, line 893: Since the verifier does over-approximate the live set, can we add a "I know what I'm doing" flag to aiecc to disable the pass and point users to it in the error message?

andrej (commented):
This is a useful sanity check for users who do manual routing, which is maybe somewhat niche but probably very useful for them. That being said, this seems to be the class of user that also is pretty "in the weeds" and knows what they're doing. 

As I understand currently, the verifier is both over-conservative and under-approximating. 

- over-conservative (the comments in the code point this out already): It assumes any packet ID anywhere in the design can arrive at any switchbox. In reality, there might be multiple disjoint partitions that never share any links/ports/switchboxes and therefore could collide with IDs from other partitions without ever causing an issue.
- under-approximating: packet IDs might come from sources others than the ones listed in the code. For example, the FastFlowLM Gemma4 decode operator writes packet IDs in the kernel core. Theoretically, these could be arbitrary dynamically computed values, so determining which IDs might be live at any time is intractable.

I think the first limitation (over-conservativeness) is more limiting, though I don't know how commonly it would cause issues -- maybe in a congested scenario it would matter? I think making the test less conservative _might_ be doable by recovering the flows (maybe reusing the AIEFindFlows infrastructure or incorporating this test in there altogether), keeping track of live IDs per-flow rather than globally.

The second case (under-approximating) is probably a little more adversarial and might not be as common (though we do see a simple case of it in the FastFlowLM design). It is not generally solvable, but then again under-approximating will just mean the error is not emitted in cases where it should be, so it's not a blocker per se.

So to conclude all my ramblings, because of all that, I think I'd like to see this whole pass gated behind a flag in aiecc so it can be disabled for the users that know what they are doing (which, if you are programming at the switchbox level, you probably are), or for it to only be a warning rather than a hard error.

PR #3517, opened 2026-08-08
[AIE] Fix pathfinder routing through non-physical crossbar hops (#2689)

### LINKED ISSUE #2689: Connectivity graph leads to invalid connections that are flagged as an error
There's an underlying incorrectness in the way connectivity is modeled that leads to invalid errors. But these errors may be fine since there should be an error but just not the error that was flagged.

Connectivity is currently specified such that all bundles/channels for a switchbox are single nodes in a graph. That means when we define the valid connectivity such as in `AIE2TargetModel::isLegalTileConnection` in the `lib/Dialect/AIE/IR/AIETargetModel.cpp`, it treats each port in the switchbox as a single node in a graph. This works ok when connections come from a neighbor to that port, and then that port is routed to another port such as:

tile(0,3) east0 -> tile(1,3) west0 -> tile(1,3) north1 -> tile(1,4) south1

Here, tile(1,3)'s west0 port is a slave that then connects to the north1 port as a master.

However, an alternate connection is possible, such as:
tile(0,3) east0 -> tile(1,3) west0 -> tile(1,3) south3 -> tile(1,3) north1 -> tile(1,4) south1

Here, south3 is a port that is treated as a single node. As such, a graph could be made as above since there is a valid connection between west0 -> south3 and south3 -> north1. However, what is not indicated is that south3 here is a master and a slave and those two roles are distinct ports in the hardware. In order for the above connection to be valid, we would also need the following:
tile(1,2) north3 -> tile(1,2) north3

It is unlikely the current software search algorithm recognizes the need for this and therefore would not create this extra connection. However, the reason this does not come up as an issue is because the Djikstra's algorithm used in pathfinder will never choose this option since it's always more expensive. So in reality, we would never encounter it even if the current structure of the graph allows it.

Where a valid error does come up is with the definition of dma as a single node in the graph. For something like trace which only has valid connection to a limited set of ports (only south direction, not east, west or north), if I wanted to connect trace to the east port, it can accomplish this in the following way:
trace0 -> dma0 -> east0

This is because trace0 cannot route to the east0 port but it can route to dma0. That is specifically dma s2mm0. Likewise, dma0 can route to east0 but that dma is actually mm2s0. Because it doesn't distinguish between s2mm0 and mm2s0, it treats this as a single node, much like how master and slave for the south0 port is treated as the same thing.

This connection is normally never chosen because it should be cheaper to route south rather than east due to this extra hop or the use of the dma0 node, but if all north-south connections are used up, then it will try to go east and it will think it can do this via this erroneous hop to the dma0 node in the graph. The correct error should be that the route is not possible, but instead, an error will be flagged where dma0 is used in two connections (the main dma0 connection that used up the north-south stream connections, and the trace0->dma0->east0 connection). It's unclear if the scenario could occur where it chooses this connection when no other dma0 use is present and thereby allowing it, but at the moment, that doesn't seem likely. As such, while this is a fundamental flaw in the modeling of the connectivity graph, it is treated as a "do no fix" because it will either never be chosen (as in the first example of the use of the south0 port) or an error will be asserted to prevent it from being true, even if that error doesn't accurately represent the "real" error.

hunhoffe:
## Summary
  
Fixes #2689. 

The routing pathfinder modeled each `(tile, port)` as a single graph node regardless of whether it was playing the crossbar's *arriving* or *departing* role. Those are physically distinct wires (e.g. a DMA channel's S2MM-write vs MM2S-read sharing a channel index, or a cardinal direction's incoming vs outgoing link), so the router could chain two intraconnect hops at the same tile with no real wire between them — e.g. compiling `aie.flow(%tile, Trace : 0, %other_tile, DMA : 0)` into `connect<Trace:0, DMA:0>` + `connect<DMA:0, East:2>`, silently dropping the data instead of routing it, with no compile error.

The fix splits each dense routing-graph node into an explicit src-role (signal entering the crossbar) and dst-role (signal leaving it) node. Intraconnect edges now only go src→dst within a tile, and wires only go dst(tile A)→src(tile B), so every hop the router selects corresponds to a real physical connection. This is confined to `AIEPathFinder`'s internal graph construction — no IR/dialect changes.

Also fixes a latent self-loop bug this change exposed in `findPaths()` (`route to self` never had a `continue`, relying on the now-false assumption that a flow's source and destination map to the same dense node).

## Testing

- New regression test for the exact #2689 repro (`test/create-flows/trace_dma_passthrough_invalid.mlir`).
- 5 existing tests updated for legitimate tie-break shifts in channel numbers / path lengths; each new route manually verified to still be a real, fully physical path (no invalid pass-throughs).
- Full `create-flows`/`create-packet-flows`/`find-flows`/`dialect/AIE` suite: 191/191 passing.


include/aie/Dialect/AIE/Transforms/AIEPathFinder.h
```diff
@@ -248,10 +248,28 @@ class Pathfinder : public Router {
...
+  // -> arriving (tile B). Collapsing both roles into one node let the
+  // pathfinder chain two intraconnect hops at the same tile with no real
+  // wire in between (issue #2689).
+  struct RoledEndPoint {
+    PathEndPoint pep;
+    bool isSrcRole;
+    bool operator<(const RoledEndPoint &rhs) const {
+      return std::tie(pep, isSrcRole) < std::tie(rhs.pep, rhs.isSrcRole);
+    }
+  };
```
andrej, line 267: Why not just update `PathEndPoint` to distinguish source from dest endpoints instead of this new struct? I'd assume any user should care. 

(also nit: verbose comment documenting code history, i.m.o. "source and destination ports are distinct wires; treat them as separate ports" would be a sufficient comment here, maybe I'm missing nuance)

PR #3518, opened 2026-08-08
clang-tidy

hunhoffe:
## Description

Rolls out clang-tidy as a blocking CI check across all 102 production C++ files (lib/, tools/), building on the single-file pilot. Bumps clang-tidy/clang-format to 20.1.0 (from 17.0.1) and adds local enforcement via a new pre-commit hook (utils/run_clang_tidy.sh).

### Check-list changes (see .clang-tidy for rationale)
- Re-enables bugprone-unchecked-optional-access, disabled since clang-tidy 17 hung 180s+ on AIEAssignBufferDescriptorIDs.cpp (llvm/llvm-project#55530); completes in ~7s under 20.1.0.
- Scopes clang-analyzer-* to four suites (following Chromium's precedent) instead of the full wildcard, and disables four sub-checks that fire on MLIR/LLVM's own internals via template instantiation, not our code.
- Excludes misc-include-cleaner (matches upstream LLVM's own .clang-tidy) and disables readability-identifier-naming repo-wide for now — it surfaced 614 pre-existing naming violations that aren't safely batch-fixable; tracked as separate follow-up work.

### Real bugs found and fixed
- FoldMulAddChainToConvOp.cpp: self-comparison bug (convMac->rhs == convMac->rhs, always true) that should have compared against convMac2, silently breaking the rhs-mismatch rejection.
- NpuDmaMemcpyNdOp::getArrayAttrMaxRanks(): declared static in AIEX.td, but OffsetSizeAndStrideOpInterface's default trampoline invokes it via cast<ConcreteOp>(op).method().
- tools/aiecc/Utils.h (CaptureStdio): unconditionally dup2'd savedErr even when its dup() had failed, unlike the already-guarded savedOut.
- tools/aiecc/aiecc.cpp: aiesim.sh's chmod failure was silently swallowed; now logged.

### Other changes
- Systematic migration from deprecated builder.create<OpTy>(...) to OpTy::create(builder, ...).
- Fixed a pkg_resources crash under modern setuptools that broke clang-tidy's own invocation.

## Checklist


.github/workflows/lintAndFormat.yml
```diff
@@ -95,16 +98,40 @@ jobs:
...
-          # aie-headers only aggregates the dialect/op/type tablegen targets;
-          # it does not pull in the Conversion pass enum header or the vendored
-          # xaiengine/*.h headers copied from third_party/aie-rt. Without these,
-          # clang-tidy fails to compile any touched file that includes
-          # aie/Conversion/Passes.h (e.g. lib/Conversion/**) or xaiengine/*.h
-          # (e.g. lib/Targets/AIERT.cpp) with a file-not-found error instead of
-          # a real lint finding.
+          # Pass/op tablegen targets not covered by aie-headers/mlir-headers --
+          # needed so clang-tidy (below) can actually parse pass source files
+          # instead of erroring on a missing generated header. Extend this
```
andrej:
```suggestion
          # needed so clang-tidy (below) can parse pass source files
          # that include generated headers. Extend this
```

andrej (approved):
Only did a superficial pass, seems good!

I think the idea is to incrementally increase the set of checked files across multiple PRs and then eventually just check all?

hunhoffe:
@andre Thanks for taking a look! I think this checks pretty much all of the core cpp files. Did you find anything it's missing?

PR #3519, opened 2026-08-08
AIE code creator skill

ypapadop-amd:
## Description

This is a skill for creating IRON kernels/designs from existing code. It has been tested with creating llama.cpp ops from C/C++ code.

## Checklist


skills/aie-code-creator/evals/evals.json
```diff
@@ -0,0 +1,128 @@
+{
+  "skill_name": "aie-code-creator",
+  "evals": [
+    {
+      "id": 0,
+      "name": "eltwise-scale-relu-multicore",
+      "prompt": "Write an IRON design that scales a bfloat16 vector of 8192 elements by 2.0 and applies ReLU, split across 4 cores on NPU2",
+      "expected_output": "A Python IRON design file and a C++ kernel file. The Python design must use the mlir-aie 1.4.0 Runtime/Program API: Runtime(seq_fn, fn_args) with fill/drain called as methods on ObjectFifoHandle objects obtained from .prod()/.cons(), and Program(device, rt, workers=[...]) (not the old Runtime()/rt.sequence()/rt.fill()/rt.start() or 2-arg Program() style). It should split the 8192-element tensor across 4 workers (data-parallel), and the kernel should perform scale-by-2.0 then ReLU on bfloat16 data, vectorized via the AIE API.",
```
andrej: Do we really want to include a version number here?

skills/aie-code-creator/references/architecture.md
```diff
@@ -0,0 +1,137 @@
...
+-->
+
+# AMD XDNA NPU Architecture Reference
+
+## Devices
+
+| Family | Architecture | NPU columns × rows (compute) | IRON device class |
+|--------|--------------|------------------------------|-------------------|
+| Phoenix / Hawk Point | **AIE2** | 4 × 4 (NPU1) | `NPU1`, `NPU1Col1`, `NPU1Col2`, `NPU1Col4` |
+| Strix / Krackan Point | **AIE2P** | 8 × 4 (NPU2) | `NPU2`, `NPU2Col1`, `NPU2Col4`, `NPU2Col8` |
```
andrej:
```suggestion
| Phoenix / Hawk Point | **AIE2** a.k.a. XDNA | 4 × 4 (NPU1) | `NPU1`, `NPU1Col1`, `NPU1Col2`, `NPU1Col4` |
| Strix / Krackan Point | **AIE2P** a.k.a. XDNA2 | 8 × 4 (NPU2) | `NPU2`, `NPU2Col1`, `NPU2Col4`, `NPU2Col8` |
```

skills/aie-code-creator/references/architecture.md
```diff
@@ -0,0 +1,137 @@
...
+
+Bottom to top:
+
+| Row | Tile | Purpose |
+|-----|------|---------|
+| 0 | **Shim** | DMA between external (DDR) memory and the AIE array |
+| 1 | **Mem** | L2 scratchpad shared between Shim and the column's compute tiles |
+| 2..5 | **Compute** | The actual cores running C++ kernels; each has a small L1 |
+
+Throughout these skills, **compute tile**, **core**, and **AIE tile** all refer to this row-2..5 tile.
```
andrej:
```suggestion
Throughout these skills, **compute tile**, **core**, and **AIE tile** all refer to these row-2..5 tiles.
```

skills/aie-code-creator/references/architecture.md
```diff
@@ -0,0 +1,137 @@
...
+  |-----------|----------------|--------------|---------------------|
+  | Mem tile | **4** | 48 | 2^17 − 1 |
+  | Compute (core) tile | **3** | 16 | 2^14 − 1 |
+  | Shim tile | **3** | 16 | 2^32 − 1 |
+
+  This drives real design decisions: a reshape needing 4 `(size, stride)` dimensions **must**
+  land on a mem-tile DMA. If a compute tile's `dims_to_stream` or a shim's tap already needs
+  3 dimensions, you have no fourth to spend — route L3→L2→L1 and put the reshape on the
+  mem-tile `split()`/`join()`/`forward()` instead. For access patterns beyond the limit, chain
+  multiple BDs or use runtime repeat counts.
```
andrej, line 136: Would a pointer to `VSHUFFLE` here be appropriate as well? DMA transforms aren't the only way to translate data, and `VSHUFFLE` can be masked by concurrent/pipelined `MAC`s.

skills/aie-code-creator/references/builtin_kernels.md
```diff
@@ -0,0 +1,230 @@
...
+| element-wise `a+b` / `a*b` | `kernels.add(...)` / `kernels.mul(...)` | **bf16 only**, `tile_size` fixed at 1024 |
+| ReLU | `kernels.relu(tile_size=1024)` | bf16 |
+| softmax / gelu / silu / swiglu / exp | `kernels.softmax`, `.gelu`, `.silu`, `.swiglu`, `.bf16_exp` | bf16, LUT-based |
+| sum / min / max over a tile | `kernels.reduce_add`, `.reduce_min`, `.reduce_max` | |
+| running max | `kernels.compute_max(dtype)` | |
+| matrix multiply `C += A*B` | `kernels.mm(dim_m, dim_k, dim_n, input_dtype, output_dtype)` | see MMUL geometry below |
+| matrix–vector `c += A*b` | `kernels.mv(dim_m, dim_k, ...)` | int16 in → int32 out only |
+| cascade-accumulated matmul | `kernels.cascade_mm(...)` | pairs with `CascadeFlow` |
+| conv2d 1×1 / 3×3 / skip / batchnorm | `kernels.conv2dk1`, `.conv2dk3`, `.conv2dk1_skip`, `.bn_*` | |
+| RGBA↔gray/hue, threshold, filter2d, bitwise | `kernels.rgba2gray`, `.threshold`, `.filter2d`, … | vision ops |
```
andrej, line 36: Does this table meaningfully improve the skill over just having the agent list the directory contents? Just asking because it feels like this could easily become outdated.

skills/aie-code-creator/references/complete_examples.md
andrej, line 1: Are these examples tested? Worried they might become outdated. Would it make sense to add these as regular examples and point to their path instead, or does that weaken the skill?

skills/aie-code-creator/references/pitfalls.md
```diff
@@ -0,0 +1,389 @@
...
+
+## ❌ Mismatched `acquire` / `release` → deadlock
+
+```python
+# BAD
+def core(of_in, of_out, k):
+    for _ in range_(N):
+        e = of_in.acquire(2)        # acquires 2
+        k(e[0], e[1], out_tmp)
+        of_in.release(1)             # releases 1 — leaks one slot per iteration
```
andrej: The comment "leaks one slot per iteration" is wrong. The mistake here is that there is not an additional `of_in.release(1)` after the loop. The following would be a perfectly valid "sliding window" pattern:

```
    for _ in range_(N):
        e = of_in.acquire(2)        # acquires 2: on first iteration, 2 new objects, on subsequent iterations 1 new object in the sliding window
        k(e[0], e[1], out_tmp)
        of_in.release(1)             # releases 1 -- advance sliding window by 1
    of_in.release(1)             # releases remaining 1
```

Remember: acquire(N) means "after this call, ensure we hold `N` objects", **not** "acquire `N` _additional_ objects".

Enabling cyclostatic/sliding window patterns are one of the reasons why the ObjectFifo acquire is "weird" like that, and I'd like to make sure AI can make use of that when there is a suitable use case for it, rather than insinuating that not releasing the same number as acquiring is always bad.

Please correct me if I'm wrong @denolf

skills/aie-code-creator/references/pitfalls.md
```diff
@@ -0,0 +1,389 @@
...
+```
+
+After `depth` iterations the FIFO is full of un-released objects; the producer blocks forever.
```
andrej: Again, I believe the loop as written above means there is just one unreleased object.

skills/aie-code-creator/references/pitfalls.md
```diff
@@ -0,0 +1,389 @@
...
+
+A `bfloat16[4096]` tile is 8 KB. Holding several of those plus the kernel stack can blow L1 (~64 KB). Symptoms: stack overflow at runtime, mysterious crashes.
+
+**Fix**: shrink `TILE`, reduce `depth`, or move large temporaries to a mem-tile ObjectFifo.
+
+When the overflow is because the *kernel* genuinely needs more simultaneous state than fits — several activations alive at once for a wide skip/concat, say — **split the work across two tiles and cascade the partial result** rather than squeezing the existing kernel by dropping precision or unrolling less. Those cuts trade away exactly the correctness or performance you're trying to establish; a tile split costs neither.
+
+---
+
+## ❌ Device name doesn't match the physical hardware → silent all-zero output
```
andrej:
```suggestion
## ❌ Device name doesn't match the physical hardware → silent all-zero output or timeout
```

skills/aie-code-creator/references/placed_api.md
```diff
@@ -0,0 +1,114 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# Low-Level (Placed) API — explicit tiles
+
+Use this only when you need explicit tile coordinates, custom routing, or features not exposed in the high-level API. For everything else, prefer `Program` / `Worker` / `ObjectFifo` from the high-level API.
```
andrej: @hunhoffe Thoughts on this? Should we drop the "placed" angle here?

skills/aie-code-creator/references/placed_api.md
```diff
@@ -0,0 +1,114 @@
...
+| Core definition | `Worker(fn, args)` | `@core(tile) def core_body()` |
+| ObjectFifo acquire | `of.acquire(n)` | `of.acquire(ObjectFifoPort.Consume/Produce, n)` |
+| Runtime | `Runtime(seq_fn, fn_args)` | `@runtime_sequence(...)` |
+| DMA | `ObjectFifoHandle.fill` / `.drain` | `npu_dma_memcpy_nd` + `dma_wait` |
+| Compilation | `Program(dev, rt, workers=[...]).resolve_program()` | wrapped in `mlir_mod_ctx()` |
+| Verification | implicit | call `ctx.module.operation.verify()` |
+
+## When the placed API is justified
+
+- You need a tile at a specific column for board-level routing reasons.
```
andrej: highlighting

skills/aie-code-creator/references/python_api.md
```diff
@@ -0,0 +1,382 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# IRON Python API Reference
+
+This reference documents **mlir-aie 1.4.0**. Version 1.4.0 reworked the `Runtime`/`Program` API (upstream commit `cd9d998ffcd`, "[dyn-seq] Rework IRON Runtime as an eager callback body"); older pinned versions (e.g. `v1.3.4`) are out of scope here — see `patterns.md` and `complete_examples.md` for the one exception, the `amd/IRON` operator-framework pattern, which still pins an older wheel and intentionally keeps the old API.
```
andrej: I think these notes will be outdated almost immediately. Also not sure it's relevant to the skill to know that the API has been reworked as compared to previous iterations (might be wrong)?

skills/aie-code-creator/references/python_api.md
```diff
@@ -0,0 +1,382 @@
...
+w = Worker(core_body, [seq_len, of_in.cons(), of_out.prod(), k])
+
+def sequence(a_in, c_out, in_h, out_h):
+    # write seq_len via offset_parameter= on a fill/drain call, then:
+    iron.sync_parameters()                # emits the host→scratchpad sync; call before workers read it
+    in_h.fill(a_in)
+    out_h.drain(c_out, wait=True)
+```
+
+Reach for `Buffer(..., use_write_rtp=True)` + `WorkerRuntimeBarrier` for the common case (matches most existing examples); use `ScratchpadParameter` when you specifically want to avoid a dedicated RTP buffer/lock.
```
andrej, line 312:
```suggestion
Reach for `Buffer(..., use_write_rtp=True)` + `WorkerRuntimeBarrier` for the common case (matches most existing examples); use `ScratchpadParameter` when you specifically want to avoid a dedicated RTP buffer/lock.

Note that `ScratchpadParameter`s depend on the full-ELF flow (`aiecc --get-full-elf`) and are not available on the Phoenix/Hawk Point architectures.
```

andrej (commented):
Thanks for contributing this. Left a couple comments. :)

skills/aie-code-creator/assets/architecture-diagram.txt
```diff
@@ -0,0 +1,52 @@
+AMD XDNA NPU Tile Architecture
+================================
```
hunhoffe, line 2: I'm good with this, but we might also be able to point agents to the targetmodel in the compiler to avoid duplicated information?

skills/aie-code-creator/assets/kernel_skeleton.cc
```diff
@@ -0,0 +1,64 @@
+//
```
hunhoffe, line 1: Again, I wonder if it would make sense (and be more maintainable) to point to assets already in the repo?

skills/aie-code-creator/references/architecture.md
```diff
@@ -0,0 +1,137 @@
+<!--
```
hunhoffe, line 1: This is a great collection of information. Would be great to advertise it on the website as a cheat sheet.

skills/aie-code-creator/references/kernel_intrinsics.md
```diff
@@ -0,0 +1,225 @@
+<!--
```
hunhoffe, line 1: Nice! I think this would also make a nice reference as a supplemental material for the program

skills/aie-code-creator/references/placed_api.md
hunhoffe: Yes. Please scrub any mentions of "placed" and "unplaced". You can do "higher-level" or "lower-level" if you must distinguish. Both high-level and low-level can be placed or unplaced.

skills/aie-code-creator/references/placed_api.md
```diff
@@ -0,0 +1,114 @@
...
+
+with mlir_mod_ctx() as ctx:
+    design()
+    assert ctx.module.operation.verify() == True
+    print(ctx.module)
+```
+
+## Key differences from high-level API
+
+| | High-level | Placed (low-level) |
```
hunhoffe: Please no placed/unplaced. It really confuses both people and AI.

skills/aie-code-creator/references/placed_api.md
```diff
@@ -0,0 +1,114 @@
...
+                              sizes=[1, 1, 1, 1024])
+            dma_wait(of_out)
+
+with mlir_mod_ctx() as ctx:
+    design()
+    assert ctx.module.operation.verify() == True
+    print(ctx.module)
+```
+
+## Key differences from high-level API
```
hunhoffe: To be honest, do you really need to talk about the high-level and low-level API as separate things at all? I believe we have enough examples to show how you can use Worker/etc with lower-level DMA patterns, etc. so it's more like a spectrum of choices. But I don't think you ever need to do @core... Worker works pretty well.

skills/aie-code-creator/references/placed_api.md
```diff
@@ -0,0 +1,114 @@
...
+| ObjectFifo acquire | `of.acquire(n)` | `of.acquire(ObjectFifoPort.Consume/Produce, n)` |
+| Runtime | `Runtime(seq_fn, fn_args)` | `@runtime_sequence(...)` |
+| DMA | `ObjectFifoHandle.fill` / `.drain` | `npu_dma_memcpy_nd` + `dma_wait` |
+| Compilation | `Program(dev, rt, workers=[...]).resolve_program()` | wrapped in `mlir_mod_ctx()` |
+| Verification | implicit | call `ctx.module.operation.verify()` |
+
+## When the placed API is justified
+
+- You need a tile at a specific column for board-level routing reasons.
+- You need a custom inter-tile flow (`flow(src, src_bundle, src_ch, dst, ...)`) that the placer doesn't generate.
```
hunhoffe: There are flows in higher-level now.

skills/aie-code-creator/references/low_level_api.md
```diff
@@ -0,0 +1,114 @@
...
+| DMA | `ObjectFifoHandle.fill` / `.drain` | `npu_dma_memcpy_nd` + `dma_wait` |
+| Compilation | `Program(dev, rt, workers=[...]).resolve_program()` | wrapped in `mlir_mod_ctx()` |
+| Verification | implicit | call `ctx.module.operation.verify()` |
+
+## When the placed API is justified
+
+- You need a tile at a specific column for board-level routing reasons.
+- You need a custom inter-tile flow (`flow(src, src_bundle, src_ch, dst, ...)`) that the placer doesn't generate.
+- You want to write a `@mem(tile)` body to script the DMA channels explicitly.
+- You're porting an existing MLIR-AIE example and want to keep its structure 1:1.
```
hunhoffe: All examples have been re-written to be high-level.

skills/aie-code-creator/references/low_level_api.md
```diff
@@ -0,0 +1,114 @@
...
+| Runtime | `Runtime(seq_fn, fn_args)` | `@runtime_sequence(...)` |
+| DMA | `ObjectFifoHandle.fill` / `.drain` | `npu_dma_memcpy_nd` + `dma_wait` |
+| Compilation | `Program(dev, rt, workers=[...]).resolve_program()` | wrapped in `mlir_mod_ctx()` |
+| Verification | implicit | call `ctx.module.operation.verify()` |
+
+## When the placed API is justified
+
+- You need a tile at a specific column for board-level routing reasons.
+- You need a custom inter-tile flow (`flow(src, src_bundle, src_ch, dst, ...)`) that the placer doesn't generate.
+- You want to write a `@mem(tile)` body to script the DMA channels explicitly.
```
hunhoffe: There are exmaples of doing custom DMA patterns in high-level IRON. If you have a use case that isn't sufficiently exposed by the high-level API, please let me know so I can fix it.

skills/aie-code-creator/references/low_level_api.md
```diff
@@ -0,0 +1,114 @@
...
+pf = PacketFlow(1, src_tile, dst_tile,
+                extra_dsts=[PacketDest(other_tile)], keep_pkt_header=False)
+rt.add_flow(pf)
+```
+
+Constructing a `Flow`/`PacketFlow` does not register it anywhere by itself — you must call `rt.add_flow(...)` on your `Runtime` instance (both share the same registration method). Without that call, `Program.resolve_program()` never sees the flow and it silently never resolves. Both resolve to `aie.flow`/packet-switching ops when the program is placed — they only declare the topology edge; you still own the `TileDma`/`Buffer`/`Lock` wiring on each end. Reach for `ObjectFifo` first; these are for the rare case where you need routing control without the full ObjectFifo abstraction.
+
+Both `Flow` and `PacketFlow` (and `CascadeFlow`, documented in `python_api.md`) implement the `Resolvable` protocol (`aie.iron.resolvable.Resolvable`) — a structural `Protocol` requiring `resolve(loc, ip)` and `tiles()`. You only need to know about it if you're writing a custom low-level primitive that should participate in `Program.resolve_program()`'s placement/resolution walk.
```
hunhoffe: I would say the resolvable override is the "advertised" method of working with low-level primitives; generally, we there isn't a reason to do a whole design in low-level primitives.

skills/aie-code-creator/references/python_api.md
hunhoffe: Agreed. The docstrings -- which are well-maintained -- are also findable using the python interpretor when the package is local, so maybe it would be most useful to tell the agent how to find information when there is a question based on the installed package?

skills/aie-code-creator/README.md
```diff
@@ -0,0 +1,42 @@
+<!--
+Copyright (C) 2026 Advanced Micro Devices, Inc.
+SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+-->
+
+# AIE Code Creator
+
+## Description
+
+This skill teaches how to write efficient designs for AMD XDNA NPUs (AIE2 in Phoenix / Hawk Point, AIE2P in Strix / Krackan Point) using the IRON (Interface Representation for hands-ON) Python API together with the C++ AIE kernels that run on the compute tiles.
```
hunhoffe: We're trying to -- somewhat -- get rid of the acronym and let it just be IRON.

skills/aie-code-creator/SKILL.md
```diff
@@ -0,0 +1,141 @@
...
+
+If the user does not specify, assume:
+
+- **Kernel source**: a built-in from `aie.iron.kernels` if one covers the op and dtype; hand-written C++ only otherwise.
+- **API level**: High-level — an `aie.iron.algorithms` template if the topology fits, else `Program` / `Worker` / `ObjectFifo`. Let the compiler place tiles.
+- **Target**: current device (auto-detect via `iron.get_current_device()`); if you must pick, default to `NPU2` (AIE2P).
+- **Data type**: `bfloat16` (best balance of accuracy and throughput on AIE).
+- **Parallelism**: multi-core (data-parallel split across all available compute tiles in 1+ columns).
+
+Drop low-level (placed) API only when the user needs explicit tile coordinates, custom routing, or hand-tuned placement.
```
hunhoffe: no placed/unplaced please!

hunhoffe (approved):
Thank you for putting this together!

Only concerns:
* [high priority for me at least] Scrub all language/mentions about "placed" and "unplaced" API -- that is not an accurate characterization.
* Maintainability

For the maintainability concern, we've talked about staging here as an initial step, and then fine-tuning in separate commits, so I'm open to merging as-is and making it more resilient/concise/fine-tuning in additional PRs.

skills/aie-code-creator/assets/kernel_skeleton.cc
ypapadop-amd: I don't think that's possible without copying a subset of the repo along the skill. As is now, you can copy the skill in your skills directory and it will work.

skills/aie-code-creator/references/architecture.md
ypapadop-amd: I'm not familiar with `VSHUFFLE`. Let me see if I can make it extract some pattern from existing code.

skills/aie-code-creator/references/builtin_kernels.md
ypapadop-amd: Reading all files, even via sub-agents (which would be per-user opt-in) would fill-up the context window; it happened to me in some more expanded algorithm rewrites. Such smaller tables and shortcuts keep the context window to the absolutely necessary.

skills/aie-code-creator/assets/architecture-diagram.txt
ypapadop-amd: What do you mean?

skills/aie-code-creator/references/complete_examples.md
ypapadop-amd: They lead to correct IRON code for the cases I tested, but I can't guarantee that all examples are correct. As I mentioned in a comment above, skills need to be self-containing so AFAIK I can't point to existing code outside the skill without additional complexity (e.g., scripts to install the skill).

I think that having some kind of test harness that is not mixed in the example code, may resolve this, but I leave it to you if it's worth it.

skills/aie-code-creator/references/kernel_intrinsics.md
ypapadop-amd: Done.

skills/aie-code-creator/references/architecture.md
ypapadop-amd: Fixed below.

skills/aie-code-creator/assets/architecture-diagram.txt
hunhoffe: The Python API has device classes which inherit information on each device type from one unified spec in mlir-aie. Each spec is called a "target model." This means that aiecc/python packages can be treated in a tool which can be consulted for information about a specific architecture. I'm not sure if a tool call is better than duplicating information here, but it's a bit hard to maintain documentation as it is so I thought I'd suggest using the internal representation of the device as a tool for agents to understand device limitations.

PR #3526, opened 2026-08-10
Fix npu2 context cache size: Strix-family HW context pool is 16, not 32

bong-water-water-bong:
## Summary

`NPU_CONTEXT_CACHE_SIZE["npu2"]` is 32, but the amdxdna driver caps concurrent HW contexts per device at **16** for the whole npu2 family (Strix / Strix Halo / Krackan) — `hwctx_limit = 16` in the driver's `npu4_regs.c` / `npu5_regs.c` / `npu6_regs.c`, and exposed live via `QUERY_RESOURCE_INFO` (`npu_task_max`). `npu1` = 6 already matches Phoenix's driver limit; 16 is the same kind of value for npu2.

## Why it matters

The HW contexts are a system-wide constrained resource. With the old default, a single process can cache 32 contexts on a 16-slot pool; once the pool saturates, `CREATE_HWCTX` fails with `MGMT_ERT_NOAVAIL` (0x2000003) and the retry-eviction loop cannot recover because other processes hold the remaining slots — a hard `RuntimeError` under parallel workers (e.g. pytest-xdist), exactly the "flaky parallel CI" case the existing comment warns about.

## Verification (Strix Halo, Ryzen AI Max)

- Driver pool confirmed empirically: 4 live server contexts + 12 more created = 16/16, 13th create → NOAVAIL.
- Parallel sweep (5 workers, 2 cached contexts each + 4 live server contexts = 14 < 16): **5/5 pass**.
- With the old default: pool saturates at 16/16 mid-run and 2/5 workers fail.

## Notes

- For multi-process workloads, the per-process cache should stay well below the pool; `XRT_CONTEXT_CACHE_SIZE` (e.g. `2`) is the knob.
- A proper long-term fix is the existing TODO: query the driver for the real pool instead of hardcoding.


jskimko:
Might be worth double-checking this number before lowering it. The 16 `hwctx_limit` (what `npu_task_max` reports) looks like how many contexts map onto the array at once; with [`temporal_only = 1`](https://github.com/amd/xdna-driver/blob/944ca35a395b7dfa37ac3f437f1da48feca7bffc/src/driver/amdxdna/npu4_family.h#L84) those slots time-share, so up to `ctx_limit = 32` can be resident. If `NPU_CONTEXT_CACHE_SIZE` is bounding *resident* contexts, then 32 might be the one that applies. E.g., the following is valid up to 32:

```cpp
std::vector<xrt::hw_context> pool;
try { for (int i = 0; i < 64; ++i) pool.emplace_back(dev, xclbin.get_uuid()); }
catch (const std::exception& e) { printf("failed at %zu\n", pool.size()); }  // 32
```

If that holds, the fix might be less about lowering the number and more about `CachedXRTRuntime` distinguishing the two caps, i.e., raising the functionality to 32. This seems to be roughly what the `TODO` right above your change is pointing at.

hunhoffe:
I've done some testing up to 32 contexts, I believe. Were you running into any errors with the current functionality? That might help us better understand why a change should be considered.

bong-water-water-bong:
Thanks for taking a look, Erika! Yes — the errors are reproducible on Strix Halo (Ryzen AI Max, npu2/npu5). With the cache at 32, parallel processes exhaust the firmware ERT pool and new context creation fails with AIE2_STATUS_MGMT_ERT_NOAVAIL (we hit EINVAL on the 14th concurrent context). Retry/backoff alone does not recover because the pool stays full of other workers' cached contexts — a per-process cap was the only robust fix we found.

The number 16 comes from the amdxdna driver itself: hwctx_limit is 16 for npu4/5/6 (and npu1=6 already matches Phoenix in the existing table), so this change just brings npu2 in line with the driver it actually runs on.

32-context testing would pass in a single sequential process — the failure only shows with concurrent processes each holding cached contexts. That is also the shape of the flaky parallel CI the existing comment mentions.

hunhoffe:
We don't have a ton of time to investigate further at the moment so I've created an issue for further investigation https://github.com/Xilinx/mlir-aie/issues/3594 and will look into getting this merged in the meantime!

hunhoffe (approved):

PR #3531, opened 2026-08-10
[aiecc] Keep the merged module's alignment when downgrading for Peano

erwei-xilinx:
> Stacked on #3528 — that PR's commit is the parent here, so review only the second commit. Rebase to `main` once #3528 lands.

## Problem

`downgradeIRForPeano` strips `, align N`. That was introduced pre-link, where the rationale still holds (retaining `align` made Peano's capped-O1 opt skip vectorizing the matmul K-loop). #3493 moved the kernel merge in-process and ran the same downgrade **again on the merged module** — and there the strip is destructive.

An `alloca`'s alignment is not an annotation opt can re-derive. Once the text is gone, re-parsing gives the type's ABI alignment, so an over-aligned stack object is silently demoted:

```llvm
-  %lin_aprox = alloca %"struct.aie::linear_approx", align 64
+  %lin_aprox = alloca %"struct.aie::linear_approx", align 4
```

`aie::linear_approx` is the LUT helper behind the softmax `exp` in an inline attention kernel. At 4-byte alignment the core misbehaves and the tile hangs — an NPU decode wedged with `ERT_CMD_STATE_TIMEOUT` after a handful of tokens. Before #3493 the kernel was merged by an external `llvm-link` and never re-downgraded, so its alignment survived.

## Fix

Strip pre-link only. The merged module keeps its `align`: the kernel arrives already annotated by its own clang, and the core module was stripped before the merge, so the vectorization rationale is unaffected.

I first tried the narrower fix of exempting only `alloca` and continuing to strip loads/stores. It is **not** sufficient — the model still failed its numerical gate (1 of 2 prompts), because the merged kernel also depends on the load/store alignment it was compiled against. Hence the whole post-link strip goes.

## Validation

Two AMD NPU2 LLM decode models that hung with this defect now pass their top-k-vs-HF-bf16 gate (`topk_passed: 2, topk_failed: 0` each) on the pinned toolchain:

| model | before | after |
|---|---|---|
| `llama32_1b_q4nx` | `decode dispatch pos7 ERT_CMD_STATE_TIMEOUT` | PASS |
| `llama32_3b_q4nx` | same | PASS |

Localized by A/B against the pre-#3493 toolchain: identical `input_physical.mlir`, identical instruction histograms, and of 27 cores only the 8 merge-mode ones differed in their object/ELF — with this single alignment line as the only semantic difference. Swapping just the pre-#3493 xclbin into an otherwise post-bump run also passes, confirming the defect is in the core code and not the instruction stream.

Keeping the alignment also produces **smaller** code here: the attention core ELF drops from 11512 to 9444 bytes (176KB vs 193KB across all cores).

## Tests

`peano_compat_merge_link.mlir` asserted `--implicit-check-not=", align "` on the merged module — precisely the behaviour that has to go — so it now pins the alignment instead, and its kernel's scratch buffer is over-aligned to make the demotion observable. Verified the new assertion fails against an unpatched `aiecc` (`LINKED-DAG: alloca [4 x float], align 64` — expected string not found) and passes with the fix. All 8 `peano_compat_*` tests pass, including `peano_compat_align_strip.mlir`, which covers the pre-link strip that is retained.

One residual risk worth a reviewer's eye: the merged module now reaches opt with the ABI-default `align` that the reprint adds back to *core* instructions. That did not regress program memory in my workload (it shrank), but the i8 matmul the original note names is the case to sanity-check.

### DIFF
--- a/test/aiecc/Inputs/peano_float_decimal_kernel.ll
+++ b/test/aiecc/Inputs/peano_float_decimal_kernel.ll
@@ -1,0 +1,47 @@
+; Copyright (C) 2026 Advanced Micro Devices, Inc.
+; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+;
+; A merge-mode ("inline") kernel artifact carrying narrow float constants in
+; every position downgradeIRForPeano has to reason about. Owned by
+; peano_compat_float_decimal_merge.mlir.
+;
+; The merge linker reprints this module with aiecc's LLVM, which spells a
+; float/half constant as a short decimal whenever that decimal round-trips in
+; the narrow type -- a spelling Peano's older parser rejects. The constants are
+; written here in the hex form so the intent is fixed whatever a given LLVM
+; prints; the reprint is what introduces the decimals.
+;
+; The (ptr, ptr) signature is aiecc's bare-pointer memref calling convention.
+
+target triple = "aie2"
+
+; Typed position: a constant array, the shape the LUT in a real attention
+; kernel takes. 2.5 is exact as a double and must survive as printed.
+@lut_f = linkonce_odr global [4 x float] [float 0x400921FA00000000,
+                                          float 0x3FBC28F5C0000000,
+                                          float 0xBF6A8292A0000000,
+                                          float 2.500000e+00], align 4
+; Same, for half, which takes its own 16-bit hex form.
+@lut_h = linkonce_odr global [2 x half] [half 0xH2F0A, half 0xH4100], align 2
+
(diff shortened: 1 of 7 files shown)


erwei-xilinx:
Superseded by #3533, which carries the same commit based on `main` directly. This one was branched on top of #3528, which would have forced a landing order (and, if merged first, dragged #3528's commit in with it). The two fixes are orthogonal — verified cherry-picking cleanly in both directions — so they can now land independently, no rebase needed.

PR #3532, opened 2026-08-10
aie-rt: zero the p_memsz - p_filesz gap when loading PT_LOAD segments

hunhoffe:
## Description

`_XAie_LoadDataMemSection` sets `SectionSize = Phdr->p_memsz` and writes that many bytes from `ElfMem + p_offset`, which holds only `p_filesz` valid ones. It substitutes a zeroed buffer *only* when `p_filesz == 0`, so a pure-data segment and a pure-bss segment are handled but the ordinary mixed `.data`+`.bss` segment (`0 < p_filesz < p_memsz`) is not: it reads past the segment's file contents and DMAs the following ELF bytes — `.comment`, `.symtab` — into tile data memory, where zero-initialised statics belong. Per the System V gABI those trailing bytes are "defined to hold the value 0"; that gap *is* how `.bss` is represented in a loadable segment, and it is the common case (27/27 Chess-built and 25/27 Peano-built cores of a real design have one).
                                                                        
The result is silent numerical corruption. In a 27-core Gemma4 decode layer the string `"Linker: LLD ..."` landed on a kernel's ping/pong selectors — a `static bool` received `'k'`, inverting the double-buffer phase for the whole run, so every call read the stale buffer. With all 12 kernels built by Peano and identical ELFs and routing, the unfixed loader matched the Chess reference on 2/3 greedy prompts (3 runs) and the fixed one on 3/3 (4 runs). Note this is not Peano-specific: Chess takes the same path and escapes only because the bytes trailing its data segment happen to be benign.
                                                                        
The fix allocates a `p_memsz` zeroed buffer whenever `p_memsz > p_filesz` and copies the initialised prefix into it, so the write loop never reads past the file contents; `p_filesz == 0` becomes the degenerate case. It is carried as a vendored patch per `third_party/patches/aie-rt/README.md`, and is not fixed upstream (`Xilinx/aie-rt` master is byte-identical here apart from error strings), so I am happy to send it there instead. `test/aiecc/bss_zero_init.mlir` is a regression test needing no hardware — it asserts the segment really is mixed so it cannot pass vacuously, and was verified to fail without the patch and pass with it.  

## Checklist

### DIFF
--- a/test/aiecc/bss_zero_init.mlir
+++ b/test/aiecc/bss_zero_init.mlir
@@ -1,0 +1,58 @@
+//===- bss_zero_init.mlir ---------------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Regression test for the aie-rt ELF loader out-of-bounds read fixed by
+// third_party/patches/aie-rt/0002-elfloader-zero-bss-gap.patch.
+//
+// A kernel with both .data and .bss links to one PT_LOAD with
+// 0 < p_filesz < p_memsz -- the ELF-spec representation of .bss.
+// _XAie_LoadDataMemSection used to write p_memsz bytes from a p_filesz-sized
+// buffer, so the ELF's own .comment/.symtab bytes were DMA'd into the tile in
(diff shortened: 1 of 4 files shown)


andrej (approved):
Fix makes sense to me while keeping intrusion minimal. Probably should also be submitted to upstream aie-rt.

hunhoffe:
@andrej Thanks for the review! I'm waiting to hear back to someone in aie-rt re: upstreaming.

PR #3535, opened 2026-08-10
ci: fix the Docker prune step that never ran in the distro wheel builds

hunhoffe:
## Description

Both distro wheel workflows guard their Docker prune step with contains(inputs.MATRIX_OS, 'ubuntu'). That idiom is correct inside the setup_base and setup_ccache composite actions, where MATRIX_OS is a declared input — but these two copies sit at workflow scope, where inputs holds only workflow_dispatch inputs. MATRIX_OS isn't one of them, so the expression tests against an empty string and the step has never executed, on any leg, on any OS. These were the only two workflow-scope uses of inputs.MATRIX_OS; every remaining reference lives in a composite action and is correct. 
  
Both now use runner.os == 'Linux', already the prevailing idiom in these files (mlirDistro.yml:331, mlirAIEDistro.yml:197). Deliberately not runner.os  != 'Windows', which is what the adjacent ccache step uses: that phrasing would also fire on the macos-12/macos-14 legs, and GitHub's macOS images ship no Docker, so it would break as soon as those commented-out matrix entries are restored. Excluding Windows is intentional rather than incidental — Docker is installed on windows-2022, so the command would run, but it serves Windows containers and prune -a would evict the image's cached Server Core / Nano Server / .NET Framework layers for no benefit, since neither wheel build uses containers on Windows (as the neighbouring comment in both files already notes). Worth flagging for review: because the step has never run, this PR's own CI is its first real test, and the plausible failure mode isn't the prune itself but a downstream step quietly depending on a Docker layer cache surviving.

## Checklist


hunhoffe:
Current CI failure has different root cause; see https://github.com/Xilinx/mlir-aie/pull/3541

atassis:
One data point on whether these copies are worth keeping once the guard is fixed.

On run 32251311800 (mlirDistro, ubuntu-24.04 x86_64), setup_base's own `Docker prune`
already reports `Total reclaimed space: 0B`. `free-disk-space` runs about 20 lines
earlier in the same composite action and prunes docker itself:

```
BEFORE CLEAN-UP:  59G used, 87G avail, 41%
  Total reclaimed space: 1.791GB
  => Docker images: Saved 1.8GiB
  => Saved 27GiB
AFTER CLEAN-UP:   31G used, 114G avail, 22%

Total reclaimed space: 0B     <- setup_base's Docker prune, immediately after
```

So the build starts with 114 GiB free on a 145 GiB disk, and the workflow-scope copies
run later still, after the build. They may well reclaim 0B too. That is one leg of one
run, so it does not settle the aarch64 leg, but it might make deleting all three the
simpler fix.

PR #3538, opened 2026-08-11
[AIE] Add BD iteration support to aie.dma_bd

jskimko:
## Description

On devices that support it, a BD can advance the address it accesses on each execution: execution `k` adds an offset of `((iteration_current + k) % iteration_size) * iteration_stride` (in elements) to its configured address. The offset advances for `iteration_size` executions (with wrapping); `iteration_current` selects the starting step. This lets a single BD cover N sub-buffers over N executions instead of needing an N-deep BD chain.

The runtime/NPU path already reaches this register through the outermost dimension of an n-dim transfer; the structural `aie.dma_bd` lowering had no way to express it:

```mlir
// runtime/NPU (already supported): outermost dim -> Wrap/StepSize, IterCurr forced to 0
aiex.npu.dma_memcpy_nd (%buf[0, 0, 0, 0][4, 1, 1, 64][64, 0, 0, 1])
    {metadata = @fifo, id = 0 : i64} : memref<256xi32>

// aie.dma_bd (this PR): the fields set directly, IterCurr included
aie.dma_bd(%buf : memref<256xi32> offset = 0 len = 64)
    {iteration = #aie.bd_iteration<size = 4, stride = 64, current = 0>}
```
Updated: previously three flat attributes: `{iteration_size = 4 : i32, iteration_stride = 64 : i32, iteration_current = 0 : i32}`. Name+semantics clash with `write_bd`, which uses raw register values.

Both program the same BD register via aie-rt:

```c
XAie_DmaSetBdIteration(XAie_DmaDesc *DmaDesc, u32 StepSize, u8 Wrap, u8 IterCurr)
```

`iteration_size` -> `Wrap`, `iteration_stride` -> `StepSize`, `iteration_current` -> `IterCurr`. ~~The attribute names match the existing `aiex.npu` op.~~ The three fields are now grouped into one `#aie.bd_iteration` attribute (true/element values) so they aren't conflated with the raw register values of the identically-named `aiex.npu.writebd` family -- mirroring `packet` / `PacketInfoAttr` vs. the flat `packet_*`.

### What's included

- **Dialect**: ~~the three `iteration_*` attributes above on `aie.dma_bd` (all optional).~~ a single optional grouped `iteration` attribute (`#aie.bd_iteration<size, stride, current>`) on `aie.dma_bd`.
- **Model Property**: a new `UsesBDIteration` target-model property (AIE2 family; AIE1 rejects). The verifier enforces the ranges the hardware needs (aie-rt has no lower-bound check and uses `value - 1`, so a `0` would underflow the field to its maximum).
- **aie-rt lowering**: calls `XAie_DmaSetBdIteration` for BDs with `iteration_size > 1`, scaling the stride from element width to 32-bit words.
- **`xaie` backend**: emits the equivalent `XAie_DmaSetBdIteration` call in the generated C++.
- **IRON**: `Bd` gains ~~the three fields~~ an `iteration` field (`BdIteration`), forwarded to `aie.dma_bd` in `TileDma.resolve()`:
  ```python
  Bd(buffer=buf, offset=0, length=64,
     iteration=BdIteration(size=4, stride=64, current=0), next="self")
  ```
- ~~**Programming example** (`programming_examples/basic/bd_iteration/`): a single self-chained S2MM BD with `iteration_size = 4` fills four slots over four executions; an `iteration_size = 1` run demonstrates the failing case (every bite lands at offset 0).~~ **Device test** (`test/npu-xrt/bd_iteration/`): a single self-chained S2MM BD with `iteration_size = 4` fills four slots over four executions; `--slots 1` demonstrates the failing case (every chunk lands at offset 0) and `--start 1` checks a nonzero `iteration_current` on device.
- **Runtime-sequence path**: the grouped attribute is rejected there; iteration stays expressed via the outermost `sizes`/`strides` dimension.                                                                                  
## Checklist

- [x] Code is formatted (`clang-format` for C++, `black` for Python - see [CONTRIBUTING](../CONTRIBUTING.md))


test/npu-xrt/bd_iteration/bd_iteration.py
jgmelber, line 1: Is this a programming example or a unit test? https://github.com/Xilinx/mlir-aie/tree/main/test/unit_tests

jgmelber: Or better: https://github.com/Xilinx/mlir-aie/tree/main/test/npu-xrt

jgmelber (approved):
Great addition! Consider where the integration test is best suited, perhaps in the npu-xrt directory

programming_examples/basic/bd_iteration/README.md
```diff
@@ -0,0 +1,62 @@
...
+    length=CHUNK,            # 64 elements per execution
+    iteration_size=n_slots,  # wrap after n_slots executions
+    iteration_stride=CHUNK,  # advance the base address by CHUNK each execution
+    acquires=[Acquire(slot_credit, value=1, greater_equal=True)],
+    releases=[Release(fill_count, value=1)],
+    next="self",             # re-run to consume the whole stream
+)
+```
+
+The BD executes four times; iteration advances its base one 64-element slot per
```
andrej: > The BD executes four times;

If I understand correctly, the BD would execute infinitely -- it is chained with `next_bd` to itself. I think it gets block on the _fifth_ iteration because it's not receiving more data from the runtime sequence. So the "end state" of the DMA is, `slot_credit` lock acquired, stalled on receiving no data.[^1]

Could you clarify this in this explanation, please?

[^1]: Side note: looking towards reconfiguration, this sort of distinction is important, because it means you now have an acquired lock and a stuck BD, which you will need to reset if you want to reconfigure -- an "unbounded" design as we've called it before.

andrej (approved):
This is great! I'm surprised this wasn't wired up yet. I guess we've only been using it from the NPU-runtime-sequence-lowering side. This is cleaner, by not mixing up the iteration size/stride with the data layout transformations.

Also really appreciate you adding a good programming example and explanation that we can point people to.

The docs and example are clear on how the iteration offset is added each iteration, but if I could make one wish I'd like to get some more clarity on how the total number of iterations is determined. As I currently understand, there are two ways of doing this:
- Infinite/unbounded by chaining one BD to itself (potentially indirectly through other BDs) via `next_bd`.
- Using `repeat_count` on a task -- allows for bounded iteration.

andrej:
Oh, and one more question, have you experimented with setting `IterationCurrent`? My understanding was that this was a register that the hardware advances on each iteration. Your explanation that setting it sets the "base iteration" makes sense, but we should probably verify that if you haven't already.

test/npu-xrt/bd_iteration/bd_iteration.py
jskimko: Moved to `test/npu-xrt`.

jskimko:
> Oh, and one more question, have you experimented with setting `IterationCurrent`?

@andrej Added `--start` to set `IterationCurrent` in the test. Observable wrapping behavior.

programming_examples/basic/bd_iteration/README.md
jskimko: Yes, `iteration_current` is a per-BD register that advances (modulo size) on each execution -- independent of any chain. In this case, the BD is blocked on the fifth execution in the same way a non-iterating BD would be.

Added into the docstring.

fifield:
I bit wordy since I had an agent write it, but it gets my concern across:

---

Reviewed at `mlir-aie` commit `a015b1d09078454909c4a52460e48a844cc502ea` (branch `bd-iteration`).

Two ops now have fields called `iteration_size`, `iteration_stride`, and
`iteration_current`. They don't mean the same thing.

## Background

A buffer descriptor on AIE2 is a block of consecutive 32-bit registers in the
tile's DMA register file, one block per BD. The words hold base address,
transfer length, the D0..D2 (or D0..D3 on a MemTile) strides and wraps, packet
header, lock acquire/release, and so on. Both ops here are ultimately writing
that same block; they just get there by different routes.

One of those words is the iteration state. It's what lets a single BD move its
own base address between executions instead of staying put, so N executions of
one BD can cover N sub-buffers rather than needing an N-deep BD chain. It packs
three fields:

- **step** — how far the base advances per execution, in 32-bit words
- **wrap** — how many executions before the base resets to the start
- **current** — which step the BD starts on

Layout and field widths vary by tile type:

| tile | word | current | wrap | step |
|---|---|---|---|---|
| core | 4 (`+0x10`) | `[24:19]` | `[18:13]` | `[12:0]` |
| mem | 6 (`+0x18`) | `[28:23]` | `[22:17]` | `[16:0]` |
| shim | 6 (`+0x18`) | `[31:26]` | `[25:20]` | `[19:0]` |

Wrap is 6 bits everywhere, hence the max of 64. Step is 13/17/20 bits, matching
`getDmaBdStepBits` in `AIETargetModel.h:690-700`.

Step and wrap are stored off by one: write `3` to get 4 executions, write `15`
to get a step of 16 words. That's so a 6-bit wrap can cover 1..64 instead of
0..63. Current is stored as-is.

So each op has to decide whether the number in the IR already has the minus-one
applied. The two ops decided differently.

## `aiex.npu.writebd`: already applied

Its attributes are the raw register fields. `AIEDmaToNpu.cpp:815-817` shifts
them into place and does no arithmetic:

```cpp
words[6] |= (op.getIterationCurrent() & 0x3f) << 23;
words[6] |= (op.getIterationSize()    & 0x3f) << 17;
words[6] |= op.getIterationStride()   & 0x1ffff;
```

The subtraction happens upstream in `BdLowering.h:180-184`
(`sub(inputSizes[3], 1)` and `biasedStride(...)`).

## `aie.dma_bd`: not applied

You write what you mean. `iteration_size = 4` is four executions. aie-rt's
`XAie_DmaSetBdIteration` does the subtraction later.

## Side by side

Four executions with a 64-byte step:

| field | `aie.dma_bd` | `aiex.npu.writebd` |
|---|---|---|
| `iteration_size` | `4` | `3` |
| `iteration_stride` | `64`, in elements | `15`, in 32-bit words minus 1 |
| `iteration_current` | `0` | `0` |

The stride differs twice over: different units *and* different bias.
`iteration_current` doesn't differ at all.

That last part is what makes this hard to remember. Two of the three fields flip
and one doesn't, so there's no rule like "the npu op is always biased" to fall
back on.

## The failure case

`0` and `1` are valid in both ops and mean opposite things.

- `aie.dma_bd`: `iteration_size = 1` disables iteration. `0` is a verifier error.
- `aiex.npu.writebd`: `iteration_size = 0` disables iteration. `1` means two
  executions.

Copy `iteration_size = 1` out of a working `writebd` example into a `dma_bd` and
you get half the transfers you wanted, with no diagnostic. Copying the other
direction is just as quiet.

## What the PR documents

`AIEOps.td:1053-1055` correctly says the new attributes hold true values and
aie-rt applies the `-1`. That part is fine.

It doesn't mention that a field of the same name behaves differently one dialect
over. Reading `AIEOps.td` alone, there's no way to find out.

Two options:

1. Note the other convention in both ops' docs and link them. Easy, but the
   mistake is still writable.
2. Rename, so the convention is visible where it's used: keep `iteration_size`
   for true counts, use something like `raw_iteration_size` for the
   register-level op. More work, but the mistake stops being writable.

## Also: dimension 3

Unrelated to the bias, the same `aie.dma_bd` op means different things depending
on its parent:

- Structural lowering (MemTile) encodes D3 and the iteration register separately.
- Task lowering treats the 4th dimension *as* iteration, on every tile type.
  See `AIEDMATasksToNPU.cpp:812-813` and the comment at `AIEXDialect.cpp:1110-1119`.

Worth pinning down how the new attributes interact with D3 before the API is
frozen.

jskimko:
@fifield These are good points.

### Same field names, two conventions

The issue is that `iteration_*` follows a different convention on `dma_bd` than on `writebd`. `dma_bd`'s other fields take **true, element-denominated values** and let aie-rt apply the register bias during lowering, e.g. `sizes`, `strides`, `len`, `offset`, whereas `writebd` takes **raw register values** (off-by-one, in 32-bit words). Options:

1. Keep `dma_bd`'s convention and document that `writebd` differs. (Maybe just for this PR? Defer "fix".)
2. Keep `dma_bd`'s convention and rename `writebd`'s fields to `raw_iteration_*`, making the difference visible at the API level. This is a breaking change.
3. Drop the true-value convention on `dma_bd` and use raw register values there instead. This is jarring/unintuitive.

### Task lowering (implicit iteration)

On the task-lowering path, the 4th `sizes`/`strides` dimension is what writes the iteration register today, with `iteration_current` hardcoded to 0. Options:

1. Ignore the `iteration_*` attributes on this path.
2. Honor them, and error out when both a 4th dimension and the attributes are given.
3. Honor them, and let them override the 4th dimension, since the attributes are the more explicit specification (and can set `iteration_current`, which the dimension form can't).

### My take

Option 2 for the first, Option 3 for the second.

Need some more votes.

**Update**:
For the first issue, I could make arguments for the other options...
- For option 1, this could be fine if we accept that `writebd` is "lower-level," so it naturally uses raw values.
- For option 3, we can break the mirroring of other fields for a cleaner separation by combining the new functionality into a single field, e.g., `iteration_register = [StepSize, Wrap, IterCurr]`, that expects raw values.

jskimko:
After much deliberation, I've arrived at the following: there will always be a semantic mismatch between the IRON / `dma_bd` element values and the register APIs' off-by-one values, so the goal is to keep the two from being conflated. We get that separation through the parameter *name*: a grouped `iteration : BdIteration` vs. the flat `iteration_*` (precedent is already established with the packet-style parameters: `packet : PacketInfoAttr` vs. `packet_*`). The concepts, however, ARE shared, so we still keep the `size`, `stride`, and `current` fields -- they're just now captured via `BdIteration`. We add documentation where appropriate, including bidirectional comments noting the difference between the two behaviors. Lastly, on the prior runtime-sequence path we disallow the grouped attribute and redirect to the previous mechanism (iteration via the outermost `sizes`/`strides` dimension).

Two things to note:
1. This PR's structural path **allows** setting the `iteration_current` field, whereas the runtime-sequence path remains fixed at 0 (no change). This can be added there in a later pass, if deemed useful.
2. There was an incorrectly large stride bound that was fixed on the register path (needed to depend on the tile type).

PR #3540, opened 2026-08-11
[aiecc] Stop stripping GEP nuw when downgrading for Peano

atassis:
## Problem

`downgradeIRForPeano` rewrites `getelementptr inbounds nuw` to `getelementptr inbounds`, but Peano parses the `nuw` form. Its `LLParser` has taken the GEP no-wrap flags since it gained `GEPNoWrapFlags`, and the released `llvm-aie` 21.0.0 accepts the spelling through both `opt` and `llc`.

The rewrite drops the flag from every objectFIFO state access for no reason: 284 GEPs across the designs I build.

## Fix

Drop the rewrite from the Peano path. `downgradeIRForChess` keeps its copy; that toolchain's LLVM is older and I have not checked it.

Also correct the doc comment, which still says LLVM 23. aiecc moved to 24 in #3491, and #3509 / #3528 added LLVM 24 rewrites to this function.

## Test

`test/aiecc/peano_compat_gep_nuw.mlir` runs the whole Peano path and checks the flag survives into `peano-compat_*.ll`. If a later Peano loses it, the test fails in Peano's own `opt`. Restoring the rewrite makes the test fail, so it gates the change rather than just recording it.

`test/aiecc` goes from 14 failures to 13 on my box; the difference is this test, and the other 13 fail identically with and without the change (no Chess, `aie-translate` or XRT locally).

Parsing checked on three Peano builds, all accepting the form: the released 21.0.0 wheel and two local `aie-public` builds.

Codegen is unchanged. I ran `opt --passes=default<O1> -inline-threshold=10` then `llc -O2 --march=aie2p` over 8 designs (MHA plus 7 GEMV shapes), keeping versus stripping the flag. The assembly is byte-identical in all 8, so this is a cleanup and not a performance change.


atassis:
One thing this surfaced, separate from the fix.

All nine `peano_compat_*` tests (the eight here plus the one I just added) assert what the downgrade *emits*: `peano-compat_*.ll` holds the hex literal, or contains no `f0x`, or carries the sized lifetime marker. That shape fails when a rewrite breaks. It cannot fail when a rewrite stops being needed, so a shim Peano has outgrown is indistinguishable from one still carrying weight. That is how the `nuw` rewrite survived.

Five of the remaining rewrites are pure LLVM 21 vs 24 skew and expire together whenever Peano rebases, with nothing to notice.

The counterpart would be to hand the raw newer spelling to Peano's own `opt` and XFAIL it:

```
// REQUIRES: peano
// XFAIL: *
// RUN: opt --passes=verify %S/Inputs/f0x_float_raw.ll -o /dev/null
```

It flips to XPASS on the rebase that makes the shim redundant, and names which one to drop.

Worth having? I am happy to add it for the constructs this function covers, or to drop the idea if you would rather the suite stay output-only.

erwei-xilinx (approved):

PR #3546, opened 2026-08-12
Make shim-DMA AxCACHE configurable via a new axcache attribute

hunhoffe:
## Description

Exposes the raw 4-bit AXI `AxCACHE` value on `aiex.npu.dma_memcpy_nd`, `aiex.npu.writebd`, and `aie.dma_bd` instead of hardcoding it.
  
`AIETargetModel::getDefaultAxCache()` is the single source of truth for the unset-attribute fallback (2, matching the historical "enable NoC upsizing" tuning). Every lowering/codegen consumer reaches it through `getAxcacheOrDefault()` on the three ops, so the fallback rule lives in exactly one place per op rather than being repeated at each backend.

**Hardware-visible change beyond the new knob:** the aie-rt (CDO /  `--convert-aie-to-transaction`) and XAIEV2 backends previously emitted `Cache=0` for every shim BD, so static shim BDs — including the objectfifo/shim_dma ones on NPU — now issue AXI bursts with `AxCACHE=2` like the runtime-sequence path always has. This unifies both halves of the compiler on one value for every architecture, AIE1 included.
  
Range is enforced via `ConfinedAttr` rather than a hand-rolled verifier, matching existing dialect idiom.
  
Tests cover a configured value reaching the packed hardware word on each backend (blockwrite word[5] on both the static and dynamic `aie-dma-to-npu` paths, the writebd on the dma_task path, and `XAie_DmaSetAxi` on the XAIEV2 path), plus the chained-BD case: BDs created when `aie-decompose-large-dma-bd` slices an oversized task BD must inherit the template's axcache, or one logical transfer ends up issuing bursts under two different cache attributes.

## Checklist


include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1222,44 +1229,29 @@ def AIE_NpuControlPacketOp: AIEX_Op<"control_packet", []> {
...
+      DefaultValuedOptionalAttr<I32Attr, "0">:$burst_length,
+      OptionalAttr<ConfinedAttr<I32Attr, [IntMinValue<0>,
+                                          IntMaxValue<15>]>>:$axcache);
   let results = (outs );
   let assemblyFormat = [{ attr-dict }];
   let hasVerifier = 1;
+  let extraClassDeclaration = [{
+    /* $axcache, or the target's default when unset. The single place the
+       fallback rule lives; every backend emitting an AXI-MM BD goes through
+       this rather than repeating the value_or. */
```
andrej:
```suggestion
    /* $axcache, or the target's default when unset. */
```

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -708,6 +710,11 @@ def AIE_NpuDmaMemcpyNdOp: AIEX_Op<"npu.dma_memcpy_nd", [
     llvm::LogicalResult verifyDynamicSizesStrides(
         const xilinx::AIE::AIETargetModel &targetModel,
         mlir::BaseMemRefType buffer);
+
+    /* $axcache, or the target's default when unset. The single place the
+       fallback rule lives; every backend emitting an AXI-MM BD goes through
+       this rather than repeating the value_or. */
```
andrej:
```suggestion
    /* $axcache, or the target's default when unset. */
```

include/aie/Dialect/AIE/IR/AIETargetModel.h
```diff
@@ -456,6 +456,15 @@ class AIETargetModel {
   virtual std::vector<std::pair<uint32_t, uint32_t>>
   getShimBurstEncodingsAndLengths() const = 0;
 
+  // Returns the default 4-bit AxCACHE value used for shim NOC DMA AXI-MM
+  // transfers when a `dma_bd`/`npu.writebd`'s `axcache` attribute is unset.
+  // Tuned to 2 to enable upsizing in the NoC. The runtime-sequence lowerings
+  // always emitted 2 while the aie-rt/CDO and XAIEV2 backends hardcoded 0;
+  // this unifies both on 2 for every architecture. No target currently needs
+  // a different value, so this is a single non-pure-virtual default rather
+  // than a per-architecture override.
```
andrej:
```suggestion
  // Tuned to 2 to enable upsizing in the NoC. No target currently needs
  // a different value, so this is a single non-pure-virtual default rather
  // than a per-architecture override.
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2600,6 +2601,10 @@ LogicalResult DMABDOp::verify() {
     return emitOpError("Burst length is only supported in Shim NOC tiles that "
                        "are connected to the memory-mapped NOC.");
 
+  if (getAxcache() && !parentTile.isShimNOCTile())
+    return emitOpError("AxCACHE is only supported in Shim NOC tiles that "
+                       "are connected to the memory-mapped NOC.");
```
```suggestion
  if (auto axcache = getAxcache()) {
    if (!parentTile.isShimNOCTile()) {
        return emitOpError("AxCACHE is only supported in Shim NOC tiles "
                       "that are connected to the memory-mapped NOC.");
    }
    if (axcache > 0xF) {
      return emitOpError("AxCache value out of 4-bit range.");
    }
  }
```

andrej (commented):
Couple suggestions, take them or leave them, they're all nits.

andrej (approved):

PR #3548, opened 2026-08-13
[AIE][AIEX] Fix BD verification gaps

xhmelon:
This implements step 0 of the roadmap in [Discussion #3502](https://github.com/Xilinx/mlir-aie/discussions/3502).

### `aie.dma_bd`
- `verify()` checked size and stride against a uniform threshold regardless of tile type. Replaced with tile-type-dependent bounds derived from the target model.
- For sub-32-bit element types, non-word-aligned size/stride combinations could be silently replaced with different hardware-expressible values during lowering. Added a 32-bit-word-alignment check based on the element width (from the buffer's memref type), size, and stride.

### `aiex.npu.writebd`
- `verify()` lacked bounds checks for padding fields. Over-width values were silently truncated to zero at compile time by bitmask. Added bounds checks for D0/D1/D2 pad widths.
- `verify()` used uniform shim-sized thresholds for size/stride instead of tile-type-dependent bounds (same issue as `aie.dma_bd`). Replaced with target-model-derived limits.

### `aiex.dma_configure_task`
- Stride/iteration bounds used `> (1 << N)` while wrap bounds used `> (1 << N) - 1`, creating an off-by-one inconsistency for the same register widths. Unified to `> (1 << N) - 1`. This did not allow illegal BDs through — out-of-range values were caught by the downstream `aiex.npu.writebd` verifier — but the diagnostic pointed at a compiler-generated op rather than the user's code.

### `AIEDMATasksToNPU`
- `d2_size` was read from the outermost-first dimension list while `d0_size`/`d1_size` used the innermost-first convention, mixing up the index. Currently harmless since `d2_size` is not wired into the BD word packing (`// TODO: D2Size`), but would become silent data corruption once that field is connected.

### Tests
7 new test files and 5 updated ones (850+ lines), covering all six defects with both negative cases (pinning the diagnostic) and positive cases (confirming legal neighbours are still accepted). Two positive cases are load-bearing: a shim BD with D2 = 2048 must stay accepted (no D2 wrap register on shim), and a memtile size of 600 must be accepted after the size-bound fix. Both were confirmed against NPU2 hardware.


lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -756,22 +756,41 @@ LogicalResult AIEX::NpuWriteBdOp::verify() {
...
-    return emitOpError("Iteration Size exceeds the [0:63] range.");
+
+  // Every value on this op is already the hardware-encoded field value (wrap
+  // fields unbiased, stepsize/iteration fields biased actual-1), so the
+  // legal encoded range for a B-bit field is simply [0, 2^B - 1] regardless
+  // of which kind of field it is. The field widths themselves are tile-type
+  // specific (e.g. on AIE2: core-tile wrap is 8-bit vs. 10-bit for mem/shim,
+  // core-tile stepsize is 13-bit, mem-tile 17-bit, shim-tile 20-bit), so
+  // derive them from the target model instead of hardcoding shim-sized
+  // constants for every tile type.
```
andrej:
```suggestion
  // legal encoded range for a B-bit field is simply [0, 2^B - 1].
```

nit

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -780,6 +799,30 @@ LogicalResult AIEX::NpuWriteBdOp::verify() {
        getD1ZeroBefore() != 0 || getD1ZeroAfter() != 0 ||
        getD2ZeroBefore() != 0 || getD2ZeroAfter() != 0))
     return emitOpError("ShimTile doesn't support zero padding.");
+
+  // Pad field widths (AIE2P ArchSpec Table 3-34, in 32-bit words): D0 is a
+  // 6-bit field (max 63), D1 a 5-bit field (max 31), D2 a 4-bit field (max
+  // 15). These are currently checked nowhere and silently masked
+  // (`& 0x3F` / `& 0x1F` / `& 0xF`) at word-packing time in AIEDmaToNpu.cpp.
```
andrej:
```suggestion
  // 15).
```

nit: that comment describes the state of code _before_ the change and therefore is misleading, since directly below we _are_ checking now

lib/Dialect/AIEX/IR/AIEXDialect.cpp
```diff
@@ -780,6 +799,30 @@ LogicalResult AIEX::NpuWriteBdOp::verify() {
...
+  // better home for this check, since its diagnostic would point at the
+  // user's own `aie.dma_bd`. But DMABDOp::verify() early-returns for BDs
+  // nested inside `aiex.dma_configure_task`, and the only case with
+  // hardware evidence for this defect is on that task path -- a check
+  // placed only in DMABDOp::verify() would miss it. NpuWriteBdOp is where
+  // both the static and task runtime paths converge, so putting it here
+  // covers both, at the cost of the diagnostic pointing at this
+  // compiler-generated op rather than the user-written BD. This check
+  // should migrate to DMABDOp::verify() once BD iteration becomes an
+  // explicit `aie.dma_bd` field and that early-return can be removed.
```
andrej: Maybe mark this with TODO or FIXME so we can find it in the future.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -805,8 +805,19 @@ struct AIEDMATasksToNPUPass
...
+        // above), matching the convention `d0size`/`d1size` below read from
+        // `sizes[0]`/`sizes[1]`; `(*dims)` is the original outermost-first
+        // list and indexing it with 2 does not name the same dimension
+        // (e.g. for a 4-element outermost-first list, `(*dims)[2]` is D1,
+        // not D2). Note that d2_size is currently a dead field regardless
+        // of this fix: AIEDmaToNpu.cpp's memtile word-packing never writes
+        // it (see its `// TODO: D2Size`), with the real D2 repeat count
+        // carried entirely by buffer_length, same as on shim tiles. Fixing
+        // the index here has no effect on lowered output today, but keeps
+        // this value correct for whenever D2Size does get wired up.
```
andrej:
```suggestion
        // TODO: Note that d2_size is currently a dead field:
        // AIEDmaToNpu.cpp's memtile word-packing never writes
        // it, with the real D2 repeat count
        // carried entirely by buffer_length, same as on shim tiles.
```

Comment mostly narrates a change/fix that will be confusing/missing context in the committed code. Please make sure to only document the _current_ code, not changes to it.

andrej (commented):
Looking good, these fixes make sense to me. I left a couple nitpicky comment suggestions that should be quick/easy to apply. Thanks for the contribution.

hunhoffe (approved):
Looks good to me once André's comments are addressed & CI is passing! Thank you for this contribution!

One note, from claude:
> One thing worth a second look: the d2size fix in AIEDMATasksToNPU.cpp reads from input_sizes[2], which runs before getHardwareStridesWraps is called — but d0size/d1size right below it read from sizes[0]/sizes[1], the post-getHardwareStridesWraps hardware-encoded array (scaled by element width). For 32-bit element types these are identical, so it's a no-op today, but for non-32-bit types they'll diverge. Since d2size is currently dead (per the // TODO: D2Size comment in AIEDmaToNpu.cpp) this doesn't affect any lowered output now, but the PR comment frames the fix as "correct for whenever D2Size does get wired up" — worth either moving this assignment down next to d0size/d1size and reading sizes[2] instead, or tightening the comment to note it's only index-correct, not encoding-correct, so it doesn't get taken further than that.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
xhmelon: Sure, applied.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
xhmelon: Sure, applied.

lib/Dialect/AIEX/IR/AIEXDialect.cpp
xhmelon: Sure, I have marked the migration sentence with TODO.

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
xhmelon: Sure. I moved the assignment (and this comment with it) per Erika's note below.

xhmelon:
Thanks both for the reviews. All of André's suggestions are applied, and CI is green now.

@hunhoffe Sure. I moved the assignment (with its comment) next to `d0size`/`d1size` so it reads `sizes[2]` from the post-`getHardwareStridesWraps` array.

Two further changes came out of CI, both narrowing the sub-32-bit granularity check:

- Dimensions with size 1 are now skipped. Such a dimension runs its loop once, so its stride is never stepped and neither its size nor its stride reaches the hardware address-generation registers.
- The innermost dimension is now taken to be the last one with size > 1, rather than the last by array position.

Several in-tree BDs use sub-32-bit element types with a multi-dimensional access pattern that collapses to a linear one [1]. There is no unified normalization for these, so each consumer copes on its own. Today they are accepted through a special case in the verifier [2] and happen to lower to registers that still produce correct data [3], but only incidentally. I think the proper fix is a normalization pass that flattens such patterns first, so the existing checks see a clean dim list. I plan to propose that in a follow-up PR.

[1] `programming_examples/basic/matrix_multiplication/cascade` (i16, `sizes = [32, 32, 1, 1]` / `strides = [32, 1, 32, 1]` after objectfifo lowering) and `test/python/dma_op.py` (i16, all dimension sizes 1).

[2] `lib/Dialect/AIE/IR/AIEDialect.cpp:2504-2513` — the innermost dimension is taken to be the last one with size > 1, rather than the last by array position, and dimensions with size 1 are skipped entirely.

[3] The cascade BD lowers through `lib/Targets/AIERT.cpp:477-497`, which scales element-granularity values to 32-bit words by array position. Its emitted registers are D0 Wrap = 0 / StepSize = 1 word, D1 Wrap = 1 / StepSize = 16 words, D2 Wrap = 32 / StepSize = 1 word. Two values are truncated on the way: the innermost size (one i16, half a word) becomes a Wrap of 0, and dim 1's stride (also half a word) truncates to 0 and is then clamped up to one word, twice what was declared. The transfer is still correct because Wrap = 0 makes the access linear, so the clamped stride is never exercised — the first truncation makes the second one irrelevant rather than the two cancelling out.

hunhoffe:
@xhmelon Thanks for more good work! Your plan for the next PR seems good too. I've submitted this to the merge queue, but feel free to remove if you think there's still more work to be done!

PR #3554, opened 2026-08-14
Add AIECC_PATH environment variable override to config.aiecc_path()

hunhoffe:
## Description

- `aie.utils.config.aiecc_path()` previously only checked the bundled MLIR-AIE `bin/` directory, then `PATH` — there was no way to point at a specific `aiecc` explicitly. This came up while unifying IRON's compilation paths, where a consumer needs to pin `aiecc` without relying on `PATH` search order.
- Adds an `AIECC_PATH` env var override, checked first, mirroring the existing `PEANO_INSTALL_DIR` precedent in `configure.py`. An `AIECC_PATH` pointing at a nonexistent file is a hard error rather than a silent fallback.
- Documents the new variable in `programming_guide/iron_configuration.md` alongside the other IRON env vars (`NPU_CACHE_HOME`, `NPU_RUNTIME`, etc.) and in the `aiecc_path()` docstring.

## Checklist

### DIFF
--- a/python/utils/config.py
+++ b/python/utils/config.py
@@ -73,8 +73,8 @@ def aiecc_path():
         return path_aiecc
 
     raise RuntimeError(
-        "Could not find aiecc. Expected it under the MLIR-AIE bin directory, "
-        "on PATH, or set via the AIECC_PATH environment variable."
+        "Could not find aiecc. Resolves in the order of the AIECC_PATH "
+        "environment variable, MLIR-AIE bin directory, then PATH."
     )
 
 
--- a/python/utils/config.py
+++ b/python/utils/config.py
@@ -50,7 +50,20 @@ def root_path():
 
 
 def aiecc_path():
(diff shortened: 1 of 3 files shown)


python/utils/config.py
```diff
@@ -60,8 +73,8 @@ def aiecc_path():
         return path_aiecc
 
     raise RuntimeError(
-        "Could not find aiecc. Expected it under the MLIR-AIE bin directory "
-        "or on PATH."
+        "Could not find aiecc. Expected it under the MLIR-AIE bin directory, "
```
jskimko: Maybe parrot the resolution order from the docstring here? E.g., something like "Resolves in the order of the AIECC_PATH environment variable, MLIR-AIE bin directory, then PATH."

andrej (approved):

python/utils/config.py
hunhoffe: addressed!

PR #3557, opened 2026-08-18
[cert] Add deterministic multi-uC runtime sequencing and verification

fifield:
## Description

Add a device-neutral CERT runtime model for deterministic multi-microcontroller execution.

- Preserve textual job and page order; treat job IDs as labels rather than priorities.
- Add target-model uC mapping, page placement, and lowering to attach_to_group.
- Use the target controller topology from #3593 for group validation and tile-to-controller mapping.
- Emit one ordered stream per actual uC group and unify explicit group 0 with device-level content.
- Form implicit cooperative pages, force the generated configure page first on uC0, and preserve placement through page legalization.
- Add aie-cert-verify for group IDs, local and remote barriers, TCT waits, preemption IDs, rendezvous membership, and firmware barrier budgets; run it in the aiecc CERT pipeline.
- Harden page sizing and splitting, including legal job-boundary cuts, local-barrier co-location, load_pdi/preempt ordering, and TCT-aware implicit boundaries.
- Make uC DMA-chain merging page-aware, ordering-safe, and safe for shared symbols.

This PR is stacked on the AIE2PS target-model foundation in #3593. The history contains six reviewable functional commits plus a small typo fix.

The independent rel_acq_sync operation in #3558 follows this PR in the stack.

## Testing

The PR adds focused lit coverage under Conversion/NpuToCert, Targets/AIETargetUcCert, dialect/AIEX, and Passes/materialize-to-cert; CI is expected to run these suites.

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIENpuToCert.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIENpuToCert.cpp
@@ -647,21 +647,18 @@ struct MergeConsecutiveCertUcDmaWriteDesSyncOps
 
     // Compute the size of the current and previous chains. If their combined
     // data size is greater than the cert page size, then we cannot merge them.
-    uint32_t prevChainSize = 0;
-    for (auto &o : prevChain.getBody().front().getOperations()) {
-      auto bdOp = dyn_cast<AIEX::CertUcDmaBdOp>(o);
-      if (!bdOp)
-        continue;
-      prevChainSize += bdOp.getLength() * sizeof(int);
-    }
-    uint32_t currChainSize = 0;
-    for (auto &o : chain.getBody().front().getOperations()) {
-      auto bdOp = dyn_cast<AIEX::CertUcDmaBdOp>(o);
-      if (!bdOp)
-        continue;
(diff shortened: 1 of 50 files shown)


lib/Dialect/AIEX/Transforms/AIENpuToCert.cpp
```diff
@@ -73,24 +94,25 @@ struct RuntimeSequenceToCertJob : OpConversionPattern<AIE::RuntimeSequenceOp> {
       newJobId = maxJobId + 1;
     }
 
-    // Create page wrapper at the same location as the runtime sequence
+    // Create the job in place of the runtime sequence, WITHOUT a page wrapper.
+    // A later "form implicit pages" step (in cert-legalize-pages) groups
+    // contiguous top-level jobs into pages; explicit cert.page ops are the only
+    // delimiters. CertJobOp legally parents under DeviceOp, so a bare
```
andrej, line 100: "CertJobOp legally parents under DeviceOp" What does that mean

andrej (approved):
I only really looked at the IR/tablegen-level changes, which look reasonable

PR #3560, opened 2026-08-18
[AIE] Track buffer free space exactly, and give each core a real data region

hunhoffe:
## Problem                                                                                                                                                       
                                                                                                                                                                   
Bank-aware allocation tracked free space as a per-bank watermark. A bump pointer cannot represent a hole, so one fixed-address buffer stranded every free byte below it.                                                                                                                                              
                                                                                                                                                                   
Underneath that: a core's own `.data`/`.rodata`/`.bss` share tile memory with its buffers, and the linker script grants exactly one region — the largest gap between the stack and the buffers. Nothing aimed for that gap, and nothing checked it, so a core that did not fit failed late with a bare `will not fit in region 'data'`.                                                                                                                                 
                                                                                                                                                                   
## Changes                                                                                                                                                       
                                                                                                                                                                   
- **Byte-granular bitmap** replaces the watermark, so holes are usable.                                                                                          
- **The data region is a placement objective.** Candidates are scored by the larrgest free run they leave, ranked above bank round-robin. Over the 200 generated designs in `allocator_properties.py`, mean largest-run/free goes **0.58 → 0.85**; bank imbalance also improves, 0.59 → 0.50. The region is stamped as `data_origin`/`data_length`, aligned, and emitted verbatim by `AIETargetLdScript`.                                                                                                                                           
- **The linker decides whether a core fits** — it knows the exact size. aiecc adds the context lld cannot have (where the region came from, what to change). Program memory gets the same treatment; it previously had no check.                                                                                   
- **`reserved_data_size`** is now only an optional floor, for a core whose needs the allocator cannot see. Absent means "leave as much as you can".                                                                                       
- **`initial_value` comes from device configuration only.** It had three delivery paths; on the linker-script path the bytes landed wherever  `*(.data*)` went while the symbol pointed at the buffer — charged to tile memory twice and never read. Memtile buffers are now covered too.                                                                                              
- **`mem_bank`** documented as it behaves: hard under bank-aware, ignored by basic-sequential, which now warns rather than silently dropping the pin.                                                                                       
- **Buffer naming split into `aie-prepare-buffers`**, since a name is a symbol in the core's object and that is separate from placement.                                                                                                      
                                                                                                                                                                   
Also fixes two latent bugs surfaced by clang-tidy re-checking touched files: `AIERT` passed an uninitialized `devGen` to `XAie_Config`, and `AIETargetXAIEV2` emitted `AieGen = ;`, both for `AIE2ps`.                                                                                                       
                                                                                                                                                                   
## Not included                                                                                                                                                  
                                                                                                                                                                   
Chess/BCF parity: that path has no data-region bound, no translated diagnostic, and no `.stack_sizes`, so the stack check no-ops there.


hunhoffe:
I'll try to get this little guy merged first https://github.com/Xilinx/mlir-aie/pull/3662, as a heads up

include/aie/Dialect/AIE/IR/AIECoreMemory.h
```diff
@@ -0,0 +1,70 @@
...
+#include "llvm/ADT/SmallVector.h"
+
+#include <algorithm>
+#include <cstdint>
+#include <utility>
+
+namespace xilinx::AIE {
+
+// A half-open [start, start + size) run of bytes.
+struct MemoryRun {
```
hunhoffe, line 33: Is there not something built-in that gives us something equivalent to this struct? Why is this needed?

include/aie/Dialect/AIE/IR/AIECoreMemory.h
```diff
@@ -0,0 +1,70 @@
...
+//
+// This is the number that decides whether a core links. AIETargetLdScript
+// hands the core compiler exactly one region for its own .data/.rodata/.bss --
+// this run -- so the allocator's reserved_data_size acceptance test and the
+// linker script must agree byte for byte. Both call this rather than sweeping
+// the intervals themselves.
+inline MemoryRun
+largestFreeRun(int64_t memSize,
+               llvm::SmallVector<std::pair<int64_t, int64_t>> occupied) {
+  llvm::sort(occupied);
```
hunhoffe, line 56: Does this sort more than strictly necessary? How do we keep this computationally efficient?

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -380,6 +380,41 @@ def AIE_CoreOp
           // Absent when aiecc cannot measure the core.
           OptionalAttr<
               ConfinedAttr<AIEI32Attr, [IntMinValue<0>]>>:$measured_stack_size,
+          // Bytes of data memory this core's own compiled sections
```
hunhoffe: this documentation is way too verbose. keep it precise but concise.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1650,14 +1685,14 @@ def AIE_BufferOp
...
-        OptionalAttr<StrAttr>:$sym_name,
-        OptionalAttr<AIEI32Attr>:$address,
-        OptionalAttr<ElementsAttr>:$initial_value,
-        OptionalAttr<AIEI32Attr>:$mem_bank,
-        DefaultValuedAttr<BoolAttr, "true">:$aligned
-  );
+  let arguments = (ins Index:$tile, OptionalAttr<StrAttr>:$sym_name,
+      OptionalAttr<AIEI32Attr>:$address,
+      OptionalAttr<ElementsAttr>:$initial_value,
+      // Pin the buffer to a specific L1 memory bank. A hard placement
```
hunhoffe: why did we add comments here? Aren't they described above?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -22,6 +28,87 @@ using namespace mlir;
...
+  return std::nullopt;
+}
+
+// The numbers every allocation scheme needs before it can place anything: how
+// much data memory the tile has, what alignment its load/store bus requires,
+// and the widest vector alignment a core access can demand. A memtile and a
+// compute tile ask the target model for these differently. A memtile is
+// reached by DMA rather than by core vector load/stores, so the bus width
+// covers it.
+struct TileMemoryLimits {
```
hunhoffe, line 46: There's a tension between aligning for vector operations in the core and aligning for BusWidth. A recent PR helped to fine tune alignment of buffers. Can you ensure that this logic is correct in light of the recent updates?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -22,6 +28,87 @@ using namespace mlir;
...
+            targetModel.getMemTileLoadStoreBusWidth()};
+  return {targetModel.getLocalMemorySize(),
+          targetModel.getComputeTileLoadStoreBusWidth(),
+          targetModel.getComputeTileMaxVectorAlignBits()};
+}
+
+// Every buffer here must already have an address: called once allocation has
+// either succeeded or is about to be reported failed, to put the memory map
+// and the final overlap/overflow checks in address order.
+static void sortBuffersByAddress(SmallVectorImpl<BufferOp> &buffers) {
```
hunhoffe, line 64: Can we use a data structure to keep things in sorted order instead of sorting all over the place?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -88,30 +175,38 @@ static bool checkAndPrintBufferOverlap(SmallVector<BufferOp> &sortedBuffers,
...
-    int64_t prevEnd = prevAddr + prev.getAllocationSize();
-    if (curAddr < prevEnd) {
-      cur.emitOpError("buffer '")
-          << cur.name() << "' at address 0x" << llvm::utohexstr(curAddr)
-          << " overlaps with '" << prev.name() << "' at address 0x"
-          << llvm::utohexstr(prevAddr) << " (size: " << prev.getAllocationSize()
-          << " bytes)";
-      return false;
+
+    // Overlap check against the closest buffer below in address order.
```
hunhoffe: Why is this overlap check necessary? Shouldn't it be correct by construction? I do want to add additional verification if needed, but not if it degrades compiler performance (time to compile) unnecessarily.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -125,25 +220,56 @@ static bool checkAndPrintOverlapStackframe(int stacksize,
   return true;
 }
 
+// A reservation the packing can't honour must fail here too, not just in
+// bank-aware: a design whose bank-aware attempt fails for an unrelated reason
+// falls back to basic-sequential, and the shortfall would otherwise surface
+// much later as a linker region-overflow error naming neither the tile nor
+// the buffers responsible.
+static bool checkAndPrintReservedData(TileOp tile, int64_t freeRun,
```
hunhoffe: Do we really need custom prints/checks or is there llvm/mlir machinery we can use here instead of custom stuff?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -184,10 +299,16 @@ static bool basicAllocation(TileOp tile) {
   device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
     if (buffer.getTileOp() == tile) {
       if (buffer.getAddress()) {
-        allocated_buffers.push_back(
-            buffer); // TODO: Right now, this ignore all buffer with only
-                     // mem_bank attribute.
+        allocated_buffers.push_back(buffer);
       } else {
+        // Basic-sequential ignores mem_bank entirely (the bit-packed bump
```
hunhoffe: This seems like burying an error. I just want a double check that this actually works/is true

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -285,25 +402,33 @@ static bool basicAllocation(TileOp tile) {
...
+    return false;
+  stampCoreDataRegion(tile, dataRun);
+  return true;
 }
 
 //===----------------------------------------------------------------------===//
 // SimpleBankAwareAllocation : round-robin each alloc over available banks
 //===----------------------------------------------------------------------===//
+namespace {
 using BankLimits = struct BankLimits {
```
hunhoffe: do we need yet another custom struct?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -312,321 +437,633 @@ static void fillBankLimits(int numBanks, int bankSize,
...
+  // it was marked, so clearing exactly its range is exact rather than
+  // approximate.
+  void markFree(int64_t start, int64_t end) {
+    assert(start >= 0 && end <= size() && start <= end &&
+           "range must lie inside the tile");
+    if (start < end)
+      occupied.reset(start, end);
+  }
+
+  // Placement for `size` bytes in [lo, hi) that leaves the largest single free
```
hunhoffe, line 436: Throughout this file, the comments are excessive.

test/aiecc/reserved_data_size_bcf.mlir
```diff
@@ -0,0 +1,43 @@
+//===- reserved_data_size_bcf.mlir ------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Chess links against a BCF, not the ldscript AIETargetLdScript emits: every
```
hunhoffe: This is.... historical and too long. Keep it short.

test/aiecc/reserved_data_size_mixed_measurable_warns.mlir
```diff
@@ -0,0 +1,34 @@
+//===- reserved_data_size_mixed_measurable_warns.mlir -----------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Only a plain relocatable object can be measured; an archive or a path that
```
hunhoffe: Comments here and in other test files are too long.

test/aiecc/reserved_data_size_undersized.mlir
```diff
@@ -0,0 +1,66 @@
+//===- reserved_data_size_undersized.mlir ----------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// A tile too small to give an auto-measured reservation the contiguous room
```
hunhoffe: Way too much prose in this file.

tools/aiecc/IRTransforms.h
```diff
@@ -376,6 +375,75 @@ inline mlir::LogicalResult checkStackSizeRequirements(
   return result;
 }
 
+// Runtime objects the link pulls in beyond link_files add ~37 bytes on a real
```
hunhoffe: Too much prose in this file.

hunhoffe (commented):
For Claude

Also note -- I think there's some documentation on memory models for cores in programming_guide/ somewhere? Let's make sure that's consistent with the changes in this branch.

andrej:
One note about how `reserved_data_size` behaves after this PR -- I think we might want to consider a follow-on to drop this attribute altogether or make it behave more intuitively, somehow (though it is not clear to me quite yet how we'd do that). Asked AI to write this:

## `reserved_data_size` can shrink the region it protects

**Setup.**

1. The generated linker script grants the core compiler **one** contiguous region for `.data`, `.rodata` and `.bss`. That region is the largest free run between the stack and the tile's buffers. The largest single run therefore decides whether a core links, not the total free memory.
2. `reserved_data_size` makes the bank-aware allocator mark that many bytes occupied before it places the unconstrained buffers. It releases them once placement finishes, and records the region as the largest free run over the stack and the final buffer addresses. The placeholder's own position does not become the region.
3. `findLeastFragmentingGap` chooses that position. It tries the two flush positions of each free hole, scores each by the largest run left behind *at that moment*, and breaks ties by lowest address.

**The effect.**

On a tile with no pinned buffers the free space at step 2 is one hole above the stack. Both flush positions leave the same run, so the tie-break puts the placeholder at the bottom of it, directly above the stack. The buffers then pack above the placeholder. Releasing it frees those bytes *below* the buffers, as an isolated hole, while the recorded region is the largest run *above* them.

The total free space is unchanged. The number of pieces increases, so the region shrinks.

`bank_aware_reserved_data.mlir`, modules one and three, differ only by `reserved_data_size = 8192` on an npu2 core tile:

| | buffer addresses | granted region |
|---|---|---|
| no reservation | 1024, 5120, 9216 | 52224 |
| `reserved_data_size = 8192` | 9216, 16384, 20480 | 40960 |

A request that the layout would satisfy anyway costs 11264 bytes of contiguous region.

aiecc writes the measured `.data` + `.rodata` + `.bss` total to `reserved_data_size` on every measurable core, and that total is usually far below the natural free run. So this is the common path, not a corner case.

**Two fixes, if we take this as a follow-up.** Place the placeholder flush against the top of memory, so its bytes merge with the region on release. Or stop releasing it, and record the placeholder itself as the region — which also makes `data_origin` mean what its name says.

atassis:
I built this branch and main and ran `test/assign-buffer-addresses/allocator_properties.py` against
both, so the numbers below come from your own suite.

Completeness is the headline and it is bigger than the description says: main solves 81 of the 200
generated designs, this branch solves 200. That carries the rewrite on its own.

Three things I would want settled before it merges.

**1. It closes the only case that had stack clearance.**

Placement now packs the first buffer flush against the stack where round-robin used to skip past it.
The signature is a default 1024-byte stack with buffers {8, 16384, 16384} on an npu1 core tile:

```
main:      lowest buffer address 16384, clearance 15360
this PR:   lowest buffer address  1024, clearance 0
```

Across the core linker scripts my builds have produced, 4874 of 4881 already have zero clearance, so
this is not a regression in the common case. What makes it worth a decision is the company it keeps:
#3680, which this branch sits on, carries "Count an unmeasured frame as 0 and report a lower bound"
and "Fall back to a measured frame for symbols the ELF does not report". Where the computed
requirement is exact, clearance buys nothing. Where it is a lower bound, clearance is the only thing
between an under-estimate and the silent corruption #3613 exists to stop.

**2. Bank round-robin is unreachable on core tiles.**

`negLargestRunLeft` is byte-valued and sits above `rrDistance` in the rank tuple, so two candidates
tie on it only when they leave exactly the same number of contiguous bytes, which a byte-granular
bitmap makes rare. Two equal buffers, nothing else on the tile:

| size each | main | this PR |
|---|---|---|
| 4096 B | bank 0, bank 1 | bank 0, bank 0 |
| 2048 B | bank 0, bank 1 | bank 0, bank 0 |
| 1024 B | bank 0, bank 1 | bank 0, bank 0 |

The `ctx.hasCore ? -runLeft : 0` ternary is the switch: the same two buffers on a memtile still land
in banks 0 and 1, because the key is constant there and round-robin survives. If the pair is a
kernel's input and output, loads the scheduler put in one bundle now target one bank.

**3. The suite cannot see (2), so the ratchet will not catch it.**

`MAX_BANK_IMBALANCE` measures byte-occupancy spread across banks, and two co-accessed buffers can
share a bank in a perfectly balanced layout. I added the fraction of same-tile buffer pairs sharing a
bank and re-ran both allocators over the 81 designs both solve, so it is like for like:

| | main | this PR |
|---|---|---|
| contiguity | 0.705 | 0.877 |
| bank imbalance | 0.238 | 0.302 |
| same-bank pairs | 0.283 | 0.479 |

Contiguity improves as designed. Co-residency is 69% worse and the imbalance metric registers only
part of it. A pair-sharing ratchet beside `MAX_BANK_IMBALANCE` would gate this in seconds with no
device. It is a proxy, since the generator models no access pattern, but it tracks what byte spread
does not.

Two smaller notes.

Ordinary buffers take their candidates from `findGap`, which returns the tightest fit, and are then
scored by largest run left. `findLeastFragmentingGap` enumerates both flush positions of every hole
precisely because an interior placement leaves strictly less contiguous space, but it only serves the
data region and pinned buffers. So for ordinary buffers the objective is a tiebreak over a candidate
set that was not generated for it. Using it there may recover contiguity while leaving room to
quantize `negLargestRunLeft`, so round-robin can still discriminate inside the band where the region
already fits.

I could not reproduce the baselines in the description: I get contiguity 0.705 to 0.877 and imbalance
0.238 to 0.302, rather than 0.58 to 0.85 and 0.59 to 0.50, and the imbalance direction inverts.
Probably a different configuration, but worth a note in the body since those numbers carry the bank
claim.

andrej:
Your point 2 @atassis is important. The fact that more buffers share the same bank may have performance implications. We need to look into this further. Thanks for validating on your end!

tools/aiecc/aiecc.cpp
```diff
@@ -915,7 +940,37 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+                    }
+                    if (log.contains("will not fit in region 'program'"))
+                      llvm::errs()
+                          << "aiecc: core " << key
+                          << ": its code exceeds the tile's program memory. "
+                             "That region covers all of the program memory, so "
+                             "only the code itself can shrink. Split the work "
+                             "across more cores, remove unused kernels from "
+                             "link_files, or build the core at a lower "
+                             "optimisation level.\n";
```
```suggestion
                             "optimization level.\n";
```

Why British?
Review all added prose. Use your technical writing skill to clean up.

tools/aiecc/IRTransforms.h
```diff
@@ -329,6 +328,40 @@ inline mlir::LogicalResult checkStackSizeRequirements(
   return result;
 }
 
+// Records what each core's linked sections occupy, and reports a data_size that
+// falls short of them. The linker already refuses a region that is too small,
+// so this catches a declaration that outruns its own core.
```
andrej: "outruns its own core" is unclear

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -380,6 +380,16 @@ def AIE_CoreOp
           // Absent when aiecc cannot measure the core.
           OptionalAttr<
               ConfinedAttr<AIEI32Attr, [IntMinValue<0>]>>:$measured_stack_size,
+          // Bytes to set aside for the core's own .data, .rodata and .bss.
+          // aie-assign-buffer-addresses turns this into a `core_data`
+          // aie.buffer, places it like any other buffer, and AIETargetLdScript
+          // emits it as the `data` MEMORY region. Absent leaves the region to
+          // whatever contiguous run the buffers leave.
```
andrej:
```suggestion
          // emits it as the `data` MEMORY region. Absent (0) leaves the region to
          // the largest remaining contiguous run after placing all buffers.
```

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -93,6 +93,24 @@ def AIEPlaceTiles : Pass<"aie-place-tiles", "DeviceOp"> {
...
+    Settles everything about a buffer that the core compilation depends on,
+    except the buffer's address.
+
+    A buffer's name becomes a symbol in the core's object, so the name must
+    exist before the core is lowered. This pass names every unnamed buffer and
+    checks buffer placement, which rejects a malformed design before any core
+    is compiled.
+
+    aie-assign-buffer-addresses assigns the addresses. It runs once the core
+    objects exist, so it can measure the space each core's own sections need.
```
andrej: Is this still true now that we measure post-link? If not, please strike

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2446,6 +2446,13 @@ LogicalResult CoreOp::verify() {
                  << "' appears in both 'link_files' and 'link_merge_files'; an "
                     "artifact must be either merged or linked, not both";
     }
+  // The core's own sections live in a `core_data` aie.buffer, so the buffer
+  // verifier covers their placement. Only the size belongs here.
+  if (auto measured = getMeasuredDataSize())
+    if (auto declared = getDataSize(); declared && *declared < *measured)
+      return emitOpError("data_size ")
+             << *declared << " is smaller than the " << *measured
+             << " bytes this core's linked sections occupy";
```
andrej, line 2455: Multiple lines but no enclosing braces is hard to read. Add enclosing braces to all blocks, even if they are semantically one line, across the entire change set. You may be able to use clang format for this.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -22,34 +29,110 @@ using namespace mlir;
...
+  OpBuilder builder = OpBuilder::atBlockTerminator(device.getBody());
+  for (auto core : llvm::to_vector(device.getOps<CoreOp>())) {
+    int64_t size = core.getDataSize().value_or(0);
+    if (size <= 0)
+      continue;
+    auto tile = cast<TileOp>(core.getTile().getDefiningOp());
+    bool present = false;
+    device.walk([&](BufferOp buffer) {
+      if (buffer.getCoreData() && buffer.getTileOp() == tile)
+        present = true;
```
andrej: In a verifier somewhere, make sure we error if there are multiple buffers that have the core data attribute set. There can only be one, which places like this in the code assume.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -785,27 +1090,92 @@ struct AIEAssignBufferAddressesPass
...
+          break;
+        case BankAwareResult::ConstraintUnsatisfiable:
+          // Can't retry under basic-sequential (see BankAwareResult).
+          emitAllocationFailure(tile, "bank-aware");
+          return signalPassFailure();
+        case BankAwareResult::OutOfMemory: {
+          // A mem_bank-only buffer would silently lose its bank under
+          // basic-sequential, even when its own pin is satisfiable and
+          // bank-aware failed for an unrelated reason. Only an address pin is
+          // safe to retry, because basic-sequential honours it.
```
andrej: honours?!

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -742,6 +1044,25 @@ static LogicalResult checkBufferScope(BufferOp buffer, DeviceOp device) {
...
+    if (auto size = core.getDataSize())
+      declared = *size;
+  InFlightDiagnostic diag = tile.emitOpError(scheme) << " allocation failed";
+  if (!declared)
+    return;
+  diag << ". Core (" << tile.getCol() << ", " << tile.getRow() << ") reserves "
+       << *declared
+       << " bytes for its static data (data_size: constant arrays such as "
+          "lookup tables and strings), which has to fit alongside this tile's "
+          "buffers. Lower data_size on the core, or shrink or move the buffers";
```
andrej: Missing period after last sentence.

tools/aiecc/aiecc.cpp
```diff
@@ -601,6 +601,28 @@ EdgeWithTypedOutput<NpuProgram> &buildNpuProgramSubgraph(
...
+    if (size_t pos = text.find("data (!RX) : ORIGIN =");
+        pos != llvm::StringRef::npos) {
+      llvm::StringRef tail = text.drop_front(pos);
+      if (size_t len = tail.find("LENGTH = 0x"); len != llvm::StringRef::npos)
+        tail.drop_front(len + strlen("LENGTH = 0x")).consumeInteger(16, length);
+    }
+  }
+  return length + over;
+}
+
```
andrej: We're just parsing the log to get this info??!?!?!?!?!?!?!

Hard-coding workDir + "/ldScripts_" + key + ".ld.script"??!?!?!?!?!!?!!!!

Including this function in the main aiecc.cpp rather than in one of the helper files?

UGH!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

tools/aiecc/StackSizeAnalysis.cpp
```diff
@@ -394,3 +394,30 @@ StackRequirementResult xilinx::aiecc::computeStackRequirement(
   llvm::sort(result.unmeasured);
   return result;
 }
+
+std::optional<int64_t>
+xilinx::aiecc::measureDataSectionBytes(llvm::StringRef elfPath) {
```
andrej, line 434: This is the proper way to measure data section size. Not by parsing the output log/string like you did in requiredDataBytes. Why didn't you just reuse this function?

andrej (commented):
...

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -785,27 +1093,92 @@ struct AIEAssignBufferAddressesPass
...
+        return signalPassFailure();
+    });
+    // A buffer's name becomes a symbol in its core's object, so every buffer
+    // needs a name before the cores are lowered.
+    OpBuilder builder = OpBuilder::atBlockTerminator(device.getBody());
+    int counter = 0;
+    device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
+      if (!buffer.hasName()) {
+        std::string name = "_anonymous";
+        name += std::to_string(counter++);
```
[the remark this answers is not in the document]
andrej: yes ...

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -487,29 +487,30 @@ struct AIEUseLockToStdLowering : OpConversionPattern<UseLockOp> {
...
+    // script or BCF. `initial_value` is device state, written by whoever
+    // configures the device -- the CDO writer, or mlir_aie_initialize_buffers
+    // on the simulator path. A definition here would place the bytes wherever
+    // `*(.data*)` lands, which costs the core's data region a copy it never
+    // reads.
     memref::GlobalOp::create(rewriter, buffer.getLoc(), symName,
                              rewriter.getStringAttr("public"), buffer.getType(),
-                             initValue, /*constant*/ false,
+                             /*initial_value=*/nullptr, /*constant*/ false,
                              /*alignment*/ nullptr);
```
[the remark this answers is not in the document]
andrej: Why was this change needed at all?

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -160,91 +263,72 @@ static bool basicAllocation(TileOp tile) {
...
+        // This scheme packs from the bottom of the tile and has no notion of
+        // banks, so it clears mem_bank: the label would name a bank the
+        // address need not fall in. A user pin reaches this only under an
+        // explicit basic-sequential request, because the fallback path refuses
+        // to drop one (see BankAwareResult::OutOfMemory).
+        if (buffer.getMemBank())
+          buffer->emitWarning("basic-sequential allocation ignores mem_bank; "
+                              "dropping the pin on: ")
+              << buffer.name();
+        buffer->removeAttr("mem_bank");
```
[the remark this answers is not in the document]
andrej: We should probably also have a verifier that errors if mem_bank and address are both set but disagree (i.e. address is not in that mem_bank).

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -284,353 +358,605 @@ static bool basicAllocation(TileOp tile) {
...
+        size == 0 ? 1 : (*addr + size - 1) / bankSize - *addr / bankSize + 1;
+    // This placement splits one run into the piece below and the piece above.
+    // Every other run stays whole, so the biggest of those is whichever of the
+    // top two this candidate does not consume.
+    MemoryRun gap = occupancy.gapAt(*addr);
+    int64_t elsewhere = gap.size == largestRun ? secondRun : largestRun;
+    int64_t leftPiece = *addr - gap.start;
+    int64_t rightPiece = gap.end() - (*addr + size);
+    int64_t runLeft =
+        std::min(std::max({elsewhere, leftPiece, rightPiece}), contiguityCap);
```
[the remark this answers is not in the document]
andrej: No -- we decided bank balancing was an important concern. Once we know there will be "enough space left" (and if no data size is reserved, zero is enough), there is no benefit in maximizing the remaining space further -- we don't use any further space! So at that point a free block that improves bank balance is preferrable. Make sure this is clear in all code documentation.

tools/aiecc/aiecc.cpp
```diff
@@ -939,10 +995,13 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
-                       out.value->get(), [&](CoreOp coreOp) -> std::string {
-                         return elfByKey.lookup(coreKey(coreOp));
-                       });
+                   auto elfFor = [&](CoreOp coreOp) -> std::string {
+                     return elfByKey.lookup(coreKey(coreOp));
+                   };
+                   if (mlir::failed(checkStackSizeRequirements(out.value->get(),
+                                                               elfFor)))
+                     return mlir::failure();
+                   return checkDataSizeRequirements(out.value->get(), elfFor);
```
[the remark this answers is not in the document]
andrej: Probably best to make data size check a separate edge instead, and make that edge conditional on its own --no-measure-data-size flag... I'd prefer that over a conditional inside an uncoditional edge.

tools/aiecc/aiecc.cpp
andrej: (Claude) Fixed in 5e7f12a. All three points:

- **Hard-coded script path.** Gone. The region size now comes from the IR. `coreDataRegion(tile, buffers)` is a new shared helper in `AIETargetShared`; the ld-script edge records its result per core key, and the failure hint reads that. `AIETranslateToLdScript` calls the same helper, so the script and the diagnostic can no longer disagree.
- **Location.** `requiredDataBytes` is deleted from `aiecc.cpp`. What survives is `parseLinkOverflowBytes` in `StackSizeAnalysis.cpp`, beside `measureDataSectionBytes`.
- **Parsing the log.** Still parsing, and I want to flag that rather than hide it. See my reply on your other comment for why.

tools/aiecc/StackSizeAnalysis.cpp
andrej: (Claude) You are right that `measureDataSectionBytes` is the proper measurement, and it is what writes `measured_data_size` on every successful link.

It cannot serve this path: a failed link writes no ELF. The object files are the only thing on disk, and measuring those is the magika bug this PR exists to fix, since `--gc-sections` has not run.

In 5e7f12a the only thing read from the log is the linker's `overflowed by N`, moved into `parseLinkOverflowBytes` in `StackSizeAnalysis.cpp`. The region size no longer comes from the script text; it comes from the IR through the new shared `coreDataRegion` helper. So `required = region + overflow`, with one number from the IR and one from the linker.

If you want the log read gone entirely, the alternative is to relink on the failure path against a permissive `data` region and then call `measureDataSectionBytes` on the result. That costs one extra link, only when the build is already failing, and would reuse the function exactly as you ask. Say the word and I will do that instead.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Added in 5e7f12a. `BufferOp::verify` now rejects a second `core_data` buffer on the same tile: "is a second core_data buffer on tile (C, R); the core's data region is one extent". It skips a buffer on a `LogicalTileOp`, where no tile is chosen yet.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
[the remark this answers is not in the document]
andrej: (Claude) Added in 5e7f12a, and it found something. `BufferOp::verify` now errors when both attributes are set and the address is outside the requested bank: "address 0x1000 lies in bank 0, but mem_bank requests bank 2".

Two consequences:
- The allocator already had this check (`mem_bank attribute is inconsistent with address attribute`). The verifier now catches it first, so I deleted the copy in `checkAndAddBufferWithAddress` and updated the two tests that expected the old message.
- A zero-sized buffer pinned at the top of the tile holds the one-past-the-end address, which is in no bank. The check skips buffers of size 0, which `bank_aware_alloc_gaps.mlir` exercises.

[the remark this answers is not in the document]
andrej: (Claude) Fixed in 5e7f12a. The walk in `basicAllocation` now clears `mem_bank` and warns for every buffer on the tile, before it branches on whether an address is present. Previously a buffer carrying both kept the bank silently.

Note that the disagreement case you and @andrej raise is now also a verifier error, so a buffer whose address and bank contradict each other never reaches this pass.

tools/aiecc/aiecc.cpp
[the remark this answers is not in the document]
andrej: (Claude) Fixed in 5e7f12a, taking @andrej's preferred shape: two separate edges rather than a conditional inside one.

`measured_stack_sizes.mlir` runs `checkStackSizeRequirements` under `--no-measure-stack-size`, and a new `measured_data_sizes.mlir` edge runs `checkDataSizeRequirements` under a new `--no-measure-data-size`. Each chains onto the previous, so either, both or neither can be skipped.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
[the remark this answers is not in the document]
andrej: (Claude) Fixed in 5e7f12a. `AIEPrepareBuffersPass` now calls `generateUniqueSymbolName(device, "_anonymous", counter)`.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -785,27 +1093,92 @@ struct AIEAssignBufferAddressesPass
...
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<func::FuncDialect>();
+    registry.insert<AIEDialect>();
+  }
+
+  void runOnOperation() override {
+    DeviceOp device = getOperation();
+    device.walk<WalkOrder::PreOrder>([&](BufferOp buffer) {
+      if (failed(checkBufferScope(buffer, device)))
+        return signalPassFailure();
```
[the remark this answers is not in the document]
andrej: (Claude) Fixed in 5e7f12a by swapping the order: the pass names every unnamed buffer first, then walks again to check scope. `checkBufferScope` can no longer reach `name()` on a buffer without `sym_name`.

tools/aiecc/Actions.h
```diff
@@ -527,8 +538,11 @@ struct ShellCommand {
         // Move off the live --progress status line before the tool's output.
         if (progress)
           llvm::errs() << '\n';
-        if (auto buf = llvm::MemoryBuffer::getFile(logPath))
+        if (auto buf = llvm::MemoryBuffer::getFile(logPath)) {
           llvm::errs() << (*buf)->getBuffer();
+          if (failureHint)
+            failureHint((*buf)->getBuffer());
```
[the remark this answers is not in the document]
andrej: (Claude) Already handled, and I confirmed it against the current code rather than assuming.

`Actions.h` captures when `(!verbose || failureHint)`, so a command carrying a hint is captured even under `--verbose`, and the capture is replayed in full in that case. The hint fires on `rc != 0 && failureHint`. No behavior change needed.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -125,25 +218,56 @@ static bool checkAndPrintOverlapStackframe(int stacksize,
...
+// can't drift apart.
+static void printMemoryMapEntry(Diagnostic &note, StringRef name,
+                                int64_t address, int64_t size, int indent,
+                                StringRef suffix = "") {
+  for (int i = 0; i < indent; ++i)
+    note << "\t";
+  note << name << " \t"
+       << ": 0x" << llvm::utohexstr(address) << "-0x"
+       << llvm::utohexstr(address + size - 1) << " \t(" << size << " bytes)"
+       << suffix << "\n";
```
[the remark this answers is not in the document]
andrej: (Claude) Already handled. `printMemoryMapEntry` computes `int64_t end = size == 0 ? address : address + size - 1;`, so a zero-sized buffer prints its own address rather than wrapping.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -6,11 +6,15 @@
...
 //===----------------------------------------------------------------------===//
 
+#include "aie/Dialect/AIE/IR/AIECoreMemory.h"
+#include "aie/Dialect/AIE/IR/AIECoreSymbols.h"
 #include "aie/Dialect/AIE/IR/AIEDialect.h"
 #include "aie/Dialect/AIE/Transforms/AIEPasses.h"
 
 #include "mlir/IR/Attributes.h"
 
+#include "llvm/ADT/BitVector.h"
```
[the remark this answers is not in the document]
andrej: (Claude) Already handled: `AIEAssignBuffers.cpp` includes `<optional>` directly at line 18.

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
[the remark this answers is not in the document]
andrej: (Claude) Investigated, and the hole is real, so I closed it in 5e7f12a.

@andrej asked why the change was needed. `AIEBufferToStandard` used to emit a `memref.global` **with** the initializer for the core that owned the buffer. That places the bytes wherever `*(.data*)` lands, which spends the core's data region on a copy nothing reads: the device gets `initial_value` from the CDO writer or `mlir_aie_initialize_buffers`, not from the ELF. Before this PR that waste was invisible because the region was leftover space. Now the region is sized and measured, so it inflates `measured_data_size` and can fail a link that should fit.

On the nested case: `checkBufferScope` permitted a buffer nested anywhere as long as it had an `initial_value`, and `collectBuffers` only walks `device.getOps<BufferOp>()`, so such a buffer got no linker-script or BCF symbol. I removed the `initial_value` escape, so a buffer nested in a core is now rejected outright. `MemTileDMAOp` stays allowed, since a memtile links no core and needs no symbol.

I checked the whole tree first: the only `aie.buffer` inside an `aie.core` anywhere is in `test/assign-buffer-addresses/bad_buffer.mlir`, which exists to test this rejection. No working design is affected.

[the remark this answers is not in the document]
andrej: (Claude) Answered in detail on the Copilot thread above. Short version: an initializer on the buffer's `memref.global` lands in the core's `.data`, consuming the data region for bytes nothing reads, since the device is initialized from the CDO writer rather than from the ELF. That was invisible while the region was leftover space; now it inflates `measured_data_size` and can fail a link that should fit.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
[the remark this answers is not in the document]
andrej: (Claude) Understood, and the documentation now says exactly that. Changed in 5e7f12a and the commit before it:

- The ranking comment in `setBufferAddress` no longer claims placement maximizes the gap. It states that criterion 2 keeps one run wide enough for the extents still to place, that `contiguityCap` holds it to that much, and that beyond the cap criterion 3 governs so buffers spread across banks.
- `programming_guide/core_data_memory.md` says free space beyond the bytes still to place serves nothing, so the ranking counts a run only up to that amount.
- Three test comments (`bank_aware_alloc_simple`, `bank_aware_vector_alignment`, `bank_aware-alloc-large-stack-address`) still described the old "maximize the free run" behavior and now describe the round-robin result.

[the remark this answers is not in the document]
andrej: (Claude) Per @andrej above, the cap is the intended behavior, not a regression: once enough space is left for what remains to place, further contiguity buys nothing and bank balance is the better use of the choice. I have rewritten the ranking comment, the programming guide and three test comments so the code no longer claims to maximize the run.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
```diff
@@ -284,353 +357,631 @@ static bool basicAllocation(TileOp tile) {
...
-    auto printbuffer = [&](StringRef name, int address, int size) {
-      note << "\t" << "\t" << name << " \t" << ": 0x"
-           << llvm::utohexstr(address) << "-0x"
-           << llvm::utohexstr(address + size - 1) << " \t(" << size
-           << " bytes)\n";
+  // Score every candidate placement and keep the best, ranked by:
+  //  1. banks touched, fewest first (each spanned bank costs DMA bandwidth);
+  //  2. largest free run left behind, biggest first;
+  //  3. round-robin distance from the cursor, nearest first (spreads for DMA);
+  //  4. slack, tightest first, to keep large clean regions available;
```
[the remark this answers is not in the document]
andrej: (Claude) Already handled. The candidate score is `ctx.hasCore ? -runLeft : 0`, so criterion 2 is identically zero on a memtile and round-robin governs there. The ranking comment states this.

tools/aiecc/aiecc.cpp
andrej: (Claude) Fixed. `optimisation` -> `optimization` in both the aiecc diagnostic and the programming guide, plus `behaviour`, `maximises`, `neighbours` and `honours` elsewhere in the change set. I swept every added line against a list of common British forms; it now returns nothing.

tools/aiecc/IRTransforms.h
andrej: (Claude) Reworded. It now reads: "The linker already refuses a region too small to hold the sections, so this catches a reservation the core never fills."

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Fixed, period added.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Fixed, `honours` -> `honors`.

lib/Dialect/AIE/IR/AIEDialect.cpp
andrej: (Claude) Done in 5e7f12a. I used clang-format as you suggested: ran it with `InsertBraces: true` over `AIEAssignBuffers.cpp`, `AIETargetLdScript.cpp`, `AIETargetShared.cpp`, `aiecc.cpp`, `IRTransforms.h`, `StackSizeAnalysis.cpp` and `Actions.h`, then restored the normal config. Every multi-line block in the change set now has braces. Rebuilt and the tests still pass.

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: (Claude) Applied your wording. The attribute comment is now the two lines you suggested.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
andrej: (Claude) Struck. You are right that it is no longer true: the allocator runs before any core object exists, and the measurement is post-link in aiecc against `measured_data_size`. Those two sentences are gone from the `aie-prepare-buffers` description.

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: (Claude) Trimmed. `data_size` is now two lines (@andrej supplied the wording), and the `core_data` paragraph on `AIE_BufferOp` went from four lines to three by dropping the restatement of what the allocator and the ld-script emitter do with it.

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: (Claude) These are gone. The per-argument comments in the `let arguments` list no longer exist, and `core_data` is described once, in the op description above.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Partly addressed, and I want to be accurate about how far.

What I did in this round: deleted or shortened the comment blocks that had gone stale against the current design, since a wrong comment costs more than a long one. The ranking-criteria block, the `largestFreeRun` header, the `splitLoweredCores` header (21 lines to 9), `checkBufferScope`, `bufferLabel`, `materializeCoreDataBuffers` and several test headers.

What I did not do: a systematic pass over every comment in the file to cut length for its own sake. If you want that, I would rather do it as its own commit so it is reviewable as pure deletion, separate from the behavior changes here.

tools/aiecc/IRTransforms.h
andrej: (Claude) Partly addressed. The `splitLoweredCores` header was the worst offender at 21 lines and is now 9, with the paragraph that described removed behavior deleted. I have not gone through the rest of the file for length. Happy to do that as a separate deletion-only commit if you want it in this PR.

test/aiecc/reserved_data_size_bcf.mlir
andrej: (Claude) This file is now `test/aiecc/data_size_bcf.mlir` (renamed with the attribute). The header is down to three lines stating what BCF does with `data_size`. The historical framing is gone.

test/aiecc/reserved_data_size_mixed_measurable_warns.mlir
andrej: (Claude) This file no longer exists under that name; the `reserved_data_size` attribute became `data_size` and the mixed-measurable case went away when the measurement moved post-link. The surviving aiecc tests have three- to five-line headers.

test/aiecc/reserved_data_size_undersized.mlir
andrej: (Claude) This file no longer exists; it was removed when the measurement moved post-link and the check became verification-only. The replacement, `test/aiecc/data_size_measured_after_link.mlir`, has a four-line header.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2446,8 +2446,37 @@ LogicalResult CoreOp::verify() {
...
+  if (originAttr && lengthAttr) {
+    int64_t origin = *originAttr;
+    int64_t length = *lengthAttr;
+    int64_t stackSize = getEffectiveStackSize();
+    // A zero-length region spans no bytes, so it has no placement to check.
+    if (length > 0) {
+      if (origin < stackSize)
+        return emitOpError("data region at 0x")
+               << llvm::utohexstr(origin) << " starts below the stack ("
+               << stackSize << " bytes)";
```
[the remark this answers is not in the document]
andrej: (Claude) Obsolete. `data_origin` and `data_length` no longer exist. The region is a `core_data` `aie.buffer`, so the ordinary buffer alignment and placement rules cover it and there is no second representation to keep aligned. The alignment you describe is applied where the region is computed, in `largestFreeRun`, using `getComputeTileMaxVectorAlignBits() / 8`.

tools/aiecc/aiecc.cpp
```diff
@@ -672,20 +668,43 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
         return populateDefaultStackSize(mod.get(), stackSize);
       });
 
-  auto &withAddresses = withDefaultStackSize.map<ModRef>(
-      "input_with_addresses.mlir",
+  // reserved_data_size auto-measurement below walks each core's link_files
+  // objects, so link_files has to exist before it runs. This edge hoists
+  // aie-assign-core-link-files out of getInputWithAddressesPipeline, where it
+  // sits after buffer address assignment. The pass analyzes the call graph
+  // over func.func and func.call, and no part of it reads addresses.
```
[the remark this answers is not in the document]
andrej: (Claude) Obsolete. The design changed after this review: `data_size` is never auto-populated by design. aiecc measures the linked ELF into `measured_data_size` and errors when `data_size` is smaller. The opt-out you asked for exists as the new `--no-measure-data-size`.

include/aie/Dialect/AIE/IR/AIECoreMemory.h
andrej: (Claude) `MemoryRun` is a plain `{start, size}` pair with `end()` and `contains()`. I am not aware of an LLVM or MLIR type that carries exactly this and is used for tile-local byte ranges; `llvm::Range`-style helpers are iterator ranges, not address extents.

That said I have not gone looking properly, and this is not something I addressed in this round. If you know of one you would prefer, name it and I will convert.

include/aie/Dialect/AIE/IR/AIECoreMemory.h
andrej: (Claude) Not addressed in this round, and I do not want to claim otherwise.

The sorts I am aware of: `placementOrder` sorts buffers once per tile; `topTwoGaps` is a linear scan, not a sort; the ld-script and BCF emitters sort by address for output order. So the per-candidate inner loop does not sort. I have not profiled it, so I cannot tell you it is optimal.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not addressed in this round. Related to your other comment on sorting. If you want the occupancy structure replaced with something that maintains order incrementally, I would rather do that as its own change with the property suite (`test/assign-buffer-addresses/allocator_properties.py`) as the guard, since it has a stress case and a determinism check that would catch a regression.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not verified against the recent alignment PR in this round, and I do not want to imply I checked it.

What the code does today: a pinned address is checked against the bus width (`tileAlignBitWidth / 8`) and only when `aligned` is set, while addresses this pass chooses use the stricter vector alignment. The data region start uses `getComputeTileMaxVectorAlignBits() / 8`, because the linker starts `.data` at its strongest section alignment.

Point me at the PR and I will reconcile the two.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not addressed in this round.

For context on cost: `checkAndPrintBufferOverlap` runs once per tile over the tile's buffers after they are sorted by address, so it is a linear neighbor scan, not quadratic. The comment above it calls it a backstop and says placements come from free space, so you are right that it should be correct by construction on the bank-aware path.

I would rather not delete it blind. If you want it gone, the honest version is to delete it and let the property suite's legality check (which verifies non-overlap independently over 200 seeded designs) be the guard.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not addressed in this round. The memory-map printing builds a `Diagnostic` note with tab-separated rows, which I am not aware of an MLIR helper for. If there is one you have in mind, name it and I will convert.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not addressed in this round, and I did not want to answer without checking, which I have not done. Flagging it as still open rather than guessing.

lib/Dialect/AIE/Transforms/AIEAssignBuffers.cpp
andrej: (Claude) Not addressed in this round. The struct at that line is the placement `Candidate`, which exists to make the five ranking keys a single `std::tie` comparison. I did not attempt to remove it.

PR #3564, opened 2026-08-19
ci: extract code coverage from lintAndFormat into a nightly workflow [MED]

atassis:
## Problem

`lintAndFormat` is one job wearing a workflow. On the current `upstream/main` YAML (checked, not
assumed): four jobs are `clang-tidy-pylint`, `formatting`, `license-compliance`, `docs-build`, all
under ~4 min combined, plus `code-coverage`, which dominates whenever its build step actually runs.
Sampled from the last 5 real runs (`gh run view --json jobs`, 2026-08-14):

| run | event | coverage job |
|---|---|---|
| 31745217330 | pull_request | 22.7 min (18.4 min in "Build and generate coverage") |
| 31752403247 | pull_request | 11.8 min |
| 31743499276 | pull_request | 3.5 min (build step skipped -- see below) |
| 31756736903 | schedule | 3.4 min (build step skipped) |

The 2026-08-04 measurement (`docs/log/2026-08/2026-08-04-ci-critical-path-per-workflow.md`) put the
build step itself at 22.23 min on a full diff; that still holds -- the 22.7 min run above is the same
shape. It's a reporting artifact, not a gate, sitting on every PR's presubmit path.

Also: `lintAndFormat` is the only PR-triggered workflow with no `concurrency` block (checked
`.github/workflows/build*.yml` -- all 8 have one). Every superseded coverage run today runs to
completion instead of getting cancelled.

Correctness finding, not just relocation: the coverage job scopes
`INSTRUMENTED_COVERAGE_FILES` to `git diff --name-only origin/main HEAD`, which is empty whenever
`HEAD == origin/main` -- i.e. on every `schedule` run today. Confirmed on run 31756736903: `Build and
generate coverage (Release)` through `Update PR with coverage results` all show `skipped`. Moving the
job as-is would carry that no-op into the new nightly trigger.

## Fix

Two commits.

1. Add the standard `concurrency` block to `lintAndFormat.yml`, same group/key pattern as
   `buildAndTestRyzenAI.yml`.
2. Move `code-coverage` into a new `.github/workflows/codeCoverage.yml` on `schedule` (daily) +
   `workflow_dispatch`, dropping the diff-scoping instead of carrying it forward:
   - Removed the `Get changed files` step and every `if: steps.changed-files...` gate -- the nightly
     run always builds.
   - Removed the `-DINSTRUMENTED_COVERAGE_FILES=...` override so it falls through to
     `CMakeLists.txt`'s own default (`lib/`, `include/`, `tools/`, `python/`), giving a genuine
     whole-repo report instead of an empty one.
   - Dropped the byte-identical duplicate `Install our python reqs` step (a leftover of the same
     if-gate as the step above it).
   - Fixed the dangling ccache key (`${{ matrix.ubuntu_version }}` / `${{
     steps.get-llvm-commit-hash.outputs.hash }}` -- neither exists in this job, both evaluated to
     empty) to `${{ runner.os }}-code-cov`.
   - Swapped `Update PR with coverage results` for an `upload-artifact` step -- there's no PR to
     comment on from a scheduled run.

## Test

Both files parse (`python3 -c "yaml.safe_load(open(f))"` on each). `grep -rn code-coverage .github/` finds the job only in
the new file. Not verifiable without CI: an actual
scheduled/`workflow_dispatch` run of `codeCoverage.yml`.


hunhoffe (approved):

hunhoffe:
Follow on PR: https://github.com/Xilinx/mlir-aie/pull/3566

PR #3565, opened 2026-08-19
[AIE] Split the placer's DMA-exhaustion diagnostic from the merge lockout [LOW]

atassis:
## Problem

`SequentialPlacer::findTileWithCapacity` scans every tile of the requested type on the whole device -- `targetCol` only orders candidates by distance, it never excludes one -- so when it returns no candidate, there are exactly two possible causes: (a) no tile of that type anywhere has the requested channels free, or (b) at least one does, but every such tile is already claimed by a different `aie.logical_tile` under `merge-logical-tiles=false`. The current message conflates both into "pin this to a column with available DMA budget, or rebalance compute peers", which is unactionable for (a): there is no column with budget, and rebalancing compute peers cannot create shim/memtile DMA channels.

I hit this row-tiling a design that fills all 16 shim MM2S channels on npu2 (8 ShimNOCTiles x 2 channels/direction). The note sent me looking for a placement fix when the real answer was fan out through a MemTile.

## Fix

Split the diagnostic. For (a), report the actual per-type channel budget and usage instead of suggesting a pin, e.g. `"no ShimNOCTile on the device has 0 input/1 output DMA channel(s) free: all 1 ShimNOCTile(s) are at 0/2 input, 2/2 output channels used"` on the single-column `shimnoc_exhaustion` test, and point at MemTile fan-out or reducing DMA-attached ObjectFIFOs. For (b), keep the "no X has sufficient DMA capacity" wording (still accurate -- the tile the search accepted really is short on channels) but replace the note with the actual cause: every tile with spare capacity already hosts another logical tile, so the fix is enabling merge or pinning one-per-column, not chasing a busier column.

Diagnostics only; no placement decision changes.

## Test

Confirmed both branches are real using the existing suite rather than adding a synthetic repro: `test_place_errors.mlir`'s `memtile_exhaustion`/`shimnoc_exhaustion` (npu1_1col, one physical tile of each type, genuinely zero spare capacity anywhere) exercise (a); `test_place_merge_logical_tiles_overflow.mlir`'s `nomerge_overflow` (4 ShimNOC columns, 5 unhinted logical tiles, `merge-logical-tiles=false`) exercises (b) -- each ShimNOCTile there still has 1 of 2 output channels free. Updated the CHECK lines in both files to the new, distinct wording per branch.

Built `aie-opt` standalone against the pinned MLIR/LLVM distro and ran every `RUN:` line in all 24 `test/place-tiles/*.mlir` files (not just the 2 touched ones) through it with `FileCheck`: 32/32 pass, including the two updated files and the 22 untouched ones. clang-format clean at both 17.0.1 and 22.1.8.

### DIFF
--- a/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
+++ b/include/aie/Dialect/AIE/Transforms/AIEPlacer.h
@@ -247,6 +247,24 @@ private:
                                              int requiredOutputChannels,
                                              AIETileType requestedType);
 
+  // Diagnosis for a failed findTileWithCapacity() call. `findTileWithCapacity`
+  // scans every tile of `requestedType` on the device -- `targetCol` orders
+  // candidates, it does not exclude any -- so a failure is either (a) no
+  // tile of this type has the requested channels free ANYWHERE (device-wide
+  // DMA exhaustion: pinning a column cannot create capacity), or (b) at
+  // least one tile has them free but is already claimed by a different
+  // non-core logical tile (only possible under merge-logical-tiles=false).
+  // `capacityExistsSomewhere` selects between the two; the totals let the
+  // caller report the real budget instead of guessing at it.
+  struct CapacityDiagnosis {
+    bool capacityExistsSomewhere = false;
+    int tilesOfType = 0;
+    int inUsed = 0, inMax = 0, outUsed = 0, outMax = 0;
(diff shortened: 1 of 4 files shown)


hunhoffe (approved):
Thank you for another contribution!

PR #3567, opened 2026-08-20
[python] hostruntime: open the NPU once per process, not once per tensor [MED]

atassis:
## Problem

`XRTTensor` opens its own `pyxrt.device(0)` when no handle is passed
(`xrtruntime/tensor.py`). That is an open, not a lookup: `get_userpf_device`
calls into the shim on every construction and returns a handle whose deleter
closes the device again, and the shim refcounts the fd, so the last handle
going out of scope closes `/dev/accel/accel0` and the next allocation reopens
it. A test or example that allocates tensors in a loop therefore opens and
closes the device in a loop.

That reopen is what fails. On the aie2-4col runner it returned `ENODEV`
mid-suite and took out 27 of `test_torch_comparison.py`'s XRTTensor cases,
while the 1127 other tests in the same job passed and `xrt-smi` found the
device at the start of the step.

#3100 moved runtime-owned insts BOs off a fresh `xrt.device(0)` for the same
ownership reason; user tensors kept it.

## Fix

`acquire_device()` in a new `xrtruntime/device.py` opens a device index once
per process and reuses the handle. `XRTTensor` uses it when no `xrt_device=`
is supplied; an explicitly supplied handle is still used as given.

`XRTHostRuntime.__init__`'s retry loop moves into the same function, so there
is one implementation of "open device 0" rather than two that drift. Retrying
at all is #2814; a failed open is not cached, so a device that comes back is
usable rather than poisoned for the process.

A handle held for the life of the process also cannot be outlived by the
buffers derived from it, which is the ownership order #3100 needed on Windows.

## Evidence

Measured on npu2 (Gorgon Point), pinned, minimum of many runs:

| | |
|---|---|
| full open/close cycle of the device | 19.0 ms |
| construction with the fd already held | 2.24 us |

Allocating 200 tensors goes from 200 device opens to 1, and from 17.67 to
14.83 us per tensor when the tensors are held.

## Test

`test/python/npu-xrt/test_xrt_device_handle.py` counts opens with
`pyxrt.device` and the XRT allocation substituted, so the real
`XRTTensor.__init__` is under test and a call site that goes back to opening
its own handle is caught here rather than as a flake on a device runner.
Reverting the one line in `tensor.py` fails it at 4 opens instead of 1.

On device: `test/python/npu/test_torch_comparison.py` and `test_tensor.py`,
569 passed.

Not covered locally: `XRTHostRuntime.__init__`. This host's device reports as
`NPU Gorgon Point 1`, which `NPU_MODELS` does not map, so that constructor
raises before and after this change here. The change to it is an extraction of
the existing loop.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -56,6 +56,7 @@ declare_mlir_python_sources(AIEPythonSources.Utils
     utils/hostruntime/tensor_class.py
     utils/hostruntime/torch_interop.py
     utils/hostruntime/xrtruntime/__init__.py
+    utils/hostruntime/xrtruntime/device.py
     utils/hostruntime/xrtruntime/hostruntime.py
     utils/hostruntime/xrtruntime/parameter_scratchpad.py
     utils/hostruntime/xrtruntime/tensor.py
--- a/python/utils/hostruntime/xrtruntime/device.py
+++ b/python/utils/hostruntime/xrtruntime/device.py
@@ -1,0 +1,116 @@
+# device.py -*- Python -*-
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+"""Opening the NPU, once per process.
+
+``pyxrt.device(index)`` is an open, not a lookup. XRT calls into the shim on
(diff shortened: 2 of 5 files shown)


hunhoffe (approved):
This is a nice improvement. Thanks!

PR #3577, opened 2026-08-20
[AIE][AIEX] Add out-of-order S2MM DMA receive channels

jskimko:
## Description

Adds out-of-order (OoO) mode to S2MM DMA channels so multiple producers can merge into one receiver's buffer without ordering constraints between sources.

> **For reviewers:** I recommend starting with `test/npu-xrt/dma_s2mm_ooo/README.md` for an overview and demonstration of the feature.

### API

Receiver -- an out-of-order S2MM channel (each packet placed by header id):

```python
from aie.iron import Bd, BdIteration, Buffer, DmaChannel

DmaChannel(
    direction=DMAChannelDir.S2MM,
    channel=0,
    bds=[Bd(buffer=buf, offset=4 * i, length=4, packet=(0, 0)) for i in range(n)],
    out_of_order=True, # pkt.out_of_order_id goes to bd.bd_id
)
```

Sender -- stamp each BD's target slot into its packet header:

```python
bds = [
    Bd(buffer=buf, offset=4 * i, length=4, packet=(0, 0), out_of_order_id=i)
    for i in range(2)
]
DmaChannel(direction=DMAChannelDir.MM2S, channel=0, bds=bds)
```

Multiple packets per BD per merge:

```python
Bd(buffer=buf, offset=0, length=4, packet=(0, 0),
   iteration=BdIteration(size=m, stride=tw))
```

Integration with `repeat_count` for multi-round merges:

```python
DmaChannel(..., out_of_order=True, repeat_count=k)  # rounds = k + 1
```

### Behaviors / invariants

- `out_of_order` is S2MM-only; rejected on an MM2S channel.
- Receive BDs must be packet-enabled (placement needs the header).
- Receive BD ids are pinned; id collisions between OoO channels on a tile are rejected, as are wrong-parity pins (memtile even/odd channel bd_id ranges).
- OoO and the task-complete token are mutually exclusive (the BD-lookup id aliases the token header bits), so OoO disables token issue; completion uses an on-chip release-only counting lock on the receive BDs.
- The hardware value for "repeat count" refers to the total packet count of the whole operation, not task-level repeat. Alternatively, one "task" is one packet. Packet count is derived from the receive BDs (sum of `BdIteration` sizes) x rounds.
- `aie-decompose-large-dma-bd` rejects an OoO BD that would need splitting (all slices would collapse into one merge slot). Might be possible with `BdIteration`+ `pad_value`, but left out of this PR.

### Tests

- Dialect verifiers: S2MM-only, packet-enable required, `out_of_order_id` range, `repeat_count` errors, runtime-task id, decompose rejection.
- `assign-bd-ids`: id pinning, channel-parity accessibility, OoO id collision.
- CDO target: header-indexed placement, unchained control word, packet-enable, iteration word, token/OoO exclusion on a shim tile, lock-carrying receive BD, MM2S id stamping.
- libxaie target: channel arming, `use_next_bd=0`, iteration, two-tile unique locals.
- `dma-tasks-to-npu`: `out_of_order_id` carried into `npu.writebd`.
- IRON python: receiver API + derived packet count; sender-side id stamping.
- Device demo (`npu-xrt`): end-to-end merge on npu2 across the receiver tile, out-of-order channel count, m-packet, non-uniform, multi-round, and receiver-backpressure modes.

## Checklist


jgmelber (approved):
Fairly quick pass through but looks solid. Nice to see many tests for this.

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -2071,6 +2072,86 @@ TileOp MemTileDMAOp::getTileOp() {
...
+          auto it = releasedByRecvBd.find(lockDef);
+          if (it != releasedByRecvBd.end() && it->second != bd.getOperation())
+            return bd.emitOpError(
+                "out-of-order S2MM prohibits inter-BD lock dependencies; "
+                "can deadlock");
+        }
+  return success();
+}
+
 LogicalResult DMAOp::verify() {
```
andrej, line 2155: As a follow-on, it would be good if you could add a verifier that `next_bd` is unused / points to only aie.end blocks when out-of-order is used.

PR #3578, opened 2026-08-20
[AIEX] aie-expand-load-pdi: reset the registers the outgoing device left [LOW]

atassis:
## Problem

`aie-expand-load-pdi` lowers each `load_pdi @device` into an empty-device PDI load, which makes
the firmware reset the partition, followed by that device's configuration writes. That reset is
the only thing clearing what the previous device configured.

A partial reset that enumerates resource kinds cannot replace it in general, because what
survives a boundary is a property of the two configurations, not of a fixed list of kinds.
`test/Passes/expand-load-pdi/switchbox_config_multiple.mlir` already shows this. Its three
devices sit on tile(0,2) and differ only in routing, and each emits two writes:

| device | writes |
|---|---|
| `@my_switchbox_device_1` | `Stream_Switch_Master_Config_AIE_Core0`, `Stream_Switch_Slave_Config_South_0` |
| `@my_switchbox_device_2` | `Stream_Switch_Master_Config_AIE_Core0`, `Stream_Switch_Slave_Config_North_0` |
| `@my_switchbox_device_3` | `Stream_Switch_Master_Config_DMA0`, `Stream_Switch_Slave_Config_South_0` |

Device 1 enables slave port South_0 and device 2 enables North_0, so neither disables the
other's. Going from device 2 to device 3 the two sets are disjoint, so nothing device 2 wrote is
cleared and its route stays live underneath device 3's. Enumerating cores and objectFIFOs would
emit nothing at either boundary here, since these devices have neither.

## Fix

`register-reset=true` takes the difference instead. The pass generates both configurations, so
the leftover set is exactly computable:

```
stale = addresses the outgoing device wrote  -  addresses the incoming device writes
```

Each stale address is restored to the reset value `lib/Dialect/AIE/Util/aie_registers_aie2.json`
documents for it, resolved through the target model's own tile classes rather than a hardcoded
map. Nothing is guessed: a boundary carrying an address the database does not know, or a write
whose address is not compile-time known, keeps its firmware reset. So does the first `load_pdi`,
which has no predecessor to undo. Mutually exclusive with `ctrl-pkt`, which configures through
control packets this pass does not read.

## Test

Three lit tests: the differential case above, the first-load case, and the mutual exclusion.

Default-mode output is byte-identical to the current pass on all four existing tests in
`test/Passes/expand-load-pdi/`, so nothing changes unless the option is set. With it set, the
multi-device test emits exactly the three restoring writes the table implies and drops two of
its three preloads.

## Scope

Not device-tested in this form. An earlier revision of this PR reset by resource kind and did
have a hardware gate behind it; this one replaces that approach after the reviews above, and the
reset set it computes is different, so that measurement does not carry over. The reachable
opcodes it can emit are the same `aiex.npu.write32` the pass already emits.


jskimko:
I've been working on something similar, but for control packet reconfiguration. In my opinion, this needs passing runtime tests since the full reset buys us a lot of leeway that bypasses issues exposed by localized resets, i.e., offline verification is not sufficient in my experience. If I had to guess, removing `@empty` would break a lot, if not all, reconfiguration examples that target two truly different resource sets (you can get away with some trivial reconfigurations on the same set).

andrej:
This is a direction many of us are very interested in as a partial reset opens the possibility of faster, more fine-grained reconfiguration. @jskimko's comments are spot-on. We need to engineer the reset sequence properly, guided by hardware tests of real workloads, to ensure users can reliably use this.

atassis:
Let me rethink current work per your feedback then, it might take some time from me for that. If you have any more ideas how to implement this better- I am ready to discuss and do them by myself

atassis:
I built this and ran it on npu2 against every reconfiguration test in the tree, both modes:

| test | default | register-reset |
|---|---|---|
| `npu-xrt/loadpdi` | pass | pass |
| `npu-xrt/scratchpad_regwrite` | pass | pass |
| `npu-xrt/scratchpad_params` | pass | pass |
| `npu-xrt/scratchpad_addr_offset` | pass | pass |

That table is not the measurement you asked for. All four are `@empty` plus one real
device, which is exactly the "trivial reconfigurations on the same set" case you said
one can get away with. Nothing in the tree exercises two truly different resource sets
on hardware, so this shows only that the mode does not break the easy path.

I tried to write the test that would answer it and stopped at a shape question. Every
on-device reconfiguration test under `test/npu-xrt` puts its `aie.runtime_sequence`
inside the single real device, which then loads `@empty` and itself. Two real devices A
and B need a sequence that dispatches against A's objectFIFO symbols and then B's, and I
do not think one device's sequence can reference another's.
`switchbox_config_multiple.mlir` puts the sequence in a separate `@main`, but that is a
pass test and I have no evidence the full-ELF path supports that shape.

So: what should the two-device on-device test look like? I would rather write the
intended shape than guess and spend your review on the guess. @jskimko, if your control
packet work needs the same harness, I am glad to build whichever half is useful to both.

One practical note for anyone else testing this: the pass option has no aiecc exposure,
so a device run needs a small CLI change on top of this branch. I can fold that in here
if that is better than a follow-up.

atassis:
A correction to my last comment, and it goes against me.

I said nothing in the tree exercises two devices on hardware and that I did not think one
device's sequence could reference another's. Both are wrong. `test/npu-xrt/reconfigure_loadpdi`
and `reconfigure_loadpdi_persistent_memtile` do exactly that -- `@main` holding a sequence of
`aiex.configure @A { aiex.run @A_seq }` over two real devices, `REQUIRES: ryzen_ai_npu2`, in the
tree since #2775. My four-test table was not "every reconfiguration test in the tree"; it missed
at least six.

With the shape in hand I went looking for what a test would actually have to prove, and found
the pass cannot currently support the claim.

`collectWrittenAddresses` models a segment as the addresses its **configuration writes** touch.
Its own comment says any other op that configures the array would under-reset the boundary --
but the code enumerates three ops and silently ignores everything else. And the pass runs in
`getExpandLoadPdiPipeline`, which is **before** `getNpuDmaLoweringPipeline`, so the sequence at
that point still holds un-lowered `aiex.npu.*` ops. Every real design has those, so the
difference is computed against an incomplete picture on any real workload.

Underneath that is the thing offline verification cannot reach at all, which is what
@jskimko was pointing at: a register the configuration never wrote but **execution** moved --
a lock counter, a DMA channel left armed. No IR pass can see it.

So the honest status is that this needs restructuring rather than a test: the differential
belongs in a separate pass after DMA lowering, where the sequence is entirely write32/blockwrite,
and the fallback needs to be a denylist rather than an allowlist. I would rather do that as its
own change than defend this one.

Separately, and worth flagging because it is the same class: the trailing-parity fix that landed
recently exists because the firmware caches the last PDI address and no-ops a repeated load. That
is firmware state the compiler does not model either.

Last thing, which I think matters more than this PR: nobody has measured what a `load_pdi`
reconfiguration costs. The transition numbers I have are hardware-context transitions, not this.
Optimizing it before that number exists is the wrong order, so I am going to go measure it first.

atassis:
@jskimko @andrej you were right that this needs runtime evidence, and going to produce it turned up something more basic: the partial reset never actually engaged on any real design, so there was nothing to run.

Starting from the ask, four things came out of it.

**It was unreachable.** `register-reset` existed only as a pass option and was never plumbed through `aiecc`, so there was no way to turn it on from the driver. That is why no runtime test existed. It is exposed now as `--register-reset`, and it errors instead of sitting silently inert if given without `--expand-load-pdis`.

**No existing test covers the case you described.** `reconfigure_loadpdi` and `reconfigure_loadpdi_persistent_memtile` both put their two devices on identical tile sets, which is the "trivial reconfiguration on the same set" you can get away with. Added `test/npu-xrt/reconfigure_disjoint_columns`: two devices sharing no tile at all (column 0 row 2 against column 1 row 4), alternating 0/1/0/1 so a stale configuration surfaces as a wrong slice rather than a hang, with a baseline arm and a `--register-reset` arm.

**Three defects, each of which alone stopped it.** The artifacts came out identical between the two arms, and the reasons stacked:

1. The expansion emits `maskwrite32` itself, 14 per segment, and `collectWrittenAddresses` treated any masked write as unreadable. Every segment was opaque, so every boundary was skipped. `NpuMaskWrite32Op::getAbsoluteAddress()` exists, so recording it makes the leftover set more complete, not less.
2. The row field was masked with `0xFF`, but it is only `columnShift - rowShift` bits wide. Every address outside column 0 decoded to row 32 and was rejected as off-array, so the disjoint-column case could never have worked.
3. A core tile carries two register modules, and the lookup named only `core`. The DMA, lock and BD registers live in the `memory` module: `0x1D000` is `DMA_BD0_0`, absent from `core` entirely. It now consults both and refuses when their overlapping offset ranges disagree.

**What remains is a design question, and it is yours rather than mine.** With all three fixed it still declines, now on `0x220004`, which is tile (0,2) offset `0x20004` and lands inside `Program_Memory`. Program memory is not a register and has no reset value, and every real design writes core program memory, so every boundary declines. Whether program-memory writes belong in the leftover set at all is a judgment about what a partial reset must restore. Leaving stale program memory looks harmless while the core's enable and reset registers are themselves restored, but that is exactly the kind of reasoning I would not want to land on inspection.

The pass now emits a remark naming why a boundary kept its firmware reset, instead of silently doing nothing.

Worth flagging what this does to offline testing, since it is your point sharpened: all eight `expand-load-pdi` lit tests pass, **including `register_reset_differential.mlir`**. The feature works on the synthetic test and declines on a real design. The lit test was not wrong, it just could not see any of the three defects above.

So your prediction that dropping `@empty` breaks reconfiguration across truly different resource sets is still untested rather than refuted, because the reset has not yet replaced `@empty` on such a design. The test to settle it is in the PR and I will post numbers once I can get the board.

jskimko:
@atassis Sorry, I haven't been getting any notifications for some reason. Here are some of my experiences:

**On the disjoint-column test.** Our reading is that this mostly exercises stale configuration, e.g., a route left live because the incoming device does not overwrite it, which seems like the case your write-difference is aimed at, so it would not surprise me if the reset arm passes it. Where we saw the full reset earn its keep was different: a resource the incoming design reused that still carried execution state from the outgoing one, like a lock at a nonzero count or a channel left armed. Nothing in the configuration writes touched those, so a difference could not have restored them, and on disjoint tiles nothing reads them anyway. So, the case that bit us sat in overlapping-but-different resource sets rather than disjoint ones.

One caveat on that, and it leans toward your own point about measuring. We hit that armed-channel case in a path that emulated the writes rather than issuing plain write32 (control packet limitation), so we cannot say it behaves the same on the load_pdi path. It may even be a trigger/pulse register that a state restore cannot express without some kind of playback/bookkeeping logic (instead of a state diff).

**On program memory.** In our runs we reconfigured across different core sets without restoring program memory and it worked, as long as the core enable and reset registers were restored so the stale program stayed dormant. I'm not too concerned about stale program memory under the assumption that the given designs are valid.

**On whether a partial reset is viable at all.** So far, it still appears that way, and I'm working to make it production-ready. FWIW, we have been working this same problem for a while, and it is not public yet nor exactly easily extractable. I'll likely release it in about a month. Rather than have you rebuild it, it may be worth holding the optimization side until then.

PR #3587, opened 2026-08-24
[AIE] Use conv_even rounding for float-only SRS

erwei-xilinx:
Cores whose only narrowing op is a float SRS had `crRnd` set to floor (0). That is a directed rounding, and it is the register **every bf16 elementwise op obeys**: an `arith.addf` on `vector<Nxbf16>` widens to an f32 accumulator and narrows back through this SRS. So bf16 add, sub and mul all rounded toward negative infinity.

The whole functional change is one literal:

```diff
-int roundingMode = hasBF16Matmul ? 12 : hasIntegerSRS ? 9 : 0;
+int roundingMode = hasBF16Matmul ? 12 : hasIntegerSRS ? 9 : 12;
```

## Evidence

npu1 (Phoenix), a 2048-element bf16 elementwise add, identical kernel, only this register differing:

| crRnd | matches numpy bf16 | signed mean error | error ≤ 0 on every element |
|---|---|---|---|
| 0 floor | 1538/2046 | −0.115 | **yes** |
| 12 conv_even | **2046/2046** | +0.0004 | no |

The bias under floor is one-sided, so it accumulates across a reduction instead of cancelling. That is the same argument this function already makes for bf16 matmul three lines above, which is why floor was already overridden there.

floor is the right answer for a *deliberate* f32→bf16 truncation and the wrong one for an arithmetic result, and this register cannot tell the two apart — it is set once per core, so whichever value is chosen applies to every float-narrowing SRS in the body. Elementwise arithmetic is the common case, so it decides the default.

**Unchanged:** integer SRS (positive_inf, 9) and bf16 matmul (conv_even, 12).

## Both generations, deliberately

The selection has never been arch-conditional — only the register *index* is (crRnd 6 on AIE2, 1 on AIE2P) — and the mode encodings are identical:

| mode | AIE2 (`aiev2_defines.h`) | AIE2P (`aie2p_defines.h`) | AIE2PS |
|---|---|---|---|
| `rnd_floor` | 0 | 0 | 0 |
| `rnd_pos_inf` | 9 | 9 | 9 |
| `rnd_conv_even` | 12 | 12 | 12 |

Splitting the two here would mean identical source computing measurably different numbers on npu1 and npu2 — a worse and far less visible failure mode than the extrapolation below, since each generation would look self-consistent on its own.

A comment at the literals now records that **AIE1 numbers the modes differently** (conv_even 6, positive_inf 2), so the enclosing `AIE2 || AIE2p` condition is load-bearing for the *values* and not just the register index. AIE1 cannot reach this code today — it has no `set.ctrl.reg` intrinsic.

## What is measured and what is not

The measurement is **npu1-only**; there is no npu2 hardware to hand, so **AIE2P is an extrapolation**. A narrow one — the argument is arithmetic rather than architectural, and nothing about AIE2P makes its f32→bf16 narrowing round differently — and it is pinned in its own check in `srs_rounding_mode_aie2p.mlir`, so if npu2 measurement ever contradicts it, that check is what fails.

`check-aie` has **not** been run locally: no working mlir-aie build on this machine. Both float checks require the patched compiler and are CI-only.

## Tests

- `srs_rounding_mode_aie2.mlir`: float-only expectation flips to conv_even (12).
- `srs_rounding_mode_aie2p.mlir`: adds a float-only SRS core, RUN line and check. Uses FileCheck captures rather than literal SSA names — the core body has its own constant for the SRS shift operand, so suffixing is ordering-dependent, and capturing additionally asserts that the value passed to `set.ctrl.reg` *is* the constant just checked to be 12.

Verified the AIE2P checks against an **unpatched** `aie-opt`: the matmul check still passes (that branch is untouched) while the new float-only check fails, confirming it is sensitive to the change rather than inert.

## How this was measured without a full rebuild

`aircc` hands `aiecc` a pure-AIE `npu.air.mlir` as a subprocess argument, and that file is not regenerated if `aiecc` is invoked on it directly. Splicing a bf16 `aievec.matmul` into each core makes the existing `hasBF16Matmul` branch select conv_even, so the untouched elementwise add could be measured under both register values with one compiler. (The matmul result must be stored — an unused one is DCE'd before the walk sees it.)


erwei-xilinx:
Addressed in a20740d20bf — all three were correct.

The two stale comments are the same defect twice: the change falsified them and left them in place. `AIECoreToStandard` still said "float SRS works better with floor", and the AIE2 test still described its float-only core as one that "should get floor rounding". Both now state what the pass does and point at where the reasoning lives.

The third was a real defect rather than a style point. The last `CHECK-FLOAT-AIE2P` line was missing its closing paren, so it was a prefix match and would have accepted any constant merely *starting* with `%c0_i32`. I did not just add the paren: both constants are now FileCheck captures. That is more robust — the core body has a second `arith.constant 0 : i32` for the SRS shift operand, so which one gets the unsuffixed name is an ordering detail — and it is also strictly stronger, since it asserts the value handed to `set.ctrl.reg` is *the same* constant just checked to be 0, which a literal name cannot express.

Verified both AIE2P prefixes still pass against the **unpatched** `aie-opt`, which is what makes that file a genuine "AIE2P is unchanged" guarantee. And since a missing-paren bug is exactly the kind that makes a test pass without asserting anything, I checked the new captures against two mutations — expecting 12 instead of 0, and swapping the two captured operands — and both fail as they should.

Still outstanding on my side: I have no working mlir-aie build here (cmake reconfigure fails on a dirty `third_party/aie-rt`), so `check-aie` has not been run locally. The AIE2 half of the rounding test needs the patched compiler and is unverified outside CI; the AIE2P half is verified as described above.

erwei-xilinx:
Correction, pushed as 22fcff02265 — the previous revision was wrong and the title has been updated to drop "on AIE2".

It changed float-only SRS on AIE2 only and left AIE2P on floor, which **introduced an architecture-dependent rounding policy that did not previously exist**. Measured against the unpatched compiler:

|  | AIE2 (npu1) | AIE2P (npu2) |
|---|---|---|
| bf16 matmul | 12 conv_even | 12 conv_even |
| float-only SRS | 0 floor | 0 floor |

Upstream is `hasBF16Matmul ? 12 : hasIntegerSRS ? 9 : 0` with no reference to the arch — only the register *index* differs (crRnd is 6 on AIE2, 1 on AIE2P). So splitting the generations was the novel thing, not keeping them together.

The caution behind the split does not hold up on inspection. The measurement is still npu1-only, but the argument is arithmetic rather than architectural — a directed rounding biases every narrowing the same way, and nothing about AIE2P makes its f32→bf16 narrowing round differently. More to the point, **the encoding is already exercised on AIE2P**: the bf16 matmul branch writes 12 to crRnd (register 1) on that part today, so this was never a guess about what 12 means there. And identical source computing different numbers on npu1 vs npu2 is a worse failure mode than a stated extrapolation, because each generation looks self-consistent in isolation.

The AIE2P float-only case is still pinned in its own check — now as the extrapolated half rather than as an unchanged-behaviour guard, so if npu2 measurement ever contradicts it, that is the check that fails.

Sensitivity checked rather than assumed: against the unpatched `aie-opt`, the AIE2P **matmul** check still passes (untouched branch) while the AIE2P **float** check now fails, where before this commit it passed. clang-format clean. `check-aie` still not run locally — no working build here.

erwei-xilinx:
Good challenge — the answer is yes, but my stated reason was circular and is now replaced (d02aa49ec10).

I had argued "12 is already emitted on AIE2P by the matmul branch, so the encoding isn't in doubt". That only establishes what mlir-aie *writes*; it says nothing about what 12 *means* on that part. Had the encoding differed, the existing matmul path would have been wrong too and I'd have been citing the bug as evidence for itself.

The real evidence is llvm-aie's per-architecture intrinsic headers, which define the values independently and agree:

| mode | AIE2 (`aiev2_defines.h`) | AIE2P (`aie2p_defines.h`) | AIE2PS | AIE1 (`versal_prod/me_defines.h`) |
|---|---|---|---|---|
| `rnd_floor` | 0 | 0 | 0 | 0 |
| `rnd_pos_inf` | 9 | 9 | 9 | **2** |
| `rnd_conv_even` | 12 | 12 | 12 | **6** |

So all three literals this code writes mean the same thing on AIE2 and AIE2P, and only the register index differs — which is what the change assumed, now on its own authority.

One thing that fell out and is worth having recorded: **AIE1 numbers the modes completely differently**, so every literal here would be wrong there. It cannot reach this code — the enclosing condition admits only AIE2/AIE2p, and AIE1 has no `set.ctrl.reg` intrinsic — but that condition is load-bearing for the *values* and not just the register index, which nothing said. There's now a comment at the literals saying so, since anyone widening that condition would otherwise have no reason to look.

Comments only in this commit; no functional change.

hunhoffe (approved):
I just want to double check that the other architectures we have tests for (e.g., phoenix) have analogous tests?

hunhoffe:
From my reviewer agent:   

> The floor default this PR removes wasn't arbitrary — it was added in #2912 specifically to fix an accumulated-bias regression in mlir-air's 42_triton_softmax_bf16 on NPU1 (256 bf16 values summed through repeated aievec.srs, where positive_inf produced ~2.2x error from one-sided bias). This PR's own argument is that floor has the same one-sided-bias problem, just in the other direction, and conv_even (round-to-nearest-even) should be unbiased and thus fine or better for that case too — but that specific downstream test lives in a different repo (mlir-air) and isn't in this repo's CI, and the PR author notes check-aie hasn't even been run locally. Worth explicitly re-running or asking the mlir-air softmax regression test against this change before merging, since it's the one known prior failure mode this exact code path has hit.

PR #3595, opened 2026-08-25
Mark the HSA discovery tests unsupported on Windows

andrej:
Tests in `test/python/test_hsa_discovery.py` were failing on Windows.

## Why they failed

`discovery.py` answers one question: is there a ROCm install on this machine, and where is its `libhsa-runtime64.so`? It looks in three places, in a fixed priority order: `$ROCM_PATH`, then a pip-installed ROCm, then `/opt/rocm`. The tests check that order. Each one creates a fake ROCm tree in a temp dir, points discovery at it, and asserts that discovery picks the right one.

The module decides at import time which file names to look for:

```python
_LIBHSA_PATTERNS = ("libhsa-runtime64.so", ...) if _IS_LINUX else ()
```

Off Linux that tuple is empty, so `find_libhsa()` returns `None` whatever sits on disk. Seven tests assert that a fabricated tree *is* found, so all seven failed. They have never passed on Windows: the test and the module arrived together in #3452, whose Windows job also reported this failure.

## Fix

ROCR exposes its AIE agent only on Linux, so discovery has no Windows behaviour to pin down. The file now declares:

```
# UNSUPPORTED: system-windows
```

lit's `LLVMConfig` registers `system-windows` from the host platform, so the suite runs on Linux and lit reports it unsupported elsewhere.

An earlier revision of this PR instead kept the tests running everywhere, by re-importing `discovery.py` under a faked `platform.system()`. The author of the test suite confirmed that Windows is out of scope, so that revision is dropped, and so is the `_discovery_as_platform` helper it built on, together with the two tests that called it.

The diff is now 2 added lines and 44 deleted ones. The remaining ten tests are unchanged from main.

### DIFF
--- a/test/python/test_hsa_discovery.py
+++ b/test/python/test_hsa_discovery.py
@@ -4,6 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 #
 # RUN: %pytest %s
+# ROCm, and hence everything discovered here, exists only on Linux.
+# UNSUPPORTED: system-windows
 
 """Discovery of a ROCm install: explicit root, pip wheel, then system."""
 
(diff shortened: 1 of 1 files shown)


test/python/test_hsa_discovery.py
```diff
@@ -7,10 +7,31 @@
 
 """Discovery of a ROCm install: explicit root, pip wheel, then system."""
 
+import importlib.util
```
hunhoffe, line 10: I often times prefer to have tests test the actual environment rather than monkeypatch, but that's just my personal preference.

hunhoffe (approved):

ypapadop-amd:
Can we disable it on Windows? There's no way to get AIE agents on Windows via ROCR currently :(

andrej:
Yes, disabling on Windows makes more sense then. I'll remove the existing "fake Linux" environment on Windows machinery as well then.

andrej:
@ypapadop-amd Revamped the PR to completely drop the fake-Linux environment and tests on Windows. Does this seem reasonable?

andrej:
Decided to just go ahead and merge to get CI green, if there are any comments happy to follow-on in another PR. :)

PR #3597, opened 2026-08-26
[aiecc] Add link-time stack-size check for peano-compiled cores [MED]

atassis:
## Problem

`CoreOp::verify()` (added in #3496) only rejects a `stack_size` that alone exceeds local memory. It has no way to see whether the compiled CODE fits the reservation. The linker places the stack directly below this tile's objectFIFO buffers with zero clearance (`AIETargetLdScript.cpp`), so a frame that overruns a correct, small `stack_size` corrupts those buffers silently instead of faulting: no crash, deterministic, surviving tiles bit-exact. Upstream has hit this three times on real kernels (#2275, #2280, #2345), each just re-guessing the constant.

The fix is already half-built and unused. The ld script has emitted a `.stack_sizes` output section since #2417 (`AIETargetLdScript.cpp:167`), but nothing ever populates or reads it: `llc` is never passed `--stack-size-section`, so the section has been empty the whole time.

## Fix

Wire the existing plumbing up to an actual link-time gate, on the peano path:

1. `llc` gets `--stack-size-section`, so each function's frame size lands in `.stack_sizes` (the ld script already collects it into the linked ELF).
2. `ShellCommand` gains `captureStdout()`: `llvm-readelf --stack-sizes` has no `-o` flag, its answer is its stdout.
3. `compile_cxx_core_function` passes `-fstack-size-section` on its Peano branch, so the kernel objects the JIT builds carry their frames too. Without this the gate sees only the core code aiecc itself compiles, which is not where the big frames are.
4. A new edge runs `llvm-readelf --stack-sizes` on the linked ELF and sums every function's frame into an upper bound (same conservative choice, for the same reason, as iree-amd-aie's own `getMaxStackSize`, `XCLBinGen.cpp:126`: without the call graph, a caller's frame could sit under any of the others). It compares that bound against `CoreOp::getEffectiveStackSize()` and fails the build with the tile and both numbers if the code does not fit.

This mirrors iree-amd-aie's `getMaxStackSizeFromExecutable` (`XCLBinGen.cpp:697`). Chess/xbridge cores are unchecked, as they are there (`if (useChess)` skips the whole peano branch); cores carrying an `elf_file` attribute are unchecked here because aiecc does not compile them at all.

## Test

Two new lit tests plus `test/aiecc/Inputs/big_stack_kernel.c`: a 2048-byte local buffer whose address escapes to a global, so the compiler cannot shrink it away, linked into a core via `link_with`.

- `cpp_stack_size_overflow.mlir`: default `stack_size` (1024), expects `not aiecc` to fail with the new diagnostic.
- `cpp_stack_size_overflow_override.mlir`: same kernel, `stack_size = 4096`, expects a clean build.

Both pass under lit. Across the rest of the in-tree suite 58 cores went through the gate and none tripped: their frame sums are 0, 32, 64 or 128 bytes against the 1024-byte default.

The failure this closes is real: with the check removed, the same design links to a valid, loadable 2212-byte ELF whose `.stack_sizes` reports 2080 bytes (32 + 2048) against the 1024 reserved.

Also confirmed, since it was previously untested: `.stack_sizes` survives this driver's own `-Wl,--orphan-handling=error` -- non-allocatable sections with an explicit but region-less SECTIONS rule (`.comment`, `.symtab`, ...) are not orphans.

## Scope

A frame is visible to the gate only if it carries a `.stack_sizes` entry, which is why item 3 exists. Measured on a design linking `aie_kernels/aie2p/layer_norm.cc`: before that change the core reported 128 bytes, the kernel's own frame invisible; after it the kernel's frame is counted. A `link_with` object built by a hand-rolled compile line rather than by `compile_cxx_core_function` is still opt-in, and Chess is unaffected. `--gc-sections` runs before the check, so the bound is over the functions that survive into the ELF, not over the object.

No in-tree design is rejected by this check. That was not true when this was first written: `programming_examples/ml/norm` at `-o layer_f32` overran the 1024-byte default, which is what `run_strix.lit` had been timing out on, and the fix was pending in #3467. #3467 has since merged -- it sets `stack_size=2048` on that Worker and rewrites `layer_norm.cc`, and the frame it produces is now 320 bytes, so the design clears the default with room to spare and clears its explicit reservation by 6x.

Re-measured against the Peano this repo's CI actually pins (`llvm-aie==21.0.0.2026080301+c9c5ecb7`, `utils/peano-requirements.txt`), the kernels behind every design #3467 added:

| kernel | functions (frame bytes) | design's `stack_size` |
|---|---|---|
| `aie2p/layer_norm.cc` | `layer_norm` 512, `layer_norm_f32` 320, `layer_norm_affine_cast` 320 | 2048 (explicit) |
| `aie2p/exp2f_vec.cc` | `exp2f_vec_f32` 128, static helper 256 | 1024 (default) |
| `aie2p/cast_f32_bf16.cc` | `cast_f32_bf16_row` 0 | 1024 (default) |
| `aie2p/dwconv1d.cc` | `dwconv1d_bf16` 0 | 1024 (default) |

Identical byte-for-byte under the divergent Peano this was developed against, so the margins here are not an artifact of one toolchain pin.

### DIFF
--- a/python/utils/compile/utils.py
+++ b/python/utils/compile/utils.py
@@ -316,6 +316,10 @@ def compile_cxx_core_function(
             "-Wno-empty-body",
             "-O2",
             "-DNDEBUG",
+            # Per-function frame sizes into .stack_sizes, so aiecc's link-time
+            # stack check covers kernel objects and not just the core code it
+            # compiles itself.
+            "-fstack-size-section",
             # Have the compiler report what it actually read, the way ninja and
             # ccache learn a translation unit's real inputs.
             "-MD",
--- a/test/aiecc/Inputs/big_stack_kernel.c
+++ b/test/aiecc/Inputs/big_stack_kernel.c
@@ -1,0 +1,12 @@
+// Deliberately oversized frame: a 2048-byte local buffer whose address
+// escapes to a global, so the compiler cannot shrink it to just its touched
+// bytes. Stands in for a real kernel (conv accumulator, im2col scratch, ...)
+// whose frame exceeds a core's declared stack_size -- see
+// cpp_stack_size_overflow.mlir.
+volatile char *escape;
+void big_stack_kernel(void) {
+  char big[2048];
+  big[0] = 1;
+  big[2047] = 2;
+  escape = big;
+}
--- a/test/aiecc/cpp_stack_size_overflow.mlir
+++ b/test/aiecc/cpp_stack_size_overflow.mlir
@@ -1,0 +1,66 @@
+//===- cpp_stack_size_overflow.mlir -----------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Link-time stack-size gate: a linked kernel whose compiled frames exceed the
+// core's declared (here, default) stack_size must fail the build instead of
+// linking clean and silently overrunning the objectFIFO buffers the linker
+// places right above the stack. See big_stack_kernel.c: its 2048-byte local
+// buffer cannot be optimized away (the address escapes to a global), so it
+// reliably reports a 2048-byte frame under --stack-size-section.
+
+// REQUIRES: peano
+
+// RUN: clang --target=aie2-none-unknown-elf -O2 -fstack-size-section -c %S/Inputs/big_stack_kernel.c -o big_stack_kernel.o
+// RUN: not aiecc --get-core-elfs %s 2>&1 | FileCheck %s
+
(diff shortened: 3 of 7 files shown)


andrej:
Related: #2347, #3586

(The issues/PRs linked in the linked issue and this PR document **at least 6 instances** of this exact stack overflow scenario causing trouble for users -- so a stack size verifier is important and useful work.)

atassis:
Ah, I see, I am closing this PR then. But still two things from here that may still be useful, @hunhoffe:
- .stack_sizes survives -Wl,--orphan-handling=error, which aiecc sets. A non-allocatable section with an explicit but region-less SECTIONS rule is not an orphan.
- It has been emitted by the ld script since #2417 but never populated: llc was never passed --stack-size-section, and objects from compile_cxx_core_function need -fstack-size-section on the Peano branch or the large frames stay invisible.

PR #3598, opened 2026-08-26
Run lit tests with lit's internal shell

andrej:
lit's ShTest raises a fatal error for execute_external=True as of LLVM 23, and the option disappears in LLVM 24.

No test uses shell features beyond what the internal shell implements, so we can just turn off execute_external.


hunhoffe (approved):
You beat me to it! Thanks!

Issue #3481, opened 2026-07-31
Establish a rounding mode convention for kernels


andrej:
### What are you trying to do?

Some kernels set their rounding mode at the start of execution for higher accuracy, such as `conv_even`.

Some recent kernel additions [added logic](https://github.com/Xilinx/mlir-aie/blob/3df743ab295b38b7ddc93e5919102a6617b2743b/aie_kernels/aie2p/mm.cc#L99-L101) for saving the old round mode on kernel entrance and restoring the previous rounding mode upon exit. This adds complexity and cycles, and my intuition is that we don't actually care enough for most kernels to do this restoration.

### Proposed solution

Establish a common "contract". I propose the following:

- Only kernels where rounding is custom/unusual/critical to the kernel should set their rounding mode at the beginning.
- For all others, assume the caller decided the rounding mode.
- Add code at the start of code initialization that initializes rounding to `conv_even` by default, as this appears to be highest-accuracy for most use cases.
- Remove current set or save/restore rounding logic in kernels.

PR #3599, opened 2026-08-26
C++ AIE kernels deduplication

efurst:
## Description

Expands the aie_kernels C++ library and extends the in-tree IRON kernel-factory layer (python/iron/kernels/) to expose the new kernels.

C++ kernels (aie_kernels/)
- Activations — ports/adds leaky_relu, sigmoid, tanh for AIE2; add, mul, sigmoid, tanh, leaky_relu, and mha (flash-attention toolkit) for AIE2P. Reworks gelu/silu/softmax and RoPE.
- Generic data-movement / BLAS — new transpose, expand (int4→bf16 dequant), mv (matrix-vector), axpy.
- Shared-op dedup — common helpers factored into aie_kernel_utils.h so eltwise/activation kernels stop copy-pasting arch-tuned inner loops.

IRON kernel factories (python/iron/kernels/)
- Adds tanh, sigmoid, leaky_relu factories (+ _ref variants) to activation.py.
- New datamovement.py with axpy, convert_copy, expand, transpose.
- Re-exported from __init__.py; test/python/test_kernels_specs.py covers the factory surface.

Docs — rewrites aie_kernels/README.md into a per-arch (generic / AIE2 / AIE2P) catalog with datatypes and coding style.

## Checklist


hunhoffe:
@atassis I wanted to make sure you saw some of the progress here to make sure there aren't duplications/conflicts! I think this takes us much closer towards deduplication and we may want to wait to integrate additional kernel fixes until after this lands.

hunhoffe:
Waiting on https://github.com/Xilinx/mlir-aie/pull/3595 for final windows failure from main, but that's not a required test so this PR is not blocked by CI at the moment 🙂

aie_kernels/aie2/sigmoid.cc
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  event1();
+
+  return;
+}
+
+extern "C" {
+
+void sigmoid_bf16(bfloat16 *restrict input, bfloat16 *restrict output,
```
hunhoffe, line 50: Can you remind me why we need an extra wrapper around the core implementation?

aie_kernels/aie2/sigmoid.cc
```diff
@@ -0,0 +1,55 @@
...
+  aie::vector<bfloat16, 32> register_0_5 = aie::broadcast<bfloat16, 32>(0.5f);
+  aie::vector<bfloat16, 32> register_1 = aie::broadcast<bfloat16, 32>(1.0f);
+  AIE_PREPARE_FOR_PIPELINING
+  AIE_LOOP_MIN_ITERATION_COUNT(32)
+  for (int i = 0; i < vector_size; i += 32) {
+    auto input = *it_in++;
+
+    aie::vector<bfloat16, 32> half_x = aie::mul(input, register_0_5);
+
+    aie::vector<bfloat16, 16> tanh_lo = getTanhBf16(half_x.extract<16>(0));
```
hunhoffe, line 33: This may be material for a separate PR, but can any of similar kernels be folded into one definition with compiler directives? If there's only 1-2 lines of difference in the innermost loops mathematical operations, I think consolidation would make sense to me.

aie_kernels/aie2/tanh.cc
```diff
@@ -0,0 +1,47 @@
...
+void tanh_bf16_vectorized(bfloat16 *restrict input_vector,
+                          bfloat16 *restrict output_vector,
+                          const int32_t vector_size) {
+  event0();
+
+  auto it_in = aie::begin_restrict_vector<32>((bfloat16 *)input_vector);
+  auto it_out = aie::begin_restrict_vector<32>((bfloat16 *)output_vector);
+
+  AIE_PREPARE_FOR_PIPELINING
+  AIE_LOOP_MIN_ITERATION_COUNT(32)
```
hunhoffe, line 25: Just to double check: if you give these functions a small amount of data, below min iter counts, do we error or support those cases cleanly?

This could also be a separate PR, but it would at minimum be good to track this in an issue, I think, if it's not handled gracefully.l

aie_kernels/aie2p/layer_norm.cc
```diff
@@ -135,7 +135,11 @@ static inline void layer_norm_f32_impl(const TIn *restrict input,
 
 extern "C" {
 void layer_norm(bfloat16 *input, bfloat16 *output, int32_t cols) {
-  layer_norm<bfloat16, 16>(input, output, cols);
+  // N=32 bf16 = 512 bits = one AIE2P vector register.  conv_even rounding
+  // matches the reference math more closely than the default floor mode for
+  // the normalize pass.
+  ::aie::set_rounding(aie::rounding_mode::conv_even);
```
hunhoffe, line 141: related: https://github.com/Xilinx/mlir-aie/issues/3481

aie_kernels/generic/expand.cc
```diff
@@ -0,0 +1,79 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <type_traits>
+
+template <typename T_in, typename T_sf, typename T_out, const int N,
+          const int G>
+void expand(T_in *in, T_out *out) {
+  // Keep vector width constant; group size can vary as a multiple of 32
+  constexpr int block_size = 32;
```
hunhoffe, line 20: should this be a `#define`?

aie_kernels/aie_kernel_utils.h
```diff
@@ -76,4 +76,27 @@ Copyright (C) 2022-2025 Advanced Micro Devices, Inc.
 #define AIE_LOOP_FLATTEN
 #endif
 
+// Runs `body` (a zero-arg lambda) `count` times.  When count >= MinIters the
+// loop is eligible for software pipelining (extra pragma macros may be passed
+// after `body`, e.g. AIE_PREPARE_FOR_POSTPIPELINING); otherwise a plain
+// no-unroll loop is emitted, avoiding invalid pipeliner assumptions for tiny
+// trip counts.
+#define VERSIONED_LOOP(MinIters, count, body, ...)                             \
```
hunhoffe, line 84: Is this used consistently? Have we double checked whether there's an analogous capability in peano? This might be note to self to pursue if there's a more generic llvm-aie solution.

aie_kernels/README.md
```diff
@@ -17,6 +17,10 @@ In some cases, the kernels are just generic C code, and will run on any family o
 | Class | Name | Coding style | Purpose | Datatypes |
 |-|-|-|-|-|
 | basic | [passThrough.cc](./generic/passThrough.cc) | AIE API | A simple memcpy operation | `uint8_t`, `int16_t`, `int32_t` |
+| data movement | [transpose.cc](./generic/transpose.cc) | AIE API | Blocked matrix transpose (4×4 / 8×8 sub-tiles, VSHUFFLE) | `bfloat16` |
```
hunhoffe, line 20: Thank you for updating this documentation!

test/python/npu/test_kernels_e2e.py
```diff
@@ -136,3 +137,574 @@ def sequence(a, b, in_h, out_h):
     # input that's the largest element.
     expected = bfloat16(TILE - 1)
     assert y[0] == expected, f"reduce_max wrote {y[0]} but max of input was {expected}"
+
+
+# ===========================================================================
+# Kernels copied back from IRON that need a bespoke hand-built design (their
```
hunhoffe: Should we align on the signature? Also, I'd like to avoid historical comments :)

hunhoffe (approved):
Hello! I think this is great. I think it can be merged as-is once tests pass, but I have some ideas for follow on/checks. I'm more looking for your opinion on these things, rather than saying action needs to be taken now.

aie_kernels/aie2/sigmoid.cc
andrej: Just chiming in here, we need this because of the ABI discussion we had in the compiler meeting with Yiannis. C++ mangles function names:

```
(buildenv) androsti@xcoradaie211:/scratch/roesti/test$ cat test.cc
void kernel(bfloat16 *in, bfloat16 *out) {
	out[0] = in[0];
	return;
}
(buildenv) androsti@xcoradaie211:/scratch/roesti/test$ $PEANO_INSTALL_DIR/bin/clang++ -std=c++20 --target=aie2ps-none-unknown-elf -c test.cc -o test.o
(buildenv) androsti@xcoradaie211:/scratch/roesti/test$ objdump -t test.o

test.o:     file format elf32-little

SYMBOL TABLE:
00000000 l    df *ABS*	00000000 test.cc
00000000 g     F .text._Z6kernelP8bfloat16S0_	00000050 _Z6kernelP8bfloat16S0_
```

The last objdump command shows that after compilation, we get a function named `_Z6kernelP8bfloat16S0_`. [Demangling](https://www.demangler.com/) that string gives `kernel(bfloat16*, bfloat16*)`.

So either in the `design.py` we have to refer to the function as `_Z6kernelP8bfloat16S0_`, or we re-declare it with extern C to get a flat `kernel`. If I understood Yiannis's approach correctly (please correct if wrong @ypapadop-amd), he re-declares what the expected function signature is (void, bf16*, bf16* in this case), then reimplements the mangling in Python, so the extern C is not necessary.

python/iron/kernels/datamovement.py
```diff
@@ -0,0 +1,159 @@
...
+from aie.iron.kernel import ExternalFunction
+from ml_dtypes import bfloat16
+
+from ._common import _default_source_path, _make_extern
+
+_AXPY_VEC = 64  # saxpy processes 64 bf16/iteration
+
+
+def axpy(tile_size: int = 1024, vectorized: bool = True) -> ExternalFunction:
+    """SAXPY kernel: ``z = a * x + y`` over bf16 tiles.
```
andrej, line 25: Not sure if this really best fits the "datamovement.py" file, since some computation is happening here? Don't feel strongly about it but it's the one kernel that I think fits least into this file name.

andrej (approved):
Looking good. Are you planning a follow-on in amd/IRON to delete those kernels there? Unfortunately, it seems to be a bit of a moving target as within the last two days we got several PRs to amd/IRON that update the kernels. One thing we could do is merge this now, remove the kernels from amd/IRON as soon as possible, then close these new PRs there and ask the authors to reopen here. Thoughts?

test/python/npu/test_kernels_e2e.py
efurst: good catch on the historical comment here - thought I had caught all the ones claude inserted, but I can remove this and take a closer look at the signatures in this file

atassis:
Three files under `aie_kernels/aie2p/` have diverged between this tree and amd/IRON in ways a signature-level diff will not show, so which copy survives is worth deciding deliberately.

`mm.cc`: here the bfp16 emulation swaps the core rounding mode to `conv_even` and restores the caller's mode, gated on `AIE_API_EMULATE_BFLOAT16_MMUL_WITH_BFP16` alone (#3442 carries the measurements). IRON drives it from a separate `-DROUND_CONV_EVEN` and never restores. The decoupling makes `GEMM(emulate_bf16_mmul_with_bfp16=True, round_conv_even=False)` expressible, which puts the bf16 to bfp16 conversion back on floor rounding; and the rounding mode is one sticky per-core register, so the missing restore leaks it to whatever runs next on that core.

`layer_norm.cc`: here the row sum accumulates in `accum<accfloat, N>` (#3410). IRON uses `vector<T, N>` with T = bfloat16, so the running sum drops low bits as the reduction grows. Embedding dims run to the thousands, so the mean is lossy before the variance is computed.

`rms_norm.cc`: IRON sets `conv_even` and does not restore it. This tree does not touch the mode.

The common thread is that rounding mode is ambient per-core state and accumulator width is a numerics contract. Neither is visible in the kind of diff dedup review naturally focuses on.

atassis:
`saxpy` truncates its scalar, and the e2e test records the cause as a framework gap.

`generic/axpy.cc:24` calls `aie::to_float<bfloat16>(a, 0)` with `a` declared `const float`. In `aie_api` at the pin on main (`bec000fd3`), the only real-scalar overload is `aie.hpp:7552`, `template <typename TR = float> auto to_float(int a, int shift = 0)`, unconstrained. The other scalar overload at 7567 requires `T` to be `cint16`/`cint32`. `TR` is given explicitly and the parameter is a non-deduced `int`, so `a` binds through a float-to-int standard conversion. `saxpy(x, y, 0.5f, z, n)` computes `z = y`. Clang does not warn by default, since that is `-Wfloat-conversion`.

`to_float` is the fixed-point-to-float conversion. The intent in the comment ("Convert to bfloat16") is a plain `bfloat16(a)` cast.

This matters beyond the one line because `test_kernels_e2e.py:210` attributes the truncation to `@iron.jit`'s CompileTime scalar plumbing and states "the kernel math itself is correct", then pins the test to `a = 3.0`. Both causes are consistent with the observed 2.5 to 2, but the kernel truncates independent of what the framework transports, so the test as written cannot ever catch it. Worth fixing the cast and testing a fractional alpha, even if the framework transport is also a separate gap.

Also note `saxpy` takes `float a` while `saxpy_scalar` takes `bfloat16 a`, which `datamovement.py` has to encode as a per-variant `a_ty`. Matching them would remove that branch.

atassis:
A few things I would pick up after this lands. None of them block.

**`mul.cc:45` states a defect that `accum.hpp` contradicts, and `rope.cc` depends on the opposite.** The comment says assigning an accumulator into a `vector<T_out>` "produces garbage at this 32-wide width". At `bec000fd3`, `accum.hpp:153` defines `operator vector<T, Elems>() const requires (accum_class_for_type_v<T> == value_class() && is_floating_point() && size() <= 32)`, whose body is `return to_vector<T>();`. At 32 lanes it is viable and identical to the explicit call. Meanwhile the new `rope_kernel_two_halves` is instantiated at N=32 and relies on exactly that implicit conversion (`rope.cc:64`, `:71`). Either the comment is wrong and gets cemented in the catalog, or `rope_two_halves` is broken. If the garbage was real, it would be worth pinning what actually caused it, since a reproducer would be an `aie_api` bug.

**`transpose.cc:23` mixes bits and bytes.** `512 / sizeof(bfloat16)` is 256 elements. `add.cc` and `mul.cc` in this same PR both state "32 bf16 = 512 bits = one AIE2P vector register", so the cap should be `512 / 8 / sizeof(bfloat16)`. Masked at the tested 32x32 and it degrades quietly above that, since `native_vector_type<bfloat16, N>` covers 64, but the constant does not say what it means.

**Rounding mode is left changed on exit.** `mv.cc:51`, `softmax.cc:113`, `mha.cc` at every entry point, and the `layer_norm` bf16 entry all set `conv_even` and never restore, while `layer_norm_f32` in the same file does not set it at all. `aie2p/cast_f32_bf16.cc:27` already has the idiom, `aie::swap_rounding(...)` plus `set_rounding(saved)`, so it is one line per kernel. The mode is a sticky per-core register, so whatever runs next on that core inherits it.

**The flash-attention denominator accumulates in bf16.** `softmax.cc:169` stores the f32 block exp-sum into a `bfloat16` `scale_buffer`, and `mha.cc` then folds `l_i = max_diff_exp * l_{i-1} + accum_exp_val` in bf16 across every KV block. Worth a measurement against an f32 sidecar before it becomes the reference attention path. Same shape as the `layer_norm` accumulator width. Separately, `softmax.cc:47` and `:127` seed the running max at `0` rather than `lowest()`; the algorithm stays self-consistent because the same `m_i` feeds both the exponentials and the fold, so this only costs conditioning.

**`expand` is unsigned, named signed.** `expand.cc:75` is `expand_int4_to_bfloat16(uint4 *a_in, ...)` instantiating `expand<uint4, ...>`, `aie::unpack` zero-extends, and there is no zero point. The kernel, the factory and the test are all consistent as unsigned; only the symbol name and the README row say `int4`. Also, `in + N/2` and `pI += block_size/2` are load-bearing on `sizeof(uint4) == 1` (`types.hpp:127` aliases `uint4_t`, which Peano defines as `unsigned _BitInt(4)`). That holds today; a `static_assert` would keep it holding.

**`mha.cc:15` includes `mm.cc`.** `mm.cc:417` to `:522` is an `extern "C"` block, so those definitions land in the mha translation unit. Any design binding both on one core gets duplicate symbols. The `-DB_COL_MAJ` coupling to the PeanoCompilationRule also lives only in a comment.

**Pragma constants that do not track their template parameter.** `add.cc` and `mul.cc` hardcode `AIE_LOOP_MIN_ITERATION_COUNT(16)` while the trip count is `N/32`; `mv.cc:59` writes `k / VEC_SIZE` where the template parameter is `r`. Every current instantiation satisfies them, so this is latent, but `mv.cc` says in its own comment that the pragma "will break the code if that is not the case", which a `static_assert` would enforce for free.

**`VERSIONED_LOOP`.** `body` is an unparenthesized macro argument, so a comma anywhere in an inlined lambda body splits into `__VA_ARGS__` and gets spliced in as pragmas; braces do not protect commas from the preprocessor. Both current call sites pass a named variable, so it is only a footgun for the next user. The doc comment could also say that `_vl_i` is not visible to the body.

**Cleanup in `mha.cc`.** `matmul_PV` (`:78`) and `rescale_O` (`:115`) carry the same hardcoded `j*64 + k*8 + l*512` triple loop, with `// VJUNG: TODO: Make this generic for every tile size` at `:127`; `rescale_O` takes an `idx_buffer` it never reads; `matmul_PV`'s first two parameters are named `Q` and `K` for what are P and V.

**Smaller ones.** `mask_bf16` (`softmax.cc:196`) writes `-INFINITY` while `mha.cc:189` writes `numeric_limits<bfloat16>::lowest()`; with `-inf` a fully masked row gives `inf - inf`. `leaky_relu` is routed through `_require_fixed_tile_size(..., 1024)` although it is LUT-free on both arches and takes the element count at runtime. The aie2p activation e2e passes at `rtol=0.128`, which is loose for a native tanh. The aie2p tanh/sigmoid/silu comments say the split into two 16-lane halves is required, but `aie::tanh` is generic over `Elems` and unrolls at `native_lanes = 16` internally, and `gelu.cc` in this PR calls it 32-wide; the aie2 files genuinely do need the split, since `getTanhBf16` takes 16 lanes. `mv.cc` needs `-DDIM_K` with no `#error` guard, while `transpose.cc` sets that precedent. `transpose.cc` has an unguarded `#define min`, ships a generic `copy` symbol, and spells "transpsoe" twice in its header. `rope.cc` and `aie_kernel_utils.h` end without a newline, and `axpy.cc` keeps a commented-out `#pragma clang loop`.

jgmelber (approved):

andrej:
We will merge this as-is so all kernel development moving forward can happen here. We'll redirect kernel changes from amd/IRON to here. I've created issue #3677 to keep track of the remaining minor cleanups discussed here that we can implement as a follow-on PR.

PR #3601, opened 2026-08-26
[iron] Reject a fill/drain that does not cover whole objectFIFO objects [MED]

atassis:
## Description

A `fill()`/`drain()` emits one shim BD; the far end of the stream receives whole
objectFIFO objects, and nothing compares the two extents. `aie.shim_dma_allocation` — the
only thing joining a runtime transfer to its fifo — carries the tile, direction and
channel and nothing about what flows through it, so no verifier can see the pair either.
A transfer that ends part-way through an object therefore builds clean and then hangs:
the receiving buffer descriptor is sized by the fifo's object type, so a partial one never
completes, its lock is never released, and the consumer's acquire never unblocks.

```
aie.objectfifo @in(...) : !aie.objectfifo<memref<300xf32>>
aie.runtime_sequence(%arg0: memref<1024xf32>) {
  aie.dma_bd(%arg0 : memref<1024xf32> offset = 0 len = 1024 ...)  <- 3.41 objects
```

This compares the two extents in `_emit_transfer`, where both are in hand, before any IR
is emitted. The element types need not agree: a DMA does not interpret its payload, and a
host buffer that is a wider view over the bytes of a differently-typed stream is an
ordinary idiom — `ml/mobilenet` reuses one i32 allocation for i8 activations and ui16 FC
data, computing every offset in i32 units by hand. Identical types compare in elements,
differing ones in bits, and a differing pair where either side reports no bit width
(block-float) is skipped rather than measured; asking those for a width is the same trap
`objectfifo_blockfloat_default_pad` already guards. The check is skipped where the extent
is not knowable here: a scalar RuntimeData carries no BD geometry, and a runtime-valued
access pattern has no constant to check. The `ir.Type` helpers live in
`python/helpers/util.py` alongside the other `ir.Type`/numpy-dtype conversions, per
@hunhoffe's review on the previous revision.

`test/python/objectfifo_transfer_covers_whole_objects.py` covers it: an aligned transfer
emits the same IR as before, a partial object on the fill side is rejected, the drain side
is rejected on its own, a wider host view of the same bytes still builds, a mismatched
view that does not divide is rejected, a `v8bfp16ebs8` fifo still resolves, and the
explicit `sizes`/`strides`/`transfer_len` form is covered in both shapes — outermost
dimension as the BD's repeat count, and an explicit `transfer_len` overriding the inner
three. The file fails on an unpatched tree. That last pair is not reachable through `tap=`
and has no user in `programming_examples`, but `test/npu-xrt/matmul_whole_array_dynamic`
passes `sizes` and `transfer_len` together and `test/npu-xrt/rolled_pingpong_passthrough`
passes `sizes` alone; both build unchanged, the former at 1, 2 and 4 columns.

Swept every `programming_examples` design that calls `fill()`/`drain()` — all 61 files —
plus the variants that change transfer geometry (`transposes` x4 strategies,
`dma_padding` x9 api/pad combinations, `mobilenet` x13 per-block and x3 chain modes).
Nothing is rejected; seven designs offer no generate-only path and are covered by the
device jobs. All 87 `test/python/*.py` behave identically apart from the new file, as do
the collectable tests under `test/python/npu*`. `test_objectfifo_pad_value.py` needs a
device to collect, so its design was replicated standalone — the padded fifo resolves,
which is the topology most likely to be mis-measured here.

## Scope

IRON-only. A design written against the raw dialect gets no protection. The natural home
for a verifier would be `NpuDmaMemcpyNdOp`, which already resolves its `metadata` symbol,
but `aie.shim_dma_allocation` carries no type information, so a verifier-level check needs
the fifo's object type plumbed onto the allocation in `AIEObjectFifoAllocate` first.

## Checklist

- [ ] `clang-tidy` passes locally for any touched file on the enabled list (no C++ touched)


hunhoffe:
Another nice fix!
  
One suggestion: `_element_width`/`_same_transfer_width` are pure `ir.Type → int | None` helpers with no ObjectFifo-specific logic — could they move to python/helpers/util.py? That module already holds the other `ir.Type/numpy-dtype` conversion helpers this file imports (`np_ndarray_type_to_memref_type`, `memref_type_to_np_dtype`), so it seems like the natural home.

atassis:
@hunhoffe Pushed a fix and rebased

atassis:
The check changed shape since your review, @hunhoffe.

Device CI rejected `ml/mobilenet`, which declares its runtime buffers as an i32 view
over byte buffers and feeds them to i8 and ui16 fifos on purpose, computing every
offset as `bytes // 4`. Comparing element widths kills 50 of its transfer sites. And
nothing separates it from the case I was after: a whole-buffer fill of 1024 bf16 into
a 256xf32 fifo has the same shape in the IR and is well formed, moving 2 objects
instead of 4. No verifier can see that one.

What is visible is a transfer ending part way through an object: the receiving BD
never completes and the consumer's acquire never unblocks. That is what the check
tests now, in bytes. Swept all 61 designs that call fill/drain, 1497 transfer sites,
nothing rejected.

Which raises where this belongs. Scope says a verifier needs the object type plumbed
onto `aie.shim_dma_allocation` first, and I have that working: the allocation records
the peer pool's `elemType` (resolved through the route, since `fifoName` is optional)
and `DMAConfigureTaskForOp::verify` checks the extent. `check-aie` is unchanged apart
from one CHECK line in `plio_test.mlir` that pinned `{plio = true}` as the whole
attribute dictionary.

They are not redundant. IRON fails at construction with a traceback at the `fill()`
call, and still applies where the allocation records nothing: a broadcast to different
object types, or an endpoint that pads into its object and so receives more than the
shim sends. The verifier covers designs that never go through IRON, which is what
iree-amd-aie and mlir-air emit.

Separate PR for the dialect side and keep this one as is, or would you rather have
only the verifier and I drop the IRON check from here?

hunhoffe:
@atassis Thank you for the thorough review! I think we want to keep both, for user friendliness. I will kick off another review

hunhoffe (commented):
On first glance, this looks good to me. I like user-friendly checks that catch errors quickly. Due to the dyn seq work, sometimes a design may not be fully paired with a corresponding runtime sequence. However, we have few examples of that now and this work could be modified to be more flexible to those cases if necessary.

I'll wait for CI and copilot to re-review and then prepare for merge.

atassis:
I am creating a set of rules now per each kernel, and trying to find out the general set of invariants for them, so those can be introduced directly in compiler. I want to migrate some checks I do manually into the compiler, but for now a lot of information to process. If you have any ideas on this topic- we could discuss in discord. Or, if your team is ok to make a brainstorm session- I can preplan it, so we can discuss what should be done next. 
I struggle a lot now because most of obvious issues are researched, some of them are directly related to my initial idea about contract, but llvm-aie team is busy, while it would take some time for me to digest my ideas for them for easier review (are simulator is too much for everyone, for example)

atassis:
Rebased onto #3683, which declares `stream_len_decoupled` and carries it to the pool. Until that
merges this PR's diff shows both; the part belonging here is `python/iron/`,
`python/helpers/util.py`, the test, and the `dma_compression` opt-in.

The check now skips a fifo whose channel (de)compresses, which is what device CI was rejecting:
`dma_compression` in its `cmp_only` arm drains 2944 i32 through a 1024 i32 object, and that ratio
is a property of the data rather than a mistake. `test/python/objectfifo_transfer_covers_whole_objects.py`
gained a case asserting the exemption.

The dialect-side verifier stays in #3672, consuming the same attribute.

hunhoffe:
@atassis -- this is blocked since it's rebased onto other PRs which aren't merged yet.

PR #3608, opened 2026-08-27
[AIEX] Add an npu-xrt device test for dma_channel_reset_for [LOW]

atassis:
### Problem

`aiex.dma_channel_reset_for` landed in #3393 with MLIR and pass-level lit coverage
only. Nothing exercises the re-arm on hardware, so a regression in the lowering would
show up as wrong data on a resident objectFIFO rather than as a failing test.

### Fix

`test/npu-xrt/dma_channel_reset_for/`: a `weights` objectFIFO of depth 1 with
compile-time `initValues`, memtile (1,1) to core (1,2), held resident across 1000
dispatches on one hardware context and re-armed each dispatch with
`dma_channel_reset_for`. The core acquires it once and holds it across all inner
iterations, so its lock and channel-queue state has to survive from one dispatch to
the next. `inputs` and `outputs` stream per dispatch and vary, so a fifo that stops
delivering surfaces as the previous dispatch's output rather than as a hang.

### Test

Run on npu2 (Strix): **1000/1000 dispatches exact**, `output[i] == input[i] + i + 1`
checked on every element of every dispatch.

Negative control, to show the test has detection power: deleting the
`dma_channel_reset_for` line from the generated MLIR and rebuilding the xclbin makes
the same binary fail at dispatch 1 with `qds_device::wait() unexpected command state`.
That is the second of the two failure modes the README documents, the one where the
mailbox channel stops accepting commands.

I ran the legs by hand (`aie2.py` to MLIR, `aiecc` to xclbin, `host_clang` equivalent,
then `./test.exe 1000`) rather than through lit, because a locally built tree resolves
tests from its own configured checkout and does not discover a new test directory
added on a branch. The `RUN:` lines are the ones CI will execute.


hunhoffe:
In general, I'd like to avoid putting documentation in READMEs in the test directories unless there is a really good reason to put it there. Features and operations, etc. should be documented in source code or in user-facing documentation, and documentation scattered throughout is harder to maintain.

Otherwise, thanks for helping ensure this features is tested!

test/npu-xrt/dma_channel_reset_for/test.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "xrt/xrt_bo.h"
+#include "xrt/xrt_device.h"
+#include "xrt/xrt_kernel.h"
+
+namespace {
+constexpr int kLength = 256;
+constexpr int kDefaultIterations = 1000;
+constexpr unsigned int kDispatchTimeoutMs = 2000;
+
+std::vector<uint32_t> loadInstructions(const std::string &path) {
```
hunhoffe, line 22: do we have helper functions for these?

hunhoffe (approved):

PR #3609, opened 2026-08-27
[AIE] Run count checks before mixed-list reads in the ODS verifier [MED]

atassis:
## Description

`myVerifyOffsetSizeAndStrideOp()` compares `op.getMixedOffsets().size()` against
`op.getMixedSizes().size()` before validating that each list's dynamic-operand count
agrees with its static list's `kDynamic` sentinel count. Both accessors call
`mlir::getMixedValues()`, which asserts
`dynamicValues.size() == count_if(staticValues, isDynamic)`. An `aiex.npu.dma_memcpy_nd`
whose `static_sizes` carries a `kDynamic` sentinel while its `$sizes` operand list is
empty therefore hits that assert during verification instead of reaching a diagnostic:
SIGABRT in an assertions build, an out-of-bounds read in release. It takes hand-written
or fuzzed generic syntax to reach, but a verifier should diagnose malformed IR, not
abort on it.

This moves the three `verifyListOfOperandsOrIntegers` calls ahead of the
mixed-offsets-vs-mixed-sizes rank check — the order
`mlir::detail::verifyOffsetSizeAndStrideOp` already uses upstream, where llvm-project's
`ViewLikeInterface.cpp` validates the same dynamic-operand counts before calling the
mixed accessors. Once the per-list checks have run, each dynamic count matches its
static list's `kDynamic` count, so the mixed accessors cannot assert. The reordering
cannot change any verdict: the checks are independent, side-effect-free, and all
required, so only which failing check reports first can differ.

Covered by `test/dialect/AIEX/bad_npu_nd_mixed_list_counts.mlir`: a generic-syntax
`aiex.npu.dma_memcpy_nd` with `static_sizes = array<i64: kDynamic, 4, 4, 4>` and an
empty `$sizes` list. Built `aie-opt` and ran it both ways — with the fix
`--split-input-file --verify-diagnostics` exits 0; without it the same binary aborts in
`StaticValueUtils.cpp:214` before emitting any diagnostic.

## Checklist

- [ ] `ruff check` and `pyright` pass locally for any touched Python file in their covered paths (no Python touched)
- [ ] `clang-tidy` passes locally for any touched file on the enabled list (file not on the enabled list)

### DIFF
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -173,11 +173,14 @@ xilinx::AIE::myVerifyOffsetSizeAndStrideOp(OffsetSizeAndStrideOpInterface op) {
   if (failed(verifyListOfOperandsOrIntegers(
           op, "stride", maxRanks[2], op.getStaticStrides(), op.getStrides())))
     return failure();
-  if (!(op.getMixedOffsets().size() == 1 && maxRanks[0] == 1) && // NOLINT
-      op.getMixedOffsets().size() != op.getMixedSizes().size())
+  // Each getMixed* call rebuilds its list, so read them once.
+  const size_t numMixedOffsets = op.getMixedOffsets().size();
+  const size_t numMixedSizes = op.getMixedSizes().size();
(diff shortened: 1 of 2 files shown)


hunhoffe:
I really don't want our tools to SIGABORT, etc. so thank you for looking into this!

PR #3611, opened 2026-08-27
[HRX] Discover pip-installed libhrx without HRX_LIBHRX

jtuyls:
## Description

Python only searched env hints and FindHRX filesystem roots, so a pip runtime was invisible unless HRX_LIBHRX was set. After those hints, look in site-packages for the conventional <package>/lib or <package>/bin layout. No package import, no dlopen, and CMake is unchanged.

## Checklist

### DIFF
--- a/programming_guide/hrx_runtime.md
+++ b/programming_guide/hrx_runtime.md
@@ -41,9 +41,14 @@ values and every HRX environment variable
 
 ## Provisioning `libhrx`
 
-Fetch, checksum-verify, and extract the pinned HRX release (coordinates in
-`utils/hrx-release.env`) with the helper script, then source the environment it
-prints:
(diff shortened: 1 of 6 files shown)


hunhoffe (approved):
This makes sense given HRX doesn't have a stable pip package name yet. Once it does, might be worth moving to the same pattern as hsaruntime/discovery.py's _wheel_root() (importlib.util.find_spec() on a known name) instead of scanning all of site-packages.

jtuyls:
> This makes sense given HRX doesn't have a stable pip package name yet. Once it does, might be worth moving to the same pattern as hsaruntime/discovery.py's _wheel_root() (importlib.util.find_spec() on a known name) instead of scanning all of site-packages.

Yes, will do that, thanks!

PR #3612, opened 2026-08-27
ci: key each ccache on what changes it, and stop caching DWARF [HIGH]

atassis:
## Problem

This repo's Actions cache is full: 9.93 GiB of the 10 GiB limit, 52 active entries.
Most of it is not reusable, for four separate reasons, and the eviction that follows
falls on the caches that would otherwise hit — `generateDocs` and `codeCoverage`
currently have zero entries surviving.

**The assert leg caches DWARF.** The `Assert` leg of `buildAndTestPythons.yml` builds
`CMAKE_BUILD_TYPE=Debug`, so ccache stores objects carrying debug info that no CI job
reads. `ccache-Assert-Linux-24.04` is 7 entries totalling 4.81 GiB, 47% of the budget,
against 43.5 MiB per entry for the `Release` leg of the same matrix.

**Three keys do not describe their contents.**

- `buildAndTestPythons.yml` keys on `build_type-os-ubuntu_version` and omits
  `python_version`, so four Python legs share a key and only the newest is restored.
- `buildAndTestMulti.yml` keys on `os-arch-assert-rtti` and omits `matrix.COMPILER`,
  which selects llvm or gcc and so determines every object. Eight ubuntu legs collapse
  onto four entries; the two compilers alternate, so a leg restores the other's cache
  about half the time and then misses everything in it.
- `generateDocs.yml` keys on the submodule hash with no `restore-keys`, so every bump
  mints a family with nothing to restore and orphans the previous one.

The comment above the first of those already states the intent the key does not
implement: "use a different key per job".

## Fix

Build the assert leg `Release` with `LLVM_ENABLE_ASSERTIONS=ON`; add `python_version`
and `matrix.COMPILER` to their keys; give the docs cache a `restore-keys` fallback.

Every other build here that wants assertions already uses Release plus assertions:
`utils/build-mlir-aie-from-wheels.sh` (what `buildAndTestRyzenAI` runs),
`build-mlir-aie.sh`, `build-llvm.sh`, `build-llvm-cross-local.sh`, and
`lintAndFormat.yml`. That leg was the only `CMAKE_BUILD_TYPE=Debug` build in CI, and
`lintAndFormat.yml`'s pre-commit cache was the only cache in the repo already setting
`restore-keys`.

## Test

`check-aie`, `check-reference-designs` and `check-programming-guide` are unchanged and
still run with assertions on. The assert leg now exercises an optimized build, which is
what ships. `LLVM_DEBUG` and `-debug-only` are gated on `NDEBUG`, which
`LLVM_ENABLE_ASSERTIONS` controls rather than the build type, so they keep working; no
test in `test/` uses `REQUIRES: asserts` or `-debug-only`.

The key changes alter no build. `compiler_check=content` already kept the shared-key
case correct rather than wrong, so what changes is hit rate, not results. Expect the
first run after this lands to be cold for the re-keyed legs.

Cache figures from `gh api repos/Xilinx/mlir-aie/actions/cache/usage` and
`.../actions/caches`, 2026-08-27.


atassis:
Once this lands the existing `ccache-Assert-Linux-24.04-*` entries (7, 4.81 GiB) no longer match any restore prefix and will age out by LRU and the 7-day expiry. Deleting them reclaims that immediately if you'd rather not wait.
You can delete them manually via next commands
```shell
gh cache list --repo Xilinx/mlir-aie --limit 100 | grep 'ccache-Assert-Linux-24.04-2026'
gh cache delete <id>      # per entry
```

hunhoffe (approved):
Looks good to me! Can you double check copilot comments before I merge?

atassis:
Fixed per copilot review

PR #3613, opened 2026-08-27
[aiecc] Validate stack_size against each core's call graph

hunhoffe:
## Description

Adds automatic stack-size validation for AIE cores. Nothing previously checked `stack_size` against actual usage — the stack sits directly below buffers with no clearance, so an undersized one silently corrupts memory at runtime rather than failing loudly (unlike a linker "out of space" error for `.data`/`.bss`).

Computes each core's real requirement by:
- Emitting a `.stack_sizes` section from every kernel and core compile (Peano; Chess already has `.stackinfo`).
- Walking the call graph across a core's `link_files` objects (`lib/Analysis/StackSizeAnalysis.{h,cpp}`), taking the max over root-to-leaf paths (not a sum — only one call chain is live at a time), including cross-object calls and a conservative indirect-call (function-pointer) heuristic.
- Validating in two places, since the two halves of the requirement become knowable at different times: an early warning once callees' subtrees are known, then a hard build failure post-compile once the core body's own frame is known too — naming the exact value to set.
  
`stack_size_override` (on a kernel's `func.func`/`Kernel`/`ExternalFunction`) is the escape hatch for symbols the analysis can't size on its own — recursion, unresolved indirect calls, Chess-compiled objects, or `link_with_mode="merge"` kernels. `--no-auto-stack-size` disables the whole check.

See `programming_guide/core_data_memory.md` for the full write-up and a lookup table of what to do for each diagnostic.

**Relates to #2347.** That issue asked for exactly this: a call-graph pass that computes maximum stack depth and errors/warns on overflow, and specifically called out indirect calls and unbounded recursion as the hard cases. This PR implements that — indirect calls are folded in conservatively (can only overcount) and a recursion cycle is a hard build failure requiring an explicit stack_size_override. The one thing it deliberately does not do is #2347's other suggestion, automatically resizing stack_size: the design here is "the compiler measures and reports, the user declares and rebuilds," so a fix can't drift silently between the source and the built binary.
  
Testing: 20 new lit tests (test/aiecc/stack_size_*, test/aiecc/default_stack_size.mlir, test/aiecc mixed_measurable_link_files_warns.mlir) plus Python unit/FileCheck coverage for the stack_size_override API surface (test/python/external_func_link_with.py, test/python/test_stack_sizes_kernel_compile.py).

## Checklist


include/aie/Analysis/StackSizeAnalysis.h
```diff
@@ -0,0 +1,151 @@
+//===- StackSizeAnalysis.h -------------------------------------*- C++ -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
```
hunhoffe: Claude: This comment is WAYYYY too long.

include/aie/Analysis/StackSizeAnalysis.h
```diff
@@ -0,0 +1,151 @@
...
+
+// One function's contribution to the call graph: its own frame size and the
+// symbols it (directly, or conservatively via an indirect-call data
+// reference) calls.
+struct StackGraphNode {
+  int64_t frameSize = -1; // -1 = not yet measured
+  llvm::SmallVector<std::string, 4> callees;
+};
+
+// The merged call graph across every object linked into one core. Symbols
```
hunhoffe: Claude: Comment too long.

include/aie/Analysis/StackSizeAnalysis.h
```diff
@@ -0,0 +1,151 @@
...
+// are looked up by name, so a callee defined in a sibling object resolves
```
hunhoffe: Claude: Comment too long.

include/aie/Analysis/StackSizeAnalysis.h
```diff
@@ -0,0 +1,151 @@
...
+// object has contributed.
+struct StackGraph {
+  llvm::StringMap<StackGraphNode> nodes;
+  llvm::StringMap<llvm::StringSet<>>
+      dataReferences; // func -> data syms it touches
+  llvm::StringMap<llvm::SmallVector<std::string, 2>>
+      dataEscapes; // data sym -> funcs stored there
+};
+
+// Pass 1: collect every function symbol `path` *defines* (not merely
```
hunhoffe: Claude: comment too long and numbered comments are hard to maintain.

include/aie/Analysis/StackSizeAnalysis.h
```diff
@@ -0,0 +1,151 @@
...
+void resolveIndirectCallEdges(StackGraph &graph);
+
+// Cycle: the requirement is unbounded, so an override is always required.
+// Unmeasurable: the same "haven't measured this yet" gap reserved_data_size
+// treats as a warning (an archive/bitcode input, or a pre-existing kernel
+// object) -- failing outright would break every such design on first
+// contact with a newer aiecc.
+enum class StackRequirementFailure { Cycle, Unmeasurable };
+
+// Result of computing one core's stack requirement: either the computed
```
hunhoffe: All of these comment blocks are too long.

include/aie/Dialect/AIE/IR/AIECoreSymbols.h
```diff
@@ -0,0 +1,46 @@
...
+#define AIE_DIALECT_AIE_IR_AIECORESYMBOLS_H
+
+#include "llvm/ADT/StringRef.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include <string>
+
+namespace xilinx::AIE {
+
+// Stack requirement aiecc's call-graph analysis computed for a core, stamped
```
hunhoffe: Claude: Both comment blocks too long.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -415,9 +415,15 @@ def AIE_CoreOp
     are always stored in the local core memory, to avoid conflicts with static data
     allocations in other cores.
 
-    The reservation is not checked against the core's actual frames, and the stack sits
-    directly below the objectFIFO buffers with no clearance, so a design with large frames
-    must set this attribute or it will overwrite them silently.
+    The stack sits directly below the buffers with no clearance, so a `stack_size` too
```
hunhoffe: Claude: comment too long.

lib/Analysis/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+#
```
hunhoffe: Claude: Do we really need this new library? Are you sure it doesn't fit into an existing category? Is this llvm/mlir best practice?

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -46,6 +47,11 @@ static StringRef getArchIntrinsicString(AIEArch arch) {
     return "aie2";
   case AIEArch::AIE2p:
     return "aie2p";
+  case AIEArch::AIE2ps:
```
hunhoffe, line 50: Note for reviewers: this with a clang-tidy note.

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -170,10 +176,25 @@ static void declareAIEIntrinsics(AIEArch arch, OpBuilder &builder) {
   case AIEArch::AIE2p:
     registerIntrinsics(getAIE2pIntrinsics(builder));
     return;
+  case AIEArch::AIE2ps:
```
hunhoffe, line 178: Note for reviewers: this with a clang-tidy note.

lib/Dialect/AIE/Transforms/AIECoreToStandard.cpp
```diff
@@ -541,8 +562,7 @@ struct AIECoreToStandardFunc : OpConversionPattern<CoreOp> {
     // The parent should be an AIE.device op.
     rewriter.setInsertionPointAfter(op->getParentOp());
 
-    std::string coreName("core_" + std::to_string(col) + "_" +
-                         std::to_string(row));
+    std::string coreName = coreFrameSymbolName(col, row);
```
hunhoffe, line 564: Claude: what assumptions does this make about symbol naming? Are they safe and consistent?

programming_examples/ml/bottleneck/bottleneck.py
```diff
@@ -208,6 +208,7 @@ def conv3x3_fn(of_wts, of_act_in, of_act_out, conv3x3, channel_offset):
                 0,
             ],
             tile=Tile(0, 3),
+            stack_size=0x1240,
```
hunhoffe: Claude: Can you use human readable integers instead of hex in the three edits in this file?

programming_examples/ml/conv2d/conv2d.py
```diff
@@ -82,7 +82,7 @@ def conv2d(
             output_channels=out_channels,
             act_dtype=np.int8,
         )
-        stack_size = 0xA00
+        stack_size = 0xA80
```
hunhoffe: Claude: use base 10 please.

python/dialects/aie.py
```diff
@@ -204,6 +204,17 @@ class external_func(FuncOp):
             the core's LLVM module with ``llvm-link`` before codegen instead of
             object-linking it.  Requires ``link_with``.  When omitted, the
             artifact is object-linked, whatever its suffix.
+        stack_size_override: Optional declared upper bound, in bytes, on the
```
hunhoffe: Claude: This seems like too much detail. Maybe a pointer to where to find more detail would be better?

test/aiecc/default_stack_size.mlir
```diff
@@ -0,0 +1,81 @@
+//===- default_stack_size.mlir -------------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// --default-stack-size stands in for the target's built-in default (1024
```
hunhoffe: Claude: too many comments in this file.

test/aiecc/mixed_measurable_link_files_warns.mlir
```diff
@@ -0,0 +1,55 @@
+//===- mixed_measurable_link_files_warns.mlir ----------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
```
hunhoffe, line 3: Claude: too many comments in this file.

test/aiecc/stack_size_absent_insufficient_error.mlir
```diff
@@ -0,0 +1,61 @@
+//===- stack_size_absent_insufficient_error.mlir ------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// stack_size is absent, and entry_a's real frame (~4160 bytes, see
```
hunhoffe: Claude: too many comments in this file.

test/aiecc/stack_size_cross_object.mlir
```diff
@@ -0,0 +1,58 @@
+//===- stack_size_cross_object.mlir --------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// Exercises the two-pass cross-object attribution (see StackSizeAnalysis.h):
```
hunhoffe: Claude: too many comments in this file.

tools/aiecc/aiecc.cpp
```diff
@@ -274,6 +274,11 @@ buildObjectSubgraph(EdgeWithTypedOutput<ModRef> &lowered,
       .arg("-O" + std::to_string(optLevel.getValue()))
       .value("--march=")
       .arg("--function-sections")
+      // Record each function's frame size in a `.stack_sizes` section (metadata
```
hunhoffe: I don't think this comment isneeded.

tools/aiecc/IRTransforms.h
```diff
@@ -185,6 +188,199 @@ collectCoreIRLinkFiles(xilinx::AIE::CoreOp coreOp, llvm::StringRef inputFile,
   return files;
 }
 
+// Clone `src` and set `stack_size = defaultStackSize` on every CoreOp that
+// doesn't already carry `stack_size` explicitly -- a design-wide way to say
```
hunhoffe: Claude: comments are too verbose throughout the changes in this file.

mkdocs.yml
```diff
@@ -245,6 +245,7 @@ nav:
     - "Section 6: ML Examples": programming_guide/section-6/README.md
     - Extra References:
       - IRON Configuration: programming_guide/iron_configuration.md
+      - Core Data Memory: programming_guide/core_data_memory.md
```
hunhoffe, line 248: Claude: This doesn't seem like the right place for this documentation, I don't think? Do you have any suggestions for how to manage this material more cohesively?

hunhoffe (commented):
Comments for claude.

include/aie/Dialect/AIE/IR/AIECoreSymbols.h
```diff
@@ -0,0 +1,40 @@
...
+inline constexpr llvm::StringLiteral kComputedStackRequirementAttrName =
+    "aiecc.computed_stack_requirement";
+
+// Name AIECoreToStandard gives the function it outlines a CoreOp's body
+// into. Shared so every reader of a compiled core object agrees on it.
+inline std::string coreFrameSymbolName(int col, int row) {
+  std::string name;
+  llvm::raw_string_ostream(name) << "core_" << col << "_" << row;
+  return name;
+}
```
andrej: This is a really good improvement to make sure those names don't drift.

jgmelber (approved):

programming_guide/core_data_memory.md
```diff
@@ -0,0 +1,158 @@
...
+whatever memory the buffers left behind. Buffer placement therefore silently
+decided whether the core would link — and it had no idea it was doing so. The
+failure modes were correspondingly opaque: a stack that quietly overwrote the
+buffers directly above it, or a linker "could not find free space" error that
+named neither the tile nor the buffers responsible.
+
+The compiler now measures and checks the **stack** region against what a
+core's call graph actually needs; this page describes the attributes and
+flags that control that check, and what to do when it fires. (The core's own
+compiled sections are reserved and validated by a separate, later change.)
```
andrej: Is this historical framing intentional? I think it's clear enough, though might be shorter if we leave off the "how it used to be" part and just describe what it does now.

programming_guide/core_data_memory.md
```diff
@@ -0,0 +1,158 @@
...
+  overrides any scheme set on the tile. Bank placement can also be requested per
+  buffer with `Buffer(mem_bank=...)`.
+
+## What to do when you hit a diagnostic
+
+**`cannot determine this core's stack requirement: ...` (error).** The call
+graph has a cycle (recursion), which is unbounded. Set `stack_size_override` on
+the affected kernel's `external_func()`/`func.func` declaration to a value large
+enough for the deepest recursion, or pass `--no-auto-stack-size` to skip the
+check entirely.
```
andrej: :+1:

programming_guide/core_data_memory.md
```diff
@@ -0,0 +1,158 @@
...
+objects and is *not* fatal; the compiler simply cannot validate `stack_size` for
+this core. If you want validation, set `stack_size_override` on the affected
+kernel so the analysis has a number to trust.
+
+**`stack_size is absent ... but this core's real requirement is N bytes; set
+stack_size = N explicitly on this aie.core ... and rebuild` (error).** The
+device default was assumed when the buffers were placed, but the finished core
+needs more. Do exactly what it says — set `stack_size = N` (or
+`Worker(stack_size=N)`) and rebuild — or pass `--no-auto-stack-size` to skip
+the check.
```
andrej: At this point I think we are certain the core needs `N`, so `--no-auto-stack-size` is bad advice because we know it will get corrupted. Suggest removing the last suggestion, or adding something like "if you know what you are doing and believe the stack size estimator is wrong (please report the bug/issue in that case!)"

tools/aiecc/aiecc.cpp
```diff
@@ -651,29 +652,70 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+        if (stackSize <= 0)
+          return ModRef(mod.get().clone());
+        return populateDefaultStackSize(mod.get(), stackSize);
+      });
+
+  // aie-assign-core-link-files moves here from inside
+  // getInputWithAddressesPipeline (it used to run after address assignment):
+  // it's a pure call-graph analysis over func.func/func.call, so nothing
+  // depends on addresses, and the stack-size check below needs its
+  // link_files output.
```
```suggestion
  // aie-assign-core-link-files is a pure call-graph analysis over 
  // func.func/func.call; the stack-size check below needs its link_files 
  // output.
```

tools/aiecc/aiecc.cpp
```diff
@@ -1664,6 +1706,154 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+// (a warning would ship a proven overflow). Like every check in this
+// analysis, it never auto-adjusts anything. `--no-auto-stack-size` skips it,
+// same as the earlier warning.
+//
+// Re-derives the early, cheap pipeline stages on a fresh parse of the input,
+// since the build graph doesn't expose its intermediate module once
+// `engine.run` has returned; the compiled core object itself is read back
+// from where the real build wrote it, not rebuilt. Returns true if any
+// core's requirement was insufficient; an `llvm::Error` reports a mechanical
+// failure of this check itself.
```
andrej: ❗

tools/aiecc/aiecc.cpp
```diff
@@ -1664,6 +1706,154 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+    if (auto err = runStage("trace pipeline",
+                            xilinx::aiecc::getTracePipeline(&context)))
+      return std::move(err);
+    {
+      auto pm = std::make_unique<mlir::PassManager>(&context);
+      pm->nest<xilinx::AIE::DeviceOp>().addPass(
+          xilinx::AIE::createAIEAssignCoreLinkFilesPass());
+      if (auto err = runStage("link-files assignment", std::move(pm)))
+        return std::move(err);
+    }
```
andrej: I'm worried about duplicating these lowering steps.

- What if these lowering pipelines change? Will the person making the change to the pipelines further up in the code remember to change the code down here as well?
- Although these are likely very fast passes, this re-does work.
- This does not use the graph abstraction, so does not explicitly capture what it depends on, and may be opaque to the user (does not show up in compilation graph, unclear how it interacts with --cut/--resume).

tools/aiecc/aiecc.cpp
```diff
@@ -1951,6 +2146,31 @@ int main(int argc, char **argv) {
...
+      for (const std::string &path : writtenOutputPaths) {
+        std::error_code ec = llvm::sys::fs::remove(path);
+        if (ec && ec != std::errc::no_such_file_or_directory)
+          llvm::errs() << "aiecc: warning: could not remove invalid output '"
+                       << path << "': " << ec.message() << "\n";
+      }
+      llvm::errs() << "aiecc: pipeline failed\n";
+      return 1;
+    }
+  }
```
andrej: Hopefully, this test ("needs a real input") should be expressible as a dependency in the graph so this handling could go away.

andrej (commented):
Thanks for tackling this, and throwing in some cleanup along the way! This will be very useful. This is a good, thorough approach with the full call graph analysis.

I'm a little worried about the way this is currently integrated with `aiecc`. The red flags for me are the duplicated compilation steps, and the fact that there is a special case for checkpoint/resume (`--cut`).

Could we integrate the stack analysis declaratively as a node in the compilation graph? An explicit "stack_analysis" node could have explicit dependencies on (a) the required intermediate MLIR and (b) the object files. If integrated this way, the analysis will only run when those prerequisites are present, while reusing the lowering steps for those prerequisites. It will also make this a visible step in the compilation graph to the user, so it shows up when compilation is visualized, and as one of the steps in `--progress`; users could even select just the stack analysis as an output, which would run the minimal pipelines needed. The node would fail if analysis fails. We would probably need to add a feature to the graph execution engine that says "if any node fails, delete all intermediates and abort."

Thoughts on this?

I did not deeply review the stack analysis logic in StackSizeAnalysis.cpp, I trust that the implementation is reasonable.

And, just for my understanding, why are we running the analysis twice? I understand we need to compile the core's "main" to definitively know all stack sizes; is the earlier "warn-only" check so we still have a handle on the IR to attach an error to?

tools/aiecc/aiecc.cpp
```diff
@@ -1951,6 +2146,31 @@ int main(int argc, char **argv) {
   if (!checkpointDir.empty())
     writeCheckpoint(cutEdges, checkpointDir, graphArgv);
 
+  // The stack_size sufficiency check needs a real, complete build (a
+  // compiled core object to read back) -- skip it for a --cut early stop or
+  // a --dry-run, neither of which produced one.
+  if (cutEdges.empty() && !dryRun) {
+    llvm::Expected<bool> insufficient =
+        checkStackSizeIsSufficient(context, getInputFilename());
```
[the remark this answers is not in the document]
andrej: This might be a symptom of the stack analysis depending on intermediates/inputs that it does not explicitly declare as a dependency in the graph (because it is not in the graph).

tools/aiecc/aiecc.cpp
```diff
@@ -790,6 +801,36 @@ buildMainGraph(mlir::MLIRContext &context, Graph &g,
...
+  auto &withMeasuredStackSizes =
+      bundle(objects.out, physical.out)
+          .join<ModRef>(
+              "measured_stack_sizes.mlir",
+              [inputFile, workDirStr, skip = noAutoStackSize.getValue()](
+                  const Node<Directory> &objs, const Node<ModRef> &physicalN,
+                  Item<ModRef> &out) -> mlir::LogicalResult {
+                out.value = ModRef(physicalN.get().get().clone());
+                if (skip)
+                  return mlir::success();
```
andrej: Just skip the entire edge if noAutoStackSize is set. Downstream use (noAutoStackSize ? physical : physicalWithMeasuredStackSizes) to distinguish. Rename this from withMeasuredStackSizes to physicalWithMeasuredStackSizes

tools/aiecc/CommandLineOptions.h
```diff
@@ -151,6 +151,17 @@ inline cl::opt<bool> noUnified(
     "no-unified",
     cl::desc("Lower each core independently, against its own clone of the "
              "design (negates --unified; the default)"));
+inline cl::opt<bool> noAutoStackSize(
+    "no-auto-stack-size",
```
andrej: Bad name, suggests we are auto-setting the stack-size. Name "no-measure-stack-size". Replace analogously everywhere else.

andrej (commented):
for my bestie claudie

tools/aiecc/aiecc.cpp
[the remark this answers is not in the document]
andrej: Gone with graph-node rewrite

tools/aiecc/aiecc.cpp
```diff
@@ -1951,6 +2146,31 @@ int main(int argc, char **argv) {
...
+  // The stack_size sufficiency check needs a real, complete build (a
+  // compiled core object to read back) -- skip it for a --cut early stop or
+  // a --dry-run, neither of which produced one.
+  if (cutEdges.empty() && !dryRun) {
+    llvm::Expected<bool> insufficient =
+        checkStackSizeIsSufficient(context, getInputFilename());
+    if (!insufficient) {
+      llvm::errs() << "aiecc: stack_size check: "
+                   << llvm::toString(insufficient.takeError()) << "\n";
+      return 1;
```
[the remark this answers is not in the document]
andrej: No need for artifact cleanup after graph-rewrite because stack analysis is upstream of those artifacts, so they never get produced in the first place

andrej (approved):

python/dialects/aie.py
```diff
@@ -244,6 +262,10 @@ def __init__(
             self.operation.attributes["link_with"] = StringAttr.get(link_with)
         if link_with_mode is not None:
             self.operation.attributes["link_with_mode"] = StringAttr.get(link_with_mode)
+        if stack_size_override is not None:
+            self.operation.attributes["stack_size_override"] = IntegerAttr.get(
+                IntegerType.get_signless(32), stack_size_override
+            )
```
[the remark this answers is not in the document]
andrej: No one's gonna set a stack size >= 2^32 you silly goose.

PR #3616, opened 2026-08-27
[AIEX] Keep the runtime BD pool off statically-assigned ids

hunhoffe:
## Description

Same bug class as the per-tile BD-id fix in https://github.com/Xilinx/mlir-aie/pull/3615, for a third holder of ids: bd_pool_init seeds the free list with every id on the tile and hands out id 0 first, ignoring any BD the static allocator already placed there.

Emit a bd_pool_reserve per statically-assigned id right after pool init. A tile with no static BDs (the common case) emits none, so existing output is byte-identical.

Includes a new standalone lit test (a shim tile with a static BD on one channel and a pool-backed runtime sequence on another) — confirmed it fails without the fix.

## Checklist

### DIFF
--- a/lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
+++ b/lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
@@ -66,21 +66,12 @@ std::string bdPoolName(uint32_t col, uint32_t row) {
 llvm::SmallVector<uint32_t, 8> staticBdIdsOnTile(AIE::DeviceOp device,
                                                  uint32_t col, uint32_t row) {
   llvm::SmallVector<uint32_t, 8> ids;
-  auto collect = [&](AIE::TileElement elem, Region &region) {
-    auto id = elem.getTileID();
+  for (AIE::DmaBody program : device.getOps<AIE::DmaBody>()) {
+    auto id = program.getTileID();
     if (id.col != static_cast<int>(col) || id.row != static_cast<int>(row))
-      return;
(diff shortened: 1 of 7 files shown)


lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
```diff
@@ -60,6 +60,32 @@ std::string bdPoolName(uint32_t col, uint32_t row) {
...
+      if (auto bdId = bd.getBdId())
+        ids.push_back(*bdId);
+    });
+  };
+  for (auto mem : device.getOps<AIE::MemOp>())
+    collect(mem, mem->getRegion(0));
+  for (auto mem : device.getOps<AIE::MemTileDMAOp>())
+    collect(mem, mem->getRegion(0));
+  for (auto mem : device.getOps<AIE::ShimDMAOp>())
+    collect(mem, mem->getRegion(0));
```
andrej: With this repetition it feels like the method to get all "used up" BD IDs from a tile belongs on `DmaBody`, so the code can be reused across both sites (PR #3615 and this one)

andrej (approved):
Approved with a thought on code deduplication I think is worth considering but not absolutely necessary

lib/Conversion/AIEXToEmitC/AIEXToEmitC.cpp
hunhoffe: Good point! I'll wait until https://github.com/Xilinx/mlir-aie/pull/3615 lands and then I'll do another commit here to address duplication.

PR #3621, opened 2026-08-27
Support out-of-order S2MM receivers on the runtime configuration path

### LINKED ISSUE #3579: Remove `bd_id_val` from `aiex.dma_configure_task`
We recently added a BD ID attribute to the DMA task:
https://github.com/Xilinx/mlir-aie/blob/760932a4abf084bfc7abdec851703b27a03c01ba/include/aie/Dialect/AIEX/IR/AIEX.td#L1320

A DMA task is a sequence of buffer descriptors (BDs).
BD IDs are a property of BDs; `aie.dma_bd` already has a `bd_id` attribute.

I don't see any benefits to duplicating this attribute on the DMA task, and it creates ambiguity:

- What if the DMA contains a single BD, and that BD also has a `bd_id` value set? What if the values disagree?
- What if the DMA contains multiple BDs -- which contained BD is the `bd_id_val` supposed to apply to?

I think this was more of an "accidental" change that we should revert; if not, let us understand and document the nuance of why this is needed more clearly.

jskimko:
## Description

#3577 added out-of-order (OoO) S2MM DMA receive channels, configured on the static tile-DMA / dataflow path. This PR extends OoO receiver support to the runtime path: an S2MM channel can now be armed out-of-order from the runtime sequence via `aiex.dma_configure_task`, including when its BDs come from the runtime dynamic BD pool. It also relocates the runtime `bd_id` to `aie.dma_bd` (closes #3579) and adds cross-program `bd_id` coordination, both of which the runtime OoO path depends on but which stand on their own (see below).

### Out-of-order receivers from the runtime sequence

- New optional `out_of_order` unit attribute on `aiex.dma_configure_task` (S2MM only): puts the channel into out-of-order mode from the runtime sequence. `aie-dma-tasks-to-npu` emits the channel OoO-enable register write.
- Supported for BDs materialized from the runtime dynamic BD pool (`aie-lower-dynamic-bd-pool`).

```mlir
aie.device(npu2) {
  %tile = aie.tile(0, 0)
  aie.runtime_sequence(%arg0: memref<4xi32>) {
    %t = aiex.dma_configure_task(%tile, S2MM, 0, <pkt_type = 0, pkt_id = 0>) {
      aie.dma_bd(%arg0 : memref<4xi32> offset = 0 len = 4) {bd_id = 0 : i32}
      aie.end
    } {out_of_order}
    aiex.dma_start_task(%t)
  }
}
```

With the dynamic BD pool the receive `bd_id` comes from the pool instead of a static attribute:

```mlir
    %bd = aiex.dma_bd_pool_pop(0, 0) : i32
    aie.dma_bd(%arg0 : memref<1024xi32> offset = 0 len = 256) bd_id_val %bd : i32
```

The runtime `bd_id` uses the separate `bd_id_val` operand form for now. A later PR should migrate it (and others) to the `custom<DynamicScalar>` spelling `bd_id = <const|%v>` that `offset`/`len` already use, so a single slot carries both the constant and the SSA case.

### Coordinate per-tile bd_id allocation across static and runtime-sequence programs

Update: dropped in favor of #3615.

- ~~New AIEX pass `aie-reserve-runtime-bd-ids` records, per tile, the compile-time `bd_id`s that `aiex.dma_configure_task` pins in the runtime sequence, as an `aiex.reserved_bd_ids` attribute on each `aie.tile`. It runs before `aie-assign-bd-ids` in the input-with-addresses pipeline.~~
- ~~The static allocator (`aie-assign-bd-ids`) seeds its per-tile generator from that attribute, and the runtime-sequence allocator (`aie-assign-runtime-sequence-bd-ids`) seeds from the tile's static BDs. Auto-assignment on either side no longer reuses an id the other program pinned; two explicit pins to the same id stay a real, diagnosed conflict.~~

### Tests

- Device (NPU2 / Strix): `test/npu-xrt/dma_s2mm_ooo` gains a runtime receiver configuration, and a new `test/npu-xrt/dma_s2mm_ooo_shim_recv` exercises a shim OoO S2MM receiver with a runtime pool `bd_id` and companion-token completion.
- lit: `out_of_order` round-trip on `dma_configure_task`; the new verifier diagnostics; OoO on the dynamic BD path; `bd_id`/`bd_id_val` mutual exclusion; and the `aie-reserve-runtime-bd-ids` reservation, both allocators seeding, and explicit-conflict diagnosis.

## Checklist


include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -1067,7 +1066,9 @@ def AIE_DMABDOp: AIE_Op<"dma_bd", [AttrSizedOperandSegments]> {
       OptionalAttr<DenseI64ArrayAttr>:$static_sizes,
       OptionalAttr<DenseI64ArrayAttr>:$static_strides,
       OptionalAttr<BDPadLayoutArrayAttr>:$pad_dimensions,
-      OptionalAttr<AIEI32Attr>:$bd_id, OptionalAttr<PacketInfoAttr>:$packet,
+      // BD id. Can be given as runtime (_val) or static; mutually exclusive.
```
hunhoffe, line 1069: Thank you for this cleanup!

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
```diff
@@ -920,6 +931,50 @@ struct AIEDMATasksToNPUPass
...
+  LogicalResult emitOutOfOrderChannelEnable(OpBuilder &builder,
+                                            DMAConfigureTaskOp op,
+                                            AIE::TileOp tile) {
+    const AIE::AIETargetModel &tm = AIE::getTargetModel(op);
+    int col = tile.getCol();
+    int row = tile.getRow();
+    int channel = op.getChannel();
+
+    uint32_t ctrlAddrLocal =
+        tm.getDmaControlAddress(col, row, channel, AIE::DMAChannelDir::S2MM) &
```
hunhoffe: should be bitwise arithmetic get folded into a helper function or something? Might be nice to codify it so magic `&` operations don't appear in multiple places.

test/bd-chains-and-dma-tasks/assign-runtime-sequence-bd-ids/reserve_static_bd_ids.mlir
```diff
@@ -0,0 +1,32 @@
+//===- reserve_static_bd_ids.mlir ------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// The runtime allocator must not reuse a bd_id the static tile program already
```
hunhoffe: I think this might actually overlap with https://github.com/Xilinx/mlir-aie/pull/3616 -- I don't feel strongly about which implementation is accepted, as long as it gets fixed!

test/npu-xrt/dma_s2mm_ooo/dma_s2mm_ooo.py
```diff
@@ -64,57 +79,32 @@
...
-MAX_REPEAT_FIELD = 0xFF
-
-# A multi-round sender walks every round from one BD whose iteration size is
-# max(ms)*rounds, which cannot exceed the BD iteration wrap of 64 (aie-rt
-# IterWrapMax + 1).
-MAX_BD_ITER = 64
+OOO_ID_MASK = 0x3F  # out-of-order id header field is 6-bit
+MAX_LOCK_VALUE = 0x3F  # AIE lock value is 6-bit
+MAX_REPEAT_FIELD = 0xFF  # DMA start-queue repeat field is 8-bit
+MAX_BD_ITER = 64  # BD iteration wrap (aie-rt IterWrapMax + 1)
```
hunhoffe, line 85: is this something that can be exposed/fetched from the python wrappers around target model?

test/npu-xrt/dma_s2mm_ooo/README.md
```diff
@@ -50,6 +50,7 @@ wrong slot.
 | Flag | Values | Meaning |
```
hunhoffe, line 50: I don't love readme's being stuck in the test directories as I feel they often get lost/outdated. Is this a utility that should be exposed and documented? If not, this is fine too.

test/npu-xrt/dma_s2mm_ooo_shim_recv/README.md
```diff
@@ -0,0 +1,54 @@
+<!---//===- README.md ---------------------------------------------*- Markdown -*-===//
+//
```
hunhoffe, line 2: Same comment re: readmes for tests.

hunhoffe (approved):
A few notes, but nice work and thanks again for the cleanup!

test/bd-chains-and-dma-tasks/assign-runtime-sequence-bd-ids/reserve_static_bd_ids.mlir
jskimko: Yeah, I saw there's some overlap with #3615. I think #3616 covers a different case.

#3615 provides runtime-reserves-static, but not static-reserves-runtime. I can drop this logic from my PR (doesn't need it) and I'll send you a patch file for the reverse edge for #3615.

test/npu-xrt/dma_s2mm_ooo/README.md
jskimko: I guess it depends on where we want to draw the line between `test/npu-xrt` and `programming_examples`. This directory is a demonstration of how ooo might be used *in practice*, but not quite a full application and not restricted to unit tests (included elsewhere).

lib/Dialect/AIEX/Transforms/AIEDMATasksToNPU.cpp
jskimko: Good call. Added `tm.getLocalDmaControlAddress`.

test/npu-xrt/dma_s2mm_ooo/dma_s2mm_ooo.py
jskimko: Not all of these are exposed, I think, so I just left them hardcoded for testing purposes.

test/npu-xrt/dma_s2mm_ooo/README.md
hunhoffe: My opinion, but I accept there can be other opinions: If it's a demonstration you intend people to read, it belongs somewhere else. Tests is pretty exclusively for quality control. If you are staging this here until a later PR when this becomes a more proper programming example, I guess that's ok, but I'd rather start it out as an experimental programming example from the beginning.

PR #3622, opened 2026-08-27
[AIEExpandLoadPdi] Keep the empty-PDI reset alternating across dispatches

erwei-xilinx:
`AIEExpandLoadPdi` replaces each `load_pdi @device` with a load of an EMPTY
device PDI, which makes the firmware reset the array, followed by explicit
write32/blockwrite configuration.

The firmware caches the PDI address, so loading the same PDI twice in a row is
a no-op. The pass already knows this — it is why the ctrl-pkt path keeps both
`ctrl_pkt_overlay` and `ctrl_pkt_overlay_copy` — and handles it by picking
`empty_ + (index % 2)`.

But `index` is the module-wide `load_pdi` index, so the alternation only holds
*within* a sequence. The host re-runs the whole runtime sequence on every
dispatch, so it has to hold across that boundary too. A sequence with an **odd**
number of resets ends on the same empty PDI it started with, so the next
dispatch's reset is swallowed and the configuration that follows lands on a
device still holding the previous run's lock and DMA state.

This appends one reset of the opposite parity to any sequence that would
otherwise end where it began.

## Impact

A single-launch ELF has exactly one reset, so it is always broken from the
second dispatch on. Multi-launch ELFs have several and mostly end on the other
parity, which is why this went unnoticed: in the mlir-air SmolVLA vision
encoder the three fused multi-launch ELFs were bit-identical across repeat
dispatches while the single-launch connector GEMM returned cosine 0.9999 on its
first call and ~0.61 on every call after it, in the same process.

Failure signature: non-deterministic between dispatches, uniform across all
32 C tiles, correct magnitude, and not explained by any missing or duplicated
data — a synchronisation race, not mis-programmed BDs. `xclbin` repeats
correctly on the identical design; only the ELF path does not.

## Testing

Parity was predicted before it was measured, 4/4:

| launches | resets | predicted | observed |
|---|---|---|---|
| 1 | 1 | BREAK | BREAK |
| 2 | 2 | OK    | OK    |
| 3 | 3 | BREAK | BREAK |
| 4 | 4 | OK    | OK    |

After the fix every parity and every shape in the repro suite repeats
correctly, and the mlir-air consumer is uniform across images
(0.999753 / 0.999752 / 0.999763, was 0.999753 / 0.9939 / 0.9939).

New lit test `test/Passes/expand-load-pdi/odd_reset_count_parity.mlir` covers
both directions: one reset gets a trailing `empty_1`, two resets get nothing
appended.

Perf is unchanged — the empty PDI is a few hundred bytes (6676.8 vs 6837.7 us
on a SmolVLA ViT layer, inside noise).

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIEExpandLoadPdi.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEExpandLoadPdi.cpp
@@ -28,6 +28,7 @@
 #include "mlir/Dialect/MemRef/IR/MemRef.h"
 #include "mlir/Pass/Pass.h"
 
+#include "llvm/ADT/MapVector.h"
 #include "llvm/ADT/StringRef.h"
 #include "llvm/Support/raw_ostream.h"
 
@@ -79,6 +80,27 @@ static AIE::DeviceOp getOrCreateCtrlPktOverlayCopy(ModuleOp moduleOp,
   return clonedDev;
 }
 
+// The empty device whose PDI load makes the firmware reset the array. `parity`
+// alternates so that two consecutive loads never name the same PDI (see
+// transformLoadPdi).
+static AIE::DeviceOp getOrCreateEmptyDevice(ModuleOp moduleOp,
+                                            AIE::AIEDevice deviceType,
+                                            unsigned parity) {
+  std::string emptyName = "empty_" + std::to_string(parity);
+  if (auto existing = moduleOp.lookupSymbol<AIE::DeviceOp>(emptyName))
+    return existing;
+
+  OpBuilder builder(moduleOp.getContext());
+  builder.setInsertionPointToStart(moduleOp.getBody());
+  auto loc = builder.getUnknownLoc();
+  auto emptyDevice = AIE::DeviceOp::create(builder, loc, deviceType,
+                                           builder.getStringAttr(emptyName));
+  emptyDevice.getRegion().emplaceBlock();
+  builder.setInsertionPointToEnd(&emptyDevice.getRegion().front());
+  AIE::EndOp::create(builder, loc);
+  return emptyDevice;
+}
+
 // Helper to transform a single load_pdi operation
 static LogicalResult transformLoadPdi(NpuLoadPdiOp loadPdiOp, ModuleOp moduleOp,
                                       unsigned index,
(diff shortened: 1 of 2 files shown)


erwei-xilinx:
Cc: @andrej

erwei-xilinx:
Independent corroboration from a second consumer: amd/Triton-XDNA#99, filed against the Triton-XDNA `NPUChain` multi-launch path on Strix Halo (npu2/aie2p, XRT 2.21.75, firmware 1.1.2.65).

The report was "a chain of one op returns correct data on its first dispatch and garbage after that; two ops are fine". That is this bug, and the parity model in the PR description predicts the rest of it — extending the reproducer to 3 and 4 ops:

| ops | resets | predicted | observed |
|---|---|---|---|
| 1 | 1 | BREAK | BREAK |
| 2 | 2 | OK | OK |
| 3 | 3 | BREAK | BREAK |
| 4 | 4 | OK | OK |

The reproducer is a chain of `C = A + B` over 98304 f32 with `BLOCK_SIZE = 1024`, `B` all zeros, so every dispatch should return `A` unchanged. Failure signature on an odd-length chain, from dispatch 1 on:

- 4000–6000 of 98304 elements wrong, scattered over ~60 of the 96 blocks, varying run to run — non-deterministic, consistent with the array retaining the previous run's lock/DMA state.
- The wrong elements are always at the *start* of a block and always satisfy `got[i] == want[i + 1024]` — a one-chunk pipeline phase skew, exactly one grid iteration.
- Independent of buffer objects: reproduces with a fresh `bo_key` and fresh host arrays on every dispatch.
- Uniform test data hides it completely — the operands have to be non-constant for the skew to be observable at all.

The aircc artifacts make the parity visible without running anything: the 1-op build emits only `bif_empty_0.bif`, the 2-op build emits `bif_empty_0.bif` and `bif_empty_1.bif`.

Rebuilding the same reproducer against this branch, the 1-op and 3-op chains are correct on all three dispatches, matching the "every parity repeats correctly" result in the PR description.

One data point on how quiet this is in practice: the affected project's example suite never caught it because its only multi-launch example is a four-op fused MLP.

andrej:
Ugh -- I'm sorry you had to waste time figuring this out. This was definitely on my radar, see https://github.com/amd/IRON/issues/143

Thanks for tackling this and taking this off my mental list! Good fix!

andrej (approved):

PR #3541, opened 2026-08-11
ci: link aarch64 distro wheel binaries with lld to fix R_AARCH64_CALL26 overflow

hunhoffe:
## Description

The `ubuntu-22.04 aarch64` legs of the MLIR Distro wheel build (mlirDistro.yml) have been consistently failing on `main` for at least the last couple days — masked from view because `upload_distro_wheels`/`smoke_test_wheels` are skipped on the `schedule`-triggered runs that normally exercise this workflow, and the `build` job itself is `continue-on-error: true`.           
                                                                                                                                     
The actual failure is a link error,:
```                                
  /usr/lib/gcc-cross/aarch64-linux-gnu/11/libstdc++.a(locale_init.o): in function __gnu_cxx::__throw_concurrence_lock_error()': 
  relocation truncated to fit: R_AARCH64_CALL26 against symbol __cxa_allocate_exception'
  collect2: error: ld returned 1 exit status                       
```

The aarch64 leg cross-links `mlir-opt`/`mlir-translate` as fully static binaries (`-static-libgcc -static-libstdc++`, `BUILD_SHARED_LIBS=OFF`). The resulting binaries are large enough that GNU `ld` fails to insert AArch64 long-branch veneers, so even calls *within* `libstdc++.a` (e.g. `__cxa_throw`) overflow the ±128MB range of the `R_AARCH64_CALL26` relocation.

Fix: link with `lld` instead, via `CMAKE_EXE_LINKER_FLAGS_INIT`/`CMAKE_MODULE_LINKER_FLAGS_INIT`/`CMAKE_SHARED_LINKER_FLAGS_INIT=-fuse-ld=lld` in `utils/mlir_wheels/setup.py`, and install the `lld` apt package alongside the existing aarch64 cross-toolchain in `setup_base`. This is the same workaround already used elsewhere in this repo's CMake configs (buildAndTestPythons.yml, lintAndFormat.yml, buildAndTestVitis.yml, buildAndTestRyzenAISw.yml) for large LLVM/MLIR link units.
                                                                   
Note: `mlirDistro.yml` only runs on `pull_request` when the PR touches that exact file, so this PR won't self-validate on the aarch64 legs unless a no-op change to `.github/workflows/mlirDistro.yml` is included, or the workflow is triggered manually via `workflow_dispatch`.   

## Checklist

- [ ] Tests pass locally, or the change is covered by CI
- [ ] Docs / docstrings updated if the change is user-facing
[...]

PR #3623, opened 2026-08-28
ci: build mlirDistro pull_request against the pinned LLVM commit

hunhoffe:
## Description

get_llvm_project_commit always live-fetched ROCm/llvm-project@amd-staging HEAD unless an explicit LLVM_COMMIT input was given — including on pull_request runs (which only fire when mlirDistro.yml itself changes). That built against an unvalidated, constantly-moving target instead of the commit already recorded in utils/clone-llvm.sh, unlike every other pinned dependency in this repo.

- schedule still live-fetches HEAD (unchanged) — it intentionally pre-builds wheels ahead of update-llvm.yml's bump search.
- pull_request / manual dispatch with no explicit LLVM_COMMIT now build the pin recorded in utils/clone-llvm.sh instead.
- Explicit LLVM_COMMIT input (used by update-llvm.yml to validate a bump candidate) is unchanged.

Actually building the pin on pull_request for the first time immediately exposed that the aarch64 leg had several real, pre-existing problems that had been silently masked by always building a fresh, different HEAD commit:

1. Stale LLVM pin. The recorded commit (56bcc187, 2026-08-01) predates upstream commit 72152fc6e2 (2026-08-26, llvm/llvm-project#218295), which added the G_VECREDUCE_FMAXIMUMNUM GlobalISel generic opcode. AArch64 backend tablegen at the old pin referenced that opcode without it being defined, so llvm-tblgen died with Could not find 'G_VECREDUCE_FMAXIMUMNUM' instruction! while parsing AArch64.td — aarch64-only, since the x86_64 leg doesn't build AArch64's full instruction/pattern set. Bumped the pin to current amd-staging HEAD (57a3bb95, 2026-08-29), which includes the fix.
2. Cross-compiler couldn't find its own linker. Once the LLVM pin was fixed, the build died with collect2: fatal error: cannot find 'ld'. Root cause: for a cross gcc, collect2's -fuse-ld=lld search doesn't include /usr/bin (where apt installs ld.lld) the way it does for a native compiler. Fixed by passing -B/usr/bin so collect2 can find it. (Two earlier hypotheses — a missing unversioned ld.lld alias, and a ccache masquerade double-wrap of the compiler path — turned out to be wrong or incomplete; the ccache fix is kept since it removes a real accidental self-referential wrap, but neither was the actual cause.)
3. Unpinned build-system deps. The aarch64 leg is the only one not routed through cibuildwheel, so it was the only one where pip wheel . resolved [build-system].requires fresh from PyPI with build isolation on. A same-day nanobind 3.0.x release broke find_package(nanobind 2.9) with no change on our side. Fixed by installing the hash-pinned requirements.lock plus mlir-native-tools up front and building with --no-build-isolation, matching what cibuildwheel's before-build already does for the other legs. Also fixed pyproject.toml's [build-system].requires, which had drifted from its requirements.in mirror (unpinned ninja/numpy/nanobind, a stray dataclasses entry); requirements.in is now generated from pyproject.toml by regenerate_lockfiles.sh instead of hand-maintained, so the two can no longer diverge.

## Checklist

### DIFF
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -81,30 +81,6 @@ runs:
         sudo apt-get install -y binutils-aarch64-linux-gnu \
           g++-aarch64-linux-gnu gcc-aarch64-linux-gnu openssl libssl-dev lld
 
-        # The `lld` metapackage here only pulls in versioned binaries
-        # (ld.lld-18, lld-18, ...) under /usr/bin; the unversioned names live
-        # under /usr/lib/llvm-*/bin, which isn't on PATH. mlir_wheels/setup.py
-        # passes plain `-fuse-ld=lld` to aarch64-linux-gnu-gcc (needed: GNU ld
-        # can't insert AArch64 long-branch veneers for the fully static
-        # mlir-opt/mlir-translate binaries), which fails to resolve without
-        # an unversioned ld.lld and dies with the misleadingly generic
-        # "collect2: fatal error: cannot find 'ld'". Register one explicitly;
-        # idempotent, so this is a no-op if a future image already provides it.
-        LLD_BIN=$(command -v ld.lld || true)
-        if [ -z "$LLD_BIN" ]; then
-          LLD_BIN=$(ls /usr/bin/ld.lld-* 2>/dev/null | sort -V | tail -1)
-        fi
-        if [ -z "$LLD_BIN" ]; then
-          echo "::error::no ld.lld binary found after installing lld"
-          exit 1
-        fi
-        sudo update-alternatives --install /usr/bin/ld.lld ld.lld "$LLD_BIN" 100
-
-        # Fail fast (seconds, not 20 minutes into the LLVM build) if
-        # aarch64-linux-gnu-gcc still can't actually link with it.
-        echo 'int main(void) { return 0; }' | \
-          aarch64-linux-gnu-gcc -fuse-ld=lld -x c - -o /tmp/ld_lld_smoke_test
-
     - name: pip install standard tools
       shell: bash
       run: pip install --require-hashes -r python/requirements_dev.lock
--- a/.github/actions/setup_base/action.yml
+++ b/.github/actions/setup_base/action.yml
@@ -81,6 +81,30 @@ runs:
         sudo apt-get install -y binutils-aarch64-linux-gnu \
           g++-aarch64-linux-gnu gcc-aarch64-linux-gnu openssl libssl-dev lld
 
+        # The `lld` metapackage here only pulls in versioned binaries
+        # (ld.lld-18, lld-18, ...) under /usr/bin; the unversioned names live
+        # under /usr/lib/llvm-*/bin, which isn't on PATH. mlir_wheels/setup.py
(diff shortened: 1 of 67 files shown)


hunhoffe:
I'm trying to unblock https://github.com/Xilinx/mlir-aie/pull/3562

atassis:
Reproduces off CI in about a minute, which may help the iteration loop:

```
docker run --rm ubuntu:24.04 bash -c '
  apt-get update -qq && apt-get install -y -qq gcc-aarch64-linux-gnu lld
  echo "int main(){return 0;}" > t.c && aarch64-linux-gnu-gcc -c t.c -o t.o
  aarch64-linux-gnu-gcc -fuse-ld=lld t.o -o t'
# collect2: fatal error: cannot find 'ld'
```

Mechanism, measured on gcc 13.3.0 / lld 18.1.3: a cross gcc's COMPILER_PATH is only the gcc-cross exec dirs plus /usr/aarch64-linux-gnu/bin. collect2 looks for the bare `ld.lld` there, and for `aarch64-linux-gnu-ld.lld` on PATH, so /usr/bin/ld.lld is reachable from neither. Patching either half alone is enough: `COMPILER_PATH=/usr/bin`, or an `aarch64-linux-gnu-ld.lld` symlink on PATH. Native gcc links fine against the same lld. `-B/usr/bin` works too and yields an aarch64 PIE with `Linker: Ubuntu LLD 18.1.3`; it also
puts `-L/usr/bin` on the link line.

And it did work. In run 33200707296 the aarch64 leg reaches `Detecting C compiler ABI info - done` and stops on the next wall instead: `find_package(nanobind 2.9)` against nanobind 3.0.1.

That one is not caused by the pin. `find_package(nanobind 2.9 CONFIG REQUIRED)` is identical at 56bcc187 and at amd-staging HEAD. The aarch64 leg is the only one that runs `pip wheel .` with build isolation, so it resolves `[build-system].requires` fresh from PyPI where nanobind is unpinned; the cibuildwheel legs get `nanobind==2.12.0` out of requirements.lock. nanobind 3.0.0 shipped 2026-08-22 and 3.0.1 on 08-27, so the scheduled builds would hit this too the moment the linker is fixed. requirements.in has also drifted from pyproject.toml on ninja, numpy and dataclasses, despite calling itself a mirror of it.

One more thing while you are in here: this is not only a PR-trigger problem. The aarch64 legs have failed on every scheduled run since #3541 merged (2026-08-12T15:54Z). Before that they were intermittent, from the CALL26 overflow that PR fixed; the last green pair, 08-12T08:33, has no -fuse-ld=lld in its cmake line. It stays invisible because `build` is `continue-on-error: true` (mlirDistro.yml:129) while `smoke_test_wheels` and `upload_distro_wheels` are `if: github.event_name != 'schedule'`, so on the 4-hourly runs the only failing job is the tolerated one and `gh run list` shows 36 of 36 green. On a pull_request the upload job does run, finds nothing to upload for aarch64, and that is the red X on #3562. The newest aarch64 wheel on the mlir-distro release is 2026-08-01; the 08-15 publish is x86_64 only.

andrej (approved):

PR #3629, opened 2026-08-29
[AIEX] Support the iteration attribute in dma_configure_task [MED]

atassis:
## Description

#3538 added `#aie.bd_iteration<size, stride, current>` to `aie.dma_bd` but rejects it
outright when the same op sits inside `aiex.dma_configure_task` (the runtime-sequence
DMA-task path), directing the author to "express iteration via the outermost
sizes/strides dimension instead". That workaround is not equivalent to what it
replaces: the outermost-dimension route (`AIEDMATasksToNPU.cpp`) hardcodes
`iteration_current = 0` unconditionally, so the register's resume-index field cannot be
set on this path at all — not degraded, absent — while the message reads as if the two
spellings were interchangeable.

This honors the attribute for a compile-time-constant BD with a static `bd_id`
(`rewriteSingleBD`), feeding `(size, stride)` through the same `getHardwareStridesWraps`
encoder the outermost-dimension path already uses so the off-by-one and word-scaling
math cannot drift between the two, and passing `current` through unbiased (matching the
structural lowering in `AIERT.cpp`). `verifyTaskBDDimensions` now costs the attribute one
ND dimension rather than rejecting it, and rejects precisely instead of blanket: a
runtime-valued BD at verify time (bd-local, so the message is immediate), and a BD given
a dynamic free-list-pool `bd_id` at lowering time (only visible once dma-tasks-to-npu
resolves it).

Four FileCheck lit tests cover it: `good-iteration-attribute.mlir` (lowers to the exact
`npu.writebd` iteration fields — size 4/stride 16/current 2 → raw 3/15/2, alone and
combined with a real 3-dimensional access pattern), plus
`dma_task_iteration_rejected.mlir`, `bad-iteration-dynamic-bdid.mlir` and
`bad-iteration-too-many-dims.mlir` for the three cases that still reject, each asserting
the exact message. 285 pre-existing lit tests under `test/dialect/` and
`test/bd-chains-and-dma-tasks/` still pass (excluding peano/XRT-gated ones, not built
here).

One pre-existing test changes. `test/Conversion/DmaToNpu/dma_to_npu_core_tile.mlir`
(added in #2824) packs `iteration_current = 42` against `iteration_size = 28`. Those
fields are encoded actual-1, so current must not exceed size, and the new bound now
rejects it — the values were arbitrary bit-packing fodder, never a legal BD. Lowered to
`iteration_current = 21` and the expected word updated accordingly
(`(21<<19)|(28<<13)|499 = 11239923`). It is the only writebd in the tree that violated
the rule; the rest of the sweep is unchanged.

## Checklist

- [ ] `ruff check` and `pyright` pass locally for any touched Python file in their covered paths (no Python touched)
- [ ] `clang-tidy` passes locally for any touched file on the enabled list (file not on the enabled list)

### DIFF
--- a/test/npu-xrt/bd_iteration_dma_task/test.cpp
+++ b/test/npu-xrt/bd_iteration_dma_task/test.cpp
@@ -1,0 +1,84 @@
+//===- test.cpp -------------------------------------------------*- C++ -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#include <cstdint>
+#include <iostream>
+#include <vector>
+
+#include "cxxopts.hpp"
+#include "test_utils.h"
+#include "xrt/xrt_bo.h"
+#include "xrt/xrt_device.h"
+#include "xrt/xrt_kernel.h"
+
+// Must match the #aie.bd_iteration attribute in aie.mlir.
+constexpr int SIZE = 4;     // iteration_size (slots)
+constexpr int START = 2;    // iteration_current
+constexpr int CHUNK = 1024; // elements per execution
+constexpr int N = SIZE * CHUNK;
+
+int main(int argc, const char *argv[]) {
+  cxxopts::Options options("bd_iteration_dma_task");
+  test_utils::add_default_options(options);
+  cxxopts::ParseResult vm;
+  test_utils::parse_options(argc, argv, options, vm);
+
+  std::vector<uint32_t> instr_v =
+      test_utils::load_instr_binary(vm["instr"].as<std::string>());
+
+  auto device = xrt::device(0);
+  auto xclbin = xrt::xclbin(vm["xclbin"].as<std::string>());
+  std::string kernelName = vm["kernel"].as<std::string>();
+  device.register_xclbin(xclbin);
+  xrt::hw_context context(device, xclbin.get_uuid());
+  auto kernel = xrt::kernel(context, kernelName);
+
+  auto bo_instr = xrt::bo(device, instr_v.size() * sizeof(int),
+                          XCL_BO_FLAGS_CACHEABLE, kernel.group_id(1));
+  auto bo_in = xrt::bo(device, N * sizeof(int32_t), XRT_BO_FLAGS_HOST_ONLY,
+                       kernel.group_id(3));
+  auto bo_out = xrt::bo(device, N * sizeof(int32_t), XRT_BO_FLAGS_HOST_ONLY,
+                        kernel.group_id(4));
+
+  uint32_t *bufIn = bo_in.map<uint32_t *>();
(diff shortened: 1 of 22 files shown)


jskimko (changes requested):
https://github.com/atassis/mlir-aie/pull/1:
- Added runtime tests.
- Extract BD iteration bounds checking to a shared helper.

LGTM. Ran a device test on my end, which should be included.

I initially left BD iteration out of the runtime/dynamic path completely in https://github.com/Xilinx/mlir-aie/pull/3538, but I might add that after this gets merged.

atassis:
@jskimko nice, thanks!

jskimko (approved):
I noticed that we need to wire `iteration=` through the shim APIs as well, but we can split that off into a new PR. This one can just carry in the basic device verification.

jskimko:
Sorry, I'll fix that failure. And I suppose I'll close out the runtime gap while I'm at it. Only the dynamic path won't be covered by BdIteration after that; at least, that's the plan.

jskimko:
@atassis Please review & merge: https://github.com/atassis/mlir-aie/pull/2. This should cover the rest of the runtime API except for decomposing large BDs. Dynamic path is rejected in this PR.

test/npu-xrt/bd_iteration_dma_task/test.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "test_utils.h"
+#include "xrt/xrt_bo.h"
+#include "xrt/xrt_device.h"
+#include "xrt/xrt_kernel.h"
+
+// Must match the #aie.bd_iteration attribute in aie.mlir.
+constexpr int SIZE = 4;     // iteration_size (slots)
+constexpr int START = 2;    // iteration_current
+constexpr int CHUNK = 1024; // elements per execution
+constexpr int N = SIZE * CHUNK;
```
jskimko:
```suggestion
constexpr int ITER_SIZE = 4;     // iteration_size (slots)
constexpr int START = 2;    // iteration_current
constexpr int CHUNK = 1024; // elements per execution
constexpr int N = ITER_SIZE * CHUNK;
```

test/npu-xrt/bd_iteration_dma_task/test.cpp
```diff
@@ -0,0 +1,84 @@
...
+  if (run.wait() != ERT_CMD_STATE_COMPLETED) {
+    std::cout << "Kernel did not complete\n";
+    return 1;
+  }
+  bo_out.sync(XCL_BO_SYNC_BO_FROM_DEVICE);
+
+  uint32_t *bufOut = bo_out.map<uint32_t *>();
+  int errors = 0;
+  for (int k = 0; k < SIZE; k++) {
+    int slot = (START + k) % SIZE;
```
jskimko:
```suggestion
  for (int k = 0; k < ITER_SIZE; k++) {
    int slot = (START + k) % ITER_SIZE;
```

PR #3630, opened 2026-08-29
[AIEPlacer] Rank unanchored non-core tiles by DMA load [LOW]

atassis:
## Problem

`SequentialPlacer::computeCentroidColumn` derives a non-core `aie.logical_tile`'s target column from
its flow peers; since #3551 that includes an already-placed non-core peer, not only a CoreTile, so a
memtile placed after its shim peer correctly anchors to that shim's column. But the first LTO in an
otherwise anchor-free chain (no CoreTile anywhere in the design) still has no peer to resolve, and
falls back to column 0. `findTileWithCapacity` ranks candidates by `(distance, load, col, row)`, so
every such LTO ties at distance 0 from column 0 and the load tiebreaker only fires once column 0's
tiles are DMA-full.

The result: N independent DMA passthrough chains (shim -> memtile -> shim, no compute tile) do not
spread across N columns, they fill each column's DMA channels before touching the next one. On
`npu1`, 4 independent chains land on 2 columns, 2 chains sharing each memtile and shim, instead of
running 4-way in parallel.

## Fix

`computeCentroidColumn` returns `std::optional<int>`, `nullopt` when there is truly no anchor (was: a
fabricated 0). `findTileWithCapacity` takes `std::optional<int> targetCol`; without one, it drops the
distance term and ranks by DMA load alone, counted only in the direction the LTO needs: summing both
directions would put a chain's read and write shims on different columns.

Gated behind `spread-unanchored-tiles` (default false, matching the `merge-logical-tiles` precedent,
#3068): with it off, an unanchored LTO keeps the historical column-0 fallback, so the default path is
provably unchanged. CoreTile placement and anchored non-core placement are untouched either way.

## Test

New `test/place-tiles/sequential_placer/test_place_spread_unanchored_tiles.mlir`: 4 independent
passthrough chains with no compute tile, and a control case with one CoreTile peer showing the option
is a no-op once a real anchor exists. Checked with `--implicit-check-not='aie.tile('` rather than
paired `CHECK-DAG`/`CHECK-NOT`: a plain `CHECK-NOT` after a `CHECK-DAG` block only excludes input
past that block's furthest match, and the printer emits tiles in reverse-numeric order, so a tile
declared before the block's matches goes unchecked by it. Verified this concretely: swapping the
`CHECK-NOT` target for a tile that is genuinely present but printed earlier still passed.

Rebased onto `upstream/main` at `b8ef72c7dd5` (post-#3551, post-#3565, post-#3591) and re-measured the
default's actual output at this base, since #3551 changed it: the two-column default above is what
`aie-opt` prints today, not a stale pre-#3551 story. Built `aie-opt` standalone against the pinned
MLIR/LLVM distro (the checked-in toolchain instance predates this base and cannot resolve a branch's
new test file) and ran every `RUN:` line in all 28 `.mlir` files under `test/` that exercise
`aie-place-tiles`: 44/44 pass, both with the option off and on. `git-clang-format` against
`origin/main` is clean on the touched `.cpp`/`.h`, verified with the exact pinned 20.1.0 binary.


hunhoffe (commented):
I'm thinking for usability reasons, perhaps this should be on by default?

atassis:
@hunhoffe agreed, and it is on by default now.

I checked the blast radius before flipping rather than after. The option only changes anything when a non-core logical tile has no column constraint *and* no centroid, so it is bounded to designs with no core anchor at all. Running every test that invokes `aie-place-tiles` with the flag forced on, 29 of 31 produced byte-identical output; the two that moved were this PR's own test and `test_place_objectfifo.mlir`, and the latter still passes because the placement it changes falls outside the `CHECK-NOT` scope it sits next to.

With the default actually flipped and rebuilt, all 31 pass. The only edit needed was in this PR's test: its first RUN line asserted that the default was "pile", so it now asserts "spread" and the explicit `spread-unanchored-tiles=false` arm keeps covering the old behaviour.

One gap worth naming: I did not run `test/python/worker_buffer_sharing.py`, which also reaches the placer, because I built without Python bindings. CI will cover it.

The flag stays, so anything that regresses can be bisected with `spread-unanchored-tiles=false` rather than a revert.

hunhoffe (approved):

PR #3641, opened 2026-08-29
[python] JIT cache: include -I paths in the cache key [HIGH]

atassis:
## Problem

`include_paths` is accepted by `compileconfig()`, stored on `CompilableDesign`, documented as
*"Extra `-I` paths forwarded to the C++ compiler"*, and round-tripped through `to_dict`/`from_dict`.
It reaches neither `_compute_recipe_hash` nor `_compute_artifact_hash`.

So two designs that differ only in which header tree their C++ kernels compile against produce the
same cache key and share a cache entry. The second build returns the first one's artifact.

`_compute_artifact_hash`'s own docstring states the invariant this breaks:

> Captures everything that can change the *output* of compilation without changing the *recipe*:
> edited C++ kernels, swapped object files, upgraded Peano / aiecc, retargeted device.

Include paths change the output and were captured by neither hash.

## Fix

Thread `include_paths` into `_compute_recipe_hash`, alongside the flag lists it belongs with, and
through `_compute_hash` -- the combiner that `__hash__` actually calls via `_compute_cache_hash`.
Adding it only to the `recipe_hash` property leaves the real cache key untouched, which is worth
noting because that is the obvious half-fix.

Hashed in order rather than `sorted()` like the adjacent `aiecc_flags`/`compile_flags`: `-I` search
order decides which header wins when two directories provide the same name, so the order is part of
the meaning.

## Test

Two unit tests in `test/python/test_compilabledesign.py` (device-free, next to the existing
`test_hash_differs_for_different_compile_flags`): one for differing paths, one for differing order.

Both fail against the unmodified files and pass with this change; the file's other 97 tests are
unaffected. `black --check` clean on all three files.

## Scope

This closes the key gap only. The adjacent risk -- a header shadowed by an `-I` reorder changing what
a *fixed* path resolves to -- is content-level and already noted as a known limitation in
`_manifest.py`'s docstring; this change does not address that, and hashing in order rather than
sorted is what keeps the door open to it.

### DIFF
--- a/python/utils/compile/jit/_hash.py
+++ b/python/utils/compile/jit/_hash.py
@@ -115,6 +115,7 @@ def _compute_recipe_hash(
     aiecc_flags: list[str] | tuple[str, ...],
     compile_flags: list[str] | tuple[str, ...],
     full_elf: bool = False,
+    include_paths: list[Path] | tuple[Path, ...] = (),
 ) -> str:
     """Hash of the "recipe": generator bytecode + CompileTime[T] kwargs + flags.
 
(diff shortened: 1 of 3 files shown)


hunhoffe (approved):
Thank you for another fix!

Scope note, not a blocker: `include_paths` isn't actually forwarded to any compiler invocation anywhere in the codebase — `compile_external_kernel` takes no include-path arg from the design; real -I flags come only from each `ExternalFunction._include_dirs`. So this closes a collision that can't yet produce a wrong artifact in practice, since the field has no effect on compilation. Still the right fix for when it's wired up. Filed #3645 to track that. LGTM.

PR #3644, opened 2026-08-29
ci: give the Windows device job a pip cache that outlives the job [MED]

atassis:
## Problem

Every run of this job re-downloads about 1.7 GB of Python wheels. `PIP_CACHE_DIR` is keyed on
`github.run_id` and `Cleanup workspace` deletes it, so on a runner with a persistent disk the cache is
empty by construction.

Two downloads dominate and neither moves between runs: the `mlir` distro wheel the build step pulls
with `pip -q download` is 961 MB and turns over only on an LLVM re-pin, and `llvm_aie` is 395 MB and
turns over only on a Peano re-pin.

On a healthy run (`33237903755`) that is about 1.13 min inside `Setup IRON environment` and 1.88 min
inside the build step. It is also where the job's tail lives: on run `33213412902` the `llvm_aie` wheel
alone took 1100 s across two resumes at 205 to 532 kB/s, putting the job at 50.3 min against a 25.2 min
median (n=25 green runs).

## Fix

Point `PIP_CACHE_DIR` at `%RUNNER_WORKSPACE%\.pip-cache`, beside the checkout rather than under it, so
the next run's `git clean -ffdx` does not take it with the work tree. The nightly `schedule` run resets
it. That is what bounds growth: pip's HTTP cache is URL-keyed with no eviction, and the rolling dev
channel publishes one or two 172 MB `mlir_aie` wheels a day, so the ceiling is roughly a day of that on
top of the 1.7 GB working set.

## Evidence

Wheel sizes and transfer times are from the two runs named above. Index pages come back
`Cache-Control: max-age=0, private, must-revalidate` with an ETag, so pip has to revalidate and cannot
resolve a stale dev wheel out of cache; wheel bodies are immutable per URL. actionlint reports no new
findings against `main`.

I cannot run this job, so the saving is unmeasured. The byte counts and the two log timings are what I
can stand behind.

## Two questions

The per-run pip cache is also the pattern in `buildAndTestRyzenAI.yml` and `buildRyzenWheels.yml`. I
scoped this to the Windows job because it is the serialized one, but I would rather extend it than
leave the convention split, if you prefer that.

Is there room for about 2 GB on the runner? That is the one thing I cannot check from here.


hunhoffe:
I'll give this a closer look tomorrow, since Windows is our slowest CI piece, but I'll need to be careful that jobs that bump dependencies in the pip cache reset it both when they run and when they merge.

PR #3648, opened 2026-08-30
[python] Report why the NPU is unavailable instead of only the device name [MED]

atassis:
## Problem

On a machine with a working NPU, `iron.arange(1024, dtype=np.int32, device="npu")` -- the example at
the top of the README -- fails with:

```
ValueError: Unsupported device: npu
```

`_probe_xrt()` catches the `ImportError` from `import pyxrt`, discards it, and returns `False`;
`tensor_factory` then selects `CPUOnlyTensor`, whose `DEVICES = ["cpu"]`. By the time the rejection
is raised the only facts in scope are the string `"npu"` and a list `["cpu"]`, so the message cannot
mention XRT. It reads as a typo in the device string.

One bool is answering five questions that fail independently and have different fixes: is there NPU
hardware, is a driver bound, is XRT userspace installed, do its Python bindings import *in this
interpreter*, and is the toolchain present. The device-presence code that could distinguish them
(`/dev/accel/accel0`, `xrt-smi examine` in `xrtruntime/hostruntime.py`) sits behind the pyxrt import,
runs only after `pyxrt.device(0)` has already thrown, is `logger.debug`, and its central check is
Linux-only in a package that ships 272 `win_amd64` wheels.

## Fix

`aie/utils/probe.py` answers the five separately, each returning a reason and a remedy rather than a
bool. Cheap stages (platform, hardware, driver) read sysfs and run eagerly; `import pyxrt` stays
deferred, preserving what the old probe was right to protect. `Check.ok` is tri-state so a Linux-only
probe on Windows reports "not determined here" instead of "absent".

`_probe_xrt()` now delegates to it, so there is one implementation. The two rejection sites quote the
reason. `python -m aie.doctor` prints the report (`--json` for non-Python consumers).

## Test

Same call, same machine, before and after:

```
- ValueError: Unsupported device: npu
+ ValueError: Unsupported device: npu (CPUOnlyTensor is in use: pyxrt is not importable:
+ No module named 'pyxrt'. pyxrt ships with XRT, not on PyPI, and is built against one
+ Python version. Install/build XRT for 3.12.13, or run under the interpreter its pyxrt
+ was built for.)
```

`python -m aie.doctor` on the same host, which had XRT installed for a different interpreter:

```
[  ok] hardware: NPU Gorgon Point 1 (/dev/accel/accel0)
[  ok] driver: amdxdna bound, firmware 1.1.2.64
[  ok] runtime: XRT userspace found (/usr/bin/xrt-smi)
[FAIL] bindings: pyxrt is not importable: No module named 'pyxrt'
```

`device="cpu"` is unchanged; `device="gpu"` still reports plainly, without an NPU explanation.
black, ruff and pyright are clean on all four files.

## Scope

Hardware and driver identity come from sysfs on Linux, which needs no userspace runtime. Elsewhere
they fall back to parsing `xrt-smi examine`, so Windows is covered whenever XRT is installed. The
parser is exercised against real `xrt-smi examine` output; I have no Windows host, so that platform
is covered by construction rather than by test. `Check.ok` stays tri-state for the one remaining
case -- no sysfs and no xrt-smi -- where neither source can answer and reporting `False` would be a
wrong claim about the machine rather than a missing probe.


hunhoffe:
This is a very user-friendly capability, I really like this idea! It would be good to make sure this is documented in an appropriate place in the programming guide (maybe quick setup too?) and would be great to ensure it works for HSA and HRX runtimes as well.

atassis:
Per documentation - good point, I'll try to include documentation changes in similar PRs going forward.
I want the compiler to fail early, and to fail loudly. A lot of the debugging and investigation I've done recently could have been avoided if these problems had been caught at compile time.
So right now I'm thinking not just about how to collect evidence, but also about how to make the compiler as informative as possible. The idea is a set of rules applied to every kernel: common invariants plus checks specific to particular type instantiations.
Each step here should make the research loop faster - my earlier PRs on compile speed were aimed at the same goal.

hunhoffe (approved):

PR #3652, opened 2026-08-31
Support tracing with multiple `aiex.run`s

andrej:
The trace lowering appends a trace buffer to each `aie.runtime_sequence` that is traced.

This currently does not work together with `aiex.run`:

The trace lowering does not update the `aiex.run` call sites when it appends the trace buffer to the runtime sequence. These call sites are therefore one argument short.

You also cannot pass the trace buffer by hand at the `aiex.run` call site. The trace lowering emits an address patch that names its buffer by argument index. Inlining moves that patch into a caller with a different argument list, so the index then names the wrong buffer.

Two calls of the same design write to the same offset. The second call overwrites the data of the first.

This PR fixes that:

- `aiex.npu.address_patch` takes a new `buffer` operand, which references the runtime sequence argument that holds the buffer (or a `memref.subview` of one). The operand replaces the fixed `arg_idx` in the early stages of the lowering. A new pass `-aie-resolve-address-patch-buffers` replaces the `buffer` with a concrete `arg_idx` and offset, after `-aie-materialize-runtime-sequences` inlines the calls.
- The new pass `-aie-fuse-trace-buffers` enables tracing across multiple reconfigured designs. It gives each runtime sequence exactly one trace buffer. If that runtime sequence calls other runtime sequences, the pass splits the buffer and gives each call site a `memref.subview` of it. The pass records these slices as `aie.trace_slices`, so we can recover which portion of the trace buffer belongs to which design after execution.
- `parse_trace` takes a device name. Two sub-designs often occupy the same tiles, so a scan of the whole module merges their event assignments. `get_trace_slices` and `parse_trace_slices` split a shared buffer per design.

One supporting change: all lowering paths derive the address patch index by counting memref arguments. A scalar argument travels in the instruction stream and occupies no host buffer slot, so the block argument number over-counts it.


lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
```diff
@@ -136,16 +134,21 @@ struct AIEInsertTraceFlowsPass
...
     //     perturbs the indices of the data arguments.
     //   - Reuse-output: trace data is written into the tail of the last
     //     existing argument (an output buffer), saving a host buffer. No new
     //     argument is added; the offset skips past the output data.
-    int traceArgIdx;
-    int traceBufferOffset = 0; // in bytes
+    //
+    // The patch names the buffer by SSA value. An index holds only within this
+    // sequence. `aiex.run` may inline this sequence into a caller with a
+    // different argument list.
```
andrej: "An index holds only within this sequence."

I think this is confusing, unless I am missing something. After this refactor, we no longer are handling _any_ indices here, so talking about them is misleading.

"The patch" --> "npu.address_patch"?

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -269,6 +269,50 @@ def AIEMaterializeRuntimeSequences : Pass<"aie-materialize-runtime-sequences", "
...
 }
 
+def AIEFuseTraceBuffers : Pass<"aie-fuse-trace-buffers", "mlir::ModuleOp"> {
+  let summary = "Give every traced sequence a slice of one fused trace buffer";
+  let description = [{
+    Trace lowering appends a trace buffer to each traced `aie.runtime_sequence`.
+    It does not update the `aiex.run` call sites. These call sites are therefore
+    one argument short.
+
+    This pass enables tracing across designs that `aiex.configure` reconfigures.
```
andrej: Does `aiex.configure` actually matter here, or is the pass really only concerned with `aiex.run`? For example, if rather than re-configuring multiple designs, we just re-ran the same design (one aiex.configure, multiple aiex.runs) a couple of times, this would still help us separating out the trace buffers, right? 

Clarify.

lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
```diff
@@ -569,13 +559,44 @@ struct AIEInsertTraceFlowsPass
...
+    // claim.
+    int traceBytesClaimed = bufferSizeBytes;
+    for (auto &[col, shimInfo] : shimInfos)
+      for (auto &chanDesc : shimInfo.channels)
+        traceBytesClaimed = std::max(traceBytesClaimed, chanDesc.bufferOffset -
+                                                            traceBufferOffset +
+                                                            bufferSizeBytes);
+    if (appendedTraceArg)
+      appendedTraceArg.setType(MemRefType::get(
+          {traceBytesClaimed}, IntegerType::get(device.getContext(), 8)));
```
andrej: Format: use curly braces _everywhere_ in the changeset (ONLY in the changed lines, not the originals), even if the contained block is semantically only one line.

lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
```diff
@@ -569,13 +559,44 @@ struct AIEInsertTraceFlowsPass
...
+                     appendedTraceArg
+                         ? appendedTraceArg.getArgNumber()
+                         : runtimeSeq.getBody().getNumArguments() - 1)),
+             builder.getNamedAttr("offset",
+                                  builder.getI64IntegerAttr(traceBufferOffset)),
+             builder.getNamedAttr("size",
+                                  builder.getI64IntegerAttr(traceBytesClaimed)),
+             builder.getNamedAttr(
+                 "dedicated", builder.getBoolAttr(bool(appendedTraceArg)))}));
+
```
andrej: Can we make sure this is a properly documented and defined attribute in the tablegen definition for `aie.runtime_sequence`?

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
```diff
@@ -0,0 +1,232 @@
...
+  llvm::DenseMap<Operation *, SmallVector<TraceSlice>> slices;
+  llvm::DenseSet<Operation *> done;
+
+  void runOnOperation() override {
+    ModuleOp mod = getOperation();
+    for (auto device : mod.getOps<AIE::DeviceOp>())
+      for (auto seq : device.getOps<AIE::RuntimeSequenceOp>())
+        if (failed(fuse(seq)))
+          return signalPassFailure();
+  }
```
andrej: Would a RewritePattern be more idiomatic here?

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
```diff
@@ -0,0 +1,232 @@
...
+    if (!needsFusing) {
+      slices[seq.getOperation()] = ownSlices(seq, ownBuffer, /*base=*/0);
+      return success();
+    }
+
+    if (ownBuffer && !cast<BoolAttr>(ownBuffer.get("dedicated")).getValue())
+      return seq.emitOpError(
+          "trace.host_config reuse_output_buffer=true cannot be combined with "
+          "aiex.run calls into traced designs: their trace slices would have "
+          "to grow the output buffer. Use a separate trace buffer instead");
```
andrej: I think technically there's nothing stopping us from growing the output buffer. If easy to implement, either grow the output buffer, or adjust the wording here so it doesn't seem like growing the output buffer is a blocker -- just say it is not supported.

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
andrej, line 1: - Make aie.trace_buffer a proper attribute that is defined in tablegen rather than kTraceBufferAttr here. Add a test that ensures if we get input IR that manually sets one of its buffers to be the trace buffer, that this is respected (across passes, and in the trace parsing utils in Python as well -- they should pick up this attribute to know which buffer to read if they aren't already).
- For the slices, would it make sense for these to be proper SSA ops (also defined in tablegen, under the trace namespace)? At a minimum, I think the attribute type should also be defined in tablegen properly, because it is an interface: The Python parse tracing utils also uses it to know which subslices to read, so it should be well-defined in one spot, the tablegen.

lib/Dialect/AIEX/Transforms/AIEResolveAddressPatchBuffers.cpp
```diff
@@ -0,0 +1,84 @@
...
+    : xilinx::AIEX::impl::AIEResolveAddressPatchBuffersBase<
+          AIEResolveAddressPatchBuffersPass> {
+
+  void runOnOperation() override {
+    AIE::DeviceOp device = getOperation();
+    SmallVector<NpuAddressPatchOp> patches;
+    device.walk([&](NpuAddressPatchOp op) {
+      if (op.getBuffer())
+        patches.push_back(op);
+    });
```
andrej: This feels like it's not idiomatic. Rewriting/replacing an op seems very much like a standard RewritePattern.

python/utils/trace/parse.py
```diff
@@ -35,6 +36,78 @@
 
 NUM_EVENTS = 8  # number of events we can view per trace
 
+TRACE_SLICES_ATTR = "aie.trace_slices"
```
andrej: Should go away and be fetched from the MLIR Python binding once defined as a proper IR component in tablegen.

python/utils/trace/parse.py
```diff
@@ -35,6 +36,78 @@
...
+
+def _device_name(device_op):
+    sym_name = device_op.sym_name
+    return sym_name.value if sym_name is not None else None
+
+
+def get_trace_slices(mlir_module_str):
+    """Return the trace buffer's layout, as recorded on the dispatched sequence.
+
+    Sub-designs run through `aiex.configure` share one trace buffer, and
```
andrej: Again, is `aiex.configure` really the relevant part here or is it `aiex.run`?

python/utils/trace/parse.py
```diff
@@ -35,6 +36,78 @@
...
+    `-aie-fuse-trace-buffers` records which byte range belongs to which design.
+    Returns a list of dicts with ``device``, ``sequence``, ``offset`` and
+    ``size``, in buffer order, or ``[]`` for a single-design buffer.
+    """
+    RuntimeSequenceOp = getattr(aiedialect, "RuntimeSequenceOp")
```
andrej: Why in this weird way? Why not aiedialect.RuntimeSequenceOp?

python/utils/trace/parse.py
```diff
@@ -35,6 +36,78 @@
...
+                {
+                    "device": entry["device"].value if "device" in entry else None,
+                    "sequence": (
+                        entry["sequence"].value if "sequence" in entry else None
+                    ),
+                    "offset": entry["offset"].value,
+                    "size": entry["size"].value,
+                }
+                for entry in attrs[TRACE_SLICES_ATTR]
+            ]
```
andrej: We're returning on the first encountered runtime sequence? Why? If we instead need another input that specifies which runtime sequence we are parsing a trace for, add that -- and default it to "main:sequence", which is the default runtime sequence name.

lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
andrej: Claude: Correct, the index is gone from that spot. The comment now says why the operand form exists, not what it replaces:

```
// `aiex.npu.address_patch` names the buffer by SSA value. `aiex.run` may
// inline this sequence into a caller with a different argument list.
```

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
andrej: Claude: `aiex.run` is the relevant part. `aiex.configure` plays no role in the split. Two runs of one design get two regions whether or not a reconfiguration sits between them, because the pass walks the `aiex.run` call graph and slices per call site.

The pass description now says so, and so does the `get_trace_slices` docstring. The added test `test_fuse_trace_buffers.mlir` covers one `aiex.configure` with two `aiex.run`s of the same design; they land at offsets 0 and 8192.

lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
andrej: Claude: Done. Braces on every changed block in `AIEInsertTraceFlows.cpp`, `AIEFuseTraceBuffers.cpp` and `AIEResolveAddressPatchBuffers.cpp`.

lib/Dialect/AIE/Transforms/AIEInsertTraceFlows.cpp
andrej: Claude: Done. `TraceBufferAttr` is an `AttrDef` in `AIEAttrs.td` with a description, and `aie.runtime_sequence` declares it as an inherent optional attribute:

```tablegen
OptionalAttr<TraceBufferAttr>:$trace_buffer,
OptionalAttr<TraceSliceArrayAttr>:$trace_slices
```

It prints as `#aie.trace_buffer<arg_index = 1, offset = 0, size = 6144, dedicated = true>`. The passes use the generated `getTraceBufferAttr` / `setTraceBufferAttr` accessors.

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
andrej: Claude: I kept the plain walk here, and I think it is the right shape.

`fuse()` recurses into callees before it sizes a caller, memoizes each claim, and assigns offsets in call order. The pattern rewriter drives a worklist in an order it chooses, so expressing "callees first, then this caller, with a running cursor" as a pattern means carrying the same recursion and memo tables inside `matchAndRewrite` and relying on the driver not to revisit. The traversal is the algorithm, not an implementation detail.

`-aie-resolve-address-patch-buffers` is different and I did convert it; see the reply on that file.

Happy to switch this one too if you would rather have it uniform.

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
andrej: Claude: Reworded. You are right that nothing prevents it; the blocker is downstream, not here. Growing the output buffer changes its host ABI, and the host sizes that buffer from the operator signature.

Both messages now say it is unsupported:

```
runs traced designs and sets trace.host_config reuse_output_buffer=true.
Growing the output buffer to hold their slices is not supported. Use a
separate trace buffer instead
```

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
andrej: Claude: Both done.

**Attributes in tablegen.** `TraceBufferAttr` and `TraceSliceAttr` are `AttrDef`s in `AIEAttrs.td`, declared as inherent optional attributes of `aie.runtime_sequence`. The string constants are gone from both the C++ and the Python.

**Test for a hand-written attribute.** `test/python/trace_slices.py` lowers a module that sets one by hand and carries no `aie.trace` op at all:

```mlir
aie.runtime_sequence @sequence(%arg0: memref<64xi32>, %arg1: memref<777xi8>)
    attributes {trace_buffer = #aie.trace_buffer<arg_index = 1, offset = 16,
                                                 size = 761, dedicated = true>} {
```

It survives the whole pipeline and `get_trace_buffer` returns it unchanged.

**SSA ops for slices.** I left them as attributes. A slice records where bytes landed after the fact; it has no operands, no result anyone uses, and no position in the instruction stream. An op would need a home in the sequence body and would have to survive inlining, which is what the attribute avoids.

**Python reads them through the bindings.** I added CAPI accessors and nanobind wrappers rather than parsing text, so the field names live only in the tablegen.

lib/Dialect/AIEX/Transforms/AIEResolveAddressPatchBuffers.cpp
andrej: Claude: Agreed, converted. `ResolveBufferOperand : OpRewritePattern<NpuAddressPatchOp>` with `walkAndApplyPatterns`, using `notifyMatchFailure` for the cases it cannot handle.

One addition: a pattern that declines to match leaves the operand in place, and the targets cannot emit that. The pass therefore walks once afterwards and reports any patch that still names a buffer.

python/utils/trace/parse.py
andrej: Claude: Gone. The attribute is inherent on `aie.runtime_sequence`, so the generated op class exposes it and the name comes from the bindings:

```python
attr = _find_sequence(module, kernel).trace_slices
device, sequence, offset, size = _aie.trace_slice_fields(slice_attr)
```

`trace_slice_fields` and `trace_buffer_fields` are new nanobind wrappers over CAPI accessors, so the field names live only in the tablegen too.

python/utils/trace/parse.py
andrej: Claude: `aiex.run`, same as the pass. The docstring now reads:

> `-aie-fuse-trace-buffers` records one `#aie.trace_slice` per `aiex.run` call site, so a sequence that runs several designs, or one design several times, keeps their traces in separate byte ranges.

python/utils/trace/parse.py
andrej: Claude: No good reason. It was `getattr` because pyright cannot see attributes on the compiled bindings module, but `find_ops` needs the class at runtime and the direct attribute works. It is now `aiedialect.RuntimeSequenceOp` with a `pyright: ignore` on the call.

python/utils/trace/parse.py
andrej: Claude: Fixed. Taking the first was wrong: a module holds one runtime sequence per entry point and the host dispatches any of them.

`get_trace_buffer`, `get_trace_slices` and `parse_trace_slices` take a `kernel` argument in `"device:sequence"` form, defaulting to `"main:sequence"`. An unknown name raises and lists what the module holds:

```
no runtime sequence named 'main:nosuch'; found ['main:sequence', 'dev_a:seq_a', 'dev_b:seq_b']
```

The test covers reading a callee directly (`dev_b:seq_b`), which returns that sequence's own buffer and an empty slice list.

lib/Dialect/AIEX/Transforms/AIEFuseTraceBuffers.cpp
```diff
@@ -0,0 +1,207 @@
...
+      }
+    }
+  }
+
+  /// Give `seq` one trace buffer covering its own traces and those of every
+  /// sequence it calls. Pass each call site its slice.
+  LogicalResult fuse(AIE::RuntimeSequenceOp seq) {
+    if (!done.insert(seq.getOperation()).second) {
+      return success();
+    }
```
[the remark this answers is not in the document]
andrej: Cycles are out of scope and rejected by other passes because we cannot support recursion -- it would require infinitely inlining runtime sequences into themselves.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -2761,17 +2761,21 @@ def AIE_RuntimeSequenceOp : AIE_Op<"runtime_sequence", [
...
+      // If set to false, suppresses automatic insertion of the parameter-sync
+      // preamble (create_scratchpad + update_from_scratchpad + set_lock) by
+      // `--aie-lower-scratchpad-parameters`.
+      // Defaults to true when any `aiex.read_scratchpad_parameter` is present
+      // in a core in the parent device, or any `offset_parameter` attribute on
+      // a DMA BD.
+      // Set to false automatically after the preamble has been emitted.
+      OptionalAttr<BoolAttr>:$emit_parameter_sync_preamble,
+      // Which argument receives trace data, and how the trace buffer is split
+      // between this sequence and the sequences it runs.
```
jgmelber: Can this be clarified/summarized?

include/aie/Dialect/AIEX/IR/AIEX.td
```diff
@@ -1059,7 +1058,28 @@ def AIE_NpuAddressPatchOp: AIEX_Op<"npu.address_patch", []> {
...
     precedence. Only the EmitC (C++ TXN) target supports a runtime `addr_val`;
     the static binary target requires the constant `addr`.
+
+    Exactly one of the `arg_idx` attribute and the `buffer` operand names the
+    host buffer. `arg_idx` is a host-buffer index of the enclosing runtime
+    sequence. It therefore holds only in that sequence. `aiex.run` inlines a
+    callee sequence into a caller with a different argument list. A patch
+    emitted in that callee names the memref.
+    `-aie-resolve-address-patch-buffers` converts the operand to an index after
+    inlining. It adds the buffer's byte offset to `arg_plus`.
```
jgmelber: This is a bit wordy as well

include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
```diff
@@ -269,6 +269,53 @@ def AIEMaterializeRuntimeSequences : Pass<"aie-materialize-runtime-sequences", "
...
+    two traces apart. `aiex.configure` plays no part in the split: the buffer
+    follows the `aiex.run` call graph, whether the calls reconfigure the array
+    between them or repeat one design.
+
+    A sequence's own traces take the front of its buffer, then each call site in
+    turn. The pass records the layout as an `aie.trace_slices` array of
+    `#aie.trace_slice`, and the buffer itself as `#aie.trace_buffer`. A host
+    reads the slices to attribute each region to the design that wrote it.
+
+    Run after the trace pipeline and before `-aie-materialize-runtime-sequences`.
```
jgmelber, line 292: This is a good explanation and required to be this long

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: @jgmelber This PR only adds the last sentence. I tried to clarify it a bit but I think you were referring to the reformatted lines above about the scratchpad parameters, so I took the opportunity to clean these up as well. I don't know if it's much shorter, but hopefully it's clearer.

include/aie/Dialect/AIEX/IR/AIEX.td
andrej: @jgmelber I tried my best to clarify this, though it did end up a bit longer. Added two sentences on what the op does in the first place (rather than just "address patch operator.") while I was at it.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -2761,17 +2761,20 @@ def AIE_RuntimeSequenceOp : AIE_Op<"runtime_sequence", [
...
+      // this runtime sequence copying and synchronizing scratchpad parameters?
+      // Defaults to true if the scratchpad parameter feature is used anywhere
+      // (`aiex.read_scratchpad_parameter` on cores, `offset_parameter` on BDs).
+      // Set to false once emitted, so it does not get re-emitted.
+      OptionalAttr<BoolAttr>:$emit_parameter_sync_preamble,
+      // Which of the runtime sequence arguments is the trace buffer, if any?
+      OptionalAttr<TraceBufferAttr>:$trace_buffer,
+      // If this runtime sequence produces multiple traces, `trace_slices`
+      // states which slice of the trace buffer corresponds to which trace.
+      OptionalAttr<TraceSliceArrayAttr>:$trace_slices);
```
andrej: @jgmelber rewritten

jgmelber (approved):

PR #3658, opened 2026-08-31
[programming_examples] ATB config1: size converted_A by memory_bytes() and raise the stack [LOW]

atassis:
## Problem

`config1/mm_bfp_mixed.cc` allocates `bfp16ebs8 converted_A[M*K/8]` by `sizeof`, which under Peano is **1 byte** per element -- 256 B total. The stream writes that same buffer by `block_vector::memory_bytes()`, 72 B x 32 pushes = **2304 B**, a stride aie_api computes deliberately (`detail/aie2p/array_helpers.hpp`: *"Struct sizes ... don't match the actual storage in memory"*).

Each side is correct in isolation; the 2048-byte overrun exists only in their disagreement, and no type system, compiler or test owns it.

Underneath it sits a second defect: sizing the array correctly grows the frame to `0x1640`, which overruns the `0xD00` stack reservation by 2368 B.

## Fix

- Size `converted_A` by `block_vector<bfp16ebs8, 64>::memory_bytes()` rather than `sizeof`.
- Raise `stack_size` `0xD00` -> `0x1700`.

Both are needed. Fixing only the seam takes config1 from 967/1000 mismatches to **66/1000 -- still `Failed.`** That residue is the second defect, and it is why earlier stack-only attempts (967 / 944 / 935) read as refuting the stack hypothesis when in fact both defects were live and each masked the other's fix.

## Test

On device, both fixes applied: **7/7 runs PASS, 0/1000 mismatches each.**

The frame figure is disassembled from the built ELF, not inferred.

## Scope

`sizeof(bfp16ebs8) == 1` under Peano is confirmed empirically against the build headers here. The reported `== 9` under Chess is untested on this box (no `xchesscc` available) and rests on llvm-aie#1232 -- not refuted, just not independently verified by me.


programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config1/n32_core.py
```diff
@@ -186,7 +186,13 @@ def core_fn(in_a, in_b, out_c, zero, matmul):
...
                 matmul_kernel,
             ],
-            stack_size=0xD00,
+            # Peano drops the chess_storage(exN/dmN) placements above, so those
+            # block_vectors spill to the stack: with converted_A sized correctly
+            # the frame is 0x1640 (disassembled `paddxm [sp], #0x1640`), which
+            # 0xD00 silently underprovisioned. 0x1800 overflows the core's data
+            # region by 4 B; 0x1700 is the largest round value that links.
+            # Chess pins these vectors off-stack and likely needs less; untested.
+            stack_size=0x1700,
```
AngryLoki, line 195: This stack size is big (5.75 KB vs original 3.25 KiB)

When sum of ObjectFIFO buffers + BSS + stack exceeds ~64 KB, L1 overflows.

https://arxiv.org/html/2511.16041v1#S6 mentions 60.3 KiB buffer size, which with stack 3.25 KiB fits the L1. 5.75 KB likely does not*

\* the calculations in the paper are not byte-perfect, but much better solution is to prevent spill (which https://github.com/Xilinx/mlir-aie/pull/3650 already does) rather than increase stack to the largest and hoping it does not corrupts L1 data.

atassis:
Checked the budget for this config. L1 holds, per core:

| buffer | size | depth | total |
|---|---|---|---|
| `A_l1` 32x64 bf16 | 4096 B | 2 | 8192 |
| `B_l1` 64x16 v8bfp16ebs8 (9 B) | 9216 B | 2 | 18432 |
| `C_l1` 128x128 bf16 | 32768 B | 1 | 32768 |
| | | | **59392 B = 58 KiB** |

64 KiB - 58 KiB leaves 6144 B, and `0x1700` is 5888 B, so it fits with 256 B spare. The paper's
60.3 KiB is a different configuration. `--alloc-scheme=basic-sequential` would have refused the
build otherwise, and it ran 7/7 on device.

You are right that growing the stack is not the end state, and 256 B is thin. If #3650 lands the
reservation should come back down with it.

On the seam itself: your rewrite fixes it too -- `uint8_t converted_A[m_a * k / 8 * 9]` with the
`sizeof(bfp16ebs8) == 1 on Peano` comment. Two of us finding it separately is worth more than
either of us finding it once. One suggestion there: `block_vector<bfp16ebs8, 64>::memory_bytes()`
derives that 9 instead of spelling it, and a hardcoded 9 is the same species of constant that
caused the bug.

These two lines apply to the tree as it stands and carry device numbers; your rewrite deletes them
wholesale, so it costs you nothing to resolve if it lands after.

programming_examples/ml/block_datatypes/gemm_asymmetric_tile_buffering/config1/n32_core.py
hunhoffe: We've been working on improving stack size estimation and checks since this PR was first opened. It's not done yet (open bug fix PR here: https://github.com/Xilinx/mlir-aie/pull/3680) but maybe that will help?

atassis: The budget is enforced at link time here, not hoped for: `0x1800` fails to link, overflowing the core data region by 4 bytes, and `0x1700` is the largest round value that links. The measured frame is `0x1640`, disassembled from the built ELF. So an oversized stack on this design is a link error rather than silent L1 corruption, which is also what #3618 made per-target.

On device it is 7/7 runs, 0/1000 mismatches each.

I agree on the direction though: not spilling beats a bigger stack, and if #3650 lands it supersedes this entirely. Two things kept me from just closing it. The example is broken today, and `converted_A` is a separate defect from the stack: it was sized by `sizeof(bfp16ebs8)`, which is 1 under Peano, while the stream writes it by `block_vector::memory_bytes()`, 72 B x 32 pushes. Sizing that correctly on its own still left 66/1000 failing, which is what pointed at the stack as a second, larger overrun underneath the first.

hunhoffe (approved):

PR #3659, opened 2026-08-31
[AIEX] aie-expand-load-pdi: reset only the registers the outgoing device left [LOW]

atassis:
## Problem

`aie-expand-load-pdi` currently precedes every device configuration with a `load_pdi` of an empty device, purely to trigger a firmware reset. That resets the whole device to clear whatever the outgoing configuration left behind, even though the set of registers that actually differ between two known devices is computable at compile time.

## Fix

Adds an opt-in `register-reset` option. When enabled, each empty-device reset is replaced with writes restoring exactly the registers the outgoing device wrote and the incoming one does not, using reset values from the register database.

A boundary carrying an address the database does not know **keeps its firmware reset**, so an incomplete database degrades to current behaviour rather than to a silently under-reset device.

## Test

Three lit tests covering the distinct cases:

- `register_reset_differential.mlir` -- dev1 enables slave port South_0 and dev2 enables North_0, so their configs write disjoint slave-config registers and dev2 leaves dev1's port enabled. The boundary restores `0x23F114` (`Stream_Switch_Slave_Config_South_0`, reset value 0) instead of loading an empty device.
- `register_reset_first_load.mlir` -- the first load has no outgoing device to differentiate against.
- `register_reset_mutually_exclusive.mlir` -- the option's interaction with `ctrl-pkt`.

## Scope

Off by default (`register-reset=false`), so no existing pipeline changes behaviour.

### DIFF
--- a/include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
+++ b/include/aie/Dialect/AIEX/Transforms/AIEXPasses.td
@@ -561,6 +561,12 @@ def AIEExpandLoadPdi : Pass<"aie-expand-load-pdi", "mlir::ModuleOp"> {
               "If true, lower each load_pdi to an @ctrl_pkt_overlay load_pdi "
               "followed by aiex.npu.control_packet ops instead of write32/"
               "blockwrite ops with an empty-device reset.">,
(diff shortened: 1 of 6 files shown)


atassis:
Duplicate of #3578, my bad

PR #3660, opened 2026-08-31
[AIEX] Add npu.read_reg: read a device register into the runtime-sequence control stream [LOW]

atassis:
## Problem

Every existing `npu.*` op inside `aie.runtime_sequence` is write-only (`results = (outs)`), so nothing in the dialect can put a device-read value into the control stream at all. The `TXN_OPC_READ_REGS` custom-op class is already reserved in `aie/Runtime/TxnEncoding.h` but no AIEX op has ever emitted it.

## Fix

Adds `aiex.npu.read_reg(%tile, offset)`. It appends a register read to the runtime sequence's TXN control stream: the 32-bit word at `tile`'s local `offset` is read on-device and returned to the host after dispatch completes, in program order alongside every other `read_reg` in the same sequence.

`tile` is an SSA `aie.tile` value, matching `aiex.dma_channel_reset` / `aiex.core_reset` / `aiex.dma_configure_task`.

## Test

- `test/dialect/AIEX/read_reg.mlir` -- parse/print round-trip
- `test/dialect/AIEX/read_reg_invalid.mlir` -- verifier rejections
- `test/Targets/NPU/npu_read_reg_instgen.mlir` -- emitted TXN instruction stream

## Scope

This is the observability half of read-into-control, not the closed-loop half, and the TXN format is why: `RegOff`/`Mask`/`Value` are host-compiled immediates throughout, with no operand slot for a value a prior instruction produced. So a later instruction in the same dispatch cannot branch on what this reads -- the value surfaces on the host only after `ERT_CMD_STATE_COMPLETED`.

That is the same shape XDP's debug/profile/PC plugins already use via `XAie_AddCustomTxnOp(..., XAIE_IO_CUSTOM_OP_READ_REGS, ...)`.


andrej:
Is this a supported firmware instruction on npu2? Were you able to hardware test this?

atassis:
Yes on both counts now. I had not hardware tested it when I opened this, and doing so found a bug.

Tested on npu2 (Ryzen AI 9 465, NPU firmware 1.1.2.64, amdxdna 7.2.0, XRT 2.21.75) with a hand-built TXN through an ordinary `EXEC_NPU` dispatch, not XDP's separate configure transaction. A `write32` of a known word followed by a `read_regs` of the same address returns that word, for both a core-tile and a shim-tile register. Values land at the start of the `use_type::debug` BO bound to the hardware context, one 32-bit word per read, in program order; one 2-address op and two 1-address ops fill it identically. With no debug BO bound the dispatch still completes and the values are dropped.

What this branch currently emits does not work: 4 words with a u32 address and `Size=16`. That times out the dispatch (`ERT_CMD_STATE_TIMEOUT`) and leaves the context unusable, 3/3 runs. The on-wire form is `XAie_CustomOpHdr` followed by the raw `read_register_op_t`, so u32 count, 4 bytes of padding, then a u64 address: 24 bytes for a single register (`_XAie_AppendCustomOp` in aie-rt `xaie_helper.c`, `read_register_op_t` in `xaiegbl.h`). I'll push the corrected encoding.

One question on the device test I want to add under `test/npu-xrt/`: reading the value back needs `xrt_core::bo_int::create_bo(hwctx, sz, use_type::debug)`, which is exported from `libxrt_coreutil` but is not in the installed XRT headers, and every existing `test/npu-xrt/*/test.cpp` sticks to the public API. Would you rather the test declare that symbol itself, or only assert that a sequence containing the op dispatches cleanly?

andrej:
Wow, cool. This must be relatively new or I just never heard of it.

> One question on the device test I want to add under test/npu-xrt/: reading the value back needs xrt_core::bo_int::create_bo(hwctx, sz, use_type::debug), which is exported from libxrt_coreutil but is not in the installed XRT headers, and every existing test/npu-xrt/*/test.cpp sticks to the public API. Would you rather the test declare that symbol itself, or only assert that a sequence containing the op dispatches cleanly?

Which of the following?

1. The XRT version we [pin](https://github.com/Xilinx/mlir-aie/blob/435f2cbb74653b498da8362c8d5fc3e28bd0c716/utils/build_drivers.sh#L99) in `utils/build_drivers.sh` does not publicly contain the function to create the proper BO.
2. This is a private function for which no public headers exist in any version.

If the last, can we just open a PR on XRT to expose the function publicly?

atassis:
The second. bo_int.h lives only at src/runtime_src/core/common/api/bo_int.h, no install rule covers that directory, and the header still says "This is for internal use only" on XRT master today. Nothing public exposes the BO use flags either; xrt::ext::bo surfaces only access_mode. So bumping the pin would not help.

Per PR- the value is already public (`XRT_BO_USE_DEBUG` in `xrt/detail/xrt_mem.h`), only the construction path is not, and the one overload a test needs, `create_bo(const xrt::hw_context&, size_t, use_type)`, takes public types plus that enum, so it does not drag in `buffer_handle/hwctx_handle` like the rest of that header. An `xrt::ext::bo` constructor taking the use flag would probably be the more idiomatic shape. I can open it.

I'll admit that I am still thinking if another approach to the PR is viable, I am still upstreaming my work, I have 25 already prepared branches to go, around 20 more- not prepared, but already defined issues. 

I might pause for a while, seems like I have created too many, it would take a while for all sides to digest them properly.

hunhoffe:
@atassis A good portion of why we were busy this week (and making slow progress) was because we had several large PRs in flight this week (kernel consolidation, stack overflow, allocator, trace, etc.) so the slowness isn't just number of PRs, it also just happened to be a busy week.

If you have fixes or plan fixes for existing issues, please do feel free to comment on the issue so no one addresses it in the meantime.

atassis:
@hunhoffe I want to merge current PRs for now, I have a lot of PRs opened across the stack, it makes me difficult to keep it locally
For example, I have a tool chain lockfile, which cherrypicks all the branches of changes I have made so I can proceed working with fixes already applied to the stack, but it takes a lot of time to juggle them around

hunhoffe:
@atassis Yes, I understand that it's a bit painful because we aren't moving quickly! We'll keep working on reviewing PRs, and maybe next week we can start a new discussion around the roadmap, etc. so we can continue to align on goals.

atassis:
@hunhoffe sorry if sounded rushing you- not at all, just was making a point!

atassis:
@andrej it is your option 2, but the private part is narrower than it looks.

`bo_int.h` lives in `core/common/api/`, has no install rule, and says "This file is for internal use only". So it is internal in every version, not just the pinned one. What is *not* private is the thing you actually need: `XRT_BO_USE_DEBUG` and the rest of the `XRT_BO_USE_*` flags are already public and documented in the installed `xrt/detail/xrt_mem.h`. `bo_int.h` names the whole gap in its own comment on `create_bo`: "All the public xrt::bo constructors doesnt use 64 bit flags." Both `create_bo` overloads are `XRT_CORE_COMMON_EXPORT`, so the symbol links today; only a public constructor reaching the 64-bit flag path is missing.

So yes, a PR on XRT is the right fix and it is a small one. Opened it: Xilinx/XRT#10045. It adds `xrt::ext::bo::use_mode` and a `bo(hwctx, sz, use_mode)` constructor, mirroring how `access_mode` is already exposed on that class, and exposes only the 11 uses `compose_internal_bo_flags` accepts (`kmd` and `debug_queue` hit its `default:` and throw, so exposing them would only give a public route to a runtime error). The cast from the public enum to `bo_int::use_type` is guarded by a `static_assert` per enumerator, since nothing else ties the two together.

If that lands, the device test here can read the value back through the public API and I will add it. Until then the choice is a test that declares the private symbol itself, or one that only asserts a sequence containing the op dispatches cleanly. I would rather wait for the public path than put the private declaration in `test/npu-xrt/`.

PR #3669, opened 2026-09-01
aie_kernels: fix the never-compiled BIT_WIDTH==16 branch of threshold.cc [LOW]

atassis:
## Problem

`threshold.cc`'s `BIT_WIDTH == 16` branch has two compile errors: `max_va` for
`max_val`, and a `threshold_aie` call whose closing paren landed after
`maxValue`, leaving `thresholdType` as a comma-operator statement so a
7-argument template is called with 6.

Nothing in the tree builds it. `-DBIT_WIDTH=8` is the only value any build
passes, so this branch has never been compiled.

## Fix

Both typos. No change to any branch that does build.

## Test

`clang++ --target=aie2p-none-unknown-elf -fsyntax-only` on `threshold.cc` at
`-DBIT_WIDTH=8`, `16` and `32`: all three compile. Before, `16` failed.


hunhoffe:
To ensure this stays fixed, ideally this would be tested by the CI. However, this is still a step up all on it's own -- thanks! 

For this and other kernel PRs, we were waiting for https://github.com/Xilinx/mlir-aie/pull/3599 to land -- now that is has, we'll start reviewing PRs in this area again. Thank you for your patience

hunhoffe (approved):

PR #3670, opened 2026-09-01
aie_kernels: turn a missing kernel -D flag into a build error [MED]

atassis:
## Problem

A kernel that selects its element type from a caller-supplied `-D` macro takes
a silent default when the macro is absent: an unset macro expands to 0, no
branch matches, and the `#else` runs. `passThrough.cc`, `bitwiseAND.cc`,
`scale.cc` and `threshold.cc` all end in `#else // 32`, which is both "the
32-bit case" and "you forgot the flag", with no way to tell them apart.

Measured: `scale.cc` builds clean under `-DBIT_WIDTH=8` and produces the 32-bit
kernel, so the host and the kernel disagree on element size with no diagnostic.

The compiler cannot see this today. `-Wundef` fires on `#if X` but not on
`#ifdef X`, and these gates are written `#ifdef`.

## Fix

- `BIT_WIDTH` becomes a checked precondition in those four kernels: a guard
  before first use, an explicit `#elif BIT_WIDTH == 32` in place of the
  fallthrough, and an `#error` naming the legal values for anything else.
- The bottleneck `BN*`/`PARTIAL_*` selectors are tested with `defined()`,
  matching the `defined()` terms already sitting beside them in the same `#if`
  expressions (35 sites).
- `-Werror=undef` on both kernel compile paths. The Makefile path and the
  `@iron.jit` path keep independent copies of the warning list, so both need it.

## Evidence

Sweep of every `aie_kernels/**/*.cc` at `--target=aie2p-none-unknown-elf` after
the change: 38 compile clean under `-Werror=undef`, 0 break, 17 need a `-D` to
compile at all and are unchanged here.

`BIT_WIDTH` swept over unset / 8 / 16 / 32 / 7 for each of the four kernels:
every valid value compiles, unset and invalid error with the kernel's own
message. `scale.cc` errors at 8, which is correct, it implements only 16 and 32.

The same sweep found #3669: `threshold.cc`'s 16-bit branch had two compile
errors and had never been built.

## Scope

`-Wundef` is only usable with `--system-header-prefix=aie_api/`. aie_api tests
capability macros Peano never defines (`__AIE_CORE_BUILTIN_CLZ__`,
`__AIE_API_CFP_TO_FP_CONVERSIONS__`, `AIE_API_DISABLE_ALIGNMENT_ASSERTIONS`),
so without the prefix 45 of 55 kernels fail inside vendored headers.

Known gap: `aie_kernels` reads 87 macros, of which 59 are defined nowhere in
the tree. Most are conv2d/bottleneck variant selectors from a design generator.
This PR makes them visible to `-Wundef` where they were already written `#if`,
and leaves the rest as they are.


atassis:
Fixed duplicated defined issue, formatted

hunhoffe (approved):

PR #3672, opened 2026-09-01
[AIE] Reject a shim DMA task that does not cover whole objectFIFO objects [HIGH]

atassis:
## Problem

A shim DMA fill/drain whose total transfer is not a whole multiple of its
objectFIFO's object size leaves the receiving MemTile S2MM short. The consumer
then blocks forever: the design compiles clean, runs, and hangs the array with
no diagnostic from any layer.

Nothing checks it today. The objectFIFO's object size is known at
`aie.shim_dma_allocation` time and thrown away, so by the time the DMA task
exists there is nothing left to compare a transfer length against.

Found through a device hang in IRON's `strided_copy` for every
`num_aie_channels > 1`, which was first mistaken for a hardware DMA limit.

## Fix

- `aie.shim_dma_allocation` records the objectFIFO's element type, so the
  object size survives into the DMA-task ops that reference the allocation.
- `verifyTaskCoversWholeObjects` rejects a task whose measurable bytes are not
  a whole multiple of that object size.
- Both extents are counted in the bytes a buffer descriptor counts
  (`DataLayout::getTypeSize`, the unit `DMABDOp::getBufferElementTypeWidthInBytes`
  already reports), so a sub-byte element type measures the same on both sides.

The check is deliberately narrow. It runs only when both sides are static: a
runtime length or repeat count has no constant to compare, and an allocation
with no recorded element type (a control-overlay channel, a join or split, an
endpoint that pads into its object) records nothing to compare against. All of
those return success rather than guessing.

## Test

`test/dialect/AIEX/dma_task_covers_whole_objects.mlir`, covering the rejected
case and each of the cases that must still pass, and
`test/objectFifo-stateful-transform/stream_len_decoupled.mlir` for the
exemption reaching the allocation.

Built against upstream/main and run locally: both tests pass, and the suites
that exercise the touched passes, `test/objectFifo-*` and
`test/dialect/{AIE,AIEX}`, are 118 pass / 0 fail. Tests needing aiecc,
aie-translate or a multi-step RUN line were left to CI.

## Scope

This is the dialect-level counterpart to #3601, which rejects the same mistake
at the IRON API. The verifier catches it for any frontend, including
hand-written MLIR, and does not depend on the Python layer being the one that
built the design.

## Known limitation: a compressing channel has to say so

DMA compression breaks the premise, and the IR cannot tell. A channel whose DMA
(de)compresses moves a data-dependent count of bytes per object, nothing in the
dialect models compression, and the enable is a BD register written at runtime
or from inside a core. So a channel that does it declares
`stream_len_decoupled` on its fifo, which travels to the shim allocation and
exempts it.

`programming_examples/basic/dma_compression` is the case in tree: it ratio-sizes
its shim BDs to the compressed length, on the input side under decompression and
on the output side under compression, and now sets the flag on both fifos.


hunhoffe:
@atassis We updated the CI drivers so there are some CI failures due to that for the moment. However, the copilot comments and the succeeding tests should still give additional information -- hopefully enough to move forward regardless.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -2006,15 +2006,21 @@ def AIE_ShimDMAAllocationOp : AIE_Op<"shim_dma_allocation", [HasParent<"DeviceOp
...
-  );
+  let arguments = (ins SymbolNameAttr:$sym_name, Index:$tile,
+      DMAChannelDir:$channel_dir, AIEI64Attr:$channel_index,
+      // If this is set we are using the PLIO in this ShimTile
+      DefaultValuedAttr<BoolAttr, "false">:$plio,
+      OptionalAttr<PacketInfoAttr>:$packet,
+      // The object the far end of this channel receives, when the allocation
+      // was emitted from an objectFIFO. A runtime sequence sizes its buffer
+      // descriptor in the host buffer's element type; this records what the
+      // stream expects, which is otherwise lost at this symbol.
```
andrej, line 2017:
```suggestion
      // was emitted from an objectFIFO.
```

andrej (commented):
Is carrying two new attributes through the IR in various places (`elem_type` and `stream_len_decoupled` on `ShimDMAAllocation`) really the best way to do this, or could we just run the check/verifier earlier before the ObjectFifo is lowered to the ShimDMAAllocation? Although not really used, the `ShimDMAAllocation` op appears to me to potentially be usable beyond ObjectFifos, so adding ObjectFifo-specific attributes and checks on it seems not ideal.

Don't feel extremely strongly about it but I do think this verifier would be much simpler if it ran earlier, before lowering to `ShimDMAAllocation`.

hunhoffe:
Is this PR rebased on https://github.com/Xilinx/mlir-aie/pull/3683? It so, it is blocked too.

PR #3679, opened 2026-09-03
[AIEX] Look inside nested regions when materializing runtime sequences

erwei-xilinx:
`AIEMaterializeRuntimeSequences` inlines an `aiex.run` call into the calling device, and in two places it only considered the top level of the callee's body. Any runtime sequence containing an `scf.for` hits one or both. Both are reachable today through `aircc --output-format elf` on an AIR design whose `air.segment` sequence has a loop in it.

## Values defined inside a nested region

`collectReferencedSSAValues` decides whether a value used in a nested region is defined within the op being inlined:

```cpp
Operation *defOp = operand.getDefiningOp();
if (defOp && op->isProperAncestor(defOp)) return;
```

A block argument has no defining op, so this never fires for one. An `scf.for` induction variable is therefore collected as an *external* reference, reaches `copyReferencedSSAValues`, and fails its null `getDefiningOp()` check:

```
error: Referenced value is not defined by an operation
error: 'aiex.run' op expects parent op 'aiex.configure'
```

The second error is misleading — the op *is* inside an `aiex.configure`, and `ConfigureOp` is not `IsolatedFromAbove`, so the operands are legally visible. The first is the real one. Fixed by extending the "defined within" test to a block argument whose owning region is nested inside the op.

## Symbol references in nested regions

`inlineReferencedSymbolDefinitions` reads only the attributes of the op it is given, and was called once per *top-level* op of the callee body. An `aiex.dma_configure_task_for` naming a shim DMA allocation commonly sits inside an `scf.for`, so its symbol was never visited: the definition stayed in the callee while the reference travelled into the caller, and a later pass failed with

```
error: 'aiex.dma_configure_task_for' op no shim DMA allocation found for symbol @air_outD_15
```

Top-level allocations inlined correctly, which is why only the in-loop ones failed. Fixed by walking the cloned op so nested references get the same inlining and renaming.

The two are sequential: the first fix is what exposes the second, so splitting them would ship a change that turns one error into another.

## Tests

Two tests, one per fix, in the existing `test/Passes/materialize-runtime-sequences` suite:

- **`inline_loop_block_arg.mlir`** — an induction variable used in the loop body. Errors out before the fix.
- **`inline_symbols_in_loop.mlir`** — a shim DMA allocation referenced from inside a loop alongside one referenced at top level. Before the fix the pass *succeeds* but silently leaves the in-loop allocation behind in the callee, so this is a FileCheck failure rather than an error. That catches the bug where it is introduced instead of in a downstream pass, and the top-level allocation is checked alongside deliberately: it working while the in-loop one did not is what the bug looked like.

The other eight tests in the suite are unaffected.

Also verified on hardware (NPU2 / Strix), since the second fix concerns IR that is wrong rather than IR that is rejected:

- mlir-air's `int4_q4nx` matrix-vector example builds an ELF whose output is **bitwise identical** to the `golden.elf` the example ships (0 of 2048 elements differ).
- A 16-layer fused decode superkernel builds an ELF whose logits are **bitwise identical** to the xclbin built from the same module.

Validation ran against `af819a802f4` (the commit `mlir-air/utils/clone-mlir-aie.sh` pins), where `AIEMaterializeRuntimeSequences.cpp` and `test/Passes/materialize-runtime-sequences/` are byte-identical to `main`. CI is the check on `main` itself.

## Known follow-up, not addressed here

With the second fix in place, a caller that inlines many `aiex.run` calls — 23, in a decode compiled with shim-DMA BD tiling — fails with `redefinition of symbol named '__air_herd_lock_2_2'`. `allSymbolNames` is seeded once from the caller's top-level block while `previouslyInlinedSymbolMap` is local to each `matchAndRewrite`, so every call re-inlines the whole callee symbol set; that should still produce unique names, so one of the two is not holding across invocations. Stated as unresolved rather than guessed at — happy to file it separately.

### DIFF
--- a/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
@@ -182,14 +182,19 @@ collectReferencedSSAValues(Operation *op, const IRMapping &argMap,
   for (Region &region : op->getRegions()) {
     region.walk([&](Operation *nestedOp) {
       for (Value operand : nestedOp->getOperands()) {
+        // `continue`, not `return`: skipping an operand must not abandon the
+        // rest of the same operation's operands. Returning here means an op
+        // whose first operand is already mapped never has its later operands
+        // examined, so a genuine external reference beside a mapped one is
+        // silently dropped.
         if (argMap.contains(operand)) {
-          return;
+          continue;
         }
 
         // Check if defined within the parent operation.
         if (Operation *defOp = operand.getDefiningOp()) {
           if (op->isProperAncestor(defOp)) {
-            return;
+            continue;
           }
         } else if (auto blockArg = llvm::dyn_cast<BlockArgument>(operand)) {
           // A block argument has no defining op, so the check above cannot see
@@ -200,7 +205,7 @@ collectReferencedSSAValues(Operation *op, const IRMapping &argMap,
           // fail with "Referenced value is not defined by an operation".
           Operation *owner = blockArg.getOwner()->getParentOp();
           if (owner && (owner == op || op->isProperAncestor(owner))) {
-            return;
+            continue;
           }
         }
 
--- a/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
+++ b/lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
@@ -182,11 +182,6 @@ collectReferencedSSAValues(Operation *op, const IRMapping &argMap,
   for (Region &region : op->getRegions()) {
     region.walk([&](Operation *nestedOp) {
       for (Value operand : nestedOp->getOperands()) {
-        // `continue`, not `return`: skipping an operand must not abandon the
-        // rest of the same operation's operands. Returning here means an op
-        // whose first operand is already mapped never has its later operands
-        // examined, so a genuine external reference beside a mapped one is
-        // silently dropped.
         if (argMap.contains(operand)) {
           continue;
         }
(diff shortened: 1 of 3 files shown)


erwei-xilinx:
Thanks — both addressed in b2f0f1d.

**`return` vs `continue` in the nested-op walk.** Correct, and worse than the comment says: it applies to the `argMap` case *and* to the "defined within" case right below it, which this PR extended. `return` exits the walk callback rather than advancing to the next operand, so an op whose first operand is already mapped never had its later operands examined. Changed all three to `continue`.

I could not build a test that exercises it. `copyReferencedSSAValues` only clones `aie.tile` and `aie.lock`, and no op legal in a runtime sequence takes a tile or lock operand *after* another operand that would be skipped — `aiex.set_lock` is single-operand, and `aiex.npu.dma_memcpy_nd`'s variadic operands are `i64` offsets/sizes/strides, which that function rejects anyway. So the change is correct by inspection rather than covered. Since the `argMap` half is pre-existing and untested, I put it in its own commit — happy to drop it if you'd rather keep this PR to the nested-region issue.

**Anchoring the symbol checks.** Agreed on the concern, but the suggested anchor does not exist: the pass emits the rewritten caller device *without* a sym name, so `aie.device(npu2) @main` never appears in the output — it is just `aie.device(npu2) {`. That unnamed form is itself the discriminator, since the callee prints as `aie.device(npu2) @config_with_symbols`, so I anchored on it and switched the two allocations to `CHECK-DAG`.

Verified against a build of this branch: all ten tests in `test/Passes/materialize-runtime-sequences` pass, and both new tests still fail with `AIEMaterializeRuntimeSequences.cpp` reverted to `origin/main` — `inline_loop_block_arg` by erroring, `inline_symbols_in_loop` by leaving `@buffer_in_loop` behind in the callee.

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -182,14 +182,31 @@ collectReferencedSSAValues(Operation *op, const IRMapping &argMap,
   for (Region &region : op->getRegions()) {
     region.walk([&](Operation *nestedOp) {
       for (Value operand : nestedOp->getOperands()) {
+        // `continue`, not `return`: skipping an operand must not abandon the
+        // rest of the same operation's operands. Returning here means an op
+        // whose first operand is already mapped never has its later operands
+        // examined, so a genuine external reference beside a mapped one is
+        // silently dropped.
```
andrej:
```suggestion

```

lib/Dialect/AIEX/Transforms/AIEMaterializeRuntimeSequences.cpp
```diff
@@ -515,10 +532,23 @@ struct InlineRuntimeCallsPattern : RewritePattern {
...
-              rewriter, clonedOp, calleeRuntimeSequence.getOperation(), argMap,
-              previouslyInlinedSymbolMap, callerDevice, symbolDefInsertPoint,
-              allSymbolNames))) {
+      // Symbol references are not only on the top-level op: an
+      // aiex.dma_configure_task_for naming a shim DMA allocation commonly sits
+      // inside an scf.for in the callee sequence. Those definitions need the
+      // same inlining and renaming, or the reference survives into the caller
+      // device pointing at a symbol that was left behind in the callee, and a
+      // later pass fails with "no shim DMA allocation found for symbol".
+      // walk() visits clonedOp itself as well as its nested ops.
```
andrej:
```suggestion
      // Inline symbol references in all nested ops.
```

andrej (approved):

erwei-xilinx:
Thanks for looing into the PR, @andrej!

PR #3685, opened 2026-09-06
[python] Compile a design's kernels concurrently, not one after another [LOW]

atassis:
## Problem

`CompilableDesign.compile()` compiles a design's external kernels one after another:

```python
for func in external_kernels:
    if not func._compiled:
        compile_external_kernel(func, kernel_dir, target_arch)
```

They are separate translation units writing separate objects, so they have always been independent —
they were merely serialised, and a design's kernel-compile wall is the SUM rather than the max. The
same loop appears three times (the xclbin path, the full-ELF path, and `compile_mlir_module`'s
auto-build for callers that do not go through `@iron.jit`).

## Fix

One `compile_external_kernels()` helper, used by all three sites, that runs them concurrently.

Two ExternalFunctions can still collide: `compile_external_kernel` writes the source as
`<_original_name>.cc`, so two sharing an original name write the same file. Those keep their relative
order; everything else runs concurrently. `list(pool.map(...))` re-raises the first failure after the
others have finished, so a compile error cannot be swallowed by a sibling that succeeded.

Bounded by cores, since each compile is a single-threaded clang. `AIE_KERNEL_COMPILE_JOBS` overrides.

## Test

On a two-kernel design (`StaticMHA`, 24 cores) the kernel-compile step goes from 7.88 s to 2.57 s;
`insts.bin` and both kernel objects stay byte-identical. The gain is `max` instead of `sum`, so it
scales with how many kernels a design has and is nothing for a single-kernel design.

Each compile peaks near 250 MB of RSS on aie2p, so on an ordinary machine the bound is cores rather
than memory; the docstring says so and names the override.


hunhoffe:
We might need to make this dependent on "not chess", at least while chess is still used/supported.

python/utils/compile/utils.py
```diff
@@ -573,15 +573,10 @@ def _rename_symbol_in_object(object_path: str, old_name: str, new_name: str) ->
...
-    always been independent -- they were merely compiled one after another, and
-    the wall was the SUM.  On the encoder-MHA design here that is 7.9 s of the
-    17.1 s build for two kernels.
-
-    Two of them can still collide: ``compile_external_kernel`` writes the source
-    as ``<_original_name>.cc``, so two ExternalFunctions sharing an original name
-    write the same file.  Those keep their relative order; everything else runs
-    concurrently.
+    Kernels are separate translation units with separate outputs, so they compile
+    concurrently.  The exception is the source path: ``compile_external_kernel``
```
hunhoffe, line 581: Documentation needs to be updated re: chess. Since our priority is peano for the moment, I think it's ok to have that be a follow-on.

hunhoffe (approved):

hunhoffe (approved):

jgmelber (approved):

PR #3686, opened 2026-09-06
[python] Let a compiler launcher wrap the kernel compile [LOW]

atassis:
> Stacked on #3684 — the interaction this warns about only exists once the PCH does. Review that one
> first; the last commit here is the only new content.

## Problem

`compile_cxx_core_function` already emits `-MD -MF`, with a comment saying it is there so "ninja and
ccache learn a translation unit's real inputs". Nothing ever routes the compile through such a tool,
so a kernel `.cc` is recompiled whenever anything around it moves — a sibling design rebuilt, a branch
switched back, a CI job rerun — even though nothing it reads changed.

## Fix

`AIE_KERNEL_COMPILER_LAUNCHER=ccache` (or sccache, or any launcher taking the compiler as its first
argument) prefixes the compile. Unset by default: a compiler cache is a policy about the developer's
machine, not a property of the compile.

It also warns about an interaction that is otherwise invisible. **ccache silently refuses to cache any
compile that uses `-include-pch` unless `time_macros` is in `CCACHE_SLOPPINESS`** — the call is
counted "uncacheable", not a miss, and the explanation appears only under `CCACHE_DEBUG`. Set both
levers without knowing this and you get neither. The warning fires once, names the fix, and costs
nothing when the PCH is not in use.

## Test

`mha.cc`: **5.44 s miss -> 0.010 s hit** (direct mode), object identical. With the PCH as well:
1.92 s miss, 0.026 s hit.

Perturb-checked rather than assumed — editing the `.cc` misses, editing an included header misses, a
comment-only edit of a transitively included `.cc` takes a preprocessed-mode hit with the correct
object.

`pch_defines` is *not* required; verified at `time_macros` alone.


hunhoffe:
On first glance, the comments/documentation seems a little excessive -- any chance you could do a clean up to make it a bit more concise?

python/utils/compile/utils.py
```diff
@@ -24,6 +27,146 @@
...
+# into an ordinary compile.
+#
+# Set AIE_KERNEL_PCH=0 to disable.
+_PCH_ENABLED = os.environ.get("AIE_KERNEL_PCH", "1") != "0"
+
+
+# Opt-in: a compiler cache is a policy about the developer's machine, not a
+# property of the compile.  Accepts any launcher that takes the compiler as its
+# first argument (ccache, sccache).
+_COMPILER_LAUNCHER = os.environ.get("AIE_KERNEL_COMPILER_LAUNCHER", "")
```
hunhoffe, line 50: Can you make sure both of these environment variables are documented in https://github.com/Xilinx/mlir-aie/blob/main/programming_guide/iron_configuration.md?

python/utils/compile/utils.py
```diff
@@ -24,6 +27,146 @@
...
+    """Warn that ccache silently declines to cache a -include-pch compile.
+
+    The call is counted "uncacheable" rather than missed, so setting both levers
+    gets neither; ccache's own explanation only appears under CCACHE_DEBUG.
+    """
+    if _warn_launcher_pch_interaction.__dict__.setdefault("done", False):
+        return
+    if "ccache" not in os.path.basename(launcher):
+        return
+    if "time_macros" in os.environ.get("CCACHE_SLOPPINESS", ""):
```
hunhoffe, line 63: Is the use of the env also documented? including the relation to CCACHE_DEBUG?

python/utils/compile/utils.py
```diff
@@ -24,6 +27,146 @@
...
+        "use. Add time_macros to CCACHE_SLOPPINESS, or set AIE_KERNEL_PCH=0 to "
+        "trade the PCH for the cache."
+    )
+
+
+def _compiler_identity(cxx: str) -> str:
+    """Key on the compiler build, not on where it sits.
+
+    ``clang --version`` prints the llvm-aie git SHA, identical across installs of
+    one build; a path or an mtime is not (mlir-aie#3427 keyed aiecc by mtime and
```
hunhoffe, line 77: This seems a bit historical

python/utils/compile/utils.py
```diff
@@ -360,6 +527,21 @@ def compile_cxx_core_function(
         check=False,
         capture_output=True,
     )
+    if (
```
hunhoffe, line 530: I'm still not sure what the best behavior is. If the PCH got corrupted, then there will be retries for every session thereafter, which seems expensive. Would it make more sense to fail with a warning to clear the PCH out of the cache or disable this path via environment variable (with example?)

hunhoffe (commented):
I like the spirit of this quite a bit, I just want to make sure expected behavior is well-documented and clear for the user.

Issue #3508, opened 2026-08-06
IRON JIT: support 1-overlay / N-sequence designs — separable insts/overlay compilation, parameterized (emitC) sequences, and a cache model that decouples them


hunhoffe:
### What are you trying to do?

The IRON JIT layer models a design as a single monolithic `(xclbin + insts)` artifact. This works for one-shot kernels, but not for a common pattern: **one shape-agnostic overlay driving many runtime sequences**, where the on-chip dataflow (tiles, ObjectFifos, kernels) is identical and only the shim runtime sequence differs. Today such designs must recompile the whole overlay for every sequence and discard the redundant xclbins.

This issue proposes closing that gap in three connected layers, from a minimal enabler up to the ergonomic goal.

### Layer (a) — allow insts-only / overlay-only compilation

`CompilableDesign.compile()` requires `xclbin_path` and `inst_path` to be set *together* (`python/aie/utils/compile/jit/compilabledesign.py:279`):

```python
if (xclbin_path is None) != (inst_path is None):
    raise ValueError("compile(): xclbin_path and inst_path must be set together ...")
```

But the underlying driver already gates the two artifacts independently (`python/aie/utils/compile/utils.py:478-481`):

```python
if insts_path:
    args.extend(["--get-npu-insts", f"--npu-insts-name={insts_path}"])
if xclbin_path:
    args.extend(["--get-xclbin", f"--xclbin-name={xclbin_path}"])
```

### Layer (b) — parameterized sequence as a standalone artifact (dynamic emitC path)

The runtime already has a **dynamic lowering path** (`python/aie/iron/runtime/runtime.py:8-15`): a sequence body with `range_`/`if_` and runtime scalar args lowers to `scf.for`/`scf.if` → **EmitC C++ TXN**, versus the static path (Python `range`/int bounds → flat binary insts). This is exactly the mechanism for "one sequence, shape supplied at launch" — instead of N static sequences, one parameterized sequence whose offsets/bounds are runtime values.

### Layer (c) — cache model that decouples overlay and sequence

[...]

PR #3696, opened 2026-09-08
[python] JIT: allow insts-only / overlay-only compilation [LOW]

atassis:
## Description

Part of #3508 (layer (a)). `CompilableDesign.compile()` rejected an `xclbin_path` /
`inst_path` pair unless both were set, and used `explicit_paths = xclbin_path is not None`
downstream as the proxy for "explicit mode". One layer below, `compile_mlir_module()` already
gates the two `--get-*` flags independently, and `aiecc` builds each on its own: `--get-npu-insts`
alone runs a 13-edge graph and writes the insts; `--get-xclbin` alone runs 32 and writes no `.bin`.
So the coupling lived only in the Python.

That blocks the pattern #3508 is about — one shape-agnostic overlay driving many runtime
sequences. Today such a design either rebuilds and discards an identical xclbin per sequence, or
skips `compile()` entirely.

Fix: `explicit_paths` becomes "either path given"; the `.prj` scratch dir is anchored on whichever
artifact was requested; `expected_outputs` checks only what was asked for. The slot of an
un-requested artifact is `None` in the returned tuple, so the return type widens to
`tuple[Path | None, Path | None]` (propagated to `CallableDesign.compile`). `elf_path` / `pdi_path`
still require an explicit output path — the cache does not track them — and the diagnostics are
reworded but keep their existing wording prefix. Passing both paths, or neither, is unchanged.

## Test / evidence

`test_compile_mixed_explicit_paths_raises` asserted the old rejection and is replaced by
`test_compile_single_path_selects_that_artifact`, which checks both single-path modes: only the
requested `--get-*` is threaded, the other slot is `None`, and `get_artifacts()` stays `None` for a
partial build.

On device (Strix, npu2), for one design compiled three ways:

- insts built alone are **byte-identical** to the insts of a full build of the same design;
- an **overlay-only xclbin loaded with an insts-only `.bin` runs correctly** — the two halves were
  produced by separate `compile()` calls that never saw each other;
- skipping the overlay is **2.87x** faster per sequence on the 16-core `whole_array` design
  (1.32 s -> 0.46 s median over 3 alternated pairs, warm ccache). On a 1-core design it is 1.05x —
  the saving is the per-core ELF compiles, CDO, bootgen and packaging, so it scales with the
  overlay, not the sequence.

`black`, `ruff check` and `pyright` are clean on the touched files (pyright error count unchanged
against the pre-change baseline).

## Checklist


test/python/test_compilabledesign.py
```diff
@@ -1201,17 +1202,41 @@ def test_kernels_mm_mac_dims_per_arch():
     ), f"AIE2P i16/i16 mac_dims expected (4, 4, 8), got {mm_aie2p.mac_dims}"
 
 
-def test_compile_mixed_explicit_paths_raises():
-    """Passing only one of (xclbin_path, inst_path) is rejected up front."""
+def test_compile_single_path_selects_that_artifact(tmp_path, monkeypatch):
+    """Either path alone compiles just that artifact and bypasses the cache."""
+    seen = {}
+
+    def fake_compile_mlir_module(**kwargs):
```
hunhoffe, line 1209: I've found that AI especially makes fake tests and monkeypatches when real tests will work too. Is this something that could be tested more directly?

hunhoffe (approved):
Can you double check my note on tests? If there's something to do, happy to merge afterwards. If there's nothing to do, I'll merge in current state. Thanks for another contribution!

PR #3697, opened 2026-09-08
[aiecc] Add --get-npu-cpp: the runtime sequence as a C++ builder [MED]

atassis:
## Description

Part of #3508 (layer (b)), and depends on the layer (a) PR. The dynamic lowering already exists —
a runtime sequence written over runtime scalars keeps its `scf.for`/`scf.if` and lowers through
`convert-aiex-to-emitc` to a C++ TXN builder — but **`aiecc` had no edge for it**. The graph
exposes `npu_lowered.mlir` and `npu_seq_{0}.mlir` and stops; `AIETranslateNpuToCpp` was reachable
only as an `aie-translate` target. That is why
`test/npu-xrt/matmul_whole_array_dynamic/run.lit` reaches the capability through eight hand-written
pass flags and a separate `aie-translate` invocation. The capability shipped; the packaging did not.

This adds `--get-npu-cpp` (`--npu-cpp-name`, default `npu_seq_{0}.cpp`): a `perSeq` edge running
`AIETranslateNpuToCpp`. The module is cloned first — the translation runs a `PassManager` that
rewrites in place, and `perSeq` also feeds the instruction-binary path — matching the existing
`txn` edge's idiom. No CMake change: `AIETargetCppTxn.cpp` is already in `AIETargets`, which
already links `MLIRAIEXToEmitC`.

Also threads `npu_cpp_path` through `compile_mlir_module()` and `CompilableDesign.compile()` /
`CallableDesign.compile()`. It counts as an explicit output path, so `compile(npu_cpp_path=...)`
alone builds just the sequence — decoupled from any overlay, which is the point.

## Test / evidence

New `test/aiecc/cpp_npu_cpp_sequence.mlir`: a sequence taking its d2 size as a runtime argument, so
no flat binary can stand in for the artifact. It checks the builder is written, takes the runtime
size as a parameter, and that the overlay branch never runs. The test **fails without this change**
(verified by reverting and rebuilding), so it is a regression test rather than a tautology.

Equivalence, not just plausibility: the edge's output is **byte-identical** (28801 B) to what
`run.lit`'s hand-rolled `aie-opt | aie-translate --aie-npu-to-cpp` pipeline produces for the same
design. It emits exactly what the documented manual route emits.

End to end on device (Strix, npu2), two independent `aiecc` invocations — one `--get-xclbin`, one
`--get-npu-cpp`, neither producing the other's artifact — then one overlay plus one parameterized
sequence serving six shapes, 0 errors: 512x512x512, 256x512x1024, 768x512x768, 1024x512x512,
256x512x512, 512x512x1024.

`test/aiecc` and `test/Targets` show no new failures against the unmodified binary. `black`,
`ruff check`, `pyright` and `clang-format` clean on the touched files.

## Checklist


jgmelber (commented):
Please address copilot feedback, otherwise LGTM

atassis:
Addressed copilot, commited

jgmelber:
Still some failures in CI left to address in this

PR #3698, opened 2026-09-08
Kernel contracts and a generic test/benchmark harness for aie_kernels

hunhoffe:
## Summary                                                                                                            
                                                                                                                        
This adds a contract system for `aie.iron.kernels` factories — each one now declares its argument roles, a numpy reference, a tolerance, and the dtype/rounding facts a signature can't express — plus a generic harness (`aie.utils.kernel_harness`) that can build, run, and judge any contract-bearing kernel without a hand-written design. Programming examples and the kernel test suite now build kernels through this system instead of hand-rolling their own designs and tolerances, so kernel numerics live in one place instead of being restated (and drifting) across tests, benchmarks, and examples.                                                                                      
                                                                                                                        
Alongside that:                                                                                                       
- `aie.utils.benchmark` gained real statistics (median/MAD/p95/CoV) instead of just avg/min/max.                      
- `aie.utils.verify` gives kernels a small tolerance vocabulary (exact, bf16 ULPs, relative, LSB) instead of one blanket `rtol=0.128` for everything.                                                                                  
- Two new CI workflows: `staticKernelChecks.yml` compiles the whole kernel registry through Peano (nightly, and on PRs touching kernels/toolchain) and flags dropped pragmas, compile failures, and program-memory regressions as PR annotations; `benchmarkKernels.yml` runs correctness + perf on self-hosted NPU runners nightly.

Getting the harness to actually drive hardware end-to-end surfaced several real, pre-existing kernel bugs, now fixed:
- `conv2dk1_skip_init`'s uint8 entry point was an empty function (its body was commented out) — it silently returned whatever was already in the output buffer.
- `filter2d` read its carry pixel *after* the shuffle that consumes it instead of before, so every 32-pixel tile boundary blended against the wrong left neighbor.
- `rope`'s rotation rounded to bf16 twice (once per product) instead of once in the accumulator, which is fine for most inputs but lossy when two similar-magnitude products cancel.
- `conv2dk3`'s int8 build never actually compiled under Peano (a vector declaration was commented out), and `conv2dk14` sized its output tile 8x too large.

Docs cover the contract/harness system, the dtype/rounding spec, and an add-a-kernel checklist.

## Known issue (aie2 only, tracked separately)
- The new e2e suite surfaced 3 pre-existing, previously-unverified correctness issues on aie2 (Phoenix/npu1) hardware: `bf16_exp`, `softmax`, and `rgba2hue`. None of the three kernels' `.cc` sources were touched by this PR — this is the first time they've run against real aie2 hardware. All three pass cleanly (29/29 cases) on aie2p (Strix/npu2). Needs someone with aie2/npu1 access to investigate.

## Checklist


.github/actionlint.yaml
```diff
@@ -8,6 +8,7 @@ self-hosted-runner:
   labels:
     - aie2-4col      # NPU1, 4 columns
     - aie2p-8col     # NPU2, 8 columns
+    - bench          # runners reserved for benchmark timing
```
hunhoffe, line 11: TODO: need to mark runners to nightly benchmarks.

aie_kernels/aie2/conv2dk3.cc
hunhoffe, line 1009: TODO: I see two comments here that appear to be be about dropped pragmas
```
conv2dk3/2048x8/uint8: pragma dropped by the compiler
Check warning: conv2dk3/2048x8/uint8: pragma dropped by the compiler
loop not unrolled: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]
```
but when I click on more details I see what appears to be an unrelated error?
```

Run python -m pytest benchmarks/kernels/test_designs_compile.py -q -x
F
=================================== FAILURES ===================================
_________ test_design_compiles_through_cdo[passthrough/2048x16/int32] __________

case = Case(factory='passthrough', kwargs={'tile_size': 2048}, calls=16, shape=None, scalars=(), params=(), tag='', perf=True, correctness=True, data_cases=None, arch=None)
tmp_path = PosixPath('/tmp/pytest-of-runner/pytest-1/test_design_compiles_through_c0')

    @pytest.mark.parametrize("case", list(_cases()))
    def test_design_compiles_through_cdo(case, tmp_path):
        with _device_for(_ARCH):
            fn = case.fn()
            inputs = registry.inputs_for(case, "random", np.random.default_rng(0))
            design = kh.design(
                getattr(_kernels, case.factory),
                **case.harness_opts(),
                params=kh.param_values(fn, inputs),
                **case.kwargs,
            )
            try:
>               design.compile(
                    xclbin_path=str(tmp_path / "d.xclbin"),
                    inst_path=str(tmp_path / "d.bin"),
                )

benchmarks/kernels/test_designs_compile.py:65: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/mlir_aie/python/aie/utils/callabledesign.py:470: in compile
    return self.compilable.compile(
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/mlir_aie/python/aie/utils/compile/jit/compilabledesign.py:394: in compile
    compile_external_kernel(func, kernel_dir, target_arch)
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/mlir_aie/python/aie/utils/compile/utils.py:749: in compile_external_kernel
    renamed = _prefix_symbols_in_object(output_file, func._symbol_prefix)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/mlir_aie/python/aie/utils/compile/utils.py:626: in _prefix_symbols_in_object
    renamed = [s for s in _defined_symbols(object_path) if not s.startswith(pre)]
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

object_path = '/tmp/pytest-of-runner/pytest-1/test_design_compiles_through_c0/d.prj/passThroughLine_3eee3302.o'

    def _defined_symbols(object_path: str) -> list[str]:
        """Return the external symbols an object file defines, via llvm-nm.
    
        ``--defined-only`` leaves the object's *references* (memcpy, the LUT
        helpers) alone -- renaming those would break the link -- and
        ``--extern-only`` skips file-local labels, which cannot collide.
        """
>       nm = config.nm_path()
             ^^^^^^^^^^^^^^
E       AttributeError: module 'aie.utils.config' has no attribute 'nm_path'

/opt/hostedtoolcache/Python/3.12.14/x64/lib/python3.12/site-packages/mlir_aie/python/aie/utils/compile/utils.py:594: AttributeError
=========================== short test summary info ============================
FAILED benchmarks/kernels/test_designs_compile.py::test_design_compiles_through_cdo[passthrough/2048x16/int32] - AttributeError: module 'aie.utils.config' has no attribute 'nm_path'
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
1 failed in 3.32s

```

hunhoffe, line 1201: Same note about the double comments.

hunhoffe, line 1337: Same note about the duplicated comments.

aie_kernels/generic/transpose.cc
```diff
@@ -18,55 +15,67 @@
 #error Please specify matrix sizes m, n at kernel compile time using e.g., -DDIM_m=32 -DDIM_n=32.
 #endif
 
+// Element type. The transpose only moves bytes, so the unsigned integer of
```
hunhoffe: Are these defines something generalizable beyond this file?

aie_kernels/generic/transpose.cc
```diff
@@ -105,74 +114,81 @@ void transpose_4x4(bfloat16 *__restrict__ in_ptr,
 }
 
 /* As transpose_4, but transposes 8x8 submatrices. */
-void transpose_8x8(bfloat16 *__restrict__ in_ptr,
-                   bfloat16 *__restrict__ out_ptr) {
+void transpose_8x8(DTYPE *__restrict__ in_ptr, DTYPE *__restrict__ out_ptr) {
```
hunhoffe: This seems... very long and manual. Is there a better way to write this kernel?

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+
+import json
+import os
+import re
+import subprocess
+from importlib.metadata import version as pkg_version
+from pathlib import Path
+
+
+class PreflightError(RuntimeError):
```
hunhoffe: Why is this needed?

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+    pass
+
+
+# Mirrors NPU2_REGEX in utils/iron_setup.py (a setup script, not part of the
```
hunhoffe: I don't want more than one copy of this, if possible. I'd like to avoid duplication of string-based regex.

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+# wheel): the NPU2 family, whose kernels build for aie2p.
+_NPU2_NAMES = re.compile(
+    r"NPU Strix|NPU Strix Halo|NPU Krackan|NPU Gorgon Point|RyzenAI-npu[456]",
+    re.IGNORECASE,
+)
+
+
+def npu_arch(npu_name: str) -> str:
```
hunhoffe: This seems also like duplicated functionality?

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+    """``aie2p`` for the NPU2 family (Strix, Krackan, Gorgon Point), else ``aie2``."""
+    return "aie2p" if _NPU2_NAMES.search(npu_name) else "aie2"
+
+
+def _run(cmd: list[str]) -> str:
```
hunhoffe: This seems unnecesary -- like it reduces readability more than it promotes code reuse. If we added logging, that might justify it.

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+    return subprocess.run(cmd, check=True, capture_output=True, text=True).stdout
+
+
+def _ver(pkg: str) -> str:
+    try:
```
hunhoffe: Uhm. Why is this needed?

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+        return pkg_version(pkg)
+    except Exception:  # noqa: BLE001
+        return "unknown"
+
+
+def provenance(pmode: str | None = None) -> str:
```
hunhoffe: Should this -- and all other helpers -- be in python utils or python helpers? I'd like to promote reuse whenever possible.

benchmarks/kernels/_util.py
```diff
@@ -0,0 +1,94 @@
...
+    return r
+
+
+def write_rows(path: str | Path, rows: list[dict]) -> None:
+    if not rows:
+        raise ValueError("refusing to write an empty benchmark file")
+    Path(path).write_text(json.dumps(rows, indent=1))
+
+
+def preflight(required_pmode: str) -> dict:
```
hunhoffe: I think there are helpers for interactie with the runtime in a more generic fashion that is less XRT specific. We need to support XRT/HSA/HRX backends and we want to avoid duplicating any existing functionality.

benchmarks/kernels/harness.py
```diff
@@ -0,0 +1,137 @@
+# benchmarks/kernels/harness.py -*- Python -*-
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+"""Measure one registry case: correctness first, then wall time, cycles, size.
+
+The design, inputs, reference and verdict all come from
+``aie.utils.kernel_harness`` and the kernel's contract; this module only adds
```
hunhoffe: Should these functions go in aie.utils instead? Benchmarking tools might be useful for those other than just the CI.

We also definitely DON"T want a top-level benchmarks folder. All of this should fit into either: python, aie_kernels, or .github CI configuration, or masybe scripts.

benchmarks/kernels/README.md
```diff
@@ -0,0 +1,156 @@
+<!-- Copyright (C) 2026 Advanced Micro Devices, Inc.
```
hunhoffe: I want to avoid adding too many documentation files, as they are hard to manage. Does this documentation really belong here? Should it instead be in the website where the results are displayed, or similar?

benchmarks/kernels/registry.py
```diff
@@ -0,0 +1,558 @@
+# benchmarks/kernels/registry.py -*- Python -*-
```
hunhoffe: This seems like a lot of machinery for what pytest probably already provides. Shouldnt we include the cases in tests packaged with the kernels and/or wrappers and/or in the top-level test directory?

benchmarks/kernels/run.py
```diff
@@ -0,0 +1,183 @@
+#!/usr/bin/env python3
```
hunhoffe: What does this actually measure? Is it usable? Mabye this belongs in scripts or python or elsewhere -- it would be neat for a user to be able to manually trigger a benchmark over all or a set of kernels/cases for the desired set of information/metadata.

benchmarks/kernels/STACK_CHECKS.md
```diff
@@ -0,0 +1,93 @@
+<!-- Copyright (C) 2026 Advanced Micro Devices, Inc.
```
hunhoffe: Uhm why is this here? I don't want to add extra doc files scattered all over.

benchmarks/kernels/test_correctness.py
```diff
@@ -0,0 +1,77 @@
+# benchmarks/kernels/test_correctness.py -*- Python -*-
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+"""Full correctness sweep of the registry: every case x data case x seed.
+
+    pytest benchmarks/kernels/test_correctness.py -v            # everything
```
hunhoffe: This might need to be documented on the website pages about the kernels and the kernel registry.

benchmarks/kernels/test_correctness.py
```diff
@@ -0,0 +1,77 @@
...
+from aie.iron import kernels
+from aie.iron.kernels._common import _detect_arch
+from aie.utils import kernel_harness as kh
+
+from . import registry
+
+# --data-cases / --seeds are registered in conftest.py.
+
+
+def _params(config):
```
hunhoffe: Is there some of this we could take from the structure of amd/IRON? It would be nice to be somewhat aligned, if there aren't downsides to that alignment.

benchmarks/kernels/test_designs_compile.py
```diff
@@ -0,0 +1,72 @@
+# benchmarks/kernels/test_designs_compile.py -*- Python -*-
```
hunhoffe: To test the artifact cache, etc and avoid keeping a bunch of artifacts in the cache, should we compile and run one-by-one instead?

benchmarks/kernels/test_run_driver.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+import types
+
+import pytest
+
+_kernels = pytest.importorskip(
```
hunhoffe: This is not clean. Let's find a cleaner way to manage imports. I think it's ok for benchmarking suite to assume the package exists.

benchmarks/kernels/test_run_driver.py
```diff
@@ -0,0 +1,208 @@
...
+from . import (
+    _util,  # noqa: E402
+    registry,  # noqa: E402
+)
+from . import run as run_mod  # noqa: E402
+from .harness import Measurement  # noqa: E402
+
+
+@pytest.fixture
+def env(monkeypatch, tmp_path):
```
hunhoffe: I am generally very against monkeypatching. What do they mean about canary cycles? It should no-op if there's not device, because there's no point, right? Or tests should just be skipped? What makes the most sense?

benchmarks/kernels/test_run_driver.py
```diff
@@ -0,0 +1,208 @@
...
+
+def test_arch_gated_cases_are_skipped_on_the_other_device(env, monkeypatch):
+    # The fake preflight reports a Strix (aie2p); an aie2-only case is skipped.
+    aie2_only = registry.Case("passthrough", calls=2, arch="aie2")
+    monkeypatch.setattr(registry, "CASES", registry.CASES + [aie2_only])
+    assert _run(env, "--only", "^passthrough") == 0
+    assert aie2_only.name not in env.calls
+
+
+_XRT_SMI_EXAMINE = """
```
hunhoffe: We don't need this level of detail on XRT. We need to use the higher-level host runtime wrappers.

benchmarks/static/run.py
```diff
@@ -0,0 +1,235 @@
+#!/usr/bin/env python3
```
hunhoffe: This kind of thing should definitely go in python/iron/kernels, I think?

benchmarks/static/STATIC_CHECKS.md
```diff
@@ -0,0 +1,137 @@
+<!-- Copyright (C) 2026 Advanced Micro Devices, Inc.
```
hunhoffe: Same comment I've had elsewhere about documentation.

benchmarks/static/test_compile_command.py
```diff
@@ -0,0 +1,112 @@
...
+import types
+
+import pytest
+
+pytest.importorskip("aie.iron.kernels", reason="run.py imports the aie package")
+
+from . import run as run_mod  # noqa: E402
+
+WHEEL = os.path.join(os.sep, "site", "mlir_aie")
+CHECKOUT = os.path.join(os.sep, "work", "mlir-aie")
```
hunhoffe: I really don't want path manipulation. It generally means you are doing something you shouldn't e.g., if this needs aie, it should indeed be in the aie package....

benchmarks/static/test_remarks.py
```diff
@@ -0,0 +1,237 @@
+# benchmarks/static/test_remarks.py -*- Python -*-
+#
+# Copyright (C) 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+"""Host-only tests of the remark parser.
+
+The sample below mirrors records captured from a real llvm-aie
+22.0.0.2026090201 build of aie_kernels (STATIC_CHECKS.md documents the
```
hunhoffe: Why do we need a sample tied to a specifc version? This seems impossible to maintain.

docs/api/kernels.md
```diff
@@ -37,8 +43,81 @@ kernel sources these wrap, see [C++ AIE kernels](aie_kernels.md).
...
+        - sample_inputs
+        - expected
+        - input_limit
+        - upload
+        - run
+        - check
+        - judge
+        - cycles_per_call
+
+`aie.utils.bfp` is the host side of the block-floating-point kernels: the
```
hunhoffe, line 154: Do we share these utils, etc. in the programming exmaples/default runtime helpers as much as possible?

programming_examples/basic/matrix_multiplication/cascade/cascade.py
```diff
@@ -80,6 +80,9 @@ def cascade(
     zero_kernel = cascade_kernel.zero
 
     r, s, t = cascade_kernel.mac_dims
```
hunhoffe, line 82: Seems a little weird that you fetch dims different than r, s, t above?

programming_examples/basic/matrix_multiplication/single_core/single_core.py
```diff
@@ -276,11 +291,14 @@ def _run_and_verify(opts):
...
     actual = C_t.numpy().reshape(opts.M, opts.N)
 
+    rtol, atol = _tolerance(opts)
     assert_close_with_benchmark(
         actual,
         expected,
         bench=bench,
         ops=2.0 * opts.M * opts.K * opts.N,
+        float_rtol=rtol,
+        float_atol=atol,
```
hunhoffe: Is this best practice for handling tolerance? It seems kinda hard to read/parse --  like maybe you should give explicitly the data types and not the opts?

programming_examples/basic/transposes/transposes.py
```diff
@@ -42,7 +43,7 @@
 import aie.iron as iron
 import numpy as np
 from aie.helpers.taplib import TensorAccessPattern, TensorTiler2D
-from aie.iron import CompileTime, In, ObjectFifo, Out, Program, Runtime, Worker
+from aie.iron import CompileTime, In, ObjectFifo, Out, Program, Runtime, Worker, kernels
```
hunhoffe: seems weird this is the only lowercase thing? Like maybe you should have 
```
from aie.iron.kernels import kernels
```

programming_examples/ml/block_datatypes/matrix_multiplication/single_core_mixed/single_core.py
```diff
@@ -28,6 +26,7 @@
     StreamDims,
     TaskGroup,
     Worker,
+    kernels,
```
hunhoffe: Same import note I made before.

programming_examples/ml/block_datatypes/matrix_multiplication/single_core_no_tiling/single_core.py
```diff
@@ -10,31 +10,26 @@
...
 from aie.iron import (
     CompileTime,
-    ExternalFunction,
     In,
     ObjectFifo,
     Out,
     Program,
     Runtime,
     Worker,
+    kernels,
```
hunhoffe: same import note I had before

programming_examples/ml/block_datatypes/matrix_multiplication/single_core_no_tiling_mixed/single_core.py
```diff
@@ -10,21 +10,20 @@
...
     CompileTime,
-    ExternalFunction,
     In,
     ObjectFifo,
     Out,
     Program,
     Runtime,
     StreamDims,
     Worker,
+    kernels,
```
hunhoffe: Same import note

programming_examples/ml/dwconv1d/dwconv1d.py
```diff
@@ -25,40 +25,18 @@
...
-            chunk_out_ty,
-            np.int32,  # pyright: ignore[reportArgumentType]
-        ],
-        include_dirs=[config.cxx_header_path()],
-        compile_flags=[
-            f"-DDWCONV_K={kernel_size}",
-            f"-DDWCONV_BIAS={int(bias)}",
-        ],
-    )
+_TAIL_SLACK = kernels.DWCONV1D_TAIL  # the kernel's fixed aligned-load window
```
hunhoffe, line 40: Why does this need to be defined here?

programming_examples/ml/norm/norm.py
```diff
@@ -34,26 +34,22 @@
...
-_KERNEL_SPEC = {
-    "rms": ("rms_norm", _KERNEL_DIR / "rms_norm.cc"),
-    "layer": ("layer_norm", _KERNEL_DIR / "layer_norm.cc"),
-    "layer_f32": ("layer_norm_f32", _KERNEL_DIR / "layer_norm.cc"),
-    "layer_affine_cast": ("layer_norm_affine_cast", _KERNEL_DIR / "layer_norm.cc"),
+# The library factories behind each op (aie.iron.kernels.transformer).
+_NORM_FACTORY = {
+    "rms": kernels.rms_norm,
+    "layer": kernels.layer_norm,
+    "layer_f32": kernels.layer_norm_f32,
```
hunhoffe, line 53: does it make sense that this is a different kernel and not the same kernel, since they only vary by dtype?

programming_examples/ml/scale_shift/scale_shift.py
```diff
@@ -17,7 +17,6 @@
 """
 
 import argparse
-from pathlib import Path
 
 import aie.iron as iron
```
hunhoffe, line 21: These files all look very similar.... is there a way to collapse them further? I like clarity but I worry about boilerplate.

programming_examples/vision/color_detect/color_detect.py
```diff
@@ -241,54 +241,18 @@ def _design_for(opts):
 
 
 def _rgba2hue_ref(rgba_uint8):
```
hunhoffe, line 244: shouldn't the refs go with the kernels, and not with the example??

programming_examples/vision/edge_detect/edge_detect.py
```diff
@@ -256,16 +256,9 @@ def _compile_kwargs(opts):
...
-    g = rgba[..., 1].astype(np.int32)
-    b = rgba[..., 2].astype(np.int32)
-    wt_r = int(round(0.299 * (1 << 15)))  # 9798
-    wt_g = int(round(0.587 * (1 << 15)))  # 19235
-    wt_b = int(round(0.114 * (1 << 15)))  # 3736
-    y = (wt_r * r + wt_g * g + wt_b * b + (1 << 14)) >> 15
-    return np.clip(y, 0, 255).astype(np.uint8)
+    """``kernels.rgba2gray_ref`` (the kernel's Q0.15 luma), one image at a time."""
+    del height, width  # the reference is per pixel
+    return kernels.rgba2gray_ref(rgba_uint8.reshape(-1))
```
hunhoffe, line 261: this seems... silly to wrap in a function??

python/iron/kernels/_common.py
```diff
@@ -7,13 +7,177 @@
...
 _log = logging.getLogger(__name__)
 
+# Argument roles a kernel contract may assign, one per entry of ``arg_types``:
+#   "in"     a tile streamed in through an ObjectFifo, one element per call
+#   "out"    the tile the kernel writes, one element per call
+#   "param"  a small buffer filled once and held for the whole run (e.g. the
+#            1-element factor of ``scale``)
+#   "count"  the trailing element count the C++ takes at runtime
+#   "scalar" a runtime scalar constant (``leaky_relu`` alpha, ``axpy`` a)
+ROLES = ("in", "out", "inout", "param", "count", "scalar")
```
hunhoffe, line 27: how this this relate to the JIT labels that already exists? Or existing types? Why do we need all these "words"?? I feel like object or enums or traits or similar is more pythonic instead of strings. 

How does numpy/existing numerical calculations handle this sort of thing? Can we use existing library functionality?

python/iron/kernels/_common.py
```diff
@@ -7,13 +7,177 @@
...
+# What a float kernel does with NaN / inf inputs: "propagate" (the IEEE
+# result numpy computes) or "unspecified" (out of contract; not sampled).
+NONFINITE = ("propagate", "unspecified")
+# What a float kernel does with subnormal inputs: "preserve" (IEEE),
+# "flush" (treated as zero, and so is the reference) or "unspecified".
+SUBNORMALS = ("preserve", "flush", "unspecified")
+
+
+@dataclass(frozen=True)
+class KernelContract:
```
hunhoffe, line 60: this seems pretty complex. can we simplify without losing functionality?

python/iron/kernels/activation.py
```diff
@@ -23,9 +23,11 @@
 
 import numpy as np
 from aie.iron.kernel import ExternalFunction
+from aie.utils.verify import Tolerance
```
hunhoffe, line 26: shouldn't numpy or other libraries have tolerance measurements? I want to make sure we're not reinventing the wheel.

python/iron/kernels/activation.py
```diff
@@ -36,18 +38,63 @@
 
 _LUT_FIXED_TILE = 1024
 
+# The LUT-approximated kernels document rtol=0.128 (the canonical C++ testbench
+# default) on their *_ref functions below; the absolute floor and the mismatch
+# budget are what test/python/npu/test_kernels_e2e.py measured for tanh and
+# sigmoid on device.
+_LUT_TOLERANCE = Tolerance.relative(
```
hunhoffe, line 45: Can we be more precise? Shouldn't it depend on more than just a hardcoded value? Is this mathematically sound?

python/iron/kernels/activation.py
```diff
@@ -300,6 +423,20 @@ def leaky_relu_ref(x, alpha=0.01):
...
 
+def swiglu_ref(x, w1, w2):
+    """Numpy reference for [`swiglu`][iron.kernels.activation.swiglu]: ``(x * w1) * silu(x * w2)``.
+
+    ``swiglu.cc`` forms the two products in bf16, then ``silu`` of the second
+    through the tanh LUT (``0.5 * (1 + tanh(z / 2))``). The reference rounds
+    the two products to bf16 as the kernel does and computes the rest in
+    float32; LUT-approximation territory, pair with ``rtol=0.128``.
+    """
+    xf = x.astype(np.float32)
```
hunhoffe, line 449: are we using numpy to the fullest extent in these examples? And Tensor objects?

python/iron/kernels/conv.py
```diff
@@ -20,6 +27,599 @@ def _i32s(n: int) -> list:
     return [np.int32] * n
 
 
+# The conv kernels end in `(sum + 2**(scale-1)) >> scale` saturated to uint8
+# (a fused ReLU); the vector paths do the shift with the core's srs rounding
+# mode, which the kernels do not fix, so the references allow one LSB.
+_CONV_TOLERANCE = Tolerance.lsb(
```
hunhoffe: Uh why aren't they modeled? Should they be modelled?

python/iron/kernels/datamovement.py
```diff
@@ -14,9 +14,80 @@
...
+# test/python/npu/test_kernels_e2e.py with this tolerance.
+_BF16_ROUNDTRIP = Tolerance.relative(
+    0.03,
+    0.05,
+    max_mismatch_frac=0.02,
+    note="fp32 compute, one bf16 rounding; tolerance measured by test_kernels_e2e",
+)
+
+
+def axpy_ref(x, y, a):
```
hunhoffe, line 36: Do you think the Kernel object should have some notion of the CPU reference function? Because it seems weird to have these functions but not have them attached to their corresponding functions in an object-oriented, hierarchical/inheritance kind of way.

python/iron/kernels/datamovement.py
```diff
@@ -127,34 +215,93 @@ def expand(tile_size: int = 1024, group_size: int = 32) -> ExternalFunction:
...
+            subnormals="preserve",
+            tolerance=Tolerance.exact(note="data movement only; lossless"),
+            ops_per_call=0,
+        ),
     )
+
+
+# Supported dtype combinations, as data: the registry and the contract test
+# enumerate these instead of restating them.
+_declare_dtypes(
```
hunhoffe, line 323: seems weird this is module level and not maybe an object attribute or something?

python/iron/kernel.py
```diff
@@ -322,15 +343,48 @@ class ExternalFunction(Kernel):
...
+    # the staged package as two module trees, so naming the class here would
+    # make the factories' own KernelContract a different type.
+    contract: Any = None
     mac_dims: tuple
+    dims: tuple
+    stream_dims: dict
+    b_col_maj: bool
+    c_col_maj: bool
+    a_dims_from_stream: object
     zero: "Kernel"
```
hunhoffe, line 360: why are there strings everywhere?

python/iron/kernel.py
```diff
@@ -8,6 +8,7 @@
 import hashlib
 import logging
 from pathlib import Path
+from typing import Any
 
 import numpy as np
```
hunhoffe, line 13: Can we use introspection/discovery on the source tree instead of hardcoding to populate the kernel registry?

python/utils/compile/utils.py
```diff
@@ -226,62 +226,27 @@ def _make_ir_inlinable(ir_path: str, symbol_name: str) -> None:
...
-            the Chess wrapper's include search.
-        inline (bool): When True, emit inlinable LLVM IR instead of an object.
-        symbol_name (str, optional): Required when ``inline`` is True; names the
-            LLVM ``define`` for the kernel that ``_make_ir_inlinable`` rewrites
-            to ``alwaysinline`` / ``linkonce_odr``. Must match the symbol as it
-            appears in the freshly emitted IR.
+) -> list[str]:
+    """Return the compiler command line that ``compile_cxx_core_function`` runs.
+
+    Factored out so tools that want the compiler's *diagnostics* rather than
```
hunhoffe, line 241: this comment seems historical

python/utils/benchmark.py
```diff
@@ -17,6 +24,47 @@ class Stats:
     avg_us: float
     min_us: float
     max_us: float
+    median_us: float = 0.0
```
hunhoffe, line 31: Why do we need a benchmark test? I thought there were already python utils for a test case. Couldn't those just be extended and then existing utilities such as pytest can be used to create the benchmarks/test suites?

python/utils/bfp.py
```diff
@@ -0,0 +1,130 @@
...
+import numpy as np
+
+BLOCK = 8  # values per block
+BLOCK_BYTES = 9  # one shared exponent plus one mantissa per value
+_MANTISSA_SHIFT = 23 - 7 + 1  # keep 7 magnitude bits of a float32 mantissa
+
+__all__ = ["BLOCK", "BLOCK_BYTES", "encode", "decode", "quantize", "shuffle"]
+
+
+def encode(x) -> np.ndarray:
```
hunhoffe, line 37: How does this relate to existing bfp helpers? I think this is great, and the right place for it, but I want to make sure there aren't duplicates elsewhere in the code base.

python/utils/config.py
```diff
@@ -101,6 +101,27 @@ def objcopy_path():
     )
 
 
+def nm_path():
+    """Return the llvm-nm used to list the symbols a compiled object defines.
+
+    Same sourcing rule as :func:`objcopy_path`: AIE objects use the AIEngine
+    ELF e_machine, which GNU binutils nm cannot parse.
+    """
+    bundled_nm = os.path.join(root_path(), "bin", _executable_name("llvm-nm"))
```
hunhoffe, line 112: will this work on windows too?

python/utils/kernel_harness/__init__.py
```diff
@@ -0,0 +1,1139 @@
+# kernel_harness.py -*- Python -*-
```
hunhoffe, line 1: It seems crazy for this file to exist at all. Most of the functions are 1-2 line wrappers of existing functions. What purpose does it serve? Cant' we use or extend existing test case helpers that are used throughout the programming examples?

python/utils/verify.py
```diff
@@ -203,3 +205,285 @@ def assert_close_with_benchmark(
...
+            return cls.exact(note="default: integer output")
+        if dt == np.dtype(np.float32) or dt == np.dtype(np.float64):
+            return cls.relative(1e-4, note="default: float32 output")
+        if dt == np.dtype(np.float16):
+            return cls.relative(1e-2, note="default: float16 output")
+        return cls.relative(_DEFAULT_RTOL, note="default: canonical bf16/LUT rtol")
+
+
+@dataclass
+class Verdict:
```
hunhoffe, line 329: this seems extremely unnecessary.

python/utils/verify.py
```diff
@@ -203,3 +205,285 @@ def assert_close_with_benchmark(
...
+    max_abs_err: float
+    max_ulp_err: int | None
+    first_bad_index: int | None
+    detail: str
+
+    def __bool__(self) -> bool:
+        return self.ok
+
+
+def bf16_ulp_distance(a, b) -> np.ndarray:
```
hunhoffe, line 344: shouldn't this go in utils? how does this fit in with existing tolerance/verifciation?

test/python/npu/test_kernels_e2e.py
```diff
@@ -9,406 +9,220 @@
...
-    yt = iron.zeros(size, dtype=bfloat16, device="npu")
-
-    _leaky_relu_design(xt, yt, size=size, alpha=alpha)
-
-    expected = kernels.leaky_relu_ref(x, alpha=alpha)
-    frac = _bf16_close(yt.numpy(), expected)
-    assert frac < 0.02, f"leaky_relu mismatch fraction {frac:.4f}"
+# (factory name, factory kwargs, harness options). ``aie2p`` marks kernels
+# whose source exists only for that arch.
+CASES = {
```
hunhoffe: can't we use a pytest test suite or something? amd/IRON seems to use pytest effectively...

hunhoffe (commented):
Feedback from @atassis so claude can see:

Read #25. What I'd push on: KernelContract describes what a kernel computes — roles, reference, tolerance, ops — but nothing about what it needs from the machine, and that's where the bugs I've paid for live. Rounding mode is the cleanest case. It's a core register, aie_api never initialises it, and the documented default is floor. In aie_kernels/aie2p today 8 files narrow an accumulator with to_vector<bfloat16> and only cast_f32_bf16.cc and dwconv1d.cc set the mode first; cast_f32_bf16's own comment calls it "one sticky register shared by every kernel on this core". Floor vs nearest-even cost me 1.3x accuracy on a real encoder and sat well inside rtol=0.128, so no tolerance gate saw it — and a suite that builds one Worker per case in a fresh context always sees the reset default, so it stays green while a fused pipeline degrades.
Closer to PR 3: I'd gate on the weight-stream loop actually being a zero-overhead loop, not just on II. A scalar unpack I once put in an inner loop turned a 7-bundle ZOL into a 181-bundle software loop, 79 → 930 ms/token, with correct output and a green build.
What I'm chasing otherwise: what it takes to have a host/device buffer relationship with no cache maintenance and no dirty state at all — correctness that doesn't depend on anyone remembering to synchronise. On x86 the maintenance was never expressible. CLFLUSH/CLFLUSHOPT always write back and invalidate, and the only invalidate-only instruction is INVD, so the FROM_DEVICE half of direction has no instruction at line granularity. But the hazard is really about ownership — that flush must write the host's dirty lines back over the device's output, and if the host never wrote the buffer there are none. The cost half turns out to be one bool: drm_gem_shmem_object.map_wc, which amdxdna hardcodes false and is the only drm/accel driver that does. Flipped as a single-bool diff I get 0/20 stale in both directions against 20/20 on write-back. Not a global flip though — host reads off a WC mapping run 0.06 GB/s at every size, ~120x under the writes, so it wants to be per-BO.
Found it because our own fused path leaves the scratch arena synced in neither direction, and every weight and both KV caches live there. The decode going nondeterministic read like a race downstream of the first bad weight, not like a missing flush.

python/utils/kernel_harness/bench.py
```diff
@@ -0,0 +1,410 @@
...
+
+    ``CallableDesign.compile`` with explicit ``xclbin_path`` and ``inst_path``
+    bypasses the on-disk cache by contract and keeps its intermediates in
+    ``<stem>.prj/`` next to the xclbin, so nothing about the cache layout has
+    to be guessed or deleted; the cached build the timed runs use is untouched.
+
+    Returns (seconds, xclbin_bytes, insts_bytes, sum_core_elf_bytes).
+    """
+    build = workdir / "compile"
+    build.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
hunhoffe: But how will we clean up artifacts between runs, etc.?

PR #3703, opened 2026-09-09
aiecc: list --get-<name> output selectors in --help [LOW]

atassis:
## Description

`aiecc --help` and `--help-hidden` list exactly one `--get-*` flag: the generic `--get=<name>`. None of the 14 named shorthands (`--get-xclbin`, `--get-full-elf`, `--get-scratchpad-parameters`, etc.) appear in either. The only place the full list is printed is the diagnostic for an unrecognized `--get-<name>`, i.e. you have to guess wrong to find out what exists. I spent significant time concluding `--get-scratchpad-parameters` did not exist for exactly this reason, before finding it in the source.

The cause: `applyOutputSelectorFlags()` in `CommandLineOptions.h` rewrites `--get-<name>` into the matching selector's bool and drops the token from argv *before* `llvm::cl::ParseCommandLineOptions` runs. None of the 14 shorthands is ever registered as a `cl::opt`, so `--help`/`--help-hidden` (which only enumerate registered options) never see them, even though `outputSelectors()` already holds the full `{name, artifact}` table and the unknown-selector diagnostic already prints it in full.

This factors that per-selector line format into `printOutputSelectorTable()`, used both by the existing diagnostic (unchanged output) and by a new `cl::extrahelp` block appended after the normal `--help`/`--help-hidden` output, built once from the same `outputSelectors()` table. No selector-resolution behavior changes; this only adds a static list to `--help`.

Added `test/aiecc/help_output_selectors.mlir`, checking that `--help` and `--help-hidden` both contain the new `OUTPUT SELECTORS:` block and a few representative entries. I confirmed this test fails against the current `aiecc` (no such block) and pass/fail wiring is otherwise identical to the neighboring `unknown_args_rejected.mlir`/`cpp_backend_dir_missing.mlir` tests in the same directory.

Validation I could actually do: the `cl::extrahelp` + static-initialization-order pattern (the backing `std::string` must construct before the `extrahelp` that references it) compiles, links and runs correctly in an isolated program against this project's pinned LLVM `Support` library, producing the expected appended block. I did not get a full local `aiecc` build to green: my working copy's `third_party/aie-rt` vendored patch fails to apply against the submodule state in my sandbox, unrelated to this change (unmodified files elsewhere in the tree). So I have not run the new lit test against an actual patched `aiecc` binary; I'm flagging that gap rather than claiming otherwise.

## Checklist

Notes on the checklist: the new lit test is written and confirmed to fail pre-patch, but I have not built a patched `aiecc` locally to confirm it passes post-patch (see above), so I'm leaving that box unchecked rather than claiming it. No Python file is touched. `clang-tidy`'s enabled list currently covers `tools/aiecc/aiecc.cpp` but not `CommandLineOptions.h`, so there's no touched file under its scope. `clang-format` is 22.1.8 locally (this repo's CI pins 20.1.0; past history has one file differ between the two versions) and reports no changes needed on the touched lines.

### DIFF
--- a/test/aiecc/help_output_selectors.mlir
+++ b/test/aiecc/help_output_selectors.mlir
@@ -6,10 +6,9 @@
 //===----------------------------------------------------------------------===//
 
 // The `--get-<name>` output-selector shorthands are resolved by
-// applyOutputSelectorFlags() before llvm::cl ever parses argv, so none of them
-// are registered as a cl::opt and --help/--help-hidden used to omit the whole
-// family; the only place the list appeared was the "unknown output selector"
-// diagnostic for a typo'd flag. cl::extrahelp now appends it to both.
+// applyOutputSelectorFlags() before llvm::cl ever parses argv, so none of
+// them are registered as a cl::opt and would not otherwise appear in
+// --help/--help-hidden. cl::extrahelp appends the list to both.
 
 // RUN: aiecc --help 2>&1 | FileCheck %s
 // RUN: aiecc --help-hidden 2>&1 | FileCheck %s
--- a/test/aiecc/help_output_selectors.mlir
+++ b/test/aiecc/help_output_selectors.mlir
@@ -1,0 +1,47 @@
+//===- help_output_selectors.mlir ------------------------------*- MLIR -*-===//
+//
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// The `--get-<name>` output-selector shorthands are resolved by
+// applyOutputSelectorFlags() before llvm::cl ever parses argv, so none of them
+// are registered as a cl::opt and --help/--help-hidden used to omit the whole
+// family; the only place the list appeared was the "unknown output selector"
+// diagnostic for a typo'd flag. cl::extrahelp now appends it to both.
+
+// RUN: aiecc --help 2>&1 | FileCheck %s
+// RUN: aiecc --help-hidden 2>&1 | FileCheck %s
+
+// CHECK: OUTPUT SELECTORS:
+// CHECK-DAG: --get-scratchpad-parameters  (params.txt)
+// CHECK-DAG: --get-xclbin  (aie.xclbin)
+// CHECK-DAG: --get-full-elf  (aie.elf)
+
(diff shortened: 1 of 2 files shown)


test/aiecc/help_output_selectors.mlir
```diff
@@ -0,0 +1,47 @@
...
+// Copyright (C) 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+// The `--get-<name>` output-selector shorthands are resolved by
+// applyOutputSelectorFlags() before llvm::cl ever parses argv, so none of them
+// are registered as a cl::opt and --help/--help-hidden used to omit the whole
+// family; the only place the list appeared was the "unknown output selector"
+// diagnostic for a typo'd flag. cl::extrahelp now appends it to both.
```
andrej: "used to emit", "was", "now appends"

tools/aiecc/CommandLineOptions.h
```diff
@@ -384,6 +384,35 @@ inline llvm::ArrayRef<OutputSelector> outputSelectors() {
   return table;
 }
 
+// Shared by the "unknown selector" diagnostic below and the --help text
+// (outputSelectorHelpText()), so the two can't drift apart.
```
andrej:
```suggestion
```

If you list everywhere a function gets called in a comment, that comment will be outdated in no time. I also don't see a benefit in that being in a comment.

tools/aiecc/CommandLineOptions.h
```diff
@@ -384,6 +384,35 @@ inline llvm::ArrayRef<OutputSelector> outputSelectors() {
...
+  for (const OutputSelector &s : outputSelectors())
+    os << "  --get-" << s.niceName << "  (" << s.edgeName << ")\n";
+}
+
+// `--get-<name>` is resolved by applyOutputSelectorFlags() below before
+// llvm::cl ever parses argv, so the shorthands are never registered as
+// cl::opt and are invisible to --help/--help-hidden; previously the only way
+// to see the list was to pass an unrecognized `--get-<name>` and read the
+// resulting error. cl::extrahelp appends this text after the normal --help
+// output instead.
```
andrej: "previously" ... "was"

andrej (changes requested):

PR #3711, opened 2026-09-09
Make the routing passes idempotent and lift routed designs to flows

andrej:
More minimal change factored out from #3421

The following two limitations (related to idempotency of passes) make debugging routing problems hard, and this PR fixes both:
1. Routing cannot run on its own output.
2. You cannot pass the output of --aie-find-flows back into the router (no round-tripping).

## Issue 1:
--aie-create-pathfinder-flows leaves state behind on every run. For certain broadcast flows, the original `aie.flow` op survives into the routed IR (which should be only switchboxes and wires). runOnPacketFlow keeps the `aie.packet_flow` ops it lowers. A second run lowers them again and emits a duplicate masterset on a port the first run claimed. 

This is annoying for debugging or incremental development. Say you want to route a flow, then use the produced IR and add more flows to route. The router will try to re-route the original flows that have already been routed and this will cause collisions in switchboxes, wires, etc. It would also just be nice to be able to pass an already-routed IR into aiecc without errors for debugging, without having to instruct aiecc to skip the routing phase.

## Issue 2:
--aie-find-flows leaves the switchbox configuration behind for the flows it lifts. This leaves the IR representing the flow in two forms, once as an `aie.flow` and once as a set of switchbox configurations. When you go to re-lower the flow, the router rejects it, because it collides with the existing switchbox configurations. This makes debugging hard.

It also does not recover all flows. It seeds a traversal only from the Core and DMA ports of a tile. A stream that enters the fabric elsewhere yields no flow. This means no partial flows are recovered, even though this would be useful. 

## Solution:

--aie-create-pathfinder-flows now erases the original `aie.flow` for broadcast flows as well after lowering to switchboxes, so there is only one set of IR ops representing the flow. It marks aie.packet_flow illegal after lowering, ensuring they don't remain, and removes each op it lowers. It keys the wires already present and emits only the missing ones. --aie-generate-column-control-overlay reads the is_ctrl_pkt_overlay marker and skips a device that already carries an overlay.

--aie-find-flows records the interconnect ops that implement each hop. It seeds a traversal from every interconnect source port that no upstream connection drives, which recovers PLIO entries, transit connections and runtime-steered packet routes. It reports a driven output port without an onward wire as an endpoint. It then erases the connects, packet rules, mastersets, amsels and wires it lifted, and drops each interconnect left empty. 

Two options control the recovery. keep-partial-flows=false restricts it to flows that begin and end on a core or DMA. remove-lifted=false keeps the configuration the pass lifted, which is the earlier behavior. The existing tests pass remove-lifted=false and check the same output as before.

The `aie.packet_flow` also gets a new `mask` attribute. This is needed to recover packet flows that might claim multiple IDs. We don't know exactly which IDs they were meant for after lowering, so a proper recovery now just expresses the packet flow with the mask that the lowering chose.

## Summary

> **With the new default options, a route is now either an aie.flow or switchbox configs with aie.connect ops, never both at the same time in IR.**

### DIFF
--- a/include/aie/Dialect/AIE/IR/AIEOps.td
+++ b/include/aie/Dialect/AIE/IR/AIEOps.td
@@ -826,6 +826,17 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
     so that they always get allocated with the same master, slave
     ports, arbiters and master selects (msel).
 
+    The optional attribute mask, written after the ID in the parentheses,
+    states the packet-rule mask of the flow. A slave port accepts a packet
+    when `incoming & mask == ID`, so the flow claims every ID that agrees
+    with ID on the bits the mask selects. Routing derives a mask from the
+    IDs it knows when the attribute is absent. ID must have no bit outside
+    the mask, or no packet can match.
+
+    A design may compute a packet header at run time, so which IDs reach a
+    port is not decidable here. State the mask to claim a set of IDs that
+    routing cannot see.
+
     Example:
     ```
       %01 = aie.tile(0, 1)
@@ -833,18 +844,34 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
         aie.packet_source<%01, "Core" : 0>
         aie.packet_dest<%01, "Core" : 0>
       }
+      // Claims 0x10 through 0x13: the mask ignores the low two bits.
+      aie.packet_flow(0x10 mask 0x1c) {
+        aie.packet_source<%01, "Core" : 0>
+        aie.packet_dest<%01, "Core" : 0>
+      }
     ```
   }];
 
-  let arguments = (
-    ins AIEI8Attr:$ID,
-        OptionalAttr<BoolAttr>:$keep_pkt_header,
-        OptionalAttr<BoolAttr>:$priority_route
-  );
+  let arguments = (ins AIEI8Attr:$ID, OptionalAttr<AIEI8Attr>:$mask,
+      OptionalAttr<BoolAttr>:$keep_pkt_header,
+      OptionalAttr<BoolAttr>:$priority_route);
   let regions = (region AnyRegion:$ports);
 
-  let assemblyFormat = [{ `(` $ID `)` regions attr-dict }];
+  let assemblyFormat = [{ `(` $ID (`mask` $mask^)? `)` regions attr-dict }];
   let hasVerifier = 1;
+
+  let builders = [OpBuilder<(ins "uint8_t":$ID,
+                                "mlir::BoolAttr":$keep_pkt_header,
+                                "mlir::BoolAttr":$priority_route),
+                            [{
+      build($_builder, $_state, ID, nullptr, keep_pkt_header, priority_route);
+    }]>,
+                  OpBuilder<(ins "mlir::IntegerAttr":$ID,
+                                "mlir::BoolAttr":$keep_pkt_header,
+                                "mlir::BoolAttr":$priority_route),
+                            [{
+      build($_builder, $_state, ID, nullptr, keep_pkt_header, priority_route);
+    }]>];
 
   let extraClassDeclaration = [{
     int IDInt() { return getID(); }
--- a/include/aie/Dialect/AIE/Transforms/AIEPasses.td
+++ b/include/aie/Dialect/AIE/Transforms/AIEPasses.td
@@ -325,14 +325,31 @@ def AIEFindFlows : Pass<"aie-find-flows", "DeviceOp"> {
     by aie.flow) or a packet-switched connection (represensted by
     aie.packetflow).  This pass is primarily used for testing automatic
     routing.
+
+    By default every configured connection is lifted, including partial flows
+    whose source or destination is a directional switchbox port rather than a
+    core or DMA (transit connections, PLIO/edge entries, and packet routes a
+    running design steers by packet id).  A physically routed design can
+    therefore drop its aie.switchbox configuration and recover it from the
+    lifted flows.  Set `keep-partial-flows=false` to restrict recovery to flows
+    that begin at a core/DMA and end on a core/DMA.
   }];
 
   let constructor = "xilinx::AIE::createAIEFindFlowsPass()";
-  let dependentDialects = [
-    "xilinx::AIE::AIEDialect",
+  let dependentDialects = ["xilinx::AIE::AIEDialect", ];
+
+  let options =
+      [Option<"clKeepPartialFlows", "keep-partial-flows", "bool",
+              /*default=*/"true",
+              "Also recover partial flows whose source or destination is a "
+              "directional/edge switchbox port rather than a core or DMA.">,
+       Option<
+           "clRemoveLifted", "remove-lifted", "bool", /*default=*/"true",
+           "Remove the interconnect configuration (connects, packet rules, "
+           "mastersets, amsels, wires) that was lifted into recovered flows, "
+           "leaving only configuration that could not be lifted.">,
   ];
 }
-
 
 def AIEVectorTransferLowering : Pass<"aie-vector-transfer-lowering", "DeviceOp"> {
   let summary = "Lower vector.transfer_read/write to vector.load/store for AIE";
--- a/lib/Dialect/AIE/IR/AIEDialect.cpp
+++ b/lib/Dialect/AIE/IR/AIEDialect.cpp
@@ -1974,12 +1974,18 @@ LogicalResult verifyNoDuplicateFlows(DeviceOp device) {
 }
 
 // Rejects two aie.packet_flow ops that carry the same ID between the same set
-// of sources and destinations. Declaring the flow twice makes the routing
-// problem look more congested than it is and gives the two copies conflicting
-// keep_pkt_header/priority_route settings when their attributes disagree.
+// of sources and destinations under the same mask. Declaring the flow twice
+// makes the routing problem look more congested than it is and gives the two
+// copies conflicting keep_pkt_header/priority_route settings when their
+// attributes disagree.
+//
+// The mask belongs in the key because it is half of what a flow claims: an ID
+// under two masks names two sets of packets, so the two flows carry different
+// traffic. A flow without the attribute claims its ID alone, which is what the
+// widest mask says, so both spellings key alike.
 LogicalResult verifyNoDuplicatePacketFlows(DeviceOp device) {
   using PacketFlowKey =
-      std::tuple<int, std::vector<PortKey>, std::vector<PortKey>>;
+      std::tuple<int, int, std::vector<PortKey>, std::vector<PortKey>>;
   std::map<PacketFlowKey, PacketFlowOp> packetFlowSeen;
   WalkResult result = device.walk([&](PacketFlowOp packetFlow) {
     Region &body = packetFlow.getPorts();
@@ -2011,12 +2017,17 @@ LogicalResult verifyNoDuplicatePacketFlows(DeviceOp device) {
       return WalkResult::advance();
     llvm::sort(sources);
     llvm::sort(dests);
+    int idBits =
+        llvm::Log2_32_Ceil(device.getTargetModel().getMaxPacketId() + 1);
+    int mask = packetFlow.getMask().value_or((1 << idBits) - 1);
     auto [it, inserted] = packetFlowSeen.try_emplace(
-        {packetFlow.IDInt(), std::move(sources), std::move(dests)}, packetFlow);
+        {packetFlow.IDInt(), mask, std::move(sources), std::move(dests)},
+        packetFlow);
     if (!inserted) {
       InFlightDiagnostic diag =
           packetFlow.emitOpError()
           << "duplicates an earlier packet flow; ID " << packetFlow.IDInt()
+          << " under mask 0x" << llvm::utohexstr(mask)
           << " is already declared between the same sources and destinations";
       diag.attachNote(it->second.getLoc()) << "the other packet flow is here";
       return WalkResult::interrupt();
@@ -2545,6 +2556,16 @@ LogicalResult PacketFlowOp::verify() {
     return emitOpError("must have at least one aie.packet_source");
   if (numDests < 1)
     return emitOpError("must have at least one aie.packet_dest");
+
+  // A slave port accepts a packet when `incoming & mask == ID`, so a bit set
+  // in ID and clear in the mask rejects every packet.
+  if (std::optional<uint8_t> mask = getMask()) {
+    uint8_t id = getID();
+    if ((id & *mask) != id)
+      return emitOpError("has ID 0x")
+             << llvm::utohexstr(id) << " outside mask 0x"
+             << llvm::utohexstr(*mask) << ", which no packet can match";
+  }
 
   return success();
 }
--- a/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
+++ b/lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <cstdint>
 #include <optional>
+#include <set>
 
 using namespace mlir;
 using namespace xilinx;
@@ -100,9 +101,16 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
     TileID srcSbId = {srcCoords.col, srcCoords.row};
     PathEndPoint srcPoint = {srcSbId, srcPort};
     if (analyzer.processedFlows[srcPoint]) {
+      // This FlowOp is a broadcast sibling of a flow whose route was already
+      // materialized (the analyzer merges all destinations sharing a source
+      // into one net, so the first sibling emitted connections for every
+      // destination). Erase it and report success so the erase is committed;
+      // returning failure() here would roll the erase back and leave a stale
+      // aie.flow in the routed IR, breaking idempotency (re-running the pass
+      // would try to re-route an already-routed net and fail).
       LLVM_DEBUG(llvm::dbgs() << "Flow already processed!\n");
       rewriter.eraseOp(Op);
-      return failure();
+      return success();
     }
     // std::map<TileID, SwitchSetting>
     SwitchSettings settings = analyzer.flowSolutions[srcPoint];
@@ -254,6 +262,13 @@ bool AIEPathfinderPass::findPathToDest(const SwitchSettings &settings,
   }
 
   int neighbourSourceChannel = currDestChannel;
+  // The destination may itself be a fabric input port (e.g. a partial packet
+  // flow pinned to end at a switchbox port). Such a port is never the output of
+  // any switchbox setting, so it would otherwise never match; recognize arrival
+  // when the current output feeds the final tile's destination input directly.
+  if (neighbourTile == finalTile && neighbourSourceBundle == finalDestBundle &&
+      neighbourSourceChannel == finalDestChannel)
+    return true;
   for (const auto &[sbNode, setting] : settings) {
     TileID tile = {sbNode.col, sbNode.row};
     if (tile == neighbourTile) {
@@ -341,28 +356,42 @@ static void bnbMinCover(ArrayRef<CoverCube> cubes,
   }
 }
 
+// Two cubes share an id when they agree on every bit both of them check.
+// Testing the cubes answers that without walking the ids they match, which a
+// cube with several don't-care bits has too many of.
+static bool cubesIntersect(std::pair<int, int> a, std::pair<int, int> b) {
+  return ((a.second ^ b.second) & a.first & b.first) == 0;
+}
+
 // Cover `matchIds` (a group's packet ids) with the fewest (mask, value) rules
-// that match every specified id and no `avoidIds` (other ids on the same slave
-// port). A rule matches id x iff (x & mask) == value; ids in neither set are
-// don't-cares, free to over-claim.
+// that match every specified id and no cube in `avoidCubes` (what the other
+// groups on the same slave port claim). A rule matches id x iff
+// (x & mask) == value; ids no cube claims are don't-cares, free to over-claim.
+//
+// An avoid cube states a claim the cover must not touch. A group that derives
+// its rules contributes one cube per id. A group whose flows state a mask
+// contributes that one cube, which stands for every id the mask matches --
+// including ids no flow in this design mentions.
 //
 // Based on Quine-McCluskey:
 //   1. Enumerate prime implicants: maximal cubes (a cube is one (mask, value))
-//      that match >=1 matchIds and no avoidIds.
+//      that match >=1 matchIds and no avoid cube.
 //   2. Pick the fewest of those cubes that cover all match ids (bnbMinCover).
 //   3. Tighten each chosen cube to the smallest enclosing cube of the ids it
 //      took, so it claims no more don't-cares than necessary; the one-cube case
 //      reduces to the old common-bits mask.
 static SmallVector<std::pair<int, int>>
 computeSubcubeCover(const SmallVector<int, 4> &matchIds,
-                    const llvm::SmallSet<int, 8> &avoidIds, int idBits) {
+                    ArrayRef<std::pair<int, int>> avoidCubes, int idBits) {
 
   // id space and full mask, sized from the target's packet-id width.
   const int numIds = 1 << idBits;
   const int idMask = numIds - 1;
 
   auto hitsAvoid = [&](int mask, int value) {
-    return llvm::any_of(avoidIds, [&](int o) { return (o & mask) == value; });
+    return llvm::any_of(avoidCubes, [&](std::pair<int, int> c) {
+      return cubesIntersect({mask, value}, c);
+    });
   };
 
   llvm::SmallBitVector matchMask(numIds);
@@ -545,6 +574,11 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
   DenseMap<std::pair<PhysPort, int>, DmaEndpoints> slaveFlowDmaEndpoints;
   // Aggregate of the above over all flows sharing a slave port.
   DenseMap<PhysPort, PortTraffic> slavePortTraffic;
+
+  // Packet-rule masks the flows state, keyed by the slave port the stream
+  // enters and the flow ID. One ID may reach a port under two masks, so the
+  // ID alone does not identify the claim.
+  std::map<std::pair<PhysPort, int>, int> pinnedMasks;
 
   for (auto tileOp : device.getOps<TileOp>()) {
     int col = tileOp.colIndex();
@@ -737,6 +771,9 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
                                       keep.value_or(false)))
                 traffic.spansConsumerBds = true;
             }
+            if (std::optional<uint8_t> mask = pktFlowOp.getMask())
+              pinnedMasks[{{currTile, {src.bundle, src.channel}}, flowID}] =
+                  *mask;
             // Assign "control packet flows" flag per switchbox, based on
             // packet flow op attribute
             auto ctrlPkt = pktFlowOp.getPriorityRoute();
@@ -808,6 +845,14 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
   // A map from Tile and master selectValue to the ports targetted by that
   // master select.
   std::map<std::pair<TileID, int>, SmallVector<Port, 4>> masterAMSels;
+
+  // <arbiter, msel> slots (per tile) that packet-switch configuration already
+  // present in the input IR occupies: a control overlay this run must not
+  // disturb, or any packet routing a previous run lowered. The allocator below
+  // starts from an empty state, so without this it can hand out a slot another
+  // master on the same tile holds, giving two independent flows one amsel.
+  // Kept apart from masterAMSels so this run does not re-emit those ops.
+  std::set<std::pair<TileID, int>> reservedAmsels;
 
   // Track which arbiter each port is assigned to (to prevent conflicts)
   std::map<PhysPort, int> portToArbiter;
@@ -977,7 +1022,8 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
         int hazardous = INVALID_AMSEL_VALUE;
         ArbiterHazard worst{0, 0};
         for (int amsel : candidates) {
-          if (masterAMSels.count({tileId, amsel}))
+          if (masterAMSels.count({tileId, amsel}) ||
+              reservedAmsels.count({tileId, amsel}))
             continue;
           std::optional<ArbiterHazard> hazard =
               arbiterHazard(tileId, getArbiterIDFromAmsel(amsel), flow);
@@ -1004,10 +1050,12 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
       [&](const std::map<std::pair<TileID, int>, SmallVector<Port, 4>>
               &masterAMSels,
           TileOp tileOp, int arbiter) {
-        for (int i = 0; i < numMselsPerArbiter; i++)
-          if (!masterAMSels.count({tileOp.getTileID(),
-                                   getAmselFromArbiterIDAndMsel(arbiter, i)}))
+        for (int i = 0; i < numMselsPerArbiter; i++) {
+          auto key = std::make_pair(tileOp.getTileID(),
+                                    getAmselFromArbiterIDAndMsel(arbiter, i));
+          if (!masterAMSels.count(key) && !reservedAmsels.count(key))
             return getAmselFromArbiterIDAndMsel(arbiter, i);
+        }
         tileOp->emitOpError("tile op arbiter ")
             << std::to_string(arbiter) << " has used up all its msels";
         return INVALID_AMSEL_VALUE;
@@ -1075,6 +1123,18 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
     }
     return INVALID_ARBITER_VALUE;
   };
+
+  // Seed the reserved set from the packet-switch configuration the switchboxes
+  // already carry, so this run allocates around it instead of over it.
+  for (auto swboxOp : device.getOps<SwitchboxOp>()) {
+    TileID tileId = swboxOp.getTileOp().getTileID();
+    for (auto mtset : swboxOp.getConnections().getOps<MasterSetOp>())
+      for (Value amselVal : mtset.getAmsels())
+        if (auto amselOp = amselVal.getDefiningOp<AMSelOp>())
+          reservedAmsels.insert(
+              {tileId, getAmselFromArbiterIDAndMsel(amselOp.arbiterIndex(),
+                                                    amselOp.getMselValue())});
+  }
 
   // Check all multi-cast flows (same source, same ID). They should be
   // assigned the same arbiter and msel so that the flow can reach all the
@@ -1338,11 +1398,40 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
     }
   }
 
-  // Ids on each slave port; a group covers its own and avoids the rest.
-  std::map<PhysPort, llvm::SmallSet<int, 8>> idsOnPort;
-  for (const auto &group : slaveGroups)
-    for (auto member : group)
-      idsOnPort[member.first].insert(member.second);
+  // What each group claims on its slave port, as cubes, split into the rules
+  // its flows state and the ids left for the cover to describe. A group may
+  // carry several rules, all pointing at its one arbiter slot, so a stated
+  // rule sits beside the derived ones rather than replacing them.
+  const uint32_t maxPacketId = device.getTargetModel().getMaxPacketId();
+  const int idBits = llvm::Log2_32_Ceil(maxPacketId + 1);
+  const int idMask = (1 << idBits) - 1;
+
+  SmallVector<SmallVector<std::pair<int, int>, 4>, 4> statedRules(
+      slaveGroups.size());
+  SmallVector<SmallVector<int, 4>, 4> derivedIds(slaveGroups.size());
+  for (size_t gi = 0; gi < slaveGroups.size(); ++gi) {
+    for (auto member : slaveGroups[gi]) {
+      auto it = pinnedMasks.find(member);
+      // A full-width mask selects the id alone, which the cover states just as
+      // well, so only a wider claim becomes a rule of its own.
+      if (it == pinnedMasks.end() || it->second == idMask) {
+        derivedIds[gi].push_back(member.second);
+        continue;
+      }
+      std::pair<int, int> cube = {it->second, member.second};
+      if (!llvm::is_contained(statedRules[gi], cube))
+        statedRules[gi].push_back(cube);
+    }
+  }
+
+  // Everything a group claims, for the other groups on its port to avoid.
+  auto claimsOf = [&](size_t gi) {
+    SmallVector<std::pair<int, int>, 8> claims(statedRules[gi].begin(),
+                                               statedRules[gi].end());
+    for (int id : derivedIds[gi])
+      claims.push_back({idMask, id});
+    return claims;
+  };
 
   // Realize the routes in MLIR
 
@@ -1425,7 +1514,8 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
 
     // Generate the packet rules
     DenseMap<Port, PacketRulesOp> slaveRules;
-    for (auto group : slaveGroups) {
+    for (size_t gi = 0; gi < slaveGroups.size(); ++gi) {
+      const auto &group = slaveGroups[gi];
       builder.setInsertionPoint(b.getTerminator());
 
       auto port = group.front().first;
@@ -1440,7 +1530,6 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
       for (auto member : group)
         matchIds.push_back(member.second);
 
-      uint32_t maxPacketId = device.getTargetModel().getMaxPacketId();
       for (int id : matchIds)
         if (id > static_cast<int>(maxPacketId)) {
           return mlir::emitError(tileLoc)
@@ -1448,39 +1537,74 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
                  << maxPacketId;
         }
 
-      int idBits = llvm::Log2_32_Ceil(maxPacketId + 1);
-      llvm::SmallSet<int, 8> avoidIds = idsOnPort[port];
-      for (int id : matchIds)
-        avoidIds.erase(id);
-      SmallVector<std::pair<int, int>> cover =
-          computeSubcubeCover(matchIds, avoidIds, idBits);
+      SmallVector<std::pair<int, int>> avoidCubes;
+      for (size_t oi = 0; oi < slaveGroups.size(); ++oi)
+        if (oi != gi && slaveGroups[oi].front().first == port) {
+          auto claims = claimsOf(oi);
+          avoidCubes.append(claims.begin(), claims.end());
+        }
+
+      // A stated rule claims ids beyond the ones this design routes. Two
+      // claims that share an id give the port two routes for that id, and the
+      // group that loses it has no rule left to state.
+      for (std::pair<int, int> own : claimsOf(gi))
+        for (std::pair<int, int> other : avoidCubes)
+          if (cubesIntersect(own, other)) {
+            int witness = (own.second & own.first) |
+                          (other.second & other.first & ~own.first);
+            return mlir::emitError(tileLoc)
+                   << "packet flows through " << stringifyWireBundle(bundle)
+                   << channel << " claim rule (mask 0x"
+                   << llvm::utohexstr(own.first) << ", id 0x"
+                   << llvm::utohexstr(own.second) << ") and rule (mask 0x"
+                   << llvm::utohexstr(other.first) << ", id 0x"
+                   << llvm::utohexstr(other.second)
+                   << "), which both match id 0x" << llvm::utohexstr(witness)
+                   << "; widen one mask to carry both, or route them apart";
+          }
+
+      // The group keeps the rules its flows state, and the cover describes the
+      // ids that state none. A stated rule also constrains that cover, so the
+      // two never claim one id twice.
+      SmallVector<std::pair<int, int>> cover(statedRules[gi].begin(),
+                                             statedRules[gi].end());
+      if (!derivedIds[gi].empty()) {
+        SmallVector<std::pair<int, int>> derivedAvoid = avoidCubes;
+        derivedAvoid.append(statedRules[gi].begin(), statedRules[gi].end());
+        SmallVector<int, 4> ids(derivedIds[gi].begin(), derivedIds[gi].end());
+        SmallVector<std::pair<int, int>> derived =
+            computeSubcubeCover(ids, derivedAvoid, idBits);
+        cover.append(derived.begin(), derived.end());
+      }
 
       LLVM_DEBUG({
         llvm::dbgs() << "packet cover " << stringifyWireBundle(bundle)
                      << channel << ": matchIds {";
         for (int id : matchIds)
           llvm::dbgs() << ' ' << id;
-        llvm::dbgs() << " } avoidIds {";
-        for (int id : avoidIds)
-          llvm::dbgs() << ' ' << id;
+        llvm::dbgs() << " } avoid {";
+        for (auto [m, v] : avoidCubes)
+          llvm::dbgs() << " (" << m << ", " << v << ")";
         llvm::dbgs() << " } ->";
         for (auto [m, v] : cover)
           llvm::dbgs() << " rule(" << m << ", " << v << ")";
         llvm::dbgs() << '\n';
       });
 
+      // A stated rule takes part in the same constraints as a derived one, so
+      // both checks cover every group.
       for ([[maybe_unused]] int id : matchIds)
         assert(llvm::any_of(cover,
                             [&](std::pair<int, int> c) {
                               return (id & c.first) == c.second;
                             }) &&
-               "subcube cover misses a match id");
-      for ([[maybe_unused]] int id : avoidIds)
+               "packet rule cover misses a match id");
+      for ([[maybe_unused]] std::pair<int, int> other : avoidCubes)
         assert(llvm::none_of(cover,
                              [&](std::pair<int, int> c) {
-                               return (id & c.first) == c.second;
+                               return cubesIntersect(c, other);
                              }) &&
-               "subcube cover over-claims an avoid id");
+               "packet rule cover claims another group's ids");
 
       Value amsel = amselOps[slaveAMSels[group.front()]];
 
@@ -1515,8 +1639,9 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
         return failure();
       }
 
-      // The cover avoids this port's avoidIds by construction; this catches a
-      // conflict only against rules from another source (e.g. hand-authored).
+      // The cover avoids what the other groups claim by construction; this
+      // catches a conflict only against rules from another source (e.g.
+      // hand-authored).
       for (auto rule : rules.getOps<PacketRuleOp>()) {
         auto verifyMask = rule.maskInt();
         auto verifyValue = rule.valueInt();
@@ -1633,7 +1758,14 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
     }
   }
 
+  // Remove the packet_flow ops now that they have been lowered into
+  // masterset/amsel/packet_rules on the switchboxes. Leaving them in the IR
+  // would break idempotency: re-running this pass on its own routed output
+  // would re-lower them and emit duplicate masterset/packet_rules (illegal
+  // double-drivers on ports already claimed by the first lowering).
+  target.addIllegalOp<PacketFlowOp>();
   RewritePatternSet patterns(&getContext());
+  patterns.insert<AIEOpRemoval<PacketFlowOp>>(device.getContext());
 
   if (failed(applyPartialConversion(device, target, std::move(patterns))))
     return failure();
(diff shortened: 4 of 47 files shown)


lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
```diff
@@ -385,26 +386,42 @@ bool Pathfinder::addFixedConnection(SwitchboxOp switchboxOp) {
   int row = switchboxOp.rowIndex();
   TileID coords = {col, row};
   auto &sb = graph[std::make_pair(coords, coords)];
+  // Validate every connect against the original connectivity before reserving
+  // any resource, so a broadcast (one source fanning out to several dests) does
+  // not invalidate its own sibling connects mid-scan. A destination port may be
+  // driven only once; a source port may legally recur across a broadcast.
+  llvm::SmallVector<std::pair<int, int>, 8> reserved;
+  llvm::SmallDenseSet<int, 8> claimedDsts;
   for (ConnectOp connectOp : switchboxOp.getOps<ConnectOp>()) {
```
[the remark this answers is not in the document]
andrej: Do not lift before routing, but yes, do ingest existing used resources and make sure you do not collide. This is critical.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -141,93 +154,210 @@ class ConnectivityAnalysis {
...
+    }
+    if (keepMaterialized)
+      continue;
+    // A packet endpoint becomes a logical packet flow, and the pass reclaims
+    // its physical configuration.
+    if (maskValue.mask != 0) {
+      for (Operation *op : c.usedOps)
+        if (op)
+          consumed.insert(op);
+      auto flowOp = PacketFlowOp::create(rewriter, loc, maskValue.value,
```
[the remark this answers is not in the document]
andrej: Yes, this is critical. However it might start getting into some limitations of what `aie.packet_flow` can express vs. what you can express at lower levels (more expressive).

Please consider all options and, without making any changes yet, reply with those options and what you think the best one is. I'm thinking along  the lines of (a) add a "mask" attribute to packet_flow, so that we can (try to) capture the same set of (multiple) IDs as the underlying lowering at the high-level `packet_flow`, (b) reject lifting those `packet_flow`s that apply to more than one ID (this does NOT mean rejecting any packet flows that cross at one particular switchbox or even a sub-string of shared links going out the same master set if they can alter be distinguished by ID again), (c) create multiple packet_flow ops, one for each ID, and allow the lowering to merge them together onto the same resources using a mask that matches both.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -141,93 +154,210 @@ class ConnectivityAnalysis {
...
 
+// The endpoint of a lifted flow: a flow-endpoint op directly, or the tile
+// owning a directional switchbox / shim-mux port (partial flow).
+static Value resolveEndpointTile(Operation *op) {
+  if (auto sb = dyn_cast<SwitchboxOp>(op))
+    return sb.getTile();
+  if (auto sm = dyn_cast<ShimMuxOp>(op))
+    return sm.getTile();
+  if (op->hasTrait<IsFlowEndPoint>() && op->getNumResults() == 1)
+    return op->getResult(0);
```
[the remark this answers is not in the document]
andrej: I think there may be an Interface for tiles/endpoints that we could use here instead -- look into that.

lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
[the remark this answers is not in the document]
andrej: (Claude) Done — ingested, not lifted.

`addFixedConnection` now also reserves the destination port of every `aie.masterset` in the switchbox, alongside the circuit `ConnectOp` reservations. A masterset fixes only its output port, so the source rows stay free: its inputs arrive through arbiters that packet flows share.

The arbiter/msel side is reserved too. `runOnPacketFlow` seeds a `reservedAmsels` set from the `aie.masterset`/`aie.amsel` pairs the switchboxes already carry, and both allocators (`getNewUniqueAmsel` and `getNewUniqueAmselPerArbiterID`) skip a slot that set holds. It is kept apart from `masterAMSels` so this run does not re-emit the existing ops.

Nothing is lifted before routing.

One gap I did not close: I reserve amsels and master output ports, but not the rule slots on a slave port. Two flows entering the same slave port still get their rules computed only from the IDs this run knows about, so a rule already present on that port is invisible to `computeSubcubeCover`. That is the same information the multi-ID question in `AIEFindFlows.cpp` is about, so I would rather settle that thread first than guess at it here. Say the word if you want it covered in this PR.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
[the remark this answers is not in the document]
andrej: (Claude) No changes made. Options below, then my recommendation.

First, a related defect in the same lines that is independent of the expressiveness question. The branch is chosen by `maskValue.mask != 0`, but the accumulated mask is not a packet/circuit discriminator: a rule may legitimately carry mask 0 (it accepts every ID), and such a path takes the circuit branch and becomes an `aie.flow`. The discriminator should be whether the traversed path used a `PacketRuleOp`, which `usedOps` already records. I have left it alone since it sits in this thread, but it is a one-line change and orthogonal to (a)/(b)/(c).

**(a) mask attribute on `aie.packet_flow`.** The op stores the hardware pair directly, so lifting is exact and one rule stays one op.
- The `via` PR (#3421) already implements this, so choosing it means promoting that piece down into this PR.
- It makes `ID` mean "value" and gives two flows with the same ID but different masks distinct meanings, so every consumer that reads `IDInt()` sees a partial truth.
- It bypasses `computeSubcubeCover`, so a pinned mask can over-claim an ID that a flow added later wants, and nothing detects that.

**(b) refuse to lift a multi-ID path.** Keep that switchbox materialized and lift everything else.
- The machinery exists: `emitFlows` already keeps control overlays materialized through `keepMaterialized`.
- No dialect change, and it cannot lose information.
- Recovery becomes partial, so a design using ID-masked broadcast keeps physical configuration and does not fully round-trip.
- "Multi-ID" has to mean "admits more than one ID the design actually uses", not "mask has a don't-care bit", or it rejects far too much.

**(c) one `aie.packet_flow` per ID, merged again by the lowering.**
- `aie.packet_flow` keeps its current meaning, and every existing consumer already understands a single-ID flow.
- The merging already exists: `computeSubcubeCover` derives one rule covering a set of IDs on a slave port, which is exactly the inverse of the split.
- The risk is the merge guarantee. If lowering does not put the N flows back on one amsel and one rule, a design that had one rule gets N and can exhaust the rule budget on that port.
- Enumeration must be over the IDs the design uses on that slave port, which the traversal can collect from the `PacketRuleOp`s. Enumerating what the mask admits is 2^k and not an option.

**Recommendation: (c) as the target, (b) as the immediate default.**

A mask is a lowering artifact, not design intent. (a) encodes it in the source dialect and leaves two legal spellings of one thing, which every consumer then has to handle. (c) keeps the high level saying what the design means, one flow one ID, and puts the packing back where it already lives.

I would land (b) now because it is strictly conservative and unblocks the idempotency work: nothing is lost, some configuration just stays physical. Then do (c) as a follow-up once the merge guarantee is verified, at which point (b)'s refusal fires in fewer and fewer cases.

I would avoid (a) here unless a mask turns out to be genuine intent for some user. Note this cuts against #3421, which needs the mask attribute for pinning a lifted straight-line packet section. If you want (c) in this PR, we should decide separately whether the mask attribute still earns its place there or whether via-pinned packet segments can carry per-ID flows too.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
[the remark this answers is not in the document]
andrej: (Claude) Done — used the interface, as you suggested.

`TileElement` is the one: it exposes `getTile()`, and `Interconnect` extends it, so a switchbox or shim-mux port resolves the same way as an `aie.shim_dma` or `aie.core`. `resolveEndpointTile` is now three lines with no op-by-op `dyn_cast` chain, and `TileOp` is the only special case since it is the tile rather than an element of one.

Every op carrying `IsFlowEndPoint` is covered: `aie.tile` by the special case, and `aie.core`, `aie.shim_dma`, `aie.mem` and `aie.memtile_dma` through `TileElement`.

This also fixes the crash Copilot reported in the sibling comment. I confirmed the failure first on a hand-written `aie.flow(%t21, Core : 0, %shim_dma, DMA : 0)`, which aborts in `--aie-create-pathfinder-flows` with exit 134 on the `cast<TileOp>`. Lifting `test/find-flows/shim.mlir` and routing the result now succeeds.

`test/find-flows/shim.mlir` changes accordingly: the recovered flow names `tile(2, 0)` rather than `shim_dma_2_0`. Worth a look, since it is a user-visible change to what the pass emits, though it is the spelling the router has always required.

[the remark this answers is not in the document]
andrej: (Claude) Confirmed and fixed. I reproduced the abort first on a hand-written `aie.flow(%t21, Core : 0, %shim_dma, DMA : 0)`: `--aie-create-pathfinder-flows` exits 134 on the `cast<TileOp>`.

Taking the first of your two options, emitting the endpoint's owning tile, since `aie.flow` names tiles by definition and teaching both routing paths to accept tile elements would widen what a flow may name.

`resolveEndpointTile` now returns `TileElement::getTile()`. `Interconnect` extends `TileElement`, so switchbox and shim-mux ports resolve through the same call, and `TileOp` is the only special case. That covers every `IsFlowEndPoint` op: `aie.core`, `aie.shim_dma`, `aie.mem` and `aie.memtile_dma`.

Lifting `test/find-flows/shim.mlir` and routing the result now succeeds. The test's expectation changed from `shim_dma_2_0` to `tile(2, 0)`.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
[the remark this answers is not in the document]
andrej: (Claude) Both halves of this are real, and I am deliberately not changing them yet — the repo owner asked in the sibling thread for options rather than a patch, and I have replied there with (a) a `mask` attribute, (b) refusing to lift a multi-ID path, and (c) one `packet_flow` per ID re-merged by the lowering. My recommendation is (c) as the target with (b) as the immediate default.

Recording your two points so neither is lost:

The discriminator is wrong. `maskValue.mask != 0` does not separate packet from circuit, because a rule may carry mask 0 and accept every ID; that path currently becomes an `aie.flow`. The fix is to test whether the traversed path used a `PacketRuleOp`, which `usedOps` already records. This is independent of (a)/(b)/(c) and I will make it once the thread settles.

The ID loss is the expressiveness question. Your `aie.rule(30, 0, ...)` example in `test_create_packet_flows5.mlir` is right: mask 30 leaves bit 0 a don't-care, so the rule accepts 0 and 1, while `packet_flow(0)` claims only one of them.

One correction to "preserve every ID represented by the accumulated mask": a mask with k don't-care bits admits 2^k IDs, so that set cannot be enumerated in general. The enumeration has to be over the IDs the design actually uses on that slave port, which the traversal can collect from the `PacketRuleOp`s.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -274,16 +450,72 @@ struct AIEFindFlowsPass
...
+    };
+    for (auto switchOp : d.getOps<SwitchboxOp>())
+      cleanupInterconnect(switchOp.getConnections());
+    for (auto shimMuxOp : d.getOps<ShimMuxOp>())
+      cleanupInterconnect(shimMuxOp.getConnections());
+
+    // Wires reference interconnect results and are regenerated by routing, so
+    // drop them, then remove any interconnect left fully empty.
+    for (auto wireOp : llvm::make_early_inc_range(d.getOps<WireOp>()))
+      wireOp.erase();
```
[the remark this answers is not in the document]
andrej: (Claude) Confirmed and fixed, though by a coarser rule than you proposed.

Erasure is no longer unconditional. The pass first collects the interconnects that still hold configuration after the lifted ops are removed, then erases a wire only when neither of its endpoints is one of them. That covers both cases you named: control-overlay paths that `emitFlows` skips, and the partial connects left in place under `keep-partial-flows=false`.

The empty-interconnect sweep had to follow. A surviving wire keeps its operands alive, so an empty switchbox that a surviving wire names is now kept rather than erased; erasing it would leave the wire with a dangling operand.

I did not go as far as recording wires per traversed path and erasing only those exclusively covered by consumed routes. Reachability from a retained interconnect is coarser: a wire between a retained switchbox and a fully lifted one is kept even though no retained configuration needs it. That errs toward keeping a wire, and the router's wire generation now dedupes against what is already present, so a redundant wire is absorbed rather than duplicated. If you have a case where the coarser rule keeps a wire that actually causes trouble, I will switch to per-path tracking.

lib/Dialect/AIE/Transforms/AIEPathFinder.cpp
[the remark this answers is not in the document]
andrej: (Claude) Fixed, following the repo owner's direction in the sibling thread: ingest the existing state as fixed resources, do not lift before routing.

Two reservations were added.

`addFixedConnection` now reserves the destination port of every `aie.masterset` in the switchbox, alongside the circuit connects. A masterset fixes only its output port, so the source rows stay free: its inputs arrive through arbiters that packet flows share. This stops the circuit pathfinder, which does not read packet ops, from routing a flow onto an output a masterset owns.

`runOnPacketFlow` seeds a `reservedAmsels` set from the `aie.masterset`/`aie.amsel` pairs already present, and both `getNewUniqueAmsel` and `getNewUniqueAmselPerArbiterID` skip a reserved slot. It is kept apart from `masterAMSels` so the existing ops are not re-emitted.

The rule-slot part of your comment is not covered. Rules on a slave port are still derived only from the IDs this run knows about, so a rule already on that port is invisible to `computeSubcubeCover`. That needs the same ID bookkeeping as the multi-ID lifting question, so I have left it until that thread settles rather than guessing.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1304,8 +1326,23 @@ void AIEPathfinderPass::runOnOperation() {
...
-  // Populate wires between switchboxes and tiles.
+  // Populate wires between switchboxes and tiles. Re-running the pass on
+  // already-routed IR must not duplicate the wires a previous run emitted, so
+  // key the wires already present and emit only the missing ones.
   builder.setInsertionPoint(d.getBody()->getTerminator());
+  llvm::DenseSet<std::tuple<Value, int, Value, int>> existingWires;
+  for (auto wire : d.getOps<WireOp>())
+    existingWires.insert(
+        {wire.getSource(), static_cast<int>(wire.getSourceBundle()),
+         wire.getDest(), static_cast<int>(wire.getDestBundle())});
```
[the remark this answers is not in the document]
andrej: (Claude) Confirmed and fixed, as you suggested: both orientations now go into the set.

`recordWire` inserts the adjacency under both operand orders and reports whether it was absent under either. Seeding uses it for every existing `aie.wire`, and the emit path only creates a wire when it comes back absent. A reverse-oriented wire already in the IR is therefore recognised, and the run no longer adds a duplicate in canonical order.

Thanks for the two concrete spellings — `test/find-flows/shim.mlir:34` against `test/Targets/AIEFlowsToJSON/simple.mlir:53` made it easy to confirm both orders really do occur.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -141,93 +154,210 @@ class ConnectivityAnalysis {
...
+    if (keepMaterialized)
+      continue;
+    // A packet endpoint becomes a logical packet flow, and the pass reclaims
+    // its physical configuration.
+    if (maskValue.mask != 0) {
+      for (Operation *op : c.usedOps)
+        if (op)
+          consumed.insert(op);
+      auto flowOp = PacketFlowOp::create(rewriter, loc, maskValue.value,
+                                         nullptr, nullptr);
```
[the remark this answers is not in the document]
andrej: (Claude) Confirmed and fixed.

The recovered `PacketFlowOp` now carries `keep_pkt_header`. Before creating it, the pass scans `usedOps` for the `MasterSetOp` whose destination matches the endpoint port and takes its `keep_pkt_header` attribute. That is the master set driving the flow's destination, which is where lowering stores the flag, so the value survives the erase that follows.

`priority_route` stays null. It is a routing directive consumed during lowering rather than a property of the recovered route, so re-deriving it from physical configuration would be inventing intent. Say if you disagree.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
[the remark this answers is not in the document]
andrej: (Claude) Implemented as option (a). Your undecidability argument settles it — a core can build a header at run time, so any scheme that reasons about "the ids a design uses" is unsound. Options (b) and (c) both do that, so both are out.

`aie.packet_flow` takes an optional `mask`, written after the ID: `aie.packet_flow(0x10 mask 0x1c)`. The verifier rejects an ID with a bit outside its mask, since `incoming & mask == ID` can never hold then.

On your question about the subcube cover: yes, and it composes rather than bypassing it. The cover already reasons over the whole id space rather than over the ids in the design, so nothing has to be enumerated. The one change is the type of the avoid set: it held single ids, and it now holds cubes. A single id is the degenerate cube `(idMask, id)`, so this is a strict generalization. Two cubes are tested by whether they agree on every bit both check:

    ((v1 ^ v2) & m1 & m2) == 0

That answers "do these claims share an id" in constant time without walking what a mask matches.

A group whose flow states a mask contributes that one cube and skips the cover. Its cube joins the avoid set of every other group on the slave port, so derived rules are fenced away from it. Prime-implicant maximality uses the same predicate, so a derived cube grows only until it would touch a stated claim.

Two consequences you asked for fall out. The two cover assertions now hold for every group, in place of the earlier bypass that switched them off for masked groups. And overlapping claims are reported rather than mis-routed: the error names both rules and an id they share, and it fires symmetrically, since the group that loses an id to someone else's mask is the one with no cover left.

Lifting is lossless as a result: `--aie-find-flows` recovers the `(mask, id)` pair the rules hold, so routing the recovered design rebuilds the same rules.

One defect fixed along the way, the one I flagged earlier in this thread: the packet/circuit branch tested `maskValue.mask != 0`, which is not a discriminator, since a rule may carry mask 0. It now tests whether the traversed path used a `PacketRuleOp`.

Tests: `packet_flow_mask.mlir` (a stated rule beside derived ones on one port), `find_flows_packet_mask.mlir` (mask survives lifting and re-routing into the same rules), `packet_flow_mask_conflict.mlir` and `packet_flow_mask_bad_id.mlir` (the two diagnostics).

On multiple ids: agreed, and I did not add it. Lifting can only ever recover the pair, so a list would be sugar that lowers to the same thing. It also needs a stated rule for the case where the enclosing cube over-claims — ids 1 and 2 enclose to mask 0, which matches everything — so it is worth its own discussion rather than riding along here.

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -824,26 +824,53 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
...
+    states the packet-rule mask of the flow. A slave port accepts a packet
+    when `incoming & mask == ID`, so the flow claims every ID that agrees
+    with ID on the bits the mask selects. Routing derives a mask from the
+    IDs it knows when the attribute is absent. ID must have no bit outside
+    the mask, or no packet can match.
+
+    A design may compute a packet header at run time, so which IDs reach a
+    port is not decidable here. State the mask to claim a set of IDs that
+    routing cannot see.
+
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -100,9 +99,16 @@ struct ConvertFlowsToInterconnect : OpConversionPattern<FlowOp> {
     TileID srcSbId = {srcCoords.col, srcCoords.row};
     PathEndPoint srcPoint = {srcSbId, srcPort};
     if (analyzer.processedFlows[srcPoint]) {
+      // This FlowOp is a broadcast sibling of a flow whose route was already
+      // materialized (the analyzer merges all destinations sharing a source
+      // into one net, so the first sibling emitted connections for every
+      // destination). Erase it and report success so the erase is committed;
+      // returning failure() here would roll the erase back and leave a stale
+      // aie.flow in the routed IR, breaking idempotency (re-running the pass
+      // would try to re-route an already-routed net and fail).
```
andrej:
```suggestion
      // destination). Erase it and report success so the erase is committed.
```

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -826,26 +826,53 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
     so that they always get allocated with the same master, slave
     ports, arbiters and master selects (msel).
 
+    The optional attribute mask, written after the ID in the parentheses,
+    states the packet-rule mask of the flow. A slave port accepts a packet
+    when `incoming & mask == ID`, so the flow claims every ID that agrees
+    with ID on the bits the mask selects. Routing derives a mask from the
+    IDs it knows when the attribute is absent. ID must have no bit outside
+    the mask, or no packet can match.
```
andrej: Reword this to: When the attribute is absent, the router decides the mask that matches the given ID and rejects known IDs from other routes. The ID must have ...

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -325,15 +325,32 @@ def AIEFindFlows : Pass<"aie-find-flows", "DeviceOp"> {
...
     aie.packetflow).  This pass is primarily used for testing automatic
     routing.
+
+    By default every configured connection is lifted, including partial flows
+    whose source or destination is a directional switchbox port rather than a
+    core or DMA (transit connections, PLIO/edge entries, and packet routes a
+    running design steers by packet id).  A physically routed design can
+    therefore drop its aie.switchbox configuration and recover it from the
+    lifted flows.  Set `keep-partial-flows=false` to restrict recovery to flows
+    that begin at a core/DMA and end on a core/DMA.
```
andrej, line 332:
```suggestion
    By default every configured connection is lifted, including partial flows
    whose source or destination is a switchbox port other than a core or DMA. 
    Set `keep-partial-flows=false` to restrict recovery to flows
    that begin at a core/DMA and end on a core/DMA.
```

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -325,15 +325,32 @@ def AIEFindFlows : Pass<"aie-find-flows", "DeviceOp"> {
...
+  let options =
+      [Option<"clKeepPartialFlows", "keep-partial-flows", "bool",
+              /*default=*/"true",
+              "Also recover partial flows whose source or destination is a "
+              "directional/edge switchbox port rather than a core or DMA.">,
+       Option<
+           "clRemoveLifted", "remove-lifted", "bool", /*default=*/"true",
+           "Remove the interconnect configuration (connects, packet rules, "
+           "mastersets, amsels, wires) that was lifted into recovered flows, "
+           "leaving only configuration that could not be lifted.">,
```
andrej, line 346: Do we still need this option? Can we drop it?

include/aie/Dialect/AIE/Transforms/AIEPasses.td
```diff
@@ -325,15 +325,32 @@ def AIEFindFlows : Pass<"aie-find-flows", "DeviceOp"> {
...
   let constructor = "xilinx::AIE::createAIEFindFlowsPass()";
-  let dependentDialects = [
-    "xilinx::AIE::AIEDialect",
+  let dependentDialects = ["xilinx::AIE::AIEDialect", ];
+
+  let options =
+      [Option<"clKeepPartialFlows", "keep-partial-flows", "bool",
+              /*default=*/"true",
+              "Also recover partial flows whose source or destination is a "
+              "directional/edge switchbox port rather than a core or DMA.">,
```
andrej:
```suggestion
              "Also recover flows that do not start or end at a core or DMA.">,
```

lib/Dialect/AIE/IR/AIEDialect.cpp
```diff
@@ -1974,12 +1974,18 @@ LogicalResult verifyNoDuplicateFlows(DeviceOp device) {
...
-// keep_pkt_header/priority_route settings when their attributes disagree.
+// of sources and destinations under the same mask. Declaring the flow twice
+// makes the routing problem look more congested than it is and gives the two
+// copies conflicting keep_pkt_header/priority_route settings when their
+// attributes disagree.
+//
+// The mask belongs in the key because it is half of what a flow claims: an ID
+// under two masks names two sets of packets, so the two flows carry different
+// traffic. A flow without the attribute claims its ID alone, which is what the
+// widest mask says, so both spellings key alike.
```
```suggestion
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -254,6 +262,13 @@ bool AIEPathfinderPass::findPathToDest(const SwitchSettings &settings,
   }
 
   int neighbourSourceChannel = currDestChannel;
+  // The destination may itself be a fabric input port (e.g. a partial packet
+  // flow pinned to end at a switchbox port). Such a port is never the output of
+  // any switchbox setting, so it would otherwise never match; recognize arrival
+  // when the current output feeds the final tile's destination input directly.
+  if (neighbourTile == finalTile && neighbourSourceBundle == finalDestBundle &&
+      neighbourSourceChannel == finalDestChannel)
+    return true;
```
andrej: I can't make sense of this. Please explain in plain terms.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -341,28 +356,42 @@ static void bnbMinCover(ArrayRef<CoverCube> cubes,
   }
 }
 
+// Two cubes share an id when they agree on every bit both of them check.
+// Testing the cubes answers that without walking the ids they match, which a
+// cube with several don't-care bits has too many of.
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -341,28 +356,42 @@ static void bnbMinCover(ArrayRef<CoverCube> cubes,
...
-// port). A rule matches id x iff (x & mask) == value; ids in neither set are
-// don't-cares, free to over-claim.
+// that match every specified id and no cube in `avoidCubes` (what the other
+// groups on the same slave port claim). A rule matches id x iff
+// (x & mask) == value; ids no cube claims are don't-cares, free to over-claim.
+//
+// An avoid cube states a claim the cover must not touch. A group that derives
+// its rules contributes one cube per id. A group whose flows state a mask
+// contributes that one cube, which stands for every id the mask matches --
+// including ids no flow in this design mentions.
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -737,6 +771,9 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
                                       keep.value_or(false)))
                 traffic.spansConsumerBds = true;
             }
+            if (std::optional<uint8_t> mask = pktFlowOp.getMask())
+              pinnedMasks[{{currTile, {src.bundle, src.channel}}, flowID}] =
+                  *mask;
```
andrej: Throughout: Use braces around code blocks, even for semantically single lines of code. This line, while semantically one, spreads across two, making the control flow more error prone. There's a clang-format feature to add braces everywhere. Apply it, but only to the lines changed in this branch.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -809,6 +846,13 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
   // master select.
   std::map<std::pair<TileID, int>, SmallVector<Port, 4>> masterAMSels;
 
+  // <arbiter, msel> slots (per tile) that packet-switch configuration in the
+  // input IR already occupies. The allocator below starts from an empty state,
+  // so without this it can hand out a slot another master on the same tile
+  // holds, giving two independent flows one amsel. Kept apart from
+  // masterAMSels so this run does not re-emit those ops.
```
andrej:
```suggestion
  // input IR already occupies. Kept apart from
  // masterAMSels so the allocator run does not re-emit those ops.
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1338,11 +1397,40 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
...
 
-  // Ids on each slave port; a group covers its own and avoids the rest.
-  std::map<PhysPort, llvm::SmallSet<int, 8>> idsOnPort;
-  for (const auto &group : slaveGroups)
-    for (auto member : group)
-      idsOnPort[member.first].insert(member.second);
+  // What each group claims on its slave port, as cubes, split into the rules
+  // its flows state and the ids left for the cover to describe. A group may
+  // carry several rules, all pointing at its one arbiter slot, so a stated
+  // rule sits beside the derived ones rather than replacing them.
```
andrej:
```suggestion
  // its flows state and the ids left for the cover to describe.
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1440,47 +1529,81 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
...
+      SmallVector<std::pair<int, int>> avoidCubes;
+      for (size_t oi = 0; oi < slaveGroups.size(); ++oi)
+        if (oi != gi && slaveGroups[oi].front().first == port) {
+          auto claims = claimsOf(oi);
+          avoidCubes.append(claims.begin(), claims.end());
+        }
+
+      // A stated rule claims ids beyond the ones this design routes. Two
+      // claims that share an id give the port two routes for that id, and the
+      // group that loses it has no rule left to state.
```
andrej: Please reword this more clearly.
Is a "stated rule" a pre-existing rule in the design, not generated by this router?
And just to make sure, this will not reject a pre-existing design that decides to fan-out the same ID to multiple ports, right?

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1633,7 +1757,14 @@ AIEPathfinderPass::runOnPacketFlow(DeviceOp device, OpBuilder &builder,
     }
   }
 
+  // Remove the packet_flow ops now that they have been lowered into
+  // masterset/amsel/packet_rules on the switchboxes. Leaving them in the IR
+  // would break idempotency: re-running this pass on its own routed output
+  // would re-lower them and emit duplicate masterset/packet_rules (illegal
+  // double-drivers on ports already claimed by the first lowering).
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1663,8 +1794,32 @@ void AIEPathfinderPass::runOnOperation() {
...
 
-  // Populate wires between switchboxes and tiles.
+  // Populate wires between switchboxes and tiles. Re-running the pass on
+  // already-routed IR must not duplicate the wires a previous run emitted, so
+  // key the wires already present and emit only the missing ones.
   builder.setInsertionPoint(d.getBody()->getTerminator());
+  llvm::DenseSet<std::tuple<Value, int, Value, int>> existingWires;
+  // aie.wire is a physical connection, so the IR may spell one adjacency in
+  // either operand order. Key both orders, or a reverse-oriented wire already
+  // present goes unnoticed and this run adds a duplicate in canonical order.
```
andrej:
```suggestion
  // either operand order.
```

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
```diff
@@ -1663,8 +1794,32 @@ void AIEPathfinderPass::runOnOperation() {
...
+    bool absent =
+        !existingWires.contains({source, sourceBundle, dest, destBundle}) &&
+        !existingWires.contains({dest, destBundle, source, sourceBundle});
+    existingWires.insert({source, sourceBundle, dest, destBundle});
+    existingWires.insert({dest, destBundle, source, sourceBundle});
+    return absent;
+  };
+  for (auto wire : d.getOps<WireOp>())
+    recordWire(wire.getSource(), static_cast<int>(wire.getSourceBundle()),
+               wire.getDest(), static_cast<int>(wire.getDestBundle()));
```
andrej: mentioned earlier, but throughout: use braces around new code blocks

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -41,11 +44,17 @@ using PortConnection = struct PortConnection {
 using PortMaskValue = struct PortMaskValue {
   Port port;
   MaskValue mv;
+  // The interconnect ops that implement this hop (a circuit ConnectOp, or a
+  // packet PacketRuleOp + MasterSetOp + AMSelOp). Recorded so the flows that
+  // consume them can be lifted out of the physical IR.
```
andrej:
```suggestion
  // packet PacketRuleOp + MasterSetOp + AMSelOp). Erased after lifting.
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+}
+
+// The tile an endpoint of a lifted flow belongs to.
+//
+// aie.flow names tiles: --aie-create-pathfinder-flows casts both endpoints of
+// every flow to TileOp. A tile element such as an aie.shim_dma therefore has to
+// resolve to its owning tile, or the flow this pass emits crashes the router it
+// is meant to feed. TileElement covers every element that can end a flow, and
+// Interconnect extends it, so a switchbox or shim-mux port resolves the same
+// way.
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+      if (op->hasAttr("is_ctrl_pkt_overlay") ||
+          (rulesParent && rulesParent->hasAttr("is_ctrl_pkt_overlay"))) {
+        keepMaterialized = true;
+        break;
+      }
+    }
+    if (keepMaterialized)
+      continue;
+    // A packet endpoint becomes a logical packet flow, and the pass reclaims
+    // its physical configuration.
```
andrej:
```suggestion
    // A packet endpoint becomes a logical packet flow, and the pass erases
    // the lowered physical configuration.
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+    //
+    // The rules decide this, not the mask: a rule may carry mask 0, which
+    // accepts every id, and a circuit path accumulates the same mask 0.
```
andrej:
```suggestion
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+    bool isPacket = llvm::any_of(c.usedOps, [](Operation *op) {
+      return isa_and_nonnull<PacketRuleOp>(op);
+    });
+    // The traversal reaches one endpoint more than once when a broadcast
+    // re-converges on it, and the three seeds can arrive at one route from
+    // different directions. Those repeats describe one flow, and
+    // DeviceOp::verify rejects a flow declared twice.
```
andrej: Which three seeds? Please clarify. What do you mean by a broadcast re-converging? Are you just talking about broadcast in general?

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+                      isPacket ? maskValue.mask : 0);
+    if (key && !seen.insert(*key).second)
+      continue;
+    if (isPacket) {
+      for (Operation *op : c.usedOps)
+        if (op)
+          consumed.insert(op);
+      // The lowering stores keep_pkt_header on the master set that drives the
+      // destination. Carry it onto the recovered flow, or re-lowering strips
+      // the header and misroutes the packet downstream.
```
andrej:
```suggestion
      // destination. Carry it onto the recovered flow.
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+            keepPktHeader = ms.getKeepPktHeaderAttr();
+      // The rules along the path accept every id that agrees with value on the
+      // bits mask selects. Which of those a running design sends is not
+      // decidable here, so state the pair and let routing rebuild the same
+      // rules.
+      //
+      // A full-width mask selects one id, which the id states on its own.
+      // Leaving the attribute off keeps such a flow mergeable with the others
+      // that share its route, the way routing found it.
+      IntegerAttr mask;
```
andrej, line 407:
```suggestion
      //
      // A full-width mask selects exactly one id; in flow IR we express this by leaving off the mask.
      // The router may merge such flows.
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+      PacketSourceOp::create(rewriter, loc, srcTile, srcBundle, srcChannel);
+      PacketDestOp::create(rewriter, loc, destTile, destPort.bundle,
+                           destPort.channel);
+      rewriter.restoreInsertionPoint(ip);
+      continue;
+    }
+    // A section that lifts no interconnect configuration carries no routing: it
+    // is the physical wire between two retained nodes. The wire is implicit and
+    // the switchboxes at both ends stay explicit, so a flow emitted for it
+    // would only fight their fixed connections on re-routing.
```
andrej: Slop. Just say that if there is no switchbox configuration between two endpoints, no flow for it needs to be lifted.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -267,27 +313,131 @@ static std::optional<FlowKey> tryGetFlowKey(Operation *srcOp, Port srcPort,
...
+    if (c.usedOps.empty())
+      continue;
+    // A flow seeded from a fabric entry that terminates on a real endpoint of
+    // the same tile never leaves that tile. It is a dead or orphan interconnect
+    // connection, such as an undriven shim-mux write path, so the pass leaves
+    // it alone.
```
andrej: Same here, simplify, reword the comment

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -298,42 +448,80 @@ static void findFlowsFrom(TileOp op, ConnectivityAnalysis &analysis,
...
+// upstream interconnect connection.  Ports driven by a core/DMA tile are
+// already covered by findFlowsFrom; ports driven by an upstream interconnect
+// are covered by that interconnect's own source, so both are skipped here to
+// avoid emitting a flow twice.
+static void findFlowsFromInterconnect(Operation *switchOp,
+                                      ConnectivityAnalysis &analysis,
+                                      OpBuilder &rewriter,
+                                      bool keepPartialFlows, int idMask,
+                                      FlowKeySet &seen,
+                                      llvm::DenseSet<Operation *> &consumed) {
```
andrej, line 476: Why does this need to be its own function? At a high level, whether I'm recovering a flow seeded from cores/DMAs or just intermediate points seems algorithmically the same -- just a different starting point. Can we not reuse more logic? Is this special handling needed at all?

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
```diff
@@ -343,21 +531,104 @@ struct AIEFindFlowsPass
...
+                                  clKeepPartialFlows, idMask, seen, consumed);
     }
+
+    if (!clRemoveLifted)
+      return;
+
+    // Every recovered flow makes the interconnect ops it traversed redundant;
+    // drop exactly those, leaving any configuration that could not be lifted
+    // (e.g. an unreachable connect) in place.  Leaf routing ops go first;
+    // amsels that lose all users and now-empty rule containers follow.
```
andrej:
```suggestion
    // (e.g. an unreachable connect) in place. 
```

Restates code

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
```diff
@@ -155,6 +155,16 @@ struct AIEGenerateColumnControlOverlayPass
         }
         continue;
       }
+      // A device that already carries a control overlay -- e.g. a physically
+      // routed design recompiled through aiecc -- must not receive a second
+      // one, or the regenerated control flows would double-drive its shim
+      // control ports during routing.
```
```suggestion
      // A device that already carries a control overlay must not receive a second
      // one.
```

include/aie/Dialect/AIE/IR/AIEOps.td
andrej: (Claude) Done. The paragraph now reads:

> The optional attribute mask, written after the ID in the parentheses, states the packet-rule mask of the flow. A slave port accepts a packet when `incoming & mask == ID`, so the flow claims every ID that agrees with ID on the bits the mask selects. When the attribute is absent, the router decides the mask that matches the given ID and rejects known IDs from other routes. The ID must have no bit outside the mask, or no packet can match.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
andrej: (Claude) Applied.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
andrej: (Claude) Applied.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
andrej: (Claude) Not yet. It has one structural consumer.

`aie-translate --aie-flows-to-json` reads both representations of a route at once. `translateSwitchboxes` iterates `SwitchboxOp` for the per-switchbox demand counts and `total_path_length`. `translateCircuitFlows` and `translatePacketFlows` iterate `FlowOp`/`PacketFlowOp` and then walk the `ConnectOp`s inside those switchboxes to trace each route. Neither half alone produces the JSON.

This branch makes that hybrid impossible by default: a route is now either an `aie.flow` or a chain of `aie.connect`, never both. So `remove-lifted=false` is the only producer left of the IR that translator reads. 21 lit tests pass it and then run `--aie-flows-to-json`, and `tools/aie-routing-command-line/mlir2json.sh` does the same. Before this branch the stale-`aie.flow` bug supplied that IR by accident.

The option therefore names the one sanctioned exception to the invariant, for the one consumer that needs both sides of it.

The fix that would retire it is to stop making the translator re-derive a route. `ConnectivityAnalysis::traverse` already walks the same graph properly: it follows `aie.wire` instead of guessing the neighbour by coordinates, handles shim muxes through one code path, and records the ops of every hop in `PacketConnection::usedOps`. The translator reimplements that walk, and needs the switchboxes only because it does. If a recovered flow carried its route, the whole JSON would follow from the flows alone. That needs a representation for "the route this flow takes", which the dialect deliberately does not have. I would rather file that than grow this PR.

Separately: `keep-partial-flows=false` stands on much weaker ground. Its only user is `mlir2json.sh`, and it exists to work around `AIEFlowsToJSON` keying routes on `(source tile, source port)` and counting demand from flow endpoints. Say the word and I will drop it and put the filter in the translator instead.

lib/Dialect/AIE/IR/AIEDialect.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Neither could I, so I measured it. The check is gone.

I instrumented it to record every time it fires and ran the full lit suite (1489 tests). It never fired. I then disabled it and re-ran: the routing suites are unchanged and the whole suite shows only the two failures this machine always has.

Reading it back with that result, `neighbourSourceBundle` is always the opposite of `currDestBundle`, so the check can only fire when a flow's destination names a switchbox *input* port. `--aie-find-flows` never emits such a flow: a partial endpoint is a driven *output* port with no onward wire, and a tile endpoint is a Core or DMA port. The case the comment describes does not arise.

Removed the check and the comment.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied, with `clang-format`'s `InsertBraces` restricted to the lines this branch changed:

```
git clang-format --style='{BasedOnStyle: LLVM, AlwaysBreakTemplateDeclarations: Yes, InsertBraces: true}' <base> -- <routing files>
```

I did not put it in `.clang-format`, since that would reformat the whole tree on the next full run. Tell me if you want it there.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Reworded, and answers to both questions.

The comment now reads:

```
// Groups on one slave port carry different destination sets, so two of
// them must not claim the same id. A mask written on an aie.packet_flow
// claims every id it matches, including ids no flow in this design
// mentions, so the overlap does not show in the ids alone.
```

> Is a "stated rule" a pre-existing rule in the design, not generated by this router?

Yes. It is the `mask` attribute on an `aie.packet_flow` in the input IR, collected into `pinnedMasks`. The router derives the other claims itself, one cube per id, and hands them to `computeSubcubeCover`. I dropped the term from the comment, since it was carrying that distinction implicitly.

> this will not reject a pre-existing design that decides to fan-out the same ID to multiple ports, right?

It will not. A group is a set of `(slave port, id)` pairs that share a slave port **and** the same destination set. A fan-out is a single `(slave, id)` entry whose destination set has several members, so it forms one group, and the check compares a group only against *other* groups on the same port.

The error fires when two ids on one slave port go to different destination sets and their claims overlap. For example `packet_flow(8 mask 28)` to A and `packet_flow(9)` to B from the same port: rule (28, 8) matches 8 through 11 and rule (31, 9) matches 9, so id 9 has two destinations and the port cannot serve both. That is a genuine ambiguity, not a fan-out.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIECreatePathFindFlows.cpp
andrej: (Claude) Applied across the branch. See the reply on the earlier brace comment for the exact command.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Cut down to two lines rather than deleted outright:

```
// aie.flow names tiles, so an endpoint that is an element of a tile, such as an
// aie.shim_dma, has to resolve to the tile that owns it.
```

I kept that much because the constraint is the reason the function exists and it is not visible from the code: the router casts both endpoints of every flow to `TileOp`, so returning a `ShimDMAOp` result here aborts it. I reproduced that abort earlier in this review. Delete it if you still want it gone.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Reworded to name them:

```
// The same endpoint is reached twice when a broadcast splits and rejoins,
// and the tile, switchbox and shim-mux seeds can each reach one route.
// DeviceOp::verify rejects a flow declared twice.
```

"Three seeds" meant the three loops in `runOnOperation`: `findFlowsFrom` over every `TileOp`, then `findFlowsFromInterconnect` over every `SwitchboxOp` and over every `ShimMuxOp`.

On broadcast, I mean the case where the split paths meet again, not broadcast in general. `getConnectionsThroughSwitchbox` returns every destination port for one source port, so a broadcast puts several paths on the worklist; when two of them arrive at the same endpoint, `traverse` reports that endpoint twice. A broadcast that stays split is unaffected. The claim predates this branch, so I kept it rather than drop a case I have not disproved.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied:

```
// A fabric entry that ends on its own tile is an orphan connection, such as
// an undriven shim-mux write path.
```

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) The traversal is already shared: both seeds end in `traverse` and then `emitFlows`. The split is only in how a seed is produced, and there the two cases genuinely differ.

`findFlowsFrom` knows the flow's source before it starts, namely the tile plus the bundle and channel it iterates. `findFlowsFromInterconnect` starts at an input port and has to walk one wire upstream to work out what the source is, with three outcomes:

- upstream is a tile (PLIO; Core and DMA it skips) -- the source is that tile's port;
- an upstream interconnect drives the wire -- skip, this port sits mid-route;
- nothing drives the wire -- the source is this interconnect's own tile.

That mid-route skip is load-bearing, and `seen` cannot replace it. A flow seeded at an intermediate switchbox gets that switchbox as its source, so its `FlowKey` differs from the full flow's key and the duplicate survives as a second, shorter, wrong flow.

You are right that `findFlowsFrom` is close to redundant, though. A core or DMA reaches its switchbox through a wire, so the interconnect scan would classify it under the first outcome and derive the identical source. The `Core`/`DMA` `continue` exists only because `findFlowsFrom` ran first. Three things block a straight merge:

- the interconnect scan runs only under `clKeepPartialFlows`, so unifying changes what `keep-partial-flows=false` recovers unless the gate moves;
- `dropIntraTile` is keyed to which function runs. It has to key to "the seed is a fabric entry", or a core-to-DMA loopback on one tile gets dropped;
- flow emission order changes, which churns the `find-flows` lit tests.

The shape I would aim for is one seed enumeration yielding `(srcTile, srcBundle, srcChannel, startPort, isFabricEntry)`, with `findFlowsFrom` deleted and the tile case falling out of the upstream classification. That removes the duplicate enumeration and keeps the classification, which is irreducible. Tell me whether to do it here or as a follow-up.

One defect I have **not** changed, in the same function: the first branch takes `upOp->getResult(0)` for any op with `IsFlowEndPoint`, but that trait also sits on `aie.shim_dma`, `aie.core`, `aie.mem` and `aie.memtile_dma`. `resolveEndpointTile(upOp)` handles all of them and is already in the file. Say the word and I will fix it.

lib/Dialect/AIE/Transforms/AIEFindFlows.cpp
andrej: (Claude) Applied.

lib/Dialect/AIE/Transforms/AIEGenerateColumnControlOverlay.cpp
andrej: (Claude) Applied.

include/aie/Dialect/AIE/Transforms/AIEPasses.td
andrej: This will be addressed in #3421. Keeping it for now

include/aie/Dialect/AIE/IR/AIEOps.td
```diff
@@ -826,26 +826,50 @@ def AIE_PacketFlowOp: AIE_Op<"packet_flow", [SingleBlockImplicitTerminator<"EndO
...
-        OptionalAttr<BoolAttr>:$keep_pkt_header,
-        OptionalAttr<BoolAttr>:$priority_route
-  );
+  let arguments = (ins AIEI8Attr:$ID, OptionalAttr<AIEI8Attr>:$mask,
+      OptionalAttr<BoolAttr>:$keep_pkt_header,
+      OptionalAttr<BoolAttr>:$priority_route);
   let regions = (region AnyRegion:$ports);
 
-  let assemblyFormat = [{ `(` $ID `)` regions attr-dict }];
+  let assemblyFormat = [{ `(` $ID (`mask` $mask^)? `)` regions attr-dict }];
```
jskimko: `aie.packet_flow(0x10 mask 0x1c)` reads strangely to me. Some other examples:
```
aie.objectfifo @in0_mem_1(%mem_tile_0_1, {%tile_0_2}, 2 : i32)
aie.dma_bd(%arg1 : memref<4xi32> offset = 0 len = 4)
aie.masterset(South : 0, %0)
aie.rule(24, 24, %1) {is_ctrl_pkt_overlay}
```

I'd vote for at least `=` and softer preference for `,`. No preference on inside `()` or `attr-dict`.

andrej: agreed, `, mask = ...` looks better, I made the change

andrej:
@hunhoffe on Slack:
> I took a quick look and it seems reasonable to me!

which I will take as an approval and merge

PR #3718, opened 2026-09-11
[AIEX] Carry address_patch's arg_plus at its real 64-bit width [HIGH]

atassis:
## Description

`aiex.npu.address_patch` declares `arg_plus` as `I32`, but the field is 64-bit on
the wire. The consumer casts the 12-word DDR_PATCH op to aie-rt's
`XAie_CustomOpHdr` followed by `patch_op_t`, whose `argplus` is a `u64` at words
10-11 -- the same word-pair form as `regaddr` at 6-7 and `argidx` at 8-9. Word 11
was commented "reserved" and hardcoded to zero.

So a buffer offset at or past 2^32 was truncated and the BD patched with a wrong
address, with no diagnostic at any layer. I hit this on a 48-layer decode design
with a 9.8 GB arena: every buffer below 2^32 was written, and every buffer above
it read back exactly zero. Bisecting by layer count brackets the boundary to
78 MB below and 82 MB above 4,294,967,296.

The op now takes `AnyTypeOf<[I32, I64]>`, `getConstantInt64Operand` matches
without narrowing, and `txn_append_address_patch` writes both halves.
`buildArgPlusValue` widens only when the offset does not fit, so a design that
fits in 32 bits emits the identical `i32` constant and no existing IR or test
moves. The new test pins both words at `0xFFFFFFFF`, 2^32 and 8 GiB + 0x100,
plus an `i32` operand for the narrow path.

Buffer resolution and TXN re-import narrowed the widened value back to 32 bits, so
`createConstantArgPlus` now picks the narrowest width that holds the offset on
every path; tests cover the resolve-buffers and round-trip paths.

## Known limitations

`buildArgPlusValue`'s runtime path is `i32` throughout. A value not known at
build time cannot be range-checked there, and the static TXN target already
refuses a non-constant `arg_plus`, so a dynamic sequence on the C++ TXN builder
is the one path that still truncates.

Deriving the word offsets also exposes that `action` is written to word 5 where
the struct puts it at word 4, the alignment pad. Both are zero for every op
emitted here, so it is inert, and I left it alone.

## Checklist

On clang-tidy: this change adds no findings -- the same 10 `misc-static-assert`
hits appear on unpatched `main` in my setup (9 in `AIEXDialect.cpp`, 1 in
`BdLowering.cpp`), all on `assert()` calls the change does not touch, so I could
not get a clean local run to tick that box against.

### DIFF
--- a/include/aie/Dialect/AIEX/IR/AIEX.td
+++ b/include/aie/Dialect/AIEX/IR/AIEX.td
@@ -1045,7 +1045,7 @@ def AIE_NpuAddressPatchOp
     : AIEX_Op<"npu.address_patch", [AttrSizedOperandSegments]> {
   let summary = "address patch operator";
   let arguments = (ins UI32Attr:$addr, Optional<I32>:$addr_val,
-      OptionalAttr<I32Attr>:$arg_idx, I32:$arg_plus,
+      OptionalAttr<I32Attr>:$arg_idx, AnyTypeOf<[I32, I64]>:$arg_plus,
       Optional<AnyMemRef>:$buffer);
   let results = (outs );
   let assemblyFormat = [{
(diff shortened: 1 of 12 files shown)


andrej (approved):
Verified the firmware accepts larger arg_plus values, so this is good.

PR #3727, opened 2026-09-11
[python] ExternalFunction: key the digest on include content, not directory mtime [HIGH]

atassis:
## Description

`ExternalFunction`'s memoization digest keyed each include directory on the
directory's own mtime. That mtime moves when an entry is created, deleted or
renamed, but not when a file under it is rewritten in place — which is the common
case a cache key has to catch, and what an editor that truncates-and-rewrites a
header does. An edited header therefore left the digest unchanged and the cached
kernel object was reused against stale source.

The digest now hashes the contents of every regular file under each include
directory, read in chunks so a large header set does not sit in memory. It also
adds the source file's own directory, which `compile_external_kernel()` already
appends to the include path — so a same-directory sibling include (for example
`aie_kernels/aie2p/mm.cc` including `zero.cc`) was previously covered by no
hashed directory at all unless the caller declared it separately.

Unreadable or missing paths hash to a marker rather than raising, so a permission
error degrades to a cache miss instead of failing the build.

`test/python/test_kernels_memoization.py` covers the in-place rewrite, the
sibling-include case, and that an unchanged tree still hits.

## Checklist

### DIFF
--- a/python/iron/kernel.py
+++ b/python/iron/kernel.py
@@ -83,6 +83,32 @@ def _maybe_collapse_to_match(arg, expected_ty):
     return memref.collapse_shape(exp_mr, arg, reassociation)
 
 
+_DIGEST_CHUNK = 1 << 20  # streamed read bound for _hash_dir_contents
+
+
+def _hash_dir_contents(d: str) -> str:
+    """Recursive content digest of every regular file under ``d``.
+
+    Replaces a directory mtime, which moves on create/delete/rename but not
+    on an in-place rewrite (e.g. an editor that truncates-and-rewrites a
+    header) -- the common case a cache key needs to catch.
+    """
+    h = hashlib.sha256()
+    try:
+        paths = sorted(p for p in Path(d).rglob("*") if p.is_file())
+    except OSError:
+        return "<missing>"
+    for p in paths:
+        h.update(str(p.relative_to(d)).encode())
+        try:
+            with open(p, "rb") as fh:
+                while chunk := fh.read(_DIGEST_CHUNK):
+                    h.update(chunk)
+        except OSError as exc:
+            h.update(f"<unreadable:{type(exc).__name__}>".encode())
+    return h.hexdigest()
+
+
 class BaseKernel(Resolvable):
     """Base class for AIE core functions that resolve to a func.func declaration.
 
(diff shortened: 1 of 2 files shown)


jgmelber (approved):

hunhoffe:
Note: please don't merge yet, I'd like a chance to review.

hunhoffe (changes requested):
Thanks you -- the `src_dir` fix is a good catch that I'd like to see land!
  
My concern is that the digest hashes the wrong set of files, and that the motivating case doesn't require content hashing to fix.

1. This keys on `-I` roots, not the include closure.

ccache does content-hash headers — but in direct mode it hashes the headers actually included, learned from the previous compile's dependency manifest. CMake and Ninja likewise key on the real include graph via -MD depfiles. None of them hash every file under every `-I` root.

That distinction matters a lot here because of what the roots are `. _include_dirs()` returns [`config.cxx_header_path()`] (python/iron/kernels/_common.py:74), and `cxx_header_path()` is [`<root>/include`](python/utils/config.py:104) — which per the install rules in CMakeLists.txt:411-440 contains `include/aie`, `include/aie-c`, and `aie_kernels`. So compiling a kernel that includes two or three headers now digests the
  entire MLIR-AIE dialect header tree: `.td` files, `OpInterfaces`, and the tablegen-generated `.inc/.gen` output from the build directory. None of that is reachable from a kernel `.cc`.
  
2. The generated files make this a false-miss source.

Because `${PROJECT_BINARY_DIR}/include/aie` is installed into the same root, the tablegen output is inside the hashed set. Rebuilding mlir-aie regenerates those files, which moves the digest, which invalidates every cached kernel object — even though no kernel source and no actually-included header changed. For anyone developing on the compiler itself, that's a persistent cold cache. Same effect at smaller scale from `aie_kernels/` sharing the root: editing any one kernel busts the digest for all of them.
  
3. The motivating case doesn't need content hashing.

From the description:
> That mtime moves when an entry is created, deleted or renamed, but not when a file under it is rewritten in place

Right — directory mtime doesn't move. But file mtime does. An editor's truncate-and-rewrite, a `git checkout`, any in-place save all bump `st_mtime_ns` on the file. A per-file `(relpath, st_mtime_ns, st_size)` digest is stat-only, needs no reads, and catches every scenario named in the PR body and the test comments. Content hashing only additionally covers mtime-preserving writes (`cp -p, rsync --times, touch -r`), which seems like a thin reason to read the whole tree on every run of the path whose entire job is to be fast.
  
4. `rglob` doesn't recurse into symlinked directories, so a symlinked subtree under an include root contributes nothing to the digest — a silent false hit, which is the same bug class this PR is fixing. Install and build trees symlink include dirs fairly often, so I don't think the patch is sound even at its current cost.

5. Error handling is coarser than intended. The except `OSError` wraps the whole `sorted(...)`, so a single unreadable file mid-traversal collapses the entire directory to "<missing>" — indistinguishable from an absent directory, and it discards the traversal work already done.

6. A permissions error means that the cache may always be cold; I'd say it's more user friendly to percolate up that failure than to hide it and accept performance degradation without notifying the user.

PR #3728, opened 2026-09-11
[AIE] Take three hardcoded bounds from the target model, and bind them [MED]

atassis:
## Description

Three places that hardcode or cannot reach a value the target model already owns.

`AIETargetBCF` computed the end of the reserved span from a literal rather than
from the model, which is the same class of defect as the program-region size fixed
in #3618 — a constant that is right for one target and unchecked for the others.

The DMA buffer-descriptor field widths (`getDmaBdStepBits`, `getDmaBdWrapBits`,
`getDmaBdIterBits`) and `getAddressGenGranularity` had no Python binding, so a
design that needs to respect a BD bound had to hardcode it. Same for the per-tile
BD count, which is not uniform — on AIE2 a MemTile has 48 while a CoreTile and a
ShimNOCTile have 16 each — so a caller must name the tile type rather than assume
one number.

Follows the precedent of #3657 and #3496, which moved `verifyStridesWraps`' field
widths and the core stack-size default onto `AIETargetModel`.

## Checklist


jgmelber (approved):

atassis:
Hi
I have addressed a CI issue here

PR #3734, opened 2026-09-11
[AIEX] Wait for DMA task-queue space instead of dropping transfers

hunhoffe:
## Description

A shim DMA channel's task queue is four deep. Pushing onto a full one does not stall or error: the hardware drops the push, that transfer never runs, and whatever was waiting on it waits forever. There is no backpressure, and nothing in the IR says whether it will happen, because it depends on how fast the consumer drains.
  
On Strix, a sequence of 14 shim MM2S pushes returns `ERT_CMD_STATE_TIMEOUT` 5 runs out of 5. With this series it completes 5 out of 5.
  
This models the queue and emits a new `aiex.npu.maskpoll` before any push that could land on a full one, blocking until `Task_Queue_Size` reports a free slot. It is on by default; `--no-enforce-dma-queue-depth` turns it off.
  
The retirement rule lives in one place (`DmaQueueModel.h`) and all three lowering paths use it, so they cannot drift about the same hardware: `dma_start_task`, `npu.dma_memcpy_nd`, and the dynamic BD pool. Awaiting a token retires that token and everything queued ahead of it, which matters -- counting pushes against awaits reports designs that are perfectly bounded.
  
Runtime-bound loops are the awkward case, since there is no trip count to count against. `analyzeLoopQueue` runs the body against the model until the state repeats, which terminates because a guarded push caps occupancy at the queue depth. Loops are taken whole at the top level so the straight-line prefix hands its exact state to the loop, and a nested loop is not settled once per enclosing level.
  
Where the target reports no pollable occupancy register (AIE2PS, xcve3858), enforcement warns with an attached note rather than declining quietly.
  
Also here: `aiex.npu.maskpoll` on the C++ TXN path, a correction to the BD-reuse diagnostics, an opt-in `warn-unsafe-bd-reuse`, and aiecc diagnostics now reaching Python callers on a build that succeeded.
  
Cost is in the noise. 4-column GEMM 512³, min of 200 pinned dispatches, A/B/A/B: 741.8/743.2/746.0/745.9us enforced against 746.3/738.3/755.4/731.5us unenforced, where the unenforced spread alone is 24us.
  
### Notes for reviewers
  
Two things worth arguing with.
  
`rolled_loop.mlir` now pins `enforce-queue-depth=false` on both sides. A runtime-bound loop has to be guarded while a hand-unrolled n=2 visibly cannot overflow, so the two stopped being byte-identical. The test asserts BD register equivalence, not guard equivalence, but it is a real narrowing of what it covers.
  
Enforcement is on by default rather than opt-in. Of 45 in-tree designs that lower, one gets a poll, and it is a true positive. The argument for the default is that the failure it prevents is silent.

## Checklist


hunhoffe:
@copilot+gpt-6-astra can you address review findings, do a thorough quality audit (and fix any issues you find), and address the CI error?

hunhoffe:
@copilot+gpt-6-astra  can you address review findings, do a thorough quality audit (and fix any issues you find), and address the CI error?

hunhoffe:
@copilot+gpt-6-astra can you address previous review findings?

hunhoffe:
@copilot+gpt-6-astra can you address review feedback and CI failures, including formatting?

hunhoffe:
@copilot+gpt-6-astra please address the review feedback and update the branch.

hunhoffe:
@copilot Can you help me draft an update title and CONCISE and CLEAR PR body (I'll do the actually update) and update the branch as well, resolving conflicts as necessary?

hunhoffe:
@copilot+gpt-5.6-sol resolve the merge conflicts in this pull request

andrej (approved):
Seems good at a glance.

(How) does this handle "implicitly synced" tasks? Example: 

1. Push task A to fill inputs (_without_ emitting a TCT).
2. Push task B to drain the output.
3. The core at the other end of these transfers produces output for task B only after receiving all input from task A.

As authors of the program, we can reason that task B completing always means task A also completed. We know the core imposes a dependency from B onto A, so syncing on B "implicitly" also syncs on A. 

But that same reasoning that we as authors have would be quite hard to establish in the compiler (would need to trace through core code and lock dependencies; not sure if always decidable, almost certainly out of scope for this PR). 

How do we handle that case? Do we under- or over-approximate, or do we error and expect the user to opt out with the new compiler flag?

andrej:
Actually re-thinking this -- is it that you do not need to be able to decide the above, you just emit a maskpoll for before any new task submission? But then what do you need the modeling code for?

hunhoffe:
@andrej Thanks for the review!

Implicit-sync cases just show up as outstanding occupancy: it's a safe over-approximation, so no error/opt-out is needed, worst case is (I believe) a redundant poll. On the "why not always poll" question — you're right that it would sidestep the approximation question entirely and eliminate most of the control-flow analysis (and most of what review has been finding bugs in). The two things that would still need it: register-less targets (still need to decide when a warning is earned instead of firing on every push) and warn-unsafe-bd-reuse, which is a different hazard. 

We could definitely implement something correct that is an unconditional check; the queue-modeling is purely meant to be an optimization and to make sure this new logic perturbs the emitted instructions only when needed.

hunhoffe:
@copilot+gpt-5.6-sol can you look into CI failure and resolve conflicts and update?

PR #3740, opened 2026-09-15
aie_kernels: finish IRON kernel migration 

efurst:
## Description

  Extends the shared `aie_kernels/` library so a single source of truth can back
  IRON's operator set, and adds the kernel-factory + test coverage each new kernel
  needs. Follow-on to #3599.

  **Runtime element-count variants** — the eltwise/activation kernels hardcoded the element count. Adds _size entry points that read it at runtime, alongside the unchanged fixed-1024 symbols:
  - eltwise_{add,mul}_bf16_vector_size (aie2 + aie2p): a scalar tail handles a size not divisible by the vector step.
  - {silu,gelu,relu}_bf16_size (aie2 + aie2p): forward the runtime size the inner compute already accepts. New aie2p/relu.cc — aie2 had ReLU, aie2p did not.
    
  **Norm variants**
  - `rms_norm_eps` (runtime epsilon, e.g. Gemma's 1e-6) alongside the fixed
    `rms_norm`; new `aie2/rms_norm.cc` (AIE2 lacks an f32 vector multiply, so it
    applies the reciprocal-RMS scale as a bf16 pair accumulated in f32).
  - New `aie2/layer_norm.cc` (aie2p had one, aie2 did not), f32-accumulator
    row-sum to match the aie2p numerics.
    
  **Other kernels**
  - `mm_fused` (+ `mm_fused_mmul.h`, `activations.h`): a fused GEMM kernel that
    accumulates over K in an f32 L1 accumulator with an activation epilogue.
  - `rope`: moved `aie2p/rope.cc` → `generic/rope.cc` (compiles for both arches)
    and added the two-halves scalar tail; updated the rope example + READMEs.
  - `mask_bf16` added to `aie2/softmax.cc` (aie2p already had it).
  
  **Factories + tests** — `aie.iron.kernels` factories, spec-table entries, and
  end-to-end hardware tests for the new symbols (`add_sized`/`mul_sized`/
  `relu_sized`/`silu_sized`/`gelu_sized`, `rms_norm`/`rms_norm_eps`/`layer_norm`,
  `rope`). `mm_fused` has no single-call factory (three composable symbols driven
  by a tiled loop nest).
  
  ## Checklist


andrej (approved):
Looks good to me!

PR #3759, opened 2026-09-17
Add q4nx to bfp16ebs8 dequantization kernel

andrej:
Dequantizes one q4nx block (32 out-features x 256 in-features; 512B scales, 512B mins, 4096B of 4-bit codes) and emits it as bfp16ebs8 in the order a blocked GEMM reads its B operand. Block geometry arrives as -D flags so one kernel serves every problem shape.

https://github.com/amd/IRON/pull/204 depends on this


jgmelber (approved):

aie_kernels/generic/q4nx_dequant.cc
```diff
@@ -0,0 +1,113 @@
...
+#include <aie_api/aie.hpp>
+
+#if !defined(Q4NX_M_TILE) || !defined(Q4NX_K_TILE) || !defined(Q4NX_GROUP) ||  \
+    !defined(Q4NX_CT_K) || !defined(Q4NX_S) || !defined(Q4NX_T)
+#error "q4nx_dequant.cc needs its geometry -D defined"
+#endif
+
+namespace {
+
+constexpr int M_TILE = Q4NX_M_TILE;
```
hunhoffe, line 23: nit: Are these redefinitions really needed?

aie_kernels/generic/q4nx_dequant.cc
```diff
@@ -0,0 +1,113 @@
...
+constexpr int GROUP = Q4NX_GROUP;
+constexpr int CT_K = Q4NX_CT_K;
+constexpr int SS = Q4NX_S;
+constexpr int TT = Q4NX_T;
+
+// n rows in one vector: 128 nibbles is one 512-bit load and the file stores 16
+// n per k, so a load spans 8 k.
+constexpr int PR = 16;
+
+constexpr int SCALES = M_TILE * K_TILE / GROUP;
```
hunhoffe, line 34: nit: Are there any divisibility checks?

hunhoffe (approved):
😄

aie_kernels/generic/q4nx_dequant.cc
andrej: Copilot flagged that too, adding them, thanks for catching

aie_kernels/generic/q4nx_dequant.cc
```diff
@@ -0,0 +1,113 @@
...
+        // Indexed k*8 + n. The transpose makes it n*8 + k, so the 8
+        // values sharing an exponent are 8 k for one n. It must precede
+        // the conversion, which fuses them.
+        auto up = aie::concat(z01.first, z23.first, z45.first, z67.first);
+        auto dn = aie::concat(z01.second, z23.second, z45.second, z67.second);
+
+        aie::accum<accfloat, TILE_VALUES> a_lo(aie::transpose(up, TT, SS));
+        aie::accum<accfloat, TILE_VALUES> a_hi(aie::transpose(dn, TT, SS));
+        s_lo << a_lo.template to_vector<bfp16ebs8>();
+        s_hi << a_hi.template to_vector<bfp16ebs8>();
```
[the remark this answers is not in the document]
andrej: (Claude) Done, but with `floor` rather than `conv_even`.

The mode is now selected explicitly and restored on exit, following the `rms_norm.cc` / `layer_norm.cc` `swap_rounding` + `set_rounding` convention.

On the choice of mode: `floor` is what a core powers up in, and since nothing in this kernel previously set the mode, `floor` is what the IRON operator and its downstream Gemma use case have been getting all along. The operator's tests are byte-exact against a host reference that truncates, so switching to `conv_even` would change numerics end to end. Making that change is a deliberate accuracy decision, separate from fixing the sticky-state bug you correctly identified, so this commit fixes only the latter.

For the record, there is a real argument for `conv_even` here, and it is not the one in `mm_fused.cc`. This kernel does no reduction, so truncation costs it at most 1 ulp per weight with no accumulation. But the weights it emits are the B operand of a GEMM, and a systematic downward bias in B does accumulate over that GEMM's K reduction. Worth revisiting; it needs an accuracy measurement on the model, not just a kernel change.

Verified: 255 byte-exact operator tests still pass against a from-source build, confirming the explicit `floor` reproduces the prior behaviour exactly.

aie_kernels/generic/q4nx_dequant.cc
```diff
@@ -0,0 +1,113 @@
...
+constexpr int M_TILE = Q4NX_M_TILE;
+constexpr int K_TILE = Q4NX_K_TILE;
+constexpr int GROUP = Q4NX_GROUP;
+constexpr int CT_K = Q4NX_CT_K;
+constexpr int SS = Q4NX_S;
+constexpr int TT = Q4NX_T;
+
+// n rows in one vector: 128 nibbles is one 512-bit load and the file stores 16
+// n per k, so a load spans 8 k.
+constexpr int PR = 16;
```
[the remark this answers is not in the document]
andrej: (Claude) Agreed, added:

```cpp
static_assert(TT == 8, "bfp16ebs8 shares one exponent across 8 values");
static_assert(SS == 8, "the column unroll and the interleave tree are 8 wide");
static_assert(PR * SS == 128, "one 512-bit load holds 128 nibbles");
static_assert(M_TILE % PR == 0, "rows must divide into PR-row groups");
static_assert(K_TILE % CT_K == 0, "k tile must be a multiple of the k slice");
static_assert(CT_K % SS == 0, "k slice must be a multiple of s");
static_assert(GROUP % SS == 0, "a group must not split a k step");
static_assert(K_TILE % GROUP == 0, "k tile must hold whole groups");
```

The `GROUP % SS` one is the invariant your comment did not mention but that I think is the easiest to violate: one scale and one min are loaded per `i`, so the `s` columns of a step have to fall inside a single quantization group.

This matches how `mm_fused.cc` guards its own geometry flags.

aie_kernels/generic/q4nx_dequant.cc
```diff
@@ -0,0 +1,113 @@
...
+constexpr int BLOCKS_PER_RUN = (CT_K / SS) * TT;
+constexpr int BLOCKS_PER_KSLICE = (M_TILE / TT) * BLOCKS_PER_RUN;
+// bfp16ebs8 pointer arithmetic counts bytes, not blocks (llvm-aie#1232).
+constexpr int BLOCK_BYTES = TT + 1;
+
+} // namespace
+
+extern "C" {
+
+void q4nx_dequant_bfp(const uint8_t *__restrict qw, bfp16ebs8 *__restrict out) {
```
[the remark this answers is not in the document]
andrej: (Claude) The coverage exists, but in IRON rather than here, and I would rather not duplicate it.

`iron/operators/flm/dequant/` has 255 tests that are byte-exact against a host reference, covering exactly the three things you list: the q4nx payload layout (scales, mins and 4-bit codes), the `min + scale * code` formula, and the bfp16/GEMM block ordering including the transpose and exponent grouping. They run on device across every shape the operator supports. A wrong `ks` or row stride fails them immediately; that is how the `uint4` byte-stepping bug in this kernel was originally caught.

That split is also the established convention for this class of kernel. `mm_fused.cc`, the GEMM this kernel feeds, and `expand.cc`, the existing q4nx dequant, both have zero tests in this repository and are tested from IRON. Of the 73 kernels in `aie_kernels`, most have no test here at all.

Reproducing a device-level harness here would mean reimplementing the reference and the layout model, giving two things to keep in sync for no additional signal. If the project wants `aie_kernels` tests to become mandatory, that seems worth doing as its own change across the directory rather than for this one kernel.

aie_kernels/generic/q4nx_dequant.cc
andrej: (Claude) `SS` and `TT` genuinely have to be renamed: `S` and `T` collide with template parameters in the aie2p built-in headers, which is what the `Q4NX_` prefix on the flags is for.

Given that, I kept the other four for uniformity and type safety, so the indexing math reads `M_TILE` rather than mixing `Q4NX_M_TILE` with `SS`. Happy to drop the four that do not strictly need it if you would prefer the macros used directly.

aie_kernels/generic/q4nx_dequant.cc
andrej: (Claude) Added, see the reply on the Copilot thread above for the full list.

aie_kernels/generic/q4nx_dequant.cc
andrej: @hunhoffe  I think the purpose of this is to avoid name collisions/accidental overrides, since `M_TILES` is relatively generic and a macro might expand, e.g. in the AIE API headers. There is precedent for this in the other kernels that use macros for compile-time parametrizations, such as [in the FLM GEMM](https://github.com/Xilinx/mlir-aie/blob/a74cde78e8ac15b91fc0561e2342e0da90adb4ad/aie_kernels/generic/mm_fused.cc#L48-L65).

Decided to leave as is for these reasons

PR #3760, opened 2026-09-17
aie_kernels: add flash-attention prefill and channels-last dwconv1d (aie2p), de-duplicate arch-copy kernels

hunhoffe:

## Description

Two new AIE2P kernels extracted from a Gemma4 IRON+Peano port, plus removal of a class of duplication in `aie_kernels/` that the new conv kernel made visible.                                                                                    
                                                                                                                                  
## New kernels                                                                                                                  
                                                                                                                                  
**`aie2p/flash_attn_prefill.{h,cc}`** — flash-attention *prefill* with online softmax; `aie2p/mha.cc` covers only the decode side. Two geometries — global (head_dim 512, chunk 8) and sliding-window (256, chunk 16) — from one templated header and one TU, under distinct `attn_*` / `swa_*` prefixes. `PrefillGeom<DH>` is specialized per head_dim and carries only the four steps genuinely tuned per variant (`attn_qk`/`attn_fv` decompose 1×1 vs 2×2, `calculate_l` one accumulator vs two with `mac_elem_16_accuracy_safe`, `reorder_s` global-only); the rest is shared templates.                                                                         
                                                                                                                                  
Entry points are split per step so a core body can drive the round and block loops with an acquire point per object — the precondition for ObjectFifo Q/K/V and O legs. They carry no locks: synchronization and the core-to-query-grid index math belong to the design.                                                                                                
                                                                                                                                  
**`aie2p/dwconv1d_channels_last.cc`** — depthwise conv1d emitting one timestep across all C channels, per-channel taps, optional clamp. A depth-K ObjectFifo is itself the sliding window. Templated on tap and channel count.                                                                  
                                                                                                                                  
## The two dwconv1d kernels                                                                                                     
                                                                                                                                  
`dwconv1d.cc` → `dwconv1d_channels_first.cc` (entry point and `-D` flags likewise). Layout is the whole distinction between them — it picks the vectorization axis, the tap representation, and which is faster — and a bare `dwconv1d.cc` said none of that.                                

README gains a *Choosing a depthwise conv1d* section: channels-last retires 12.2 MACs/instruction-slot vs 4.2, because `sliding_mul` spends half its vector on the window halo and rebuilds each tap's operand with a `vshift`. That is **not** a reason to prefer it — it only pays when the data is already channels-last, and costs C×K resident weights, a compile-time C, and program memory linear in C. Both stay.                                  

## De-duplication                                               
                                 
Four kernels existed as near-identical `aie2/` + `aie2p/` copies differing only by an arch-implied constant: `zero.cc` (store width 256/512), `add.cc`/`mul.cc` (`vec_factor` 16/32), `conv2dk1_i8.cc` (`NUM_ACC`/`MMUL_M` 8/4 vs 4/8). That is what `generic/` plus `__AIE_ARCH__` is for — `generic/activations.h` already uses the pattern — so each collapses to one file. `_kernel_source()` already falls through `<arch>/` to `generic/`, so no Python changes needed.

Also: `mm_bfp_mixed.cc` kept a private `zero_vectorized` beside `zero.cc`, and the combos macros were byte-identical across three matmuls (`mv.cc` keeps its own — genuinely different, N is 1). Moving `zero.cc` additionally repairs `generic/mm_fused.cc`, whose `#include "zero.cc"` could only resolve with an arch `-I` that `_include_dirs()` does not supply.

## Checklist


hunhoffe:
@copilot+gpt-6-astra please address review findings. Also, there was recently a large rework of kernel code and testing in main. Please update this branch to fit the new model, including tests for the new kernels.

hunhoffe:
@copilot+gpt-5.6-sol please address recent review and fix CI issues too

hunhoffe:
@copilot+gpt-5.6-sol please address review comments and CI failures

hunhoffe:
@copilot+gpt-5.6-sol please fix CI errors

hunhoffe:
@copilot can you please fix current workflow failure here: https://github.com/Xilinx/mlir-aie/actions/runs/36031835104/job/107742229115#step:7:1840 We want to run no matter the power mode, just record it throughout.

hunhoffe (commented):
Looks good to me! Further cleanups coming in additional PRs.

jgmelber (approved):

PR #3584, opened 2026-08-24
[utils] Add prefix_symbols_in_object() for bulk symbol renaming

hunhoffe:
## Description

IRON's operator-fusion feature hand-rolls a bulk "prefix every defined symbol in this object" operation (nm to list symbols, then a single objcopy --redefine-syms= pass) to avoid collisions when linking several independently-compiled kernel objects into one fused module. Upstream only had the single-symbol version (_rename_symbol_in_object).

Add a public prefix_symbols_in_object(object_path, prefix), plus the config.nm_path() companion to config.objcopy_path(), so IRON can later call into mlir-aie instead of maintaining its own copy. Ported as a single cross-platform Python implementation (no sh/awk, no os.name branch) rather than IRON's two-branch version, but preserves its most important property: nm's exit status is always checked before objcopy ever runs, so a failing nm can't silently produce an empty rename map.

This PR also adds public ObjectFile(..., symbol_prefix=...) / .bind(...) APIs, exposes the shared handle through Kernel.object_file and ExternalFunction.object_file, and migrates linalg companion bindings to it. Compiled-object prefixing uses content-validated metadata for cache recovery. Lit Python/pytest commands use per-test cache namespaces so unrelated parallel tests cannot share mutable compilation artifacts, while RUN lines within one test retain cache reuse. These changes support consistent symbol bindings and reliable testing of the bulk-prefix operation; higher-level IRON fusion orchestration remains out of scope.

Design credit for the operation this replaces goes to IRON's _prefix_symbols (iron/common/compilation/base.py): original implementation by André Rösti.

This is part of the deduplication effort between mlir-aie and IRON repos.

## Checklist

- [ ] Tests pass locally, or the change is covered by CI
- [ ] Docs / docstrings updated if the change is user-facing
- [ ] Code is formatted (`clang-format` for C++, `black` for Python — see [CONTRIBUTING](../CONTRIBUTING.md))
- [ ] `ruff check` and `pyright` pass locally for any touched Python file in their covered paths (see [CONTRIBUTING](../CONTRIBUTING.md#linting-python))
- [ ] `clang-tidy` passes locally for any touched file on the enabled list (see [CONTRIBUTING](../CONTRIBUTING.md#static-analysis-for-c-clang-tidy))

PR #3774, opened 2026-09-18
[utils] Resolve LLVM binutils from both bundled toolchains, not just one

hunhoffe:

## Description

Follow-up to #3584.                                                                                                                                                                                                
                                                                                                                                                                                                                     
`objcopy_path()`/`nm_path()` searched only the MLIR-AIE `bin/` directory, then `PATH` under the bare name. But the MLIR-AIE and Peano trees are complementary: `llvm-objcopy` ships only in the former, `llvm-nm` and `llvm-ar` only in the                                                                                                                                       
  latter. So on a stock checkout `nm_path()` raises `Could not find llvm-nm` with `llvm-nm` sitting in the Peano `bin/` next door, and distro spellings like `llvm-nm-18` go unseen. Downstream consumers carry their own lookup to cope.                                                                                                                                       
                                                                                                                                                                                                                     
Resolution is now one `_find_llvm_tool()`: env override (`AIE_OBJCOPY_PATH`, `AIE_NM_PATH`, `AIE_AR_PATH`, mirroring `AIECC_PATH`) → both bundled `bin/` dirs → `PATH` incl. versioned spellings. Candidates that fail to exec are skipped for a later one; if all are broken the first is returned anyway so the caller sees that tool's own error. Adds `ar_path()`.                                                                                                                                                               
                                                                                                                                                                                                                     
**Rationale correction.** The claim that GNU binutils can't parse AIE objects holds only for `objcopy`, and not for the stated reason. On a real `aie2p` object, GNU Binutils 2.42 `nm`, `ar` and `objdump` all read it fine; only `objcopy` fails. The object isn't malformed — it's clean ELF32/LE/REL, and `objcopy -I elf32-little --redefine-sym=...` succeeds. Only auto-detection fails, since no BFD backend claims `e_machine 0x108`. So `llvm-objcopy` is a hard requirement (it needs no hint), while `llvm-nm`/`llvm-ar` are preferences. Docstrings and the `CMakeLists.txt` comment now say this.                                                                                                                                                          
                                                                                                                                                                                                                     
**Tests.** Both files previously stubbed out `compile_cxx_core_function`, `nm_path`, `objcopy_path` and `subprocess.run`, so nothing touched an object file. Now a real kernel is compiled with Peano and inspected with the real `llvm-nm`; discovery runs in a real subprocess over real executables on a real `PATH`; failures use real broken inputs. Gated `# REQUIRES: peano`, no NPU, ~38s. This caught a bug in this change — `shutil.which` returns only the *first* `PATH` match, leaving no fallback past a broken tool — and is what  disproved the GNU-nm claim. 

## Checklist


hunhoffe:
@copilot+gpt-6-astra please address the CI failure and review comments.

andrej (approved):

